PR #48, opened 2019-03-22
Project Management Docs

LamarMoore:
Added first drafts of some documents which could constitute a basis for governance of software endeavors within the DO group.


governance/ProjectManagementCommittee-Terms.md
```diff
@@ -0,0 +1,31 @@
...
+The Project Management Committee operates on the following basis:
+
+ * The Committee meets at quarterly in person.
+ * The Committee can be reached via a mailing list.
+ * The Committee will elect a chair who should be a senior member within the group or the division head.
+
+Members
+-------
+
+ - Sean Langridge (Chair)
```
LamarMoore: As mentioned before this should help with senior management buy-in

governance/CodingStandards.md
```diff
@@ -0,0 +1,100 @@
+# Coding Standards for Disordered Materials Software Projects
+
+## Fortran Standards (Taken from Fortran 90 Standards for JULES)
```
LamarMoore: These standards probably need to be expanded.

governance/CodingStandards.md
```diff
@@ -0,0 +1,100 @@
...
+    =  'This is a very very very very very very very '  &
+    // 'very very very very very very very very very '  &
+    // 'long character assignment'
+```
+- Short and simple Fortran statements are easier to read and understand than long and complex ones. Where possible, avoid using continuation lines in a statement.
+- Avoid putting multiple statements on the same line. It is not good for readability.
+- Each program unit (module, subroutine, function etc.) should follow a structure similar to the templates supplied in [Standard code templates](http://jules-lsm.github.io/coding_standards/standard_code_templates/index.html). The intended behaviour of the unit should be clearly described in the header.
+- Comments should start with a single ! at beginning of the line. A blank line should be left before (but not after) the comment line. The only exception is for one line comments which can be indented within the code or placed after the statement.
+- Each subroutine, function and module should be in a separate file. Modules may be used to group related variables, subroutines and functions.
+## C++ Standards (Taken from Mantid C++ Standards)
```
LamarMoore: As stated here, taken from Mantid. Might be an idea to remove this reference to Mantid but also to include anything else you feel should be there (or remove anything which shouldn't).

governance/TechnicalSteeringCommittee-Terms.md
```diff
@@ -0,0 +1,36 @@
...
+ * Acts as an arbitration/triage point for old, controversial and large issues.
+ * If the committee cannot come to a majority consensus, then the issue will be raised at the PMC.
+
+Members
+-------
+
+ - Tristan Youngs (Chair)
+ - Daniel Bowron
+ - Lamar Moore
+ - Catherine Jones (Suggested from SCD)
```
LamarMoore: We could probably use another technical expert. I would nominate either Anthony Lim or Nick Draper (but you may not want a contractor, however he is very experienced both in project management and technical aspects and would be  great asset).

PR #98, opened 2019-11-19
Forcefields Py4OH, Py5 and NTf2 by Ludwig group

johannabusch:
Forcefields Py4OH, Py5 and NTf2 by Ludwig group are added


johannabusch:
Looks good! Suggestions:

1) Rename forcefield source files from, e.g., py5_ludwig.cpp to py5.cpp - since they're in a directory of their own, no need to qualify them with "_ludwig" anymore.
2) Remove the determineAtomTypes() function completely (from .h and .cpp) - typing is now handled by the NETA definitions, so we don't want to define this function at all.
3) Descriptive text in header files really should say "Sole responsibility of JB." :)

PR #123, opened 2020-01-17
Updated description to match GUI

sabrina-gaertner:
Updated atom type names to match GUI.


trisyoungs:
Not merging, as this highlights a bug (new atomtypes should be named after element symbol not element name) rather than an error in the documentation.

PR #217, opened 2020-02-12
PointerArray based on std::vector

rprospero:
A much smaller proof of concept of moving the template classes onto standard containers.  It only touches a single file, since PointerArray is left as a facade around std::vector.

To keep things simple, no changes were made to the class interface, though the following improvements could be made:

* items() is unused and could be removed
* at(), value(), and operator[]() are all identical and two of those could be removed.

### DIFF
--- a/src/templates/pointerarray.h
+++ b/src/templates/pointerarray.h
@@ -41,6 +41,7 @@ template <class T> class PointerArray : std::vector<T*>
 	public:
 	using std::vector<T*>::begin;
 	using std::vector<T*>::clear;
+	using std::vector<T*>::data;
 	using std::vector<T*>::end;
 	using std::vector<T*>::erase;
 	using std::vector<T*>::operator[];
@@ -73,6 +74,11 @@ template <class T> class PointerArray : std::vector<T*>
 	{
 		return operator[](index);
 	}
+	T** items()
+	{
+		return data();
+	}
+
 
 	/*
 	 * Item Management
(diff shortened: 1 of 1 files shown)


src/templates/pointerarray.h
```diff
@@ -34,121 +36,48 @@
  * A contiguous pointer array class designed to be as lightweight as possible.
  * Order of items must not be important, as the arrangement of pointers is subject to change.
  */
-template <class T> class PointerArray
+template <class T> class PointerArray : std::vector<T*>
```
LamarMoore: Is it necessary to use std::vector at all? The methods this class implements can all be found in the STL. This should probably be removed.

LamarMoore: If you are worried about the effects on all of the files which use this. You could just do `using PointerArray = std::vector<T*>` which will solve that problem.

rprospero: I had considered that.  I was worried that not keeping the facade would lead to changing too many files at once, but it's worth at least trying.  I'll see how big of a commit it would be to just make the switch.

LamarMoore: https://en.cppreference.com/w/cpp/language/type_alias

rprospero: PR #240 performs this kind of switch.  I kept it in a separate PR and branch in case we decide that it's too large of a change.

rprospero:
This has been superseded by #240.

PR #240, opened 2020-03-03
Replace PointerArray<T> with type alias to std::vector<T*>

rprospero:
This completely removes the PointerArray class and replaces it with std::vector.  Since the method names were not identical between the two classes, this required some in place editing where PointerArray had been used.

As port of the modernization, all the locations where the nItems() member function was used to set the limits on a for loop has been replaced with a proper range based for loop.

This PR is based on the feedback that @LamarMoore provided on pr #217.  At most, only one of these two requests should be merged.


src/gui/viewer_render.cpp
```diff
@@ -371,9 +371,8 @@ void BaseViewer::postRedisplay()
 	PointerArray<BaseViewer> updateStack;
 	createUpdateStack(updateStack);
 
-	for (int n=0; n<updateStack.nItems(); ++n)
+	for (auto viewer : updateStack)
```
LamarMoore: For the range based loops you would be better off using `auto &` or `auto *` if you are using pointers so it is clear what is happening.

src/neta/connection.cpp
```diff
@@ -152,10 +152,9 @@ int NETAConnectionNode::score(const SpeciesAtom* i, RefList<const SpeciesAtom>&
 
 	// Get directly connected atoms about 'i', excluding any that have already been matched
 	RefDataList<const SpeciesAtom, int> neighbours;
-	const PointerArray<SpeciesBond>& bonds = i->bonds();
-	for (int n=0; n<bonds.nItems(); ++n)
+	for (auto b : i->bonds())
 	{
-		const SpeciesAtom* j = bonds.at(n)->partner(i);
+		const SpeciesAtom* j = b->partner(i);
```
LamarMoore: `const auto *` modernize as you go along. Give a better name than `j` so you can have clarity.

src/neta/connection.cpp
```diff
@@ -191,10 +190,10 @@ int NETAConnectionNode::score(const SpeciesAtom* i, RefList<const SpeciesAtom>&
 			// Now have a match, so break out of the loop
 			break;
 		}
-		if (atomScore == NETANode::NoMatch) for (int n=0; n<allowedAtomTypes_.nItems(); ++n)
+		if (atomScore == NETANode::NoMatch) for (auto t : allowedAtomTypes_)
 		{
 			// Evaluate the neighbour against the atom type
-			int typeScore = allowedAtomTypes_.at(n)->neta().score(j);
+			int typeScore = t->neta().score(j);
```
LamarMoore: auto

src/gui/viewer_render.cpp
rprospero: I've clarified where in the latest update the auto is working with pointers or references.

src/neta/connection.cpp
rprospero: The latest update improves the names and attempts to modernize the code as it's changed.

trisyoungs:
PointerArray can be completely removed rather than aliased to std::vector - pointless to retain the naming.

Something is badly broken, however, as all the system tests involving molecules with bonds now fail.

rprospero:
I've isolated the commit where the test failure was introduced.  Just
tracking down the exact source of the bond issue.

DanNixon:
As it is I think this is a good improvement.

As for what to do with the raw pointers, it is probably worth looking at moving interfaces to take references rather than pointers. Smart pointers would also work but I think you can avoid needing them based on what was described in the last meeting (all objects are created before any actual processing happens and are not before until exit).

LamarMoore:
@rprospero is there an issue for addressing the use of pointer?

rprospero:
I've created issue #244 to catalog the move to smarter pointers and #245 to track the move to STL containers.

Issue #245, opened 2020-03-10
Move to standard STL Containers


rprospero:
Currently, the code base uses several custom container classes.  It may be possible to replace these with standard STL classes.  This will make it easier to on board new coders, lessen the code that needs to be maintained, and open up the possibility to use the STL library for further simplifications.

This will inherently be tied in with issue #244, since the current containers are also responsible for managing the memory at this time.

- [x] Array - could be std::vector
- [x] Array2d  - possibly an std::vector wrapper?
- [x] DynamicArray - could be std::deque
- [x] List - could be std::list
- [x] ParentList - Subclass of list that doesn't appear to be used anywhere in the code.
- [x] OrderedPointerArray - possibly an std::set, to keep the values ordered - #246
- [x] OrderedDAOArray - use whatever is chosen for OrderedPointerArray - #246
- [x] OrderedList - probably use std::set
- [x] OrdredPointerDataArray - Possible a std::vector of tuples.  Only used twice in the whole code base.
- [x] OrderedPointerList - use whatever is chosen for OrderedPointerArray
- [x] Pair - An std::pair of *optional* values.  This might require a little thought if we don't move to c++17.
- [x] PointerArray - probably an std::vector #240 
- [ ] RefList - Possibly a good choice to be replaced by std::list
- [x] RefDataList - Possibly a good choice to be replaced by std::list of tuples.

rprospero:
Github jumped the gun on #948

PR #246, opened 2020-03-10
Use STL containers in the Cell class

rprospero:
The overarching goal of this PR is to move to using `STL` containers in the Cell class, instead of the custom containers and raw array pointers.  A crucial side effect of this is that the `OrderedPointerArray` and `OrderedDAOArray` classes are no longer used and can be removed, checking off two boxes on issue #245.

`OrderedPointerArray` was replaced with `std::set`, since that gives us the ordering automatically.  The first ugly catch with `std::set` was that it didn't support indexed access, so every for loop had to be converted to a range or iterator based loop.  This is a bit trickier for some of the loops which take larger strides through the data, especially since the set iterator doesn't support comparison, so it's not possible to test if we've moved past the end of the container.  This shows up in still using a loop counter and manually advancing the iterator with `std::advance`.


rprospero:
This PR also paves the way for a large PR that starts starts adding smart pointers to the code base.

src/classes/energykernel.cpp
```diff
@@ -127,17 +129,16 @@ double EnergyKernel::energy(Cell* centralCell, Cell* otherCell, bool applyMim, b
 	// Loop over central cell atoms
 	if (applyMim)
 	{
-		for (i = start; i < centralAtoms.nItems(); i += stride)
+		for (i = start; i < centralAtoms.size(); i += stride)
```
DanNixon: Could these iterator variables have better names?

DanNixon: You could rewrite this using iterators rather than indices (e.g. https://godbolt.org/z/4dQvoN), that should avoid the need for `std::advance` below.

src/classes/energykernel.cpp
```diff
@@ -315,14 +315,16 @@ double EnergyKernel::energy(const Atom* i, Cell* cell, int flags, ProcessPool::D
 	// Get start/stride for specified loop context
 	int start = processPool_.interleavedLoopStart(strategy);
 	int stride = processPool_.interleavedLoopStride(strategy);
+	std::advance(other, start);
 
 	if (flags&KernelFlags::ApplyMinimumImageFlag)
 	{
 		// Loop over other Atoms
-		if (flags&KernelFlags::ExcludeSelfFlag) for (j=start; j<nOtherAtoms; j += stride)
+	  if (flags&KernelFlags::ExcludeSelfFlag) for (j=start; j<nOtherAtoms; j += stride)
```
DanNixon: Unrelated to the changes here, but you may consider `std::bitset` as the more modern way of storing single bit flags.

src/classes/energykernel.cpp
rprospero: I looked at using iterators.  Unfortunately, there is no guarantee that the stride divides evenly into the number of items, which leads to undefined behaviour (e.g. adding one more element to the vector in your example https://godbolt.org/z/M46KNu).  Now, this can be solved by using `<= a.end()` instead of `!= a.end()` in the example.  Unfortunately, `set` iterators aren't comparable, so `!=` is all we get.

src/classes/energykernel.cpp
rprospero: I'll add a ticket for it.

src/classes/energykernel.cpp
DanNixon: The other option would be to use a `std::vector` and perform a `std::sort` before the items are used. This has the same runtime complexity as what you are currently doing but I'm not sure how well it fits into how `Cell` operates in Dissolve.

I'll let you decide on which you end up using, but if `std::set` is kept can the variable names be improved so it is more obvious what is happening.

trisyoungs: 'indexI' and 'indexJ'?

src/classes/cell.h
```diff
@@ -22,17 +22,21 @@
...
 #include "templates/vector3.h"
-#include "templates/reflist.h"
-#include "templates/ordereddaoarray.h"
-#include "templates/orderedpointerarray.h"
 
 // Forward Declarations
 class Box;
 class CellNeighbour;
 
+template <class T> struct IndexComparator {
```
trisyoungs: Can we add a short comment here, rather than having the code 'just appear'?

src/classes/energykernel.cpp
trisyoungs: Also, 'auto centralAtom' rather than 'auto central'

src/classes/cell.h
rprospero: Following Dan's comments above, I'll be switching to sorted vectors over sets, so my next push should remove the need for this struct entirely.

src/classes/energykernel.cpp
trisyoungs: When talking about changing variable names there is also the context of "the science" to consider. Most all computational chemistry types would expect energetic functions to be between atoms referred to as 'i' and 'j', so I would not want to lose that meaning within the functions (if indeed it is those variable names that are the subject of this discussion!).

trisyoungs: In regard to the use of std::vector and std::sort, this is potentially inefficient unless one keeps track of how / when the Cell contents are modified, and re-sorts accordingly. Obviously possible, but perhaps undesirable to have to implement.

src/classes/cell.h
trisyoungs: Made a comment in my review just now about this - need to be careful not to unnecessarily sort the vector, because those functions are heavily used.

src/classes/energykernel.cpp
DanNixon: > but perhaps undesirable to have to implement.

Boost provides a ready to go alternative: https://www.boost.org/doc/libs/1_72_0/doc/html/boost/container/flat_set.html

rprospero: The OrderedPointerArray performed needed O(N) for each insertion, requiring O(N²) to create the whole array.  If we create the array, then sort it, that'll only be O(log(N)), which should be faster than what we had before.

As for std::set, it has a tree underneath it that doesn't have the best cache locality.  Some rudimentary bench-marking has shown that this is important and that sticking with vector may be the better option.

Obviously, I'll also check the vector implementation to ensure that we don't have any regressions in speed.

trisyoungs: No arguments regarding the simplicity of the OrderedPointerArray class, just saying that atoms within Cells are a dynamic quantity during the running of the code (i.e. are subject to change as the simulation progresses) and so any sorting must be continually performed.

DanNixon:
@rprospero Is your intention to eventually replace `ordered_vector` with something like `boost::flat_set`?

trisyoungs:
@DanNixon Much as I know how good Boost is, it's a massive dependency to add for those users (who will exist) who need to compile it on their own machines / HPC clusters. I would rather avoid if possible.

DanNixon:
@trisyoungs This is something that still needs to be discussed, but you shouldn't be forced to aim for zero dependencies. There are tools (e.g. Conan) that make this trivial now.

rprospero:
I've moved from std::set to to a sorted wrapper on std::vector.  Each element is sorted on insertion.  This wrapper is in [ordered_vector](https://github.com/rprospero/dissolve/blob/stl_cell/src/templates/orderedvector.h).  This class should provide a base to solve most instances of OrderedPointerArray.  I've gone through some benchmarking again and, while there's nothing impressive happening versus the develop branch, the slowest runs of this PR have been strictly faster than the fastest runs of the main branch.

rprospero:
@trisyoungs I've changed the name

PR #250, opened 2020-03-18
Pipelines

trisyoungs:
Add single pipeline for develop branch build / deploy.

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -181,12 +181,11 @@ jobs:
       with:
         name: artifacts-osx
         path: damage/
-  updatecontinuous:
+  publish:
     runs-on: ubuntu-latest
     needs: build
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    if: contains(github.ref, 'develop') && github.event_name == 'push'
     steps:
     - uses: actions/checkout@v1
     - name: Download Artifacts (Linux)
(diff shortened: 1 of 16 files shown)


DanNixon:
This isn't quite what I had in mind.
There is a lot of duplication between pipelines and no clear separation between different steps for each platform.

If GH Actions doesn't let you reuse pipeline fragments then I think we'd be better off moving to Azure Pipelines.

trisyoungs:
Only a first step! The duplication between workflows, and the incoherence between OS-specific steps, can largely be removed by splitting into local actions, each dealing with a specific OS.

.github/workflows/develop.yml
```diff
@@ -0,0 +1,114 @@
...
+        path: installers/
+    - name: Upload Artifacts (OSX)
+      if: matrix.os == 'macos-latest'
+      uses: actions/upload-artifact@v1
+      with:
+        name: artifacts-osx
+        path: damagefiles/
+  publish:
+    runs-on: ubuntu-latest
+    needs: build
```
DanNixon: You likely don't need this, if a previous job fails the entire pipeline does.

.github/workflows/develop.yml
```diff
@@ -0,0 +1,114 @@
...
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-16.04, windows-latest, macos-latest]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Build (Linux)
+      if: matrix.os == 'ubuntu-16.04'
```
DanNixon: Things like this do not help readability at all. See what we do in Scipp.

.github/workflows/osx-serial-gui.yml
```diff
@@ -3,10 +3,10 @@ name: OSX (GUI/Serial)
 on:
   push:
     branches:
-    - develop
-    - master
-    paths:
-    - 'src/**'
+    - NOdevelop
```
DanNixon: If this file is not used simply remove it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,166 @@
+name: Release Deployment
```
DanNixon: How is this different from `develop.yml`?

ci/osx/build_serial_gui
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# Script to build Dissolve GUI (and serial) versions on OSX systems.
+# Intended to be called as a job within GitHub Actions.
+
+#
+# Configuration
+#
+# -- Enforce verbosity for commands and exit on first failure
+set -ev
```
DanNixon: Also add `x` here to make build logs a bit more useful.

ci/ubuntu/build_16.04_serial_gui
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
```
DanNixon: What is specific about Ubuntu 16.04 here? If nothing can the file be renamed.

Issue #244, opened 2020-03-10
Move away from Raw Pointers


rprospero:
Currently, the code base uses a large number of raw pointers, which brings a host of problems

* Memory safety cannot be guaranteed by the compiler.  We might attempt to access freed pointers or leak memory.  Diligent programmers can avoid this, but it's better to have this work performed by the compiler
* It slows down onboarding new team members.  Any given pointer could either be an owning pointer or merely an observing pointer and the difference is not immediately clear.

We would prefer to move onto either smart pointers or references, if possible.

This will be a fairly sweeping change through the code base, so care must be taken.  It will also likely be heavily intertwined with the move to standard containers, since many of these pointers are currently managed by the custom containers.

rprospero:
One place that becomes a problem with moving onto purely references is that there are many places where the classes would introduce reference cycles.  For example, the Atom class currently has a [pointer](https://github.com/projectdissolve/dissolve/blob/459f6dcba8ab092ebb0bf32257afae5429ec6cc1/src/classes/molecule.h#L61) to the molecule that contains it, while the Molecule class has an [array of pointers](https://github.com/projectdissolve/dissolve/blob/459f6dcba8ab092ebb0bf32257afae5429ec6cc1/src/classes/molecule.h#L61) to each atom it contains.

This becomes an issue because, unlike pointers, references must be set in the constructor and cannot be changed.  So the Molecule must be passed into the constructor for the Atom, but that Atom's constructor required the reference to the Molecule.  It might be possible to tie the knot by creating one child object inside the constructor of the parent, but it will require careful though.  This is compounded by the fact that the Configuration class maintains a master array of all of the atoms and molecules, which must also be populated with references.

It's possible that using std::reference_wrapper might enable us to avoid some of these issues, but it also may be an indication that smart pointers might be the more intuitive solution for some of these issues.

trisyoungs:
A further potential complication is that the workflow discussed at the previous TSC assumed adherence to a basic [Create Data] -> [Run Simulation on Data] -> [Analyse Data] but there are other use cases where the simulation could actively modify the data (creating / deleting molecules would be a prime example).

Adam is probably correct that at least some smart pointers will have to be used.

LamarMoore:
Instead of explicitly deleting memory you can overwrite it and (for example) resize a vector to either enlarge or shrink. In the short term maybe using smart pointers is the best strategy? Would be interested to see that PR as it may allow me to get more insight into what is happening with the memory.

[...]

PR #254, opened 2020-03-24
Use shared_ptr for Molecule class

rprospero:
The shortest explanation for this PR is that it converts every Molecule pointer into a std::shared_ptr<Molecule>.  The reason for the class based conversion is that the memory for all of the molecules is owned by the Configuration class.  All other pointers are merely observing the Molecules maintained by Configuration.  This PR switches from the custom DynamicArray container to a standard std::deque container for the molecules and uses shared pointers instead of raw ones.  This change then propagates through the rest of the code base, removing all raw Molecule pointers.

Every change fits into one of three categories.
- Changing the type of a variable
- Converting to a range or iterator based loop
- Converting a DynamicArray method call to the corresponding deque method.  This occasionally requires using an algorithm.

This is part of issue #244


DanNixon:
@rprospero Is it possible to use some of the benchmarking tools @trisyoungs mentioned this morning to measure any performance difference?

src/modules/benchmark/process.cpp
```diff
@@ -182,11 +182,11 @@ bool BenchmarkModule::process(Dissolve& dissolve, ProcessPool& procPool)
 					}
 
 					// Loop over target Molecules
-					for (int n = 0; n<targetMolecules.nItems(); ++n)
+					for (int n = 0; n<targetMolecules.size(); ++n)
 					{
 						// Get Molecule index and pointer
 						auto molId = targetMolecules[n];
-						Molecule* mol = cfg->molecule(molId);
+						auto mol = cfg->molecule(molId);
```
DanNixon: Are we certain this is not being optimized out if `mol` is not used?

trisyoungs: Not at all! The question is how to do "useful work" with the object without biasing the benchmarking of the distributor. Ultimately I suppose it doesn't matter if this boils down to an empty loop, because the real work that we're interested in is done elsewhere.

DanNixon: You don't actually want to do any additional work (this would skew the benchmarks), rather just prevent the compiler form optimizing away the lookup.
See https://github.com/google/benchmark#preventing-optimization and https://github.com/google/benchmark/blob/master/include/benchmark/benchmark.h#L307-L310

On a somewhat related note, I'd rather we use Google benchmark for things like this rather than reinventing the wheel.

trisyoungs: Happy for Google Benchmarks to be used - the stuff I'd written was basic and high level to simply test whether sweeping changes to algorithms etc. made significant differences on real-world input examples, rather than drill down into the performance of specific functions.

Issue #257, opened 2020-03-31
Remove use of ListItem base class


rprospero:
The ListItem class is needed for every class that is stored in
the custom List container.  Removing references to ListItem
allows us to progressively remove our dependence on the List
class.

- [x] ~src/base/charstring.h~
- [x] src/base/geometry.h
- [x] src/base/plainvalue.h
- [x] src/base/processgroup.h
- [x] src/classes/atomtype.h
- [x] src/classes/atomtypedata.h
- [x] src/classes/braggreflection.h
- [x] src/classes/changedata.h
- [x] src/classes/configuration.h
- [x] src/classes/coordinateset.h
- [x] src/classes/isotopedata.h
- [x] src/classes/isotopologue.h
- [x] src/classes/isotopologues.h
- [x] src/classes/isotopologueset.h
- [x] src/classes/isotopologueweight.h
- [x] src/classes/kvector.h
- [x] src/classes/masterintra.h
- [x] src/classes/pairpotential.h
- [x] src/classes/partialset.h
- [x] src/classes/sitereference.h
- [x] src/classes/sitestack.h
- [x] src/classes/species.h
- [x] src/classes/speciesatom.h
- [x] src/classes/speciesinfo.h
- [x] src/classes/speciesring.h
- [x] src/classes/speciessite.h
- [x] src/data/elementcolours.h
- [x] src/data/ff.h
- [x] src/data/ffangleterm.h
- [x] src/data/ffatomtype.h
- [x] src/data/ffbondterm.h
- [x] src/data/ffimproperterm.h
- [x] src/data/ffparameters.h
- [x] src/data/fftorsionterm.h
- [x] src/data/isotopes.h
- [x] src/data/sg/generator.h
- [x] src/expression/node.h
- [x] src/genericitems/item.h
- [ ] src/gui/charts/chartblock.h
- [ ] src/gui/charts/charthotspot.h
- [x] src/gui/configurationtab.h
- [x] src/gui/gizmo.h
- [x] src/gui/layertab.h
- [x] src/gui/moduletab.h
- [x] src/gui/referencepoint.h
- [x] src/gui/render/primitive.h
- [x] src/gui/render/primitiveassembly.h
- [x] src/gui/render/primitiveinstance.h
- [x] src/gui/render/renderable.h
- [x] src/gui/render/renderablegroup.h
- [x] src/gui/render/textformat.h
- [x] src/gui/render/textfragment.h
- [x] src/gui/render/textprimitive.h
[...]

PR #258, opened 2020-04-01
Remove ListItem from ElementColour class

rprospero:
This removes ListItem as an ancestor of the ElementColour class.

A couple of changes that were **not** made were

* replacing the `colour_` raw array with an std::array.  This is because the value's primary use is as a GLFloat pointer expected by the OpenGL API and it would be better to avoid an impedence mismatch.
* Replacing the `ElementColour*` values with shared pointers or references.  The current architecture relies on global variables updated within the ElementColour constructor and modifying the setup would significantly expand the scope of the PR.  I will create a new issue to look at clearing up this part of the code.

### DIFF
--- a/src/data/elementcolours.cpp
+++ b/src/data/elementcolours.cpp
@@ -23,14 +23,14 @@
 #include "base/messenger.h"
 
 // Static Singletons
-Array< List<ElementColour> > ElementColours::coloursByElementPrivate_;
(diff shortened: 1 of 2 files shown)


rprospero:
This checks another box off issue #257

rprospero:
In light of these comments, I feel that this PR has been superseded by #260.

PR #260, opened 2020-04-02
Simplify element colours

rprospero:
As noted by Tristan, the ElementColours class was based on the Isotopes class and contained a large amount of infrastructure that wasn't necessary for this class.  This commit PR greatly simplifies the class structure, in addition to removing the dependency on the custom list class, as per issue #257.

### DIFF
--- a/src/data/elementcolours.cpp
+++ b/src/data/elementcolours.cpp
@@ -20,206 +20,137 @@
 */
 
 #include "data/elementcolours.h"
-#include "base/messenger.h"
 
-// Static Singletons
-std::vector< std::vector<ElementColour*>> ElementColours::coloursByElementPrivate_;
-
-/*
- * Isotopic Neutron Scattering Data
- */
-
-// Constructor
-ElementColour::ElementColour(int z, float r, float g, float b, float a) : ElementReference(z)
-{
-	// Set the colour data
-	colour_[0] = r;
-	colour_[1] = g;
-	colour_[2] = b;
-	colour_[3] = a;
-
-	// Add this isotope to its parent element's list
-	ElementColours::registerElementColour(this, z);
-}
-
-// Assignment Operator
-ElementColour& ElementColour::operator=(const ElementColour& source)
-{
-	colour_[0] = source.colour_[0];
-	colour_[1] = source.colour_[1];
-	colour_[2] = source.colour_[2];
-	colour_[3] = source.colour_[3];
-
-	return (*this);
-}
-
-// Return colour
-const float* ElementColour::colour() const
-{
-	return colour_;
-}
-
-/*
- * Element Colours
- */
-
-// Return colour data, grouped by element
-std::vector<ElementColour*>& ElementColours::coloursByElement(int Z)
-{
-	// Has the master array been initialised yet? If not, do it now, before the Sears data is constructed
-	if (coloursByElementPrivate_.empty())
-	{
-		coloursByElementPrivate_.resize(Elements::nElements());
-	}
-
-	/* Copied from Aten */
-
-	static ElementColour colourData[] = {
-	//	Z	El	R	G	B	A
-		{ ELEMENT_XX,	0.5,	0.5,	0.5,	1.0 },
-		{ ELEMENT_H,	0.87,	0.87,	0.87,	1.0 },
-		{ ELEMENT_HE,	1.0,	0.784,	0.784,	1.0 },
-		{ ELEMENT_LI,	0.647,	0.165,	0.165,	1.0 },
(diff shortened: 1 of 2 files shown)


src/data/elementcolours.cpp
```diff
@@ -20,212 +20,137 @@
...
 }
 
 // Return colour for specified Element
 const float* ElementColours::colour(Element* el)
 {
-	return coloursByElement(el->Z()).first()->colour();
+	return colours_[el->Z()].data();
 }
+
+const std::vector<std::array<float, 4>> ElementColours::colours_ = {
```
LamarMoore, line 35: could you just use a `std::tuple<float, float, float, float>` instead of `std::array<float, 4>`

src/data/elementcolours.h
```diff
@@ -22,48 +22,20 @@
...
-	// Return colour data for specified Element
-	static List<ElementColour>& coloursByElement(int Z);
+	static const std::vector<std::array<float, 4>> colours_;
 
 	public:
-	// Register specified ElementColour to given Element
-	static void registerElementColour(ElementColour* ec, int Z);
 	// Return colour for specified Z
 	static const float* colour(int Z);
+
```
LamarMoore, line 38: Why not just overload the `operator[]` in the class instead of using a function. Also why return a pointer instead of a const ref? You could just do something like:
```cpp
const tuple<float, float, float, float> &operator[](size_t index) const {
  return colours_[index];
}
```
If you wanted to be more expressive you could have a type called `colour`.

```cpp
struct Colour
{
float red;
float blue;
float green;
float alpha;
};

class ElementColours  
{
private:
  std::vector<Colour> colours; //you can initialise in a very similar way to the way you are now
public:
  const Colour &operator[](size_t index) const { return colours[index]; }
  const Colour &colourOfElement(const Element &elem) const {...}
}
```

src/data/elementcolours.cpp
trisyoungs: Unless the data contained in the tuple can be returned explicitly as a float[4], then the OpenGL routines will choke on it. The return of GLfloat* is for compatibility with OpenGL calls which demand a pointer to an array[4] of GLfloat (== float).

trisyoungs:
I would argue to keep this as a (simple) class for the following reasons:

1. Potential extension to use different colouring schemes in future via derivation from a suitable base class.
2. Providing editability of the data in-class to keep things neat (users can be very passionate about their element colours!).

This could be construed as needless future-proofing, I grant you, but those possibilities may well arise.

LamarMoore:
@trisyoungs , couldn't you store the element colouring as some external configuration file or csv? The you can edit and use different maps without having to keep making classes. Any thoughts @DanNixon ?

rprospero:
As Tristan mentioned, I did use arrays because the data is always immediately passed to OpenGL, who needs that specific data format.

I do agree that we might be better off with just a namespace, though, even with the future proofing.  Since the values are all global, having a class doesn't really get us much.  Even if we later want to add loading or editing colours, that could could simply be a new function in the namespace.

My latest commit removes the class entirely.

DanNixon:
FWIW I think [Lamar's suggestion](https://github.com/projectdissolve/dissolve/pull/260#discussion_r402283238) is a good option. I'd advise against using arrays if you are using this purely for OpenGL (you can cast that sturct to `float *`) (although the use of OpenGL is a potential issue that we'd have to discuss at some point).

trisyoungs:
OK, for now let's proceed with the namespace route.  @LamarMoore I've always tried to avoid loading in external data files for the most part - having a single binary to install / distribute is a lot cleaner IMO.  @DanNixon Are you suggesting moving to Vulkan at some point, or some other GL?

DanNixon:
> Are you suggesting moving to Vulkan at some point, or some other GL?

No, a platform/GPU agnostic library (i.e. one that presents a single API and consumes whatever API the system presents (be it D3D, OpenGL, GLES, Metal, Vulkan, etc...).
I'd give examples of such libraries but I'm still evaluating some of them because we will eventually have to do the same in Mantid.

This isn't a high priority, Mac support might break in the next couple of releases but other platforms *should* keep running with OpenGL a bit longer.

A secondary advantage is getting to use a modern graphics API rather than a legacy mess that is OpenGL, which is also a quick win for graphics performance.

trisyoungs:
OK, be interesting to hear what you settle on as the best solution.

For the present PR, if nobody has anything further to add I'll close and merge shortly.

rprospero:
As per @DanNixon's comment, I have replaced the vector with a constexpr static array.  As per @LamarMoore's much earlier comment, I've created a structure to label the individual colour components, instead of just using std:array.

PR #261, opened 2020-04-03
Remve ListItem from ForcefieldBondTerms

rprospero:
The ultimate goal of this PR was to remove the ListItem dependency from `ForcefieldBondTerm`, as part of issue #257.  As part of that process, this replaces the List of `ForcefieldBondTerm` with a std::vector.

Since the class was largely contained in this instance, it was possible to use a direct class and references, instead of needing a vector of `shared_ptr`.  The only price for this is that it's not possible to return NULL if bond term being searched for isn't present.  The appropriate solution hear would be to return an optional reference, but that will have to wait until either we add Boost or we upgrade to c++17.  In the meantime, I've put in a local `optional` to make the code clearer and to make it easier to find the locations that need updating when the dependencies change.

There's of Terms classes that will be updated in a later PR after this one, but I thought it was worth hammering out just one class at first and then use the final strategy on all of the other classes.

### DIFF
--- a/src/data/ff.cpp
+++ b/src/data/ff.cpp
@@ -19,8 +19,6 @@
 	along with Dissolve.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#include <algorithm>
-#include <functional>
 #include "data/ff.h"
 #include "data/ffangleterm.h"
 #include "data/ffatomtype.h"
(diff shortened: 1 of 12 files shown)


src/data/ff.cpp
```diff
@@ -340,11 +342,13 @@ bool Forcefield::assignIntramolecular(Species* sp, int flags) const
...
-
-		bond->setForm(term->form());
-		bond->setParameters(term->parameters());
+		auto opt_term = bondTerm(typeI, typeJ);
+		if (std::get<1>(opt_term)) {
+			return Messenger::error("Failed to locate parameters for bond %i-%i (%s-%s).\n", i->userIndex(), j->userIndex(), typeI->equivalentName(), typeJ->equivalentName());
+		}
+		const ForcefieldBondTerm& term = std::get<0>(opt_term);
+		bond->setForm(term.form());
+		bond->setParameters(term.parameters());
```
trisyoungs, line 351: To me this seems a little convoluted. As this is the only place in the code where the search is made, an alternative approach could be to modify the Forcefield::bondTerm() function to become applyBondTerm(), accepting a reference/pointer to the SpeciesBond as an additional argument.  The function can then simply return a bool indicating whether the term was found or not.

DanNixon: > As this is the only place in the code where the search is made, an alternative approach could be to modify the Forcefield::bondTerm() function to become applyBondTerm(),

Mixing responsibility like this between classes makes code difficult to read and understand.

trisyoungs: Fair comment - just thinking out loud.

trisyoungs:
Probably no need to benchmark as the search and application of forcefield terms such as these is never performed in a time-critical section of the code. Typically it is only done interactively through the GUI when setting up an input file.

src/data/ff.cpp
rprospero: Also, much of the convolution is simply due to the poor quality `optional` substitute.  Using a proper optional, the only things that change from the original code are the type of the `term` variable and the conditional being `if (!term.has_value())` instead of `if (!term)`.  The remainder of the code would have remained the same.

src/data/ff.cpp
```diff
@@ -175,9 +177,7 @@ ForcefieldAtomType* Forcefield::atomTypeById(int id, Element* element) const
 // Add bond term
 void Forcefield::addBondTerm(const char* typeI, const char* typeJ, SpeciesBond::BondFunction form, double data0, double data1, double data2, double data3)
```
LamarMoore, line 178: why `const char *` and not `const std::string &`?

src/data/ff.cpp
```diff
@@ -205,12 +205,14 @@ void Forcefield::addImproperTerm(const char* typeI, const char* typeJ, const cha
 }
 
 // Return bond term for the supplied atom type pair (if it exists)
-ForcefieldBondTerm* Forcefield::bondTerm(const ForcefieldAtomType* i, const ForcefieldAtomType* j) const
+optional<const ForcefieldBondTerm&> Forcefield::bondTerm(const ForcefieldAtomType* i, const ForcefieldAtomType* j) const
```
LamarMoore, line 208: What is this function doing? Is the name appropriate? Should it be `findBondTerm`? If so, you might be able to implement standard behaviour. What happens when `std::find` does not find something? That way you remove the confusion of the optional if an end iterator is returned. We should discuss this @trisyoungs @rprospero @DanNixon

src/data/ff.cpp
```diff
@@ -340,11 +342,13 @@ bool Forcefield::assignIntramolecular(Species* sp, int flags) const
 		ForcefieldAtomType* typeJ = determineTypes ? determineAtomType(j) : atomTypeByName(j->atomType()->name(), j->element());
 		if (!typeJ) return Messenger::error("Couldn't locate object for atom type named '%s'.\n", j->atomType()->name());
 
-		ForcefieldBondTerm* term = bondTerm(typeI, typeJ);
-		if (!term) return Messenger::error("Failed to locate parameters for bond %i-%i (%s-%s).\n", i->userIndex(), j->userIndex(), typeI->equivalentName(), typeJ->equivalentName());
-
-		bond->setForm(term->form());
-		bond->setParameters(term->parameters());
+		auto opt_term = bondTerm(typeI, typeJ);
```
LamarMoore, line 345: Couple quesitons:
- Could this be implemented using a range-based for loop?
- Can all of the checks occur in a free function? So you don't confuse all of these checks with what is actually supposed to be happening. (Readability)

src/data/ffbondterm.h
```diff
@@ -52,7 +52,7 @@ class ForcefieldBondTerm : public ListItem<ForcefieldBondTerm>
 
 	public:
 	// Return if this term matches the atom types supplied
-	bool matches(const ForcefieldAtomType* i, const ForcefieldAtomType* j);
+	bool matches(const ForcefieldAtomType* i, const ForcefieldAtomType* j) const;
```
LamarMoore, line 55: minor: `isMatch(...)` probably a better name?

src/data/ff.cpp
trisyoungs: Having now woken up and thought about this, I would strongly argue for keeping optional<T> as the return value. This seems like a much neater solution than returning an iterator, which will require exposure of end() from lots of places, and is much closer to the intention of this function and many others.

Agree that it should be named findBondTerm(), but don't really see the benefit in making it compliant with other STL search functions at the expense of exposing the end iterators of all the data structures.

DanNixon: I think the API should be considered here, to me this function is actually a getter rather than a search operation, therefore `std::optional<T>` is a sensible return type and a neat way of handling the requesting thing not being present.
For the same reason I also think `find` should not be in the function name.

rprospero: Returning an iterator has a second issue in that the functions are public, but the vectors being iterated over are private, so the calling function might not be able to access the end iterator to perform the comparison.

src/data/ff.cpp
rprospero: My most recent commit uses the variadic templates to clean up this code and make it more universal across the different types.

src/data/ffbondterm.h
rprospero: I have switched over to that name, as well as renaming the various `bondTerm`, `angleTerm` style accessor to read as `getBondTerm`, which makes the meaning clearer.

rprospero:
I've pushed several changes from this morning.  This commit uses the variadic templates to unify the four different term types into a single code base and return type.  I've also renamed the various *Term functions to get*Term, which might make them a bit clearer.  Similarly, matches has been named to isMatch.

The code is still on optionals, since the recent comments seem to be leaning in that direction.  As mentioned above, the returning an iterator brings some problems with it, since the function itself is public, but the end of the iterator is private.  As mentioned on the call, the optional could be replaced with an exception that is caught in the GUI code.  However, the code base currently has no explicit exceptions, so adding exceptions to the code seems like a decision that shouldn't be made lightly, even if this is a natural example.

PR #266, opened 2020-04-16
No List class for ForcefieldAtomType

rprospero:
Remove the ListItem dependency from the ForcefieldAtomType class.

This is mostly just a changing of types, but there are two non-intuitive parts.

1) Storing ForcefieldAtomType in a vector requires that it supports the copy constructor which, eventually, requires that NETANode supports a default constructor.

2) `atomTypesByElementPrivate_` was changed from `Array< RefList<ForcefieldAtomType>` to `std::vector<std::vector<std::reference_wrapper<ForcefieldAtomType>>>` .  The `reference_wrapper` allows us to store references in the vector.  The price for this is that the type inference doesn't see through the `reference_wrapper`, so we need to manually declare the types at some points.

### DIFF
--- a/src/data/ff.cpp
+++ b/src/data/ff.cpp
@@ -139,8 +139,15 @@ ForcefieldAtomType* Forcefield::atomTypeByName(const char* name, Element* elemen
 	for (int Z=startZ; Z<=endZ; ++Z)
 	{
 		// Go through types associated to the Element
-		for(ForcefieldAtomType& type : atomTypesByElementPrivate_[Z]) if (DissolveSys::sameString(type.name(), name)) return &type;
-	}
+		auto it = std::find_if(atomTypesByElementPrivate_[Z].begin(),
+				       atomTypesByElementPrivate_[Z].end(),
+				       [&name](ForcefieldAtomType& type) {
+					 return DissolveSys::sameString(type.name(), name);
+				       });
+                if (it != atomTypesByElementPrivate_[Z].end()) {
+                  return &it->get();
+                }
+        }
 
 	return NULL;
 }
--- a/src/data/ff.cpp
+++ b/src/data/ff.cpp
@@ -93,9 +93,10 @@ ForcefieldAtomType* Forcefield::determineAtomType(SpeciesAtom* i, const std::vec
 	// Go through AtomTypes defined for the target's element, and check NETA scores
 	int bestScore = -1;
 	ForcefieldAtomType* bestType = NULL;
-	for(ForcefieldAtomType& type : atomTypes[i->element()->Z()])
+	for(const auto& typeRef : atomTypes[i->element()->Z()])
 	{
 		// Get the scoring for this type
+		auto& type = typeRef.get();
 		int score = type.neta().score(i);
 		if (score > bestScore)
 		{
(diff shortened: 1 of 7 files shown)


rprospero:
This is part of issue #257

src/neta/node.h
```diff
@@ -51,6 +51,7 @@ class NETANode : public ListItem<NETANode>
 	enum NETAResult { NoMatch = -1 };
 	// Constructor / Destructor
 	NETANode(NETADefinition* parent, NodeType type);
+	NETANode();
```
trisyoungs, line 54: Semantic / style question - instead of providing an explicit default constructor with no args, could just add default values to 'parent' and 'type' args?

LamarMoore: You can do that and delete the default constructor if you are sure that is what you want to do e.g.
 ```cpp
NETANode(NETADefiniton * parent=nullptr, NoteType type = NoteType::BasicNode);
NETANode()=delete;
```
or you can implement a default constructor:
```cpp
NETANode()=default;
```

src/data/ff.cpp
```diff
@@ -162,7 +152,7 @@ ForcefieldAtomType* Forcefield::atomTypeById(int id, Element* element) const
 	for (int Z=startZ; Z<=endZ; ++Z)
 	{
 		// Go through types associated to the Element
-		for(auto type : atomTypesByElementPrivate_.constAt(Z)) if (type->index() == id) return type;
+		for(ForcefieldAtomType& type : atomTypesByElementPrivate_[Z]) if (type.index() == id) return &type;
```
LamarMoore, line 155: Is this not an `std::find_if`

src/data/ff/uff.cpp
```diff
@@ -412,7 +412,8 @@ ForcefieldAtomType* Forcefield_UFF::determineAtomType(SpeciesAtom* i) const
 			break;
 		// Default - all elements with only one type
 		default:
-			return atomTypesByElementPrivate_.constAt(i->element()->Z()).firstItem();
+			ForcefieldAtomType& temp = atomTypesByElementPrivate_[i->element()->Z()].front();
```
LamarMoore, line 415: `const auto &` see suggestion about for getting the type from the reference

src/neta/node.h
rprospero: I went with letting the compiler generate the default constructor, instead of using adding default arguments to the existing constructor.  My suspicion is that, once the `List` class is fully removed, we can eliminate this default constructor and that will be a bit easier if it's just its own line.

src/data/ff.cpp
```diff
@@ -37,8 +37,6 @@
 // Constructor / Destructor
 Forcefield::Forcefield() : ListItem<Forcefield>()
 {
-	// Initialise the Element RefList Array
-	Elements::createElementRefListArray<ForcefieldAtomType>(atomTypesByElementPrivate_);
 }
 
 Forcefield::~Forcefield()
```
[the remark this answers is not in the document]
rprospero: I haven't include these changes in this PR, since I already have a PR waiting that removes the List class from the Forcefield class itself.  That PR includes changing Forcefield over to the default constructor and destructor.

src/data/ff.cpp
```diff
@@ -148,7 +138,7 @@ ForcefieldAtomType* Forcefield::atomTypeByName(const char* name, Element* elemen
 	for (int Z=startZ; Z<=endZ; ++Z)
```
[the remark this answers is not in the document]
rprospero: I've switched the the type inference version.  I'd originally gone with the explicit type declaration because it was a line shorter, but the inferred version does seem to be more robust.

src/data/ff.cpp
rprospero: Good point.  I've made the change in the latest commit.

PR #268, opened 2020-04-21
No list ffparameters

rprospero:
This removes the last references to the List class in the Forcefields classes.

This is mostly just converting types, changing names, and upgrading to algorithms.  The most surprising change is likely to be the `Q_DECLARE_SMART_POINTER_METATYPE` and `Q_DECLARE_METATYPE` in selectforcefieldwidget.h.  These were necessary to allow shared pointers to Forcefield to be stored in a QVariant.

### DIFF
--- a/src/classes/species.h
+++ b/src/classes/species.h
@@ -22,8 +22,6 @@
 #ifndef DISSOLVE_SPECIES_H
 #define DISSOLVE_SPECIES_H
 
-#include "base/charstring.h"
-#include "base/version.h"
 #include "classes/atomtypelist.h"
 #include "classes/coordinateset.h"
 #include "classes/isotopologue.h"
@@ -36,6 +34,7 @@
 #include "io/import/coordinates.h"
 #include "templates/dynamicarray.h"
 #include "templates/objectstore.h"
+#include <memory>
 
 // Forward Declarations
 class Box;
@@ -247,15 +246,15 @@ class Species : public ListItem<Species>, public ObjectStore<Species>
 	 */
       private:
 	// Forcefield to source terms from
-	Forcefield *forcefield_;
+	std::shared_ptr<Forcefield> forcefield_;
 	// Auto-generate missing intramolecular terms, and remove invalid ones
 	bool autoUpdateIntramolecularTerms_;
 
       public:
 	// Set Forcefield to source terms from
-	void setForcefield(Forcefield *ff);
+	void setForcefield(std::shared_ptr<Forcefield> ff);
 	// Return Forcefield to source terms from
-	Forcefield *forcefield() const;
+	std::shared_ptr<Forcefield> forcefield() const;
 	// Set whether to auto-generate missing intramolecular terms, and remove invalid ones
 	void setAutoUpdateIntramolecularTerms(bool b);
 	// Return whether to auto-generate missing intramolecular terms, and remove invalid ones
--- a/src/data/fflibrary.cpp
+++ b/src/data/fflibrary.cpp
@@ -80,6 +80,12 @@ std::shared_ptr<Forcefield> ForcefieldLibrary::forcefield(const char *name)
 		if (DissolveSys::sameString(ff->name(), name))
 			return ff;
 	}
+	auto it = std::find_if(forcefields().begin(),
+			       forcefields().end(),
+			       [&name](const std::shared_ptr<Forcefield> ff) {
+				 return DissolveSys::sameString(ff->name(), name);
+			       });
+	if (it == forcefields().end()) return nullptr;
 
-	return nullptr;
+	return *it;
 }
--- a/src/data/fflibrary.cpp
+++ b/src/data/fflibrary.cpp
@@ -73,19 +73,16 @@ std::vector<std::shared_ptr<Forcefield>> &ForcefieldLibrary::forcefields()
 }
 
 // Return named Forcefield, if it exists
-std::shared_ptr<Forcefield> ForcefieldLibrary::forcefield(const char *name)
+std::shared_ptr<Forcefield> ForcefieldLibrary::forcefield(const string name)
 {
 	for (auto &ff : forcefields())
 	{
-		if (DissolveSys::sameString(ff->name(), name))
+		if (DissolveSys::sameString(ff->name(), name.c_str()))
 			return ff;
 	}
-	auto it = std::find_if(forcefields().begin(),
-			       forcefields().end(),
-			       [&name](const std::shared_ptr<Forcefield> ff) {
-				 return DissolveSys::sameString(ff->name(), name);
-			       });
-	if (it == forcefields().end()) return nullptr;
+	auto it = std::find_if(forcefields().begin(), forcefields().end(), [&name](const std::shared_ptr<Forcefield> ff) { return DissolveSys::sameString(ff->name(), name.c_str()); });
+	if (it == forcefields().end())
+		return nullptr;
 
 	return *it;
 }
--- a/src/data/fflibrary.cpp
+++ b/src/data/fflibrary.cpp
@@ -35,7 +35,7 @@
 #include "data/ff/uff.h"
 
 // Static Members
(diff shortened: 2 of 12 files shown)


rprospero:
Note that this builds on pr #266, and shouldn't be merged until that one has.  This PR is also part of Issue #257.

src/classes/species.h
```diff
@@ -251,15 +252,15 @@ class Species : public ListItem<Species>, public ObjectStore<Species>
 	 */
 	private:
 	// Forcefield to source terms from
-	Forcefield* forcefield_;
+	std::shared_ptr<Forcefield> forcefield_;
```
LamarMoore, line 255: could this be `std::unique_ptr<Forcefield>`? instead? You can reset it or move memory into it if you want to update it. For example:
 ```cpp
std::unique_ptr<Forcefield> forcefield_;

void setForcefield(std::unique_ptr<Forcefield> &&ff) {
  forcefield_= std::move(ff);
}

// Or you could implement an operator to allow a shorthand elsewhere.
void operator=(std::unique_ptr<Forcefield> &&ff) {
  forcefield_ = std::move(ff);
}
const Forcefield &forcefield() const {
  return *forcefield_;
}
```

File: src/data/fflibrary.cpp [hunk 2]
```diff
@@ -44,38 +44,39 @@ List<Forcefield> ForcefieldLibrary::forcefields_;
...
 	return forcefields_;
 }
 
 // Return named Forcefield, if it exists
-Forcefield* ForcefieldLibrary::forcefield(const char* name)
+std::shared_ptr<Forcefield> ForcefieldLibrary::forcefield(const char* name)
 {
-	ListIterator<Forcefield> ffIterator(forcefields());
-	while (Forcefield* ff = ffIterator.iterate()) if (DissolveSys::sameString(ff->name(), name)) return ff;
+	for (auto& ff : forcefields()) {
```
LamarMoore, line 77: `std::find_if`?

src/data/fflibrary.cpp
```diff
@@ -35,7 +35,7 @@
 #include "base/sysfunc.h"
 
 // Static Members
-List<Forcefield> ForcefieldLibrary::forcefields_;
+std::vector<std::shared_ptr<Forcefield>> ForcefieldLibrary::forcefields_;
```
LamarMoore, line 38: As above, can these be `unique_ptr`? Do these need to be pointers? `RegisterForcefields()` seems to hardcode all the forcefield values. Can this be part of constructing the `ForcefieldLibrary` class? Then you can just return const references.

src/data/fflibrary.cpp
```diff
@@ -44,38 +44,39 @@ List<Forcefield> ForcefieldLibrary::forcefields_;
...
 	// If the list is empty, we haven't yet constructed the list...
-	if (forcefields_.nItems() == 0) registerForcefields();
+	if (forcefields_.empty()) registerForcefields();
 
 	return forcefields_;
 }
 
 // Return named Forcefield, if it exists
-Forcefield* ForcefieldLibrary::forcefield(const char* name)
+std::shared_ptr<Forcefield> ForcefieldLibrary::forcefield(const char* name)
```
LamarMoore, line 75: `std::string` instead of `const char *`?

src/gui/addforcefieldtermswizard_funcs.cpp
```diff
@@ -222,7 +222,7 @@ bool AddForcefieldTermsWizard::progressionAllowed(int index) const
 	switch (index)
 	{
 		case (AddForcefieldTermsWizard::SelectForcefieldPage):
-			return (ui_.ForcefieldWidget->currentForcefield());
+			return (ui_.ForcefieldWidget->currentForcefield() != nullptr);
```
LamarMoore, line 225: ` return (!ui_.ForcefieldWidget->currentForcefield());` should work.

src/gui/selectforcefielddialog.h
```diff
@@ -54,7 +54,7 @@ class SelectForcefieldDialog : public QDialog
 
 	public:
 	// Run the dialog, returning the selected Forcefield
-	Forcefield* selectForcefield(Forcefield* currentFF);
+	std::shared_ptr<Forcefield> selectForcefield(std::shared_ptr<Forcefield> currentFF);
```
LamarMoore, line 57: You should probably be returning const references instead of pointers unless there is a good reason not to.

rprospero: I have a reason, though I'm not sure that it's good enough.  The UI for the dialog contains a QListWidget for the user to select their Forcefield.  This has two consequences

1) If the user cancels the dialog, the QListWidget returns null, which `selectForcefield` returns in kind.  If we move to references, we'd either need to have some sort of optional to represent this or use exceptions for this fairly common situation.

2) Currently, the QListWidget contains the Forcefield pointers inside `QVariant`.  From what I've read, QVariant cannot hold references.  It might be possible to get around this by using a Bimap between some concrete identifiers and references and having the QListWidget contain the identifiers.

src/gui/addforcefieldtermswizard_funcs.cpp
rprospero: I've changed that back.  Wasn't sure which style would be more clear.

src/data/fflibrary.cpp
rprospero: Made the change to the function signature.  I'm also going to put in an issue about looking at the DissolveSys::sameStirng function, because it's used throughout the code base.  At the very least, it can be upgraded to support std::string as well as char*.  It might also be worth looking at how often it can be replaced with `==`

src/data/fflibrary.cpp
rprospero: Unfortunately, these do have to be pointers.  Each of the Forcefield types is its own subclass, so trying to use concrete values would require a heterogeneous vector.

src/gui/selectforcefielddialog.h
trisyoungs: The UI component could be reworked to just return the name of the selected Forcefield (which we would have to force to be unique) or an empty string. The return of the optional reference could then be based on the result of the search for the named forcefield with Forcefield::find().

DanNixon: > The return of the optional reference could then be based on the result of the search for the named forcefield with Forcefield::find().

Why add a search for no reason?

src/classes/species.h
LamarMoore: Sounds like it should be a `weak_ptr<Forcefiled>`?

src/data/fflibrary.cpp
LamarMoore: @rprospero couldn't you just use polymorphism? You could use a `dynamic_pointer_cast<>` when you need specificity. Obviously here you would use `unique_ptr`.

src/classes/species.h
rprospero: If `Species` has a weak_ptr, doesn't that require that we use `shared_ptr` in the library, instead of `unique_ptr`? I didn't think it was possible to make a `weak_ptr` from a `unique_ptr`.

If we are going with `shared_ptr` inside the vector in the ForcefieldLibrary, then I'm not sure that I see the advantage of using `weak_ptr` in the `Species` class.  It's true that `Species` doesn't need ownership, but the `weak_ptr` is locked into a `shared_ptr` before use, anyway, so it would seem that all we're doing is adding an locking overhead on each call.

src/data/fflibrary.cpp
rprospero: Sorry, I didn't make myself clear.  Using *references* is prohibited, since it would require a heterogeneous vector.  Using polymorphism and some kind of smart pointer does seem like the only way to go here, though I've made my comments above on why I think that this can not be a `unique_ptr`.

src/classes/species.h
LamarMoore: True but it makes ownership clear and also you always know whether or not the pointer goes out of scope at the point of use.

PR #279, opened 2020-05-04
Add XRay - Part 2

trisyoungs:
Present PR will...

- Rename old neutron Weights class to NeutronWeights
- Add XRayWeights class

Issue #283, opened 2020-05-06
Use auto where possible


trisyoungs:
Use auto instead of (often complicated) explicit type declarations.

See clang-tidy.

trisyoungs:
Closed by merging #290

PR #284, opened 2020-05-06
Add XRay - Part 2a

trisyoungs:
Slight digression from (but related to) addition of XRay structure factor calculation, this PR will:

- Replace custom List<T> class with standard containers in Isotopologues, IsotopologueSet, and IsotopologueCollection
- Remove ListItem<T> dependence from IsotopologueWeight, Isotopologues, and IsotopologueSet

Partial work towards #257.

~One point for discussion - to overcome the inability of QVariant to store references I have just taken pointers to the objects as per the original implementation where necessary. Adam had already encountered this, but I don't recall the exact solution (if indeed one had been decided).~

The present changes are an example of where classes can straightforwardly use std::vector to manage objects over my old List class - IsotopologueCollection and its related classes are used as storage for keyword-based data, and are not referenced as pointers. Instead, the data is retrieved when needed. So, strictly we don't need the ability to use make_shared and refer to it with a weak_ptr. The GUI, however, allows the data within IsotopologueCollection to be edited, and so this is where using weak_ptr would be useful to ensure object validity. So, the choices as I see it are:
1. Keep the neatness of a std::vector of objects, and worry about how to refer to those objects by pointer in the GUI classes.
2. Move to a vector of shared_ptr, and use weak_ptr in the GUI classes.

A third option is implemented in 5c00c3c which introduces a manager class to maintain items in a QTreeWIdget as well as associate those items with references to the underlying data.

To be merged after #279.

### DIFF
--- a/src/classes/isotopologuecollection.cpp
+++ b/src/classes/isotopologuecollection.cpp
@@ -38,9 +38,8 @@ IsotopologueCollection::~IsotopologueCollection() {}
 // Remove any sets from the collection that are empty
 void IsotopologueCollection::pruneEmptySets()
 {
-	std::remove_if(isotopologueSets_.begin(), isotopologueSets_.end(), [&](IsotopologueSet &set) {
-		return set.nIsotopologues() == 0;
-	});
+	std::remove_if(isotopologueSets_.begin(), isotopologueSets_.end(),
+		       [&](IsotopologueSet &set) { return set.nIsotopologues() == 0; });
 }
 
 // Clear all existing data
@@ -51,9 +50,8 @@ void IsotopologueCollection::add(Configuration *cfg, Isotopologue *iso, double r
 {
 	printf("TRYING TO ADD ISO %s TO CFG %s...\n", iso->name(), cfg->name());
 	// Check if a set already exists for this Configuration
-	auto it = std::find_if(isotopologueSets_.begin(), isotopologueSets_.end(), [&](IsotopologueSet &set) {
-		return set.configuration() == cfg;
-	});
+	auto it = std::find_if(isotopologueSets_.begin(), isotopologueSets_.end(),
+			       [&](IsotopologueSet &set) { return set.configuration() == cfg; });
 	if (it == isotopologueSets_.end())
 	{
 		isotopologueSets_.emplace_back(this, cfg);
@@ -66,17 +64,14 @@ void IsotopologueCollection::add(Configuration *cfg, Isotopologue *iso, double r
 // Remove the specified set from the collection
 void IsotopologueCollection::remove(IsotopologueSet *set)
 {
-	std::remove_if(isotopologueSets_.begin(), isotopologueSets_.end(), [&](IsotopologueSet &data) {
-		return &data == set;
-	});
+	std::remove_if(isotopologueSets_.begin(), isotopologueSets_.end(), [&](IsotopologueSet &data) { return &data == set; });
 }
 
 // Remove the Configuration from the collection
 void IsotopologueCollection::remove(Configuration *cfg)
 {
-	std::remove_if(isotopologueSets_.begin(), isotopologueSets_.end(), [&](IsotopologueSet &set) {
-		return set.configuration() == cfg;
-	});
+	std::remove_if(isotopologueSets_.begin(), isotopologueSets_.end(),
+		       [&](IsotopologueSet &set) { return set.configuration() == cfg; });
 }
 
 // Remove the Species from the specified set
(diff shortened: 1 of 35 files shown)


src/classes/neutronweights.cpp
```diff
@@ -174,19 +167,19 @@ void Weights::calculateWeightingMatrices()
...
 	// Loop over defined Isotopologues in our defining mixtures, summing terms from (intramolecular) pairs of Atoms
-	boundWeights_ = 0.0;
+	intramolecularWeights_ = 0.0;
 	Array2D<double> intraNorm(atomTypes_.nItems(), atomTypes_.nItems(), true);
 	Array2D<bool> intraFlag(atomTypes_.nItems(), atomTypes_.nItems(), true);
 	Array2D<bool> globalFlag(atomTypes_.nItems(), atomTypes_.nItems(), true);
 	intraNorm = 0.0;
 	globalFlag = false;
-	for (Isotopologues *topes = isotopologueMixtures_.first(); topes != NULL; topes = topes->next())
+	for (auto &topes : isotopologueMixtures_)
```
LamarMoore, line 176: `const auto &`?

src/classes/neutronweights.cpp
```diff
@@ -478,27 +469,29 @@ bool Weights::broadcast(ProcessPool &procPool, const int root, const CoreData &c
 }
 
 // Check item equality
-bool Weights::equality(ProcessPool &procPool)
+bool NeutronWeights::equality(ProcessPool &procPool)
```
LamarMoore, line 472: consider removing this method if not necessary otherwise remove the guards

src/classes/xrayweights.cpp
```diff
@@ -0,0 +1,352 @@
...
+#include "classes/xrayweights.h"
+#include "base/lineparser.h"
+#include "base/processpool.h"
+#include "classes/atomtype.h"
+#include "classes/species.h"
+#include "classes/speciesinfo.h"
+#include "genericitems/array2ddouble.h"
+#include "templates/enumhelpers.h"
+
+XRayWeights::XRayWeights()
```
LamarMoore, line 31: use constructor intializer list.

src/gui/helpers/treewidgetupdater.h
```diff
@@ -133,6 +133,101 @@ template <class T, class I> class TreeWidgetUpdater
...
+		}
+	}
+
+	// Update the children of the specified parent QTreeWidgetItem
+	TreeWidgetUpdater(QTreeWidgetItem *parentItem, const std::vector<I> &data, T *functionParent,
+			  TreeWidgetChildUpdateFunction updateChildFunction)
+	{
+		int count = 0;
+
+		for (auto dataItem : data)
```
LamarMoore, line 192: `const auto &` or `auto &` this is copying items.

src/gui/keywordwidgets/isotopologuecollection_funcs.cpp
```diff
@@ -165,17 +165,20 @@ void IsotopologueCollectionKeywordWidget::addButton_clicked(bool checked)
 
 		Species *sp = isoWeight->isotopologue()->parent();
 
-		Isotopologues *topes = set->isotopologues(sp);
+		auto data = set->getIsotopologues(sp);
```
LamarMoore, line 168: auto *

src/modules/neutronsq/functions.cpp
```diff
@@ -35,11 +35,11 @@ bool NeutronSQModule::calculateWeightedGR(PartialSet &unweightedgr, PartialSet &
 		for (typeJ = typeI; typeJ < unweightedgr.nAtomTypes(); ++typeJ)
 		{
 			double weight = weights.weight(typeI, typeJ);
-			double boundWeight = weights.boundWeight(typeI, typeJ);
+			double intraWeight = weights.intramolecularWeight(typeI, typeJ);
```
LamarMoore, line 38: auto

src/templates/broadcastvector.h
```diff
@@ -0,0 +1,79 @@
...
+	/*
+	 * Constructor-only template class which iterates over a supplied sstd::vector, broadcasting the object from master
+	 * to slave processes. The items must provide the 'broadcast()' virtual.
+	 */
+	private:
+	// Result of broadcast
+	bool result_;
+
+	public:
+	BroadcastVector(ProcessPool &procPool, int root, std::vector<T> &items, const CoreData &coreData)
```
LamarMoore, line 39: Is `ProcessPool` being modified?

src/classes/isotopologuecollection.cpp
```diff
@@ -228,13 +211,12 @@ bool IsotopologueCollection::read(LineParser &parser, const CoreData &coreData)
 bool IsotopologueCollection::write(LineParser &parser)
 {
 	// Write number of Configurations in the collection
-	if (!parser.writeLineF("%i\n", isotopologueSets_.nItems()))
+	if (!parser.writeLineF("%i\n", isotopologueSets_.size()))
```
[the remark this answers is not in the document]
trisyoungs: That's a good suggestion. LineParser is YAAC (Yet Another Archaic Class) from back in the day.

src/classes/neutronweights.cpp
trisyoungs: This, and other uses of auto throughout the code, will be addressed in a forthcoming PR (related issue #283).

src/templates/broadcastvector.h
trisyoungs: No, but the class functions themselves are lazily non-const.  Depending on the upcoming change to parallelism, it may be rewritten or may potentially be removed entirely.

PR #287, opened 2020-05-12
No list atomtypedata

rprospero:
This PR removes the ListItem dependency from the AtomTypeData class and the AtomTypeList class.  Both are now handled by references instead of pointers.

As part of this process, the `for_pairs` algorithm has been added, which simplifies handling loops which go through each combination of elements in a container.

### DIFF
--- a/src/classes/atomtypedata.cpp
+++ b/src/classes/atomtypedata.cpp
@@ -32,42 +32,42 @@
 #include <string.h>
 
 AtomTypeData::AtomTypeData(AtomType &type, double population, double fraction, double boundCoherent, int nIso)
-	: atomType_(type), exchangeable_(false), population_(population), fraction_(fraction), boundCoherent_(boundCoherent)
+    : atomType_(type), exchangeable_(false), population_(population), fraction_(fraction), boundCoherent_(boundCoherent)
 {
-	isotopes_.clear();
-	for (int n = 0; n < nIso; ++n)
-	{
-		isotopes_.add();
-	}
+    isotopes_.clear();
+    for (int n = 0; n < nIso; ++n)
+    {
+        isotopes_.add();
+    }
 }
 
 AtomTypeData::AtomTypeData(const AtomTypeData &source) : atomType_(source.atomType_), listIndex_(source.listIndex())
 {
-	(*this) = source;
+    (*this) = source;
 }
 
 // Read data through specified LineParser
 AtomTypeData::AtomTypeData(LineParser &parser, const CoreData &coreData, int listIndex)
-	: atomType_(*coreData.findAtomType(parser.argc(0))), listIndex_(listIndex)
+    : atomType_(*coreData.findAtomType(parser.argc(0))), listIndex_(listIndex)
 {
-	population_ = parser.argd(1);
-	fraction_ = parser.argd(2);
-	boundCoherent_ = parser.argd(3);
-	isotopes_.clear();
-	int nIso = parser.argi(4);
-	for (int n = 0; n < nIso; ++n)
-	{
-		IsotopeData *tope = isotopes_.add();
-	}
+    population_ = parser.argd(1);
+    fraction_ = parser.argd(2);
+    boundCoherent_ = parser.argd(3);
+    isotopes_.clear();
+    int nIso = parser.argi(4);
+    for (int n = 0; n < nIso; ++n)
+    {
+        IsotopeData *tope = isotopes_.add();
+    }
 }
 
 // Initialise Constructor
 AtomTypeData::AtomTypeData(int listIndex, AtomType &type, double population)
-	: atomType_(type), listIndex_(listIndex), population_(population)
+    : atomType_(type), listIndex_(listIndex), population_(population)
 {
-	exchangeable_ = false;
-	fraction_ = 0.0;
-	boundCoherent_ = 0.0;
+    exchangeable_ = false;
+    fraction_ = 0.0;
+    boundCoherent_ = 0.0;
 }
 
 void AtomTypeData::operator=(const AtomTypeData &source)
@@ -151,13 +151,13 @@ void AtomTypeData::finalise(double totalAtoms)
 // Remove any existing isotopes, and add only the natural isotope
 void AtomTypeData::naturalise()
 {
-	// Clear the isotopes list and add on the natural isotope, keeping the current population
-	isotopes_.clear();
-	IsotopeData *topeData = isotopes_.add();
-	topeData->initialise(Isotopes::naturalIsotope(atomType_.element()));
-	topeData->add(population_);
-	topeData->finalise(population_);
-	boundCoherent_ = topeData->isotope()->boundCoherent();
+    // Clear the isotopes list and add on the natural isotope, keeping the current population
+    isotopes_.clear();
+    IsotopeData *topeData = isotopes_.add();
+    topeData->initialise(Isotopes::naturalIsotope(atomType_.element()));
+    topeData->add(population_);
+    topeData->finalise(population_);
+    boundCoherent_ = topeData->isotope()->boundCoherent();
 }
 
 // Return if specified Isotope is already in the list
@@ -206,15 +206,14 @@ const char *AtomTypeData::atomTypeName() const { return atomType_.name(); }
 // Write data through specified LineParser
 bool AtomTypeData::write(LineParser &parser)
 {
-	// Line Contains: AtomType name, exchangeable flag, population, fraction, boundCoherent, and nIsotopes
-	if (!parser.writeLineF("%s %f %f %f %i\n", atomType_.name(), population_, fraction_, boundCoherent_,
-			       isotopes_.nItems()))
-		return false;
-	ListIterator<IsotopeData> isotopeIterator(isotopes_);
-	while (IsotopeData *topeData = isotopeIterator.iterate())
-		if (!topeData->write(parser))
-			return false;
-	return true;
+    // Line Contains: AtomType name, exchangeable flag, population, fraction, boundCoherent, and nIsotopes
+    if (!parser.writeLineF("%s %f %f %f %i\n", atomType_.name(), population_, fraction_, boundCoherent_, isotopes_.nItems()))
+        return false;
+    ListIterator<IsotopeData> isotopeIterator(isotopes_);
+    while (IsotopeData *topeData = isotopeIterator.iterate())
+        if (!topeData->write(parser))
+            return false;
+    return true;
 }
 
 /*
@@ -225,21 +224,21 @@ bool AtomTypeData::write(LineParser &parser)
 // Broadcast data from Master to all Slaves
 bool AtomTypeData::broadcast(ProcessPool &procPool, const int root, const CoreData &coreData)
 {
-	// For the atomType_, use the fact that the AtomType names are unique...
-	CharString typeName;
-	if (procPool.poolRank() == root)
-		typeName = atomType_.name();
-	procPool.broadcast(typeName, root);
-	atomType_ = *coreData.findAtomType(typeName);
+    // For the atomType_, use the fact that the AtomType names are unique...
+    CharString typeName;
+    if (procPool.poolRank() == root)
+        typeName = atomType_.name();
+    procPool.broadcast(typeName, root);
+    atomType_ = *coreData.findAtomType(typeName);
 
-	// Broadcast the IsotopeData list
-	BroadcastList<IsotopeData> topeBroadcaster(procPool, root, isotopes_, coreData);
-	// if (topeBroadcaster.failed())
-	//   Messenger("Broadcase of AtomTypeData failed");
+    // Broadcast the IsotopeData list
+    BroadcastList<IsotopeData> topeBroadcaster(procPool, root, isotopes_, coreData);
+    // if (topeBroadcaster.failed())
+    //   Messenger("Broadcase of AtomTypeData failed");
 
-	procPool.broadcast(population_, root);
-	procPool.broadcast(fraction_, root);
-	procPool.broadcast(boundCoherent_, root);
+    procPool.broadcast(population_, root);
+    procPool.broadcast(fraction_, root);
+    procPool.broadcast(boundCoherent_, root);
 }
 #endif
 
--- a/src/classes/atomtypedata.cpp
+++ b/src/classes/atomtypedata.cpp
@@ -31,16 +31,44 @@
 #include "templates/broadcastlist.h"
 #include <string.h>
 
-AtomTypeData::AtomTypeData() : ListItem<AtomTypeData>()
+AtomTypeData::AtomTypeData(AtomType &type, double population, double fraction, double boundCoherent, int nIso)
+	: atomType_(type), exchangeable_(false), population_(population), fraction_(fraction), boundCoherent_(boundCoherent)
 {
-    atomType_ = NULL;
-    exchangeable_ = false;
-    population_ = 0;
-    fraction_ = 0.0;
-    boundCoherent_ = 0.0;
+	isotopes_.clear();
+	for (int n = 0; n < nIso; ++n)
+	{
+		isotopes_.add();
+	}
 }
 
-AtomTypeData::AtomTypeData(const AtomTypeData &source) { (*this) = source; }
+AtomTypeData::AtomTypeData(const AtomTypeData &source) : atomType_(source.atomType_), listIndex_(source.listIndex())
+{
+	(*this) = source;
+}
+
+// Read data through specified LineParser
+AtomTypeData::AtomTypeData(LineParser &parser, const CoreData &coreData, int listIndex)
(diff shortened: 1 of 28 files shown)


src/modules/bragg/functions.cpp
```diff
@@ -331,12 +331,13 @@ bool BraggModule::calculateBraggTerms(ProcessPool &procPool, Configuration *cfg,
 		kVectors[m].calculateIntensities(reflections);
 
 	// Normalise intensities against number of atoms and unit cell multiplicity
-	AtomTypeData *atd1 = cfg->usedAtomTypes();
+	auto &types = cfg->usedAtomTypesList().types();
 	const double divisor = 1.0 / (nAtoms * multiplicity.x * multiplicity.y * multiplicity.z);
-	for (int i = 0; i < cfg->nUsedAtomTypes(); ++i, atd1 = atd1->next())
+	int i = 0;
+	for (auto atd1 = types.begin(); atd1 != types.end(); i++, atd1++)
```
rprospero, line 337: This should probably be another `for_pairs` instance

src/modules/bragg/functions.cpp
```diff
@@ -398,11 +399,12 @@ bool BraggModule::formReflectionFunctions(ProcessPool &procPool, Configuration *
 	// Loop over pairs of atom types, adding in contributions from our calculated BraggReflections
 	double qCentre, factor;
 	int bin;
-	AtomTypeData *atd1 = cfg->usedAtomTypes();
-	for (int typeI = 0; typeI < nTypes; ++typeI, atd1 = atd1->next())
+	auto &types = cfg->usedAtomTypesList().types();
+	int typeI = 0;
+	for (auto atd1 = types.begin(); atd1 != types.end(); typeI++, atd1++)
```
rprospero, line 404: `for_pairs` again

src/classes/atomtypelist.h
```diff
@@ -40,54 +40,54 @@ class AtomTypeList : public GenericItemBase
...
-	bool contains(AtomType *atomType, Isotope *tope);
+	bool contains(AtomType &atomType, Isotope *tope);
 	// Return number of AtomType/Isotopes in list
 	int nItems() const;
 	// Return first item in list
-	AtomTypeData *first() const;
+	const AtomTypeData &first() const;
 	// Return types list
-	const List<AtomTypeData> &types() const;
+	const std::vector<AtomTypeData> &types() const;
```
rprospero, line 80: Replacing this member with proper `begin` and `end` members would allow `AtomTypeList` to be used in STL containers and would remove the implementation detail that the base is a vector from the API

src/classes/neutronweights.cpp
```diff
@@ -190,7 +190,7 @@ void NeutronWeights::calculateWeightingMatrices()
 		const AtomTypeList &speciesAtomTypes = sp->usedAtomTypes();
 		const int nAtoms = sp->nAtoms();
 		intraFlag = false;
-		for (atd1 = speciesAtomTypes.first(); atd1 != NULL; atd1 = atd1->next())
+		for (auto atd1 = types.begin(); atd1 != types.end(); ++atd1)
```
rprospero, line 193: `for_pairs` instance

src/classes/partialset.cpp
```diff
@@ -411,49 +376,34 @@ void PartialSet::setFileNames(const char *prefix, const char *tag, const char *s
...
+	});
 
 	total_.values() += delta;
 }
 
 // Form partials from stored Histogram data
 void PartialSet::formPartials(double boxVolume)
 {
-	int n, m;
+	int n = 0, m = 0;
```
rprospero, line 394: remove unused variables n and m.  They are now completely contained within the lambda.

src/modules/rdf/functions.cpp
```diff
@@ -848,13 +819,15 @@ bool RDFModule::testReferencePartials(PartialSet &setA, PartialSet &setB, double
 {
 	// Get a copy of the AtomTypeList to work from
 	AtomTypeList atomTypes = setA.atomTypes();
-	AtomTypeData *typeI = atomTypes.first();
+	auto &types = atomTypes.types();
 	double error;
-	for (int n = 0; n < atomTypes.nItems(); ++n, typeI = typeI->next())
+
+	for (auto typeI = types.begin(); typeI != types.end(); ++typeI)
```
rprospero, line 825: This should be a `for_pairs` instance

src/templates/algorithms.h
```diff
@@ -0,0 +1,36 @@
+/*
+	*** Generic Algorithms
+	*** src/templates/pair.h
```
trisyoungs, line 3:
```suggestion
	*** src/templates/algorithms.h
```

src/templates/algorithms.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+// Perform an operation on every pair of elements in a container
+template <class Iter, class Lam> void for_pairs(Iter begin, Iter end, Lam lambda)
+{
+	int i = 0;
+	for (auto elem1 = begin; elem1 != end; elem1++, i++)
+	{
+		int j = i;
+		for (auto elem2 = elem1; elem2 != end; elem2++, j++)
```
trisyoungs, line 31: Prefix incrementers rather than postfix?

src/classes/atomtypelist.cpp
```diff
@@ -61,76 +62,68 @@ void AtomTypeList::clear() { types_.clear(); }
...
+				[&atomType](const AtomTypeData &data) { return &data.atomType() == &atomType; });
 
 	// Create new entry if one wasn't found
-	if (atd == NULL)
+	if (atd == types_.end())
 	{
-		atd = types_.add();
-		atd->initialise(types_.nItems() - 1, atomType, 0);
+		types_.emplace_back(types_.size(), atomType, 0);
+		atd = types_.end() - 1;
```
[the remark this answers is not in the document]
rprospero: Unfortunately, `types_.back()` gives us the last element, while what we need is an iterator located at the last element.  `types_.rbegin()` does give us an iterator in the position that we're looking for and we don't have to worry about the reversed iteration semantics in this instance.

Then again, it might be worth just splitting this up into two return statements so that we aren't *forced* to use an iterator when all we really wanted to do was return the value in the first place.

src/classes/atomtypedata.cpp
```diff
@@ -186,39 +197,17 @@ void AtomTypeData::setBoundCoherent(double d) { boundCoherent_ = d; }
...
 // Return referenced AtomType name
-const char *AtomTypeData::atomTypeName() const { return atomType_->name(); }
+const char *AtomTypeData::atomTypeName() const { return atomType_.name(); }
 
 /*
  * I/O
  */
 
-// Read data through specified LineParser
-bool AtomTypeData::read(LineParser &parser, const CoreData &coreData)
```
LamarMoore, line 196: If I am understanding correctly, you would ideally maintain a handle to the parser so that you could call read/write without the need to pass the reference? You may want to use `weak_ptr` or `shared_ptr` (I would prefer the `weak_ptr`)

src/classes/atomtypedata.h
```diff
@@ -35,11 +34,18 @@ class ProcessPool;
...
+class AtomTypeData
 {
 	public:
-	AtomTypeData();
+	// Constructor
+	AtomTypeData(AtomType &type);
+	// Copy Constructor
 	AtomTypeData(const AtomTypeData &source);
+	// Read data through specified LineParser
+	AtomTypeData(LineParser &parser, const CoreData &coreData, int listIndex);
```
LamarMoore, line 45: I would either take a pointer to the LineParser. We should discuss.

src/classes/atomtypedata.cpp
```diff
@@ -171,7 +182,7 @@ void AtomTypeData::setSingleIsotope(Isotope *tope)
 }
 
 // Return Isotope
-IsotopeData *AtomTypeData::isotopeData() { return isotopes_.first(); }
+IsotopeData *AtomTypeData::isotopeData() const { return isotopes_.first(); }
```
LamarMoore, line 185: Why pointer and not reference for the return type?

src/classes/atomtypelist.cpp
```diff
@@ -39,13 +40,12 @@ AtomTypeList::AtomTypeList(const AtomTypeList &source) { (*this) = source; }
 void AtomTypeList::operator=(const AtomTypeList &source) { types_ = source.types_; }
 
 // Array access operator
-AtomTypeData *AtomTypeList::operator[](int n)
+AtomTypeData &AtomTypeList::operator[](int n)
```
LamarMoore, line 43: `unsigned int` or `size_t` for index.

src/classes/neutronweights.cpp
```diff
@@ -270,25 +266,19 @@ void NeutronWeights::calculateWeightingMatrices()
...
-		ci = atd1->fraction();
-
-		atd2 = atd1;
-		for (int typeJ = typeI; typeJ < atomTypes_.nItems(); ++typeJ, atd2 = atd2->next())
-		{
-			// Skip this pair if there are no such intramolecular interactions
-			if (!globalFlag.at(typeI, typeJ))
-				continue;
+	// Normalise the boundWeights_ array, and multiply by atomic concentrations and Kronecker delta
+	for_pairs(atomTypes_.begin(), atomTypes_.end(),
```
LamarMoore, line 270: `for_each_pair` probably a better name

src/templates/algorithms.h
```diff
@@ -0,0 +1,36 @@
...
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with Dissolve.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+// Perform an operation on every pair of elements in a container
+template <class Iter, class Lam> void for_pairs(Iter begin, Iter end, Lam lambda)
```
LamarMoore, line 25: You could probably achieve something similar using `std::transform`:
```cpp
std::vector<T> someList(someLength);

std::transform(someList.begin(), someList.end()-1, someList.begin()+1, someList.end(), ...);
```

rprospero: I've looked at `std::transform`, but it still seems to be limited to O(N) algorithms and the uses of `for_each_pair` is decidedly O(N²).  Transform would only allow us to examine at most two pairs for every element.

src/classes/atomtypedata.cpp
rprospero: That will be happening in a future PR.  The AtomTypeList and AtomTypeData classes were chosen for this PR because they were blockers on switch the Isotope classes from pointers to references.

src/classes/atomtypedata.cpp
trisyoungs: Just to restate what was said in the stand-up, there is no requirement to maintain a handle to the LineParser. The underlying issue is that the AtomTypeData class is responsible for reading in its own data, one component of which is the AtomType reference, and which is not known at construction.  The way around this is to delegate responsibility for reading data up to the parent AtomTypeList (see #289).

src/classes/atomtypelist.cpp
trisyoungs: +1 `unsigned int`

src/templates/algorithms.h
trisyoungs: Looking at the documentation is does indeed state that std::transform is designed to operate on a linear list of items, whereas we want to operate on all possible pairs of elements in the list (i.e. the 2D matrix), hence the appearance of O(N²) runtime. In fact, 99% of times we only want to consider the upper-triangular part of the matrix, so the runtime is in fact O(N(N+1)/2).

src/classes/atomtypedata.cpp
```diff
@@ -31,16 +31,40 @@
 #include "templates/broadcastlist.h"
 #include <string.h>
 
-AtomTypeData::AtomTypeData() : ListItem<AtomTypeData>()
+// Constructor
```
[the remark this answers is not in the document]
rprospero: Removed the constructor and copy constructor comments

src/classes/partialset.cpp
```diff
@@ -247,26 +240,21 @@ Data1D PartialSet::boundTotal(bool applyConcentrationWeights) const
 	Data1D bound;
 	bound.initialise(boundPartials_.constAt(0, 0));
 
-	int typeI, typeJ;
-	AtomTypeData *atd1 = atomTypes_.first();
-	for (typeI = 0; typeI < nTypes; ++typeI, atd1 = atd1->next())
-	{
-		AtomTypeData *atd2 = atd1;
-		for (typeJ = typeI; typeJ < nTypes; ++typeJ, atd2 = atd2->next())
+	int typeI = 0, typeJ = 0;
```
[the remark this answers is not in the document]
rprospero: I wound up removing the variables entirely.

src/classes/neutronweights.cpp
```diff
@@ -147,27 +148,26 @@ void NeutronWeights::calculateWeightingMatrices()
 	double ci, cj, bi, bj;
 
 	// Determine atomic concentration products, bound coherent products, and full scattering weights
-	AtomTypeData *atd1 = atomTypes_.first(), *atd2;
-	for (int typeI = 0; typeI < atomTypes_.nItems(); ++typeI, atd1 = atd1->next())
-	{
-		ci = atd1->fraction();
-		bi = atd1->boundCoherent() * 0.1;
+	int typeI = 0;
```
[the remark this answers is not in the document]
rprospero: Ultimately, we didn't need the variable at all.

src/classes/partialset.cpp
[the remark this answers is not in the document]
rprospero: Again, removed the variables completely.

src/modules/refine/process.cpp
```diff
@@ -327,23 +328,17 @@ bool RefineModule::process(Dissolve &dissolve, ProcessPool &procPool)
...
-								      factor);
-					combinedRho.at(globalI, globalJ) += rho * factor;
-					combinedFactor.at(globalI, globalJ) += factor;
-					combinedCWeights.at(globalI, globalJ) += weights.concentrationProduct(i, j);
-				}
-			}
+			auto &types = unweightedSQ.atomTypes().types();
+			for_pairs(types.begin(), types.end(), [&](int i, const AtomTypeData &atd1, int j, const AtomTypeData &atd2) {
+				double globalI = atd1.atomType().index();
+				double globalJ = atd2.atomType().index();
```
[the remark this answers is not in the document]
rprospero: switched to auto

src/classes/atomtypelist.cpp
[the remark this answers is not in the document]
rprospero: I did split this into two separate returns so that the code could be more readable, if a bit less concise.

src/templates/algorithms.h
rprospero: Switch made

src/classes/neutronweights.cpp
rprospero: That name is much better

src/classes/atomtypelist.cpp
rprospero: Moved to unsigned int

src/classes/atomtypedata.cpp
```diff
@@ -31,16 +31,40 @@
...
 	boundCoherent_ = 0.0;
 }
 
-AtomTypeData::AtomTypeData(const AtomTypeData &source) { (*this) = source; }
+// Copy Constructor
+AtomTypeData::AtomTypeData(const AtomTypeData &source) : atomType_(source.atomType_), listIndex_(source.listIndex()) { (*this) = source; }
+
+// Read data through specified LineParser
+AtomTypeData::AtomTypeData(LineParser &parser, const CoreData &coreData, int listIndex) : atomType_(*coreData.findAtomType(parser.argc(0))), listIndex_(listIndex)
+{
```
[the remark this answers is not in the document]
rprospero: I've removed this constructor and made AtomTypeList responsible for adding all the elements to the list.

src/classes/atomtypedata.cpp
rprospero: I've moved the parsing of AtomTypeData entirely into the AtomTypeList.

src/classes/atomtypedata.h
rprospero: This constructor has been removed entirely and the AtomTypeData class doesn't contain any pointers or references to pointers.

src/classes/atomtypelist.cpp
```diff
@@ -255,45 +248,43 @@ AtomType *AtomTypeList::atomType(int n)
...
+AtomTypeData &AtomTypeList::atomTypeData(AtomType &atomType)
 {
-	for (AtomTypeData *atd = types_.first(); atd != NULL; atd = atd->next())
-		if (atomType == atd->atomType())
+	for (auto &atd : types_)
+		if (&atomType == &atd.atomType())
 			return atd;
-
-	return NULL;
 }
```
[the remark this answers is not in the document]
rprospero: I've switched this to optional.

PR #293, opened 2020-06-04
Move code to C++17

rprospero:
This moves us to C++17.  Besides the basic build switch, I've also gone through and moved us onto the official std::optional (which turned out to be a mandatory move) and also switch lambda function params to auto where possible.

### DIFF
--- a/src/data/formfactors.h
+++ b/src/data/formfactors.h
@@ -27,7 +27,7 @@
 #include <optional>
 #include <tuple>
 
-template <class T> using opt = std::optional<std::reference_wrapper<T>>;
+template <class T> using optional_reference_wrapper = std::optional<std::reference_wrapper<T>>;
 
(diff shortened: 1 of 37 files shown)


src/data/formfactors.h
```diff
@@ -24,9 +24,10 @@
 #include "base/enumoptions.h"
 #include "data/elements.h"
 #include "data/formfactordata.h"
+#include <optional>
 #include <tuple>
 
-template <class T> using optional = std::tuple<T, bool>;
+template <class T> using opt = std::optional<std::reference_wrapper<T>>;
```
DanNixon, line 30:
```suggestion
template <class T> using optional_reference_wrapper = std::optional<std::reference_wrapper<T>>;
```

`optional_reference_wrapper` is clearer than just `opt` but still shorter.

PR #294, opened 2020-06-08
Move to mutable CoreData in read and parse

rprospero:
Back when CoreData contained mostly pointers, a const CoreData could
still mutate the values that it pointer to.  Now that we're moving to
references, the const references are immutable.

CoreData is essentially a global index of all of the values in the
simulation.  For example, it contains a list of molecules, a list of
atoms, and a list of isotopes.  Each molecules, in turn, has a list of
pointers/references to the atoms, and each atom a pointer/reference to
an isotope.  But these pointers/references are only to the ones in the
arrays in CoreData.

I bring this up to explain that the read and parse functions still
shouldn't be mutating the CoreData, but the created values that return
from the function need to contain non-constant references to the other
members in CoreData, so CoreData itself needs to be non-const.

As a future optimisation, we may eventually eliminate CoreData
entirely, since one of its main purposes is to encapsulate the
lifetime off all of the simulation data, but the move to references
allows the compiler to handle that on its own.  This is a ways down
the road, though.

### DIFF
--- a/src/classes/atomtypelist.cpp
+++ b/src/classes/atomtypelist.cpp
@@ -300,7 +300,7 @@ void AtomTypeList::print() const
 const char *AtomTypeList::itemClassName() { return "AtomTypeList"; }
 
 // Read data through specified LineParser
-bool AtomTypeList::read(LineParser &parser, const CoreData &coreData)
+bool AtomTypeList::read(LineParser &parser, CoreData &coreData)
 {
     types_.clear();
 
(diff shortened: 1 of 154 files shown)


rprospero:
The actual commit itself doesn't do anything beyond removing the `const` qualifier on the CoreData parameter of all of the `read` and `parse` functions.  It needed to be done on all of them due to the recursive nature of `read` and `parse`.

trisyoungs:
As discussed, removing the actual CoreData object itself may not be viable as its function is to allow all descriptive information for a simulation (of which there may be more than one) to be encapsulated into a single object, rather than to provide lifetime management for data.

Issue #295, opened 2020-06-09
Remove custom Pair type


rprospero:
This can be replaced with the tuple type.  Now that we're on c++17, this even allows for automatic unwrapping.

rprospero:
This would be part of the modernisation project.

trisyoungs:
Closed by #297.

PR #297, opened 2020-06-09
Replace the custom Pair class with std::tuple

rprospero:
Not only does this lower the code we have to maintain, it also gets us nice features like structured binding and comparability with std::tie.

This solves issue #295 and also part of issue #257.


src/classes/scatteringmatrix.h
```diff
@@ -48,7 +49,7 @@ class ScatteringMatrix
      */
     private:
     // Reference pairs of AtomTypes
-    List<Pair<AtomType *, AtomType *>> typePairs_;
+    std::vector<tuple<AtomType *, AtomType *>> typePairs_;
```
DanNixon, line 52:
```suggestion
    std::vector<std::tuple<AtomType *, AtomType *>> typePairs_;
```

PR #298, opened 2020-06-19
Fix isotopologue delegate

trisyoungs:
This PR addresses an issue with the GUI that arose from merging previous work (#284) that replaced pointer usage with references in Isotopologue-related classes.

### Issue
Pointers to class objects (IsotopologueSet, Isotopologue, and IsotopologueWeight) were stored in QVariants and associated to the QTreeWidgetItem's that exposed them for editing. Following use of references and std::vector as the parent store, storing pointers is no longer performed. This broke item delegates within the table as they relied on those pointers in order to locate and make changes to the underlying objects.

### Resolution
A new delegate, CustomComboDelegate, has been introduced which uses a class member function callback to get the available items for display in the QComboBox. It achieves this by passing the QModelIndex for the activated item to the callback function - it is then the responsibility of the callback function to determine which particular item is being edited, and hence which Isotopologues should be returned.

### Alternative Approach Considered
Move to full Model/View implementation for tree/table widgets displaying custom class data for editing. This is an involved approach, and requires either the addition of several (light) new classes for each data type to be displayed, or the reworking of the data type class in order to allow better integration with Qt's Model/View framework. In the interest of simplicity, and maintaining the mantra of keeping data classes completely unpolluted by GUI-related hooks and functions, this was disregarded.


src/classes/isotopologuecollection.cpp
```diff
@@ -126,7 +126,8 @@ bool IsotopologueCollection::contains(const Configuration *cfg) const
 }
 
 // Return IsotopologueSet for the specified Configuration
-std::optional<const IsotopologueSet> IsotopologueCollection::getIsotopologueSet(const Configuration *cfg) const
+std::optional<std::reference_wrapper<const IsotopologueSet>>
```
rprospero, line 129: in src/data/formfactors.h, there's an alias the std::optional<std::reference_wrapper<T>>.  Maybe we should put that into a header so it can be used here and elsewhere?  It might just make things a little less combersome.  Plus, if the committee ever decides to allow straight references inside optionals, it'll be a single place to fix everything.

src/gui/delegates/customcombodelegate.h
```diff
@@ -0,0 +1,102 @@
...
+#include <QComboBox>
+#include <QModelIndex>
+#include <QObject>
+#include <QtWidgets/QItemDelegate>
+
+// Forward Declarations
+/* none */
+
+// Custom Combo Delegate
+template <class P> class CustomComboDelegate : public QItemDelegate
```
rprospero, line 33: Should we be subclassing QStyledItemDelegate instead of QItemDelegate?  I'm not that familiar with the subject, but it was given as the best practice in [the Qt docs](https://doc.qt.io/qt-5/qitemdelegate.html#details).  Granted, the only direct benefit I understood was stylesheet support, but I know that that can be important for accessibility.

src/gui/delegates/customcombodelegate.h
```diff
@@ -0,0 +1,102 @@
...
+    // Set initial value in editor
+    void setEditorData(QWidget *editor, const QModelIndex &index) const
+    {
+        // Grab (cast) the QComboBox
+        QComboBox *comboBox = static_cast<QComboBox *>(editor);
+
+        // Get the current text and search for it in the combo
+        QString value = index.model()->data(index, Qt::EditRole).toString();
+
+        for (int n = 0; n < comboBox->count(); ++n)
```
rprospero, line 79: Is there any reason that this for loop couldn't just be `combobox->setCurrentText(value)` ?

src/classes/isotopologuecollection.cpp
DanNixon: If there is a common place that makes sense than I think this is a good idea.

trisyoungs: Agreed. To be addressed in subsequent PR.

src/gui/delegates/customcombodelegate.h
trisyoungs: Yes we should.  Now corrected.

src/gui/delegates/customcombodelegate.h
trisyoungs: How dreadfully embarrassing... :)

PR #299, opened 2020-06-26
Remove ListItem off CharString

rprospero:
This removes the ListItem base from the CharString class.  It's part of issue #257.

### DIFF
--- a/src/base/lineparser.cpp
+++ b/src/base/lineparser.cpp
@@ -763,14 +763,14 @@ void LineParser::getAllArgsDelim(int optionMask)
     while (!endOfLine_)
     {
         // Create new, empty CharString
-        arg = new CharString;
+        CharString arg;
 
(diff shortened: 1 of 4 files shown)


src/base/lineparser.cpp
```diff
@@ -763,14 +763,14 @@ void LineParser::getAllArgsDelim(int optionMask)
...
-        if (getNextArg(optionMask, arg))
+        if (getNextArg(optionMask, &arg))
         {
             // printf("getAllArgsDelim arg=%i [%s]\n", arguments_.nItems(), arg->get());
             // Add this char to the list
-            arguments_.own(arg);
+            arguments_.push_back(arg);
         }
         else
             delete arg;
```
trisyoungs, line 776: No need to delete the arg here. I suspect that this might be causing the segfaults in the serial test pipeline.

PR #300, opened 2020-06-26
Define optional reference wrapper as separate entity

trisyoungs:
This PR removes the need for manual definition of the rather lengthy `std::optional<std::reference_wrapper<T>>` by defining it once in a separate header (src/templates/optionalref.h).

New class is named `OptionalReferenceWrapper` for consistency with other classes in the code.

### DIFF
--- a/src/templates/optionalref.h
+++ b/src/templates/optionalref.h
@@ -1,0 +1,26 @@
+/*
+    *** Optional Reference
+    *** src/templates/optionalref.h
+    Copyright T. Youngs 2012-2020
+
+    This file is part of Dissolve.
+
+    Dissolve is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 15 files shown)


src/templates/optionalref.h
```diff
@@ -0,0 +1,26 @@
...
+
+    You should have received a copy of the GNU General Public License
+    along with Dissolve.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+#include <optional>
+
+template <class T> using OptionalReferenceWrapper = std::optional<std::reference_wrapper<T>>;
```
DanNixon, line 26: Bit of a nitpick but I'd be tempted to go with `optional_reference_wrapper` to be closer to the standard library naming (since this is made up of standard library components), however I'm happy with either.

DanNixon: Ah, I wrote this after reading your comment about it in the PR description...

PR #302, opened 2020-07-10
Acid FF and Forcefield bug-hunt

trisyoungs:
This PR is rooted in adding a sulphuric acid forcefield (Kulmala 2010), but propagated the following modernisation of the Forcefield class:

- Use std::vector to store forcefield data where it wasn't before.
- Pass references to ForcefieldAtomTypes instead of pointers to term matching functions.
- Instantiate atom type / term data for forcefields in new (pure virtual) setUp() function rather than the ctor to allow better error handling.
- Rework Universal Forcefield (UFF) to use references.

At present, forcefield atom types are not assigned correctly to even simple molecules, which appears to stem from bad references to ForcefieldAtomTypes or other associated memory corruption (potentially linked to how they are stored in `Forcefield::atomTypesByElementPrivate_`.

To quickly reproduce the issue, run `dissolve -c water.txt` in the `examples/species` folder. There is no crash (at least not on my system) but memory corruption of the atom type names is clear. Running with valgrind reveals many errors of the following sort:

```
==19437== Invalid read of size 8
==19437==    at 0x595390: CharString::get() const (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x60F06A: Forcefield::determineAtomType(SpeciesAtom*, std::vector<std::vector<std::reference_wrapper<ForcefieldAtomType const>, std::allocator<std::reference_wrapper<ForcefieldAtomType const> > >, std::allocator<std::vector<std::reference_wrapper<ForcefieldAtomType const>, std::allocator<std::reference_wrapper<ForcefieldAtomType const> > > > > const&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x60F12B: Forcefield::assignAtomType(SpeciesAtom*, CoreData&) const (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x611A7F: Forcefield::assignAtomTypes(Species*, CoreData&, Forcefield::AtomTypeAssignmentStrategy) const (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x524E6C: Species::applyForcefieldTerms(CoreData&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x52CFE7: Species::read(LineParser&, CoreData&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x4AD082: Dissolve::loadInput(LineParser&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x4AD58F: Dissolve::loadInput(char const*) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x49F4A8: main (in /home/tris/src/dissolve/build/dissolve)
==19437==  Address 0x5d02230 is 48 bytes inside a block of size 320 free'd
==19437==    at 0x4C2FA0B: operator delete(void*) (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==19437==    by 0x613AEC: Forcefield::addAtomType(int, int, char const*, char const*, char const*, double, double, double, double, double) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x631256: Forcefield_SPCFw::setUp() (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619B40: ForcefieldLibrary::registerForcefield(std::shared_ptr<Forcefield>) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619C3F: ForcefieldLibrary::registerForcefields() (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619D25: ForcefieldLibrary::forcefields() (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619D48: ForcefieldLibrary::forcefield(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x52CE67: Species::read(LineParser&, CoreData&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x4AD082: Dissolve::loadInput(LineParser&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x4AD58F: Dissolve::loadInput(char const*) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x49F4A8: main (in /home/tris/src/dissolve/build/dissolve)
==19437==  Block was alloc'd at
==19437==    at 0x4C2E94F: operator new(unsigned long) (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==19437==    by 0x6132C6: Forcefield::addAtomType(int, int, char const*, char const*, char const*, double, double, double, double, double) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x63120F: Forcefield_SPCFw::setUp() (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619B40: ForcefieldLibrary::registerForcefield(std::shared_ptr<Forcefield>) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619C3F: ForcefieldLibrary::registerForcefields() (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619D25: ForcefieldLibrary::forcefields() (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x619D48: ForcefieldLibrary::forcefield(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x52CE67: Species::read(LineParser&, CoreData&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x4AD082: Dissolve::loadInput(LineParser&) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x4AD58F: Dissolve::loadInput(char const*) (in /home/tris/src/dissolve/build/dissolve)
==19437==    by 0x49F4A8: main (in /home/tris/src/dissolve/build/dissolve)
==19437== 
```

Focus of attention for me was `Forcefield::determineAtomType()` (`src/data/ff.cpp`, line 97) where some debug output statements bound the above errors.


rprospero:
Some notes as I participate in this bug hunt.  Probably just as much so I don't forget where I am as anything else.

Through using the debugger, I've determined that the `AtomType` is correctly assigned the string when the variable is allocated in `Forcefield::addAtomType`.  I've also confirmed that the pointer to the string hasn't changed when the value is printed out in `Forcefield::determineAtomType`.  So it is an issue with something writing over the memory location.

rprospero:
So I've found a bit more information.  I think that the string is overwritten when the vector is reallocated.  Initially, the vector is empty.  Then HW is added to the vector, which requires allocating a one element vector.  Everything is fine at this point.  Then OW is added, which requires expanding the vector out to two elements.  The value of HW then needs to be moved to this new memory location.  It's during this move that HW gets cobbled.  Since OW is still just being added, it comes through just fine.

I'm not sure yet why it's happening, but I suspect that it has something to do with CharString possibly having an issue with its move semantics.  The easy solution there might be to turn the name into a std::string.

Before we do that, I could test this hypothesis.  If the vector was preallocated with a size of, let's say 10, then no reallocation would be necessary and the effect I'm desribing wouldn't occur.  I'll run a test with an initial allocation and see if that clears up the problem.

rprospero:
I've confirmed that that is the issue.  adding

        atomTypes_.reserve(100);

at line 64 of src/data/ff.cpp causes the problem to go away.  Unfortunately, this is only a hack fix, as the problem would recur once the 101st atom type appeared, but it does show that the problem occurs when the value is moved to a new memory location.  Turning name into a std::string, whose move semantics have been well established, may be a good path forward.

trisyoungs:
Good work! My CharString obviously needs to go, so ForcefieldAtomType and friends can be the first place where this happens.

trisyoungs:
Redundant.

PR #303, opened 2020-07-14
Condense TableWidgetUpdater

rprospero:
As part of my work on switch AtomType over to shared pointers, I need the TableWidgetUpdater to work with vectors or shared pointers.  This requires adding a new function which is *largely* similar to the existing ones, but uses shared pointers.  To make this easier, I wanted to simplify down the class.  Now, each "Constructor" merely iterates through the container type that has been passed to it and calls the `inner_` function on each value.  The third template argument allows us to swap out a shared_ptr (or a reference or another type) for a raw pointer, if that's what the container holds.  The variadic argument allows for extra data types to be passed in, which allowed for the `TableWidgetRefDataList` updater to be merged in with the original `TableWidgetUpdater`

Eventually, once we've moved away from custom containers and onto stl containers, much of this infrastruction can go away, but we'll need it during the interim.

### DIFF
--- a/src/gui/helpers/tablewidgetupdater.h
+++ b/src/gui/helpers/tablewidgetupdater.h
@@ -36,8 +36,12 @@ template <class T, class I, typename Raw = I *, typename... Args> class TableWid
     typedef void (T::*TableWidgetRowUpdateFunction)(int row, Raw item, bool createItems, Args... args);
 
     private:
-    static void inner_(QTableWidget *table, int rowCount, Raw dataItem, T *functionParent,
-                       TableWidgetRowUpdateFunction updateRow, Args... args)
+    // For a given QTableWidget, ensure that the given dataItem is at
+    // the given rowCount index.  If there are other items in the way,
+    // remove them.  If the item isn't in the table, add it.  Finally,
+    // updateRow.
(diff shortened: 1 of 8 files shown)


src/gui/helpers/tablewidgetupdater.h
```diff
@@ -24,58 +24,62 @@
...
 // as possible
-template <class T, class I> class TableWidgetUpdater
+template <class T, class I, typename Raw = I *, typename... Args> class TableWidgetUpdater
 {
     // Typedefs for passed functions
-    typedef void (T::*TableWidgetRowUpdateFunction)(int row, I *item, bool createItems);
+    typedef void (T::*TableWidgetRowUpdateFunction)(int row, Raw item, bool createItems, Args... args);
+
+    private:
+    static void inner_(QTableWidget *table, int rowCount, Raw dataItem, T *functionParent,
```
DanNixon, line 39: It would be preferable to have a more descriptive name (also without the trailing underscore).

rprospero: We definitely need a better name there.  Does `updateItemAtIndex` make its operation clearer.  The function ensures that the item is at the correct index and then runs the update on the item.

DanNixon: :+1: Sounds good, there is enough comments to make it clear from the function body what is actually happening.

Issue #307, opened 2020-07-16
Remove ListItem from AtomType


rprospero:
As part of #257, we need to remove the ListItem dependency from AtomType and move it into vectors of share_ptr.

PR #309, opened 2020-07-17
Simple tree widget updater

rprospero:
This refactors the code in TreeWidgetUpdater to remove duplicate code. This should help with adding new container types, which will aid in completing #307.

I'm quite open to stylistic suggestions, as I'm not sure that the lambdas here are the cleanest solution.  I briefly looked at a class based solution, but that seemed to add more boilerplate without making the code cleaner, but that might be worth a second look.

### DIFF
--- a/src/gui/helpers/treewidgetupdater.h
+++ b/src/gui/helpers/treewidgetupdater.h
@@ -40,8 +40,8 @@ template <class T, class I> class TreeWidgetUpdater
 
     private:
     void innerAlgorithm_(std::function<QTreeWidgetItem *(int)> accessor, std::function<int()> counter,
-                         std::function<void(QTreeWidgetItem *)> remover, std::function<void(int, I *, bool)> updater, int count,
-                         I *dataItem)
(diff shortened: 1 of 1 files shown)


src/gui/helpers/treewidgetupdater.h
```diff
@@ -97,36 +139,7 @@ template <class T, class I> class TreeWidgetUpdater
...
-            }
-
-            // If the current child index is (now) out of range, add a new item to the parent
-            if (count == parentItem->childCount())
-            {
-                // Create new item
-                (functionParent->*updateChildFunction)(parentItem, count, dataItem, true);
-            }
-
+            updateTreeChildren(parentItem, count, dataItem, functionParent, updateChildFunction);
```
trisyoungs, line 142: Use `count++` here and remove the braces / explicit `++count`?

rprospero: That was in my original code, but I took it out for being far too satisfying to be readable.  I accept any excuse to change that back.

rprospero:
Also patched up a typo in one of the function names

PR #312, opened 2020-07-19
Use ANTLR for NETA tree generation

trisyoungs:
This PR replaces the Bison parser/generator for NETA descriptions with one in ANTLR4, and paves the way for future conversion of Bison-based grammars and the eventual removal of that dependency.

### Dependencies
ANTLR4 comes as two distinct components - a generator tool written in Java, and a runtime library specific to the target language. Recent versions of both, >= 4.7.2, are required. While the Java tool is available easily via `apt-get`, `brew`, and `choco`, only `brew` also has the C++ runtime package. This seems surprising since the current version number belies some sort of longevity, but unsurprising since the C++ runtime has only relatively recently been implemented.

To mitigate this, the CMake build system has been updated to allow:
- Use of a preinstalled runtime lib (specified via `ANTLRRUNTIME_DIR` if necessary)
- Building of the runtime library as an external project

Compiled versions for all platforms are available as a failsafe in [`projectdissolve/antlr4`](https://github.com/projectdissolve/antlr4), although none are used in the current pipelines.

### Notes

On Windows, Qt / MinGW have been upgraded to use more recent versions (5.15.1 / 8.1) in this PR, negating the need for a downgrade of `gcc` and shaving three minutes off the build, mitigating somewhat the additional ten minutes required to build the ANTLR4 runtime.

**Builds on #311 - do not merge until #311 is resolved.**


src/neta/NETAErrorListeners.cpp
```diff
@@ -0,0 +1,58 @@
+/*
```
DanNixon, line 1: Could you use the new style license headers.

src/neta/NETALexer.g4
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Directly precedes the lexer class declaration in the h file (e.g. for additional types etc.).
+@lexer::context {/* lexer context section */}
+
+// Appears in the private part of the lexer in the h file.
+@lexer::declarations {
+bool isElement(std::string symbol) { return Elements::element(symbol.c_str()).Z() != 0; }
+
+void myFooLexerAction() { /* do something*/ };
```
DanNixon, line 18: Are these needed?

DanNixon (changes requested):
Looks good, nice to see the docs updates too.

src/neta/NETAErrorListeners.cpp
trisyoungs: Missed due to this being a pretty old piece of work.  Now fixed.

src/neta/NETALexer.g4
trisyoungs: Nope!  Now removed.

DanNixon (approved):
:+1:

PR #314, opened 2020-07-24
Overhaul string handling and printing

### LINKED ISSUE #286: Use std::string
Remove use of custom CharString class, moving to std::string.

trisyoungs:
This PR touches most parts of the code with minor modifications in order to:

1. Replace use of `printf`, `sprintf` and friends with type-safe formatting using `fmtlib`
2. Remove the `CharString` class in favour of using `std::string` (or `QString`) as appropriate
3. Replace `const char*` in argument passing and return values with `std::string_view`
4. Remove other custom functions that can be replaced with STL equivalents (e.g. number conversion in `DissolveSys`)

### Details

The PR itself is too big to review as a whole, so I highlight here some specific commits that illustrate the types of changes made and are worthy of a detailed look:

1. bed29c99e1426ef6441c9f9edebb6a06dc5e7070 rewrites the `Messenger` class to utilise `fmtlib`
2. 5e6e6281646e5b5cba4cc3a6802f796159892d32 updates `LineParser` (probably the most archaic of all the files) to use `std::string`
3. 1b6c845264a81f9cfc16bf24c7484f3b8c8e594d shows the simplification of `DissolveSys`

### Notes

- Where the CLI code (which uses `std::string` and `std::string_view`) interfaces with the GUI (which uses `QString`) judicious use of either the `qPrintable` macro or the verbose `QString::fromStdString(std::string(text))` to convert between the two forms has been applied, as `QString` has no implicit conversion from  or to `std::string` or `std::string_view`.

### Related Issues

Closes #286.


.azure-pipelines/templates/build-linux-parallel.yml
```diff
@@ -8,7 +8,7 @@ steps:
       set -ex
       # -- General
       sudo apt-get update -q
-      sudo apt-get install bison ninja-build libopenmpi-dev openmpi-bin openmpi-common
+      sudo apt-get install libfmt-dev ison ninja-build libopenmpi-dev openmpi-bin openmpi-common
```
DanNixon, line 11:
```suggestion
      sudo apt-get install libfmt-dev bison ninja-build libopenmpi-dev openmpi-bin openmpi-common
```

Typo?

src/base/sysfunc.cpp
```diff
@@ -37,25 +37,25 @@ using namespace std;
 // Convert integer to string representation
 const char *DissolveSys::itoa(int n)
```
DanNixon, line 38: Is there a reason these functions still exist? The standard library has equivalents (that are likely more efficient than using `fmt` for this.

src/base/charstring.h
```diff
@@ -22,6 +22,7 @@
 #pragma once
 
 #include "templates/listitem.h"
+#include <fmt/core.h>
 
 // Character String
 class CharString
```
DanNixon, line 28: What is `CharString` still used for and how far away from being able to remove it are we?

.azure-pipelines/templates/build-linux-parallel.yml
trisyoungs: Doh!

src/base/sysfunc.cpp
trisyoungs: Absolutely none whatsoever. They are being pruned in the move to std::string.

src/base/charstring.h
trisyoungs: Awful, custom handling of strings in a few classes. It will be systematically removed following the `const char *` -> `std::string_view` work.

CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ CMAKE_POLICY(SET CMP0020 NEW)
 # Request C++17 Standard Support (Required for std::optional and auto types in lambda parameters)
 set(CMAKE_CXX_STANDARD 17)
 
+# Include Conan Packages
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+conan_basic_setup()
```
DanNixon, line 31:
```suggestion
conan_basic_setup(CMAKE_TARGETS)
```

We should be using modern targets where possible in CMake.

CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ CMAKE_POLICY(SET CMP0020 NEW)
...
+set(EXTRA_LINK_LIBS ${EXTRA_LINK_LIBS} ${CONAN_LIBS})
```
DanNixon, line 32:
```suggestion
```

Do not set things globally. Instead set it for each target with the appropriate CMake function, in this case `target_link_library(xyz INTERFACE CONAN_PKG::fmt)` where `xyz` is the target that requires fmt (I think, I have not actually tested this).

(see https://docs.conan.io/en/latest/reference/generators/cmake.html)

DanNixon (approved):
Looks good to me :+1: 

Just a couple of suggestions for improvements to the CMake config.

PR #336, opened 2020-08-06
AtomType shared ptr

rprospero:
This removes the ListItem dependency from AtomType and replaces all lists of it with vectors of shared_ptr.  `shared_ptr` was chosen over references due to the fact that many objects need to be able to reassign their AtomType.


src/main/pairpotentials.cpp
```diff
@@ -22,6 +22,7 @@
 #include "classes/atomtype.h"
 #include "genericitems/listhelper.h"
 #include "main/dissolve.h"
+#include "templates/algorithms.h"
```
trisyoungs, line 25: Is this unused?  Perhaps put in with the intention of employing `for_each_pair`, before discovering the double-loop is more complicated than it first appeared?

src/main/pairpotentials.cpp
```diff
@@ -121,28 +122,28 @@ bool Dissolve::generatePairPotentials(AtomType *onlyInvolving)
...
+        for (auto at2 = at1; at2 != coreData_.atomTypes().end(); ++at2)
         {
             // If an AtomType was supplied, only generate the pair potential if one of its AtomTypes matches
-            if (onlyInvolving && (at1 != onlyInvolving) && (at2 != onlyInvolving))
+            if (onlyInvolving && (*at1 != onlyInvolving) && (*at2 != onlyInvolving))
                 continue;
 
             // Does a PairPotential for this AtomType pair already exist?
-            PairPotential *pot = pairPotential(at1, at2);
+            PairPotential *pot = pairPotential(*at1, *at2);
```
trisyoungs, line 134: `auto`

src/main/species.cpp
```diff
@@ -69,14 +69,17 @@ void Dissolve::copyAtomType(const SpeciesAtom *sourceAtom, SpeciesAtom *destAtom
...
+    auto opt_at = findAtomType(sourceAtom->atomType()->name());
+    if (!opt_at)
     {
         at = addAtomType(sourceAtom->element());
         at->setName(sourceAtom->atomType()->name());
         at->parameters() = sourceAtom->atomType()->parameters();
         at->setShortRangeType(sourceAtom->atomType()->shortRangeType());
     }
+    else
+        at = *opt_at;
```
trisyoungs, line 82:
```suggestion

    at = *opt_at;
```

src/classes/coredata.cpp
```diff
@@ -110,13 +114,15 @@ const char *CoreData::uniqueAtomTypeName(const char *base) const
 }
 
 // Search for AtomType by name
-AtomType *CoreData::findAtomType(const char *name) const
+std::optional<std::shared_ptr<AtomType>> CoreData::findAtomType(const char *name) const
```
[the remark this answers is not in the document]
rprospero: The idea is that accidentally ignoring the null case is a compile time error.  If you pass a pointer and the caller forgets to check for null, the code compiles just fine.  If you pass the optional, then the caller must dereference it to compile, which give the psychological push to add the error handling code.  This style also allows for a convention where there are no null pointers (only optional ones), so null checking can safely be skipped at runtime.

This is admittedly a very weak benefit and I'm happy to change it if we decide not to go with this style in the project.

rprospero:
With the exception of the style question on the optional pointer, I've merged in all of the other suggested changes.

rprospero:
@trisyoungs Could you do me the favour of pulling this branch and running the test on your computer?  I cannot replicate the error that is occurring in the tests, which is making it a pain to debug.

trisyoungs:
On it now.

trisyoungs:
@rprospero all looks OK now.  Are you happy for me to go ahead and merge?

PR #337, opened 2020-08-12
Add N-term Cosine Torsion Functions

### LINKED ISSUE #326: Add CosN torsion terms
Allow Cosine series for torsion terms up to order 10.

**Additional details**
Test data provided by user.

trisyoungs:
This PR adds more flexibility when specifying torsion (dihedral) intramolecular terms for species, implementing generic `CosN` and `CosNC` forms that take some number of parameters (currently limited to 10) and which are expanded accordingly. It also addresses some bugs discovered in the NETA classes at the same time.

### Side Effects
- Storing a variable number of parameters in `SpeciesIntra` proved to be a good reason to modernise that particular part of the code, removing the raw array of doubles and replacing with a `std::vector`. This also removes the hard-coded parameter limit specified in the `MAXINTRAPARAMETERS` global. The remainder of the modernisation of `SpeciesIntra` is in progress in #313.
- Previously, adding terms to `Forcefield`s with the `add*()` functions accepted up to four parameters. These functions now accept a `const std::vector`, as do the relevant class constructors.
- As a result of the changes to `Forcefield::add*()` functions, all forcefield definitions have had to be updated to provide their parameter data in `{}` form, rather than just a sequence of `double` args.

### Additional
System test `energyforce3` has been added to ensure correctness of the new forms for three separate systems.
Closes #326.


DanNixon:
I'm sorry, but this is simply unreviewable.
There is *far* too much new code for a single PR that is seemingly not split well into atomic commits.

For things of this scale please try and split the task up.

trisyoungs:
I've reworked the PR into #340 - only 8 commits this time, and hopefully more structured. It's a big commit, but the majority of the changes are either reformatting forcefields, or adding system tests.

DanNixon:
For future reference you could have force pushed the existing branch, then you would not have to open multiple pull requests.

trisyoungs:
Hedged my bets with that one, and did the reworking in a separate branch to be on the safe side!

PR #342, opened 2020-08-14
Fix Improper Torsions

trisyoungs:
This PR follows on from #340, during which issues were found with improper torsion calculation. The present PR fixes calculation of improper terms throughout the code, and adds a system test to cover this particular aspect.

To be merged after #340.


rprospero:
Am I feeling deja-vu or does this PR have many of the same changes as the PR for the Cos-N terms?  If there's a lot of duplication, it may be better to hold off on my review until the other branch has merged so that I'm reviewing the bits that matter.

trisyoungs:
Yep - sorry.  Cos-N branch is now merged so this should have cleaned up.

trisyoungs:
@rprospero And now force-pushed.

Issue #139, opened 2020-01-25
Protect data access by Renderables when simulation thread is active


trisyoungs:
Ineffective protection of data structures written to by the main simulation thread leads to crashes in the GUI when Viewers attempt to read / update them.  Current methods are not adequate, leading to:

- ConfigurationRenderable not updating when running.
- Frequent crashes on all systems (worse on Windows and OSX) because of concurrent data access in underlying data structures linked to and used by Renderable.

PR #352, opened 2020-08-21
Possible solution to Renderable bug

rprospero:
When transform was not enabled, it was possible for a reference to the
original source from the simulation to be passed to the GUI.  This now
ensures that the GUI is always passed an independent copy.

This should hopefully solve #139.


trisyoungs:
Although this didn't appear to completely fix the issue, I still think these changes should be merged in!

PR #364, opened 2020-09-02
std::accumulate for EnergyKernel claculations

rprospero:
This PR replaces many of the trivial summation loops in EnergyKernel with std::accumulate.


trisyoungs (approved):
All looks good!

PR #365, opened 2020-09-03
Actually create NETA descriptions for atom types

trisyoungs:
This PR fixes a severe error whereby atom type descriptions (NETA) were not generated at all for any atom types, leading to incorrect typing of species.

### DIFF
--- a/src/data/ff.cpp
+++ b/src/data/ff.cpp
@@ -136,6 +136,21 @@ void Forcefield::addParameters(std::string_view name, double data0, double data1
     shortRangeParameters_.emplace_back(name, data0, data1, data2, data3);
 }
 
+// Create NETA definitions for all atom types from stored defs
+bool Forcefield::createNETADefinitions()
+{
+    auto nFailed = 0;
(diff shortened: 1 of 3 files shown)


src/data/ff.cpp
```diff
@@ -136,6 +136,21 @@ void Forcefield::addParameters(std::string_view name, double data0, double data1
     shortRangeParameters_.emplace_back(name, data0, data1, data2, data3);
 }
 
+// Create NETA definitions for all atom types from stored defs
+bool Forcefield::createNETADefinitions()
+{
+    auto nFailed = 0;
+    for (auto &atomType : atomTypes_)
+        if (!atomType.createNETA(this))
+            ++nFailed;
```
rprospero, line 145:
```suggestion
    auto nFailed = std::count_if(atomTypes_.begin(), atomTypes_.end(), [this](auto &atomType){return !atomType.createNETA(this);});
```

rprospero (commented):
One little algorithm suggestion but otherwise it looks good.

PR #366, opened 2020-09-04
Add CheckSpecies module

trisyoungs:
This PR adds a new module, `CheckSpecies`, to properly assess the atom types and forcefield parameters present in a species, particularly those applied automatically from within Dissolve.

Work includes:
- Addition of new keywords `IntegerStringVectorKeyword` and `IntegerDoubleVectorKeyword` allowing general input of a set number of integers (in this instance, atom indices) followed by a variable list of strings or doubles. The underlying information is stored in a `std::vector` of `std::tuple`, with the tuple containing further `std::vector`s of the integer and string/double data.
- Updating atom type names in the OPLS-AA forcefield, some of which were using asterisks that were interpreted as wildcards when in fact they are purely identifiers.
- Updating the NETA system test to thoroughly check applied atom types and terms from the automatic typing routines, rather than just assume that the successful application of terms means the correct application of terms.

Notes:
- The new keyword classes could be template-driven, but as they are relatively short and would require specialisations this was not performed.


src/modules/checkspecies/functions.cpp
```diff
@@ -20,3 +20,28 @@
...
+    {
+        if (fabs(source.at(n) - ref.at(n)) >= tolerance)
+        {
+            Messenger::print("  ... parameter {} is incorrect ({:.5e} vs. {:.5e} reference, delta = {:.5e}", n + 1,
+                             source.at(n), ref.at(n), fabs(source.at(n) - ref.at(n)));
+            result = false;
+        }
+    }
+
+    if (source.size() != ref.size())
```
DanNixon, line 40: This check may as well come first to avoid performing the loop.

src/modules/checkspecies/process.cpp
```diff
@@ -96,5 +96,89 @@ bool CheckSpeciesModule::process(Dissolve &dissolve, ProcessPool &procPool)
...
+                Messenger::print("Angle {}-{}-{} has the correct parameters.", indices.at(0), indices.at(1), indices.at(2));
+            else
+            {
+                Messenger::print("Parameters for angle {}-{}-{} are incorrect - see above.", indices.at(0), indices.at(1),
+                                 indices.at(2));
+                ++nAnglesFailed;
+            }
+        }
+    }
+
```
DanNixon, line 125: There looks to be a fair bit of duplication here, can this easily be avoided?

DanNixon (changes requested):
Looks good :+1: and nice to see some more testing, just some minor comments.

src/modules/checkspecies/functions.cpp
trisyoungs: From an efficiency perspective this is valid, but potentially hides other useful information - its possible for the number of parameters as well as their respective values to be incorrect, and this is all useful output to the user.

DanNixon: But if the number of parameters and number of checked values differ there is no longer any guarantee that the correct pairs of values are being checked.

src/modules/checkspecies/process.cpp
trisyoungs: With a templated function and some STL magic, yes.

src/keywords/vector_intdouble.h
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+#pragma once
+
+#include "keywords/data.h"
+#include <optional>
+#include <tuple>
+#include <vector>
+
+typedef std::vector<std::tuple<std::vector<int>, std::vector<double>>> IntegerDoubleVectorKeywordData;
```
rprospero, line 29: Wouldn't we be better off with `std::vector<std::vector<std::tuple<int, double>>`?  This would ensure that there are always an equal number of integers and doubles.  Or is that no desireable?

src/keywords/vector_intstring.h
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+#pragma once
+
+#include "keywords/data.h"
+#include <optional>
+#include <tuple>
+#include <vector>
+
+typedef std::vector<std::tuple<std::vector<int>, std::vector<std::string>>> IntegerStringVectorKeywordData;
```
rprospero, line 29: Same comment as with vector_intdouble

src/keywords/vector_intstring.h
```diff
@@ -0,0 +1,60 @@
...
+
+// Keyword with list of Tuples of Vectors
+class IntegerStringVectorKeyword : public KeywordData<IntegerStringVectorKeywordData &>
```
rprospero, line 32: After looking through the implementation classes, couldn't we have a single `IntegerTVector` template that takes the second type as a parameter?  The code for the two classes seems identical.  Or, if not a template, at least a base class that can get subclassed out for the parts that actually change?

rprospero (commented):
A couple of design comments, but otherwise everything looks good.

src/modules/checkspecies/checkspecies.h
```diff
@@ -0,0 +1,133 @@
...
+        const auto &sourceParams = intra.parameters();
+
+        // Check parameter values
+        auto result = true;
+
+        for (auto n = 0; n < std::min(sourceParams.size(), refParams.size()); ++n)
+        {
+            if (fabs(sourceParams.at(n) - refParams.at(n)) >= tolerance)
+            {
+                Messenger::print("  ... parameter {} is incorrect ({:.5e} vs. {:.5e} reference, delta = {:.5e}", n + 1,
```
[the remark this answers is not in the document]
trisyoungs: I tend to use `Messenger::error` exclusively for syntactic, programmatic, or otherwise 'hard' errors that should stop execution at that point. When testing, I prefer to think of these 'scientific' errors separately, and so just print normal information (partly for readability reasons when looking through the system test logs).

src/keywords/vector_intdouble.h
trisyoungs: The number of integers and the number of doubles (or strings) is never guaranteed to be equal, hence the v-t-v ordering.

src/keywords/vector_intstring.h
trisyoungs: Yes, as I mentioned in the PR text this could no doubt be templated to some degree. In the first instance where there are only two similar classes I went for simplicity, but if a third (fourth...) derivative is required in the future then certainly we should template.

DanNixon (approved):
:+1:

PR #367, opened 2020-09-04
Remove conversions between NULL and std::string_view

rprospero:
On my system, this results in attempting to dereference the null pointer.


rprospero:
I'm curious what compiler you've been using, because I could not get the code to pass the test suite without these changes.  It's probably just an issue of different STL implementations, but I'm still surprised that I even had this issue.

trisyoungs (approved):
Good catch!  Somewhat surprised that this didn't show up for me at all with gcc 7.5.0 on my laptop (and probably a  newer version on my desktop).

Switching `NULL` for `nullptr` (or `""`) was on my list to do after finishing the `reprint` branch, but this one got lost.

PR #371, opened 2020-09-10
Correct initial MD velocities

trisyoungs:
This PR addresses a bug where random initial velocities for atoms in the Molecular Dynamics module were not assigned consistently across processes, leading to incorrect energies and forces, and corruption of the resulting configurations.

### DIFF
--- a/src/modules/md/process.cpp
+++ b/src/modules/md/process.cpp
@@ -150,6 +150,7 @@ bool MDModule::process(Dissolve &dissolve, ProcessPool &procPool)
         if (created)
         {
             randomVelocities = true;
+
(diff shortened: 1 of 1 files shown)


src/modules/md/process.cpp
```diff
@@ -150,22 +150,26 @@ bool MDModule::process(Dissolve &dissolve, ProcessPool &procPool)
         if (created)
         {
             randomVelocities = true;
+
```
rprospero, line 153:
```suggestion
```
Do we need the blank line for clang-format?

rprospero (approved):

PR #377, opened 2020-09-15
MD, MolShake, and restart file system tests

trisyoungs:
This PR adds three new system tests:

1) `tests/md` - Covers deterministic operation of the MD module, testing forces against known reference values.
2) `tests/molshake` - Covers basic operation of the MolShake module, ensuring that it runs to completion.
3) `tests/restart` - Ensures that a basic restart file can be correctly loaded.

Some other small modifications made along the way:

1) The `tests/go` script has been updated to better handle custom arguments to individual tests.
2) Some minor cleanup of unused vars and uses of `auto` have been made.
3) Force import/export has been harmonised through classes in `src/io`

Closes #369. Closes #375.

PR #379, opened 2020-09-15
Catch std::stod out_of_range exception

### LINKED ISSUE #376: Underflows when calling std::stod()
**Describe the bug**
`LineParser::argd()` uses `std::stod()` in order to convert the text string to a `double` value, but underflows can cause exceptions (implementation dependent).

This behaviour is a minor annoyance on some systems, and could be mitigated through catching the exception and handling it better, or may be corrected as part of a more in-depth replacement of the `LineParser` class with a dedicated parsing library.

Reproducible on OpenSuSE Leap 15.2 (gcc 7.5.0) with numbers containing exponents less than ~ -308.

trisyoungs:
This PR adds a try/catch around the `std::stod()` call in `LineParser::argd()` in order to avoid bailouts when reading in small, but legitimate, floating point numbers.

The new behaviour is to raise a warning to the user, but return `0.0` as the result (no exception is raised).

Closes #376.

**Uses the system test introduced in PR #377 - not to be merged until #377 is merged.**

### DIFF
--- a/src/base/lineparser.cpp
+++ b/src/base/lineparser.cpp
@@ -25,8 +25,6 @@
 #include "base/sysfunc.h"
 #include "templates/enumhelpers.h"
 #include <limits>
-#include <stdarg.h>
-#include <string.h>
 
 LineParser::LineParser(ProcessPool *procPool)
 {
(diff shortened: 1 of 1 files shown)


DanNixon (approved):
:+1:

PR #381, opened 2020-09-17
Update Pipelines

trisyoungs:


trisyoungs:
@DanNixon Annoyingly simple in the end. Don't install any npm stuff explicitly in order to fix stale bugs - just do vanilla `npm install` before running `hugo` and it's all fine. Not sure how extensive the list of emojis is in GHFM, but if there was a `:pissedoffface:` I would use it!

DanNixon:
:man_facepalming: At least it works now.

DanNixon (approved):

PR #383, opened 2020-09-17
Replace NULL with nullptr

trisyoungs:
This PR sweeps the code to remove uses of `NULL` and replace with `nullptr` (or where appropriate `""`).

One exception to this is the `sglib` code in `src/data/sg` remains untouched.


DanNixon (approved):
:+1:

PR #387, opened 2020-09-21
Enable type safe early returns in custom for_each_pair

rprospero:
This creates a second version of `for_each_pair` called `for_each_pair_early`.  Using this algorithm requires a bit more overhead, in that

1) The lambda must be given an explicit return type of EarlyReturn<T>, where T is the return type
2) The lambda must return EarlyReturn<T>::Continue at the end of the function to continue to the next step.

In return for the extra overhead and hassle, this version both allows for skipping sections of the inner loop and also performing early returns.


DanNixon (approved):
:+1:

trisyoungs (approved):
Looks good, now that I understand it!

PR #388, opened 2020-09-22
Replace old file headers with SPDX style headers

### LINKED ISSUE #386: Use SPDX license identifiers in file headers
See: https://spdx.dev/ids/

Using this format the header included in source files can be reduced to (or at least something like):
```
// Copyright (c) 2020 Project Dissolve and contributors
```

You could use either [GPL-3.0-only](https://spdx.org/licenses/GPL-3.0-only.html) or [GPL-3.0-or-later](https://spdx.org/licenses/GPL-3.0-or-later.html).

DanNixon:
Fixes #386.


trisyoungs:
I think this gives you the crown for removal of most LoC. :1st_place_medal:

PR #406, opened 2020-10-19
Refactor CLI option handling

trisyoungs:
This PR removes custom code handling CLI option parsing, and replaces it with one utilising the CLI11 library. Benefits of this change are less duplication of code (serial/parallel and GUI versions all now use the same class to parse options), better maintainability through use of a standard library, the ability to use stacked short as well as long options, and more consistent / expected handling of arguments.

Summary of work:
- New class `CLIOptions` to handle option parsing for any executable target in the codebase.
- Existing options reimplemented for the most part, but some have been harmonised or changed to different, more descriptive names.
- Updated docs and system tests.

### DIFF
--- a/.azure-pipelines/templates/build-windows-serial-gui.yml
+++ b/.azure-pipelines/templates/build-windows-serial-gui.yml
@@ -13,14 +13,11 @@ steps:
     inputs:
       versionSpec: '3.x'
   - script: |
-      choco install -y antlr4 ninja winflexbison unzip
+      choco install -y antlr4 winflexbison unzip
     displayName: 'Install Prerequisites'
   - script: |
       python -m pip install aqtinstall
       python -m aqt install --outputdir $(Build.BinariesDirectory)\\Qt ${{ parameters.qtver }} windows desktop win64_msvc2019_64 -m qtcore qtgui qtwidgets qtprintsupport
-      ls $(Build.BinariesDirectory)\Qt
-      ls $(Build.BinariesDirectory)\Qt\${{ parameters.qtver }}
-      ls $(Build.BinariesDirectory)\Qt\${{ parameters.qtver }}\msvc2019_64
     displayName: 'Install Qt'
   - powershell: |
       $ErrorActionPreference = 'Stop'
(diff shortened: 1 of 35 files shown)


src/dissolve-gui.cpp
```diff
@@ -12,131 +13,21 @@
 #include <stdlib.h>
 #include <time.h>
 
-int main(int args, char **argv)
+int main(int nArgs, char **args)
```
rprospero, line 16: Any particular reason behind this rename?  I'm mostly familiar with the `argc` `argv` standard, but it seems that we're moving away from than instead of toward it.  Is this just a newer naming system or is there some other constraint.

rprospero (approved):
I fully support moving to a proper argument parsing library.  Besides my bikeshedding about parameter names, this is all perfect.

Also, if you want to feel a bit of envy about how easy this could be, [optparse-applicate](https://github.com/pcapriotti/optparse-applicative#quick-start) makes writing these absolutely beautiful.  I'm sure that C++ will have something like this by C++40.

src/dissolve-gui.cpp
trisyoungs: I got a bit keen here, and renamed the args away from the accepted standard. It was late!  I'll revert them back to what would be expected.

DanNixon (approved):
Looks good :+1: 

The last Windows build did not get to the build step so I restarted those to see what the latest error is.

trisyoungs:
I've restarted them again as the Windows build failed with the old winflexbison/SourceForge issue.  The OSX builds fail at the `brew update` step - are you seeing this elsewhere on other projects?

DanNixon:
> are you seeing this elsewhere on other projects?

Not many other projects use Sourceforge in this way. Honestly the best solution is to avoid having to use it.

I haven't actually looked at the OS X failure yet, will take a look now.

.azure-pipelines/templates/build-windows-serial-gui.yml
```diff
@@ -57,7 +59,7 @@ steps:
       $env:LIB += "$(Build.BinariesDirectory)\ftgl-build\src;"
       mkdir build
       cd build
-      conan install -s compiler="gcc" --remote=conan-center ../
+      conan install --remote=conan-center -s compiler.libcxx=libstdc++11 -s compiler=gcc -s compiler.version=7 ../
```
DanNixon, line 62: Do these need to be set here or can they be in the profile above?

trisyoungs: @DanNixon Moved them to the profile above, but it may all be irrelevant - turns out that there is a bug in MinGW 8.1 (fixed in 9.X) on windows that mean a FS header and C++17 don't play nicely together, and this breaks the `CLI11` stuff. Rather than go backwards and revert to MinGW 7.3 perhaps the mention of using the VS compilers this morning was quite timely.

PR #407, opened 2020-10-19
Update/fix isotopologue editing (GUI)

### LINKED ISSUE #401: Update use of Isotopologue class and its data
The `Isotopologue` class archaically keeps an ordered list of all atom types in the parent `Species` and their associated `Isotope`s, but the class is only ever accessed as a means to query which `Isotope` should be used for a given `AtomType`.  This can be greatly simplified by just storing any `AtomType` / `Isotope` tuples for which the `Isotope` is not the natural one.

Class Requirements:
- [x] Remove dead functions (e.g. `update()`, `isotope(int)`)
- [x] Remove initialisation / upkeep of ordered `AtomType`/`Isotope` vector
- [x] Modify `atomTypeIsotope()` to return the natural isotope if the queried `AtomType` is not present

GUI Requirements:
- [ ] ~Update SpeciesTab in GUI to allow addition of single `AtomType`/`Isotope` lines to an `Isotopologue` in the relevant table.~
- [ ] ~Update SpeciesTab in GUI to allow removal of `AtomType`/`Isotope` lines within an `Isotopologue` in the relevant table.~
- [ ] ~Removing an `AtomType` from the ForcefieldTab should remove any definitions from current `Isotopologue`s.~

trisyoungs:
This PR addresses some usability and editability issues in the GUI regarding Isotopologues, simplifies their use within the code, and fixes a few bugs discovered along the way.

Closes #401.

### DIFF
--- a/src/classes/coredata.cpp
+++ b/src/classes/coredata.cpp
@@ -62,8 +62,7 @@ int CoreData::nAtomTypes() const { return atomTypes_.size(); }
 // Return core AtomTypes list
 std::vector<std::shared_ptr<AtomType>> &CoreData::atomTypes() { return atomTypes_; }
 
-// Return core AtomTypes list (const)
-const std::vector<std::shared_ptr<AtomType>> &CoreData::constAtomTypes() const { return atomTypes_; }
+const std::vector<std::shared_ptr<AtomType>> &CoreData::atomTypes() const { return atomTypes_; }
 
 // Return nth AtomType in list
 std::shared_ptr<AtomType> CoreData::atomType(int n) { return atomTypes_[n]; }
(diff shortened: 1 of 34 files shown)


rprospero (approved):
Nicely done.  My only comment is that we may need to add an issue for the modernization upgrade to remove all of the const accessors.  From my reading, it seems like it's legal to overload the member function based on constness, so you can have a 

```
class Cat {
  private: int tail_;
  public:
    int& tail() {return tail;}
    const int& tail() const {return tail;}
}
```

This would mean that there would only bee on function call to remember and it would do the "right thing" at the call site.  Hopefully, this could mean letting the compiler handle the const correctness most of the time, instead of leading it up to the humans.

However, this is an issue of style that may be a later discussion.

trisyoungs:
@rprospero Completely agree. I've been changing these as and when I find them, but I'll add an issue to do a sweep throughout the code.

PR #410, opened 2020-10-30
During MPI, only output from master process unless explicitly requested by the user

### LINKED ISSUE #384: Don't write output from all MPI processes by default
**Describe the bug**
When launching `dissolve-mpi` the default behaviour is for all processes to write out all information all the time, whereas a more sensible default would be for only the master process to write.  The expected behaviour can be achieved through the CLI switch `-m`.

rprospero:
The default behaviour the user would likely prefer is to only see output from the master process, instead of seeing a great deal of duplicate information from all processes.  This inverts the documented behaviour of the "-m" command line switch, but A) the documentation is updated with the new behaviour and B) the old documentation was misleading (the switch was actually a no-op).

This should closes #384.


trisyoungs (commented):
I'm afraid this clashes with work I did as part of the `new-cli-handling` PR - see second commit in [https://github.com/projectdissolve/dissolve/pull/406/commits](https://github.com/projectdissolve/dissolve/pull/406/commits). Because of the ongoing windows issues, #384 hasn't been closed yet because the PR is still open.

trisyoungs (approved):
Nice.  Must have removed the original parallel logic when I moved to `fmtlib`.

PR #412, opened 2020-10-31
Add Padua & Canongia Lopes Ionic Liquids FF

### LINKED ISSUE #201: Implement up-to-date CLDP-IL forcefield

trisyoungs:
This PR adds the Padua and Canongia Lopes complete ionic liquids forcefield, as found at https://github.com/agiliopadua/ilff.

Aside from the addition of the relevant forcefield data, some additions to the main code were also made:

1. Improper terms now re-use the defined torsion functional forms, rather than re-defining identical functions.
2. A `TotalCharge` option was added to the `CheckSpecies` module.

The `neta` system test has been reorganised into a series of tests focussing on individual forcefields, with anion- and cation-specific tests added for the present forcefield. The new tests can be found in `tests/ff`.

Documentation has been added describing the various implementation details of the forcefield, and sets a precedent for the addition of future pages for each built-in forcefield in Dissolve.

Closes #201.

**Depends on #411 - do not merge until #411 is merged.**


src/classes/speciestorsion.h
```diff
@@ -97,8 +97,12 @@ class SpeciesTorsion : public SpeciesIntra
     double fundamentalFrequency(double reducedMass) const;
     // Return type of this interaction
     SpeciesIntra::InteractionType type() const;
+    // Return energy for specified angle and functional form, given supplied parameters
+    static double energy(double angleInDegrees, int form, const std::vector<double> &params);
```
rprospero, line 101: Could the type of `form` be `SpeciesTorsion::TorsionFunction` instead of `int`?  It'd make the code both clearer and more type safe.

src/classes/speciestorsion.h
```diff
@@ -97,8 +97,12 @@ class SpeciesTorsion : public SpeciesIntra
...
     // Return energy for specified angle
     double energy(double angleInDegrees) const;
+    // Return force multiplier for specified angle and functional form, given supplied parameters
+    static double force(double angleInDegrees, int form, const std::vector<double> &params);
```
rprospero, line 105: Same as above.

rprospero (approved):
Just one small question about our types.

src/classes/speciestorsion.h
trisyoungs: I agree, but the current implementation of the various intramolecular terms relies on inheriting from a base `SpeciesIntra` class, which provides storage for common types of data across any interaction type. One of these is the functional form, which is generically stored as an `int`, hence the signatures of the functions you highlight here and below. This does, however, permit the `energy()` and `force()` functions here to take the `int` `form()` from either a `SpeciesTorsion` or a `SpeciesImproper`.

I'm all for changing this relationship to remove the `int` storage of the functional form, but it's too big a job for this PR.

PR #421, opened 2020-11-04
Prevent running simulation in the GUI if no file name has been saved.

### LINKED ISSUE #409: GUI can run a simulation without a valid input filename
**Describe the bug**
When creating a simulation from scratch in the GUI, no checks are made that the input file has been saved when attempting to run the simulation. This means that the generated restart filename ends up as ".restart", and cannot potentially be written to as the path presumably defaults to the location of the executable, which will probably be the read-only install location.

Solution: Check that the input file has been saved before running.  Alternatively, offer an option to not save the input file and run the simulation without writing restart or heartbeat files?

Screenshot of error on OSX:
Dissolve2

rprospero:
This will close #409.

A couple of comments on the commit.  First, I've refactored a good bit of menu_simulation.cpp to reuse  more code when possible.  In this combined code, I've also added a check that there is a proper project file before running the simulation.  If there is not one, an error message is popped in the try and the simulation is not started.

A final change is in the CLI handling of dissolve-gui.cpp, fixing a regression that I mentioned in the stand-up.  Previously, it was possible to start the GUI with no file given on the command line.  This was because there was a check if a file name had been given before running `dissolveWindow.openLocalFile`.  When switching to the new CLI, this check was removed and `openLocalFile` was called directly on the file name.  Within `openLocalFile`, there is a safety check that confirms that the file isn't empty and bails out of the program if it is.  Therefore, to resume our original behaviour, I've re-implemented the original check.

Two possible changes that might be added atop this
1. Display an error dialog if trying to run the simulation with no file, instead of just an error in the message centre.
2. Possible take advantage of the new <filesystem> library for the input file, though this is probably a large change to warrant its own PR.

### DIFF
--- a/src/dissolve-gui.cpp
+++ b/src/dissolve-gui.cpp
@@ -68,7 +68,8 @@ int main(int args, char **argv)
     DissolveWindow dissolveWindow(dissolve);
 
     // If an input file was specified, load it here
-    if (!dissolveWindow.openLocalFile(options.inputFile().value_or(""), options.restartFilename().value_or(""),
+    if (options.inputFile() &&
+        !dissolveWindow.openLocalFile(options.inputFile().value_or(""), options.restartFilename().value_or(""),
                                       options.ignoreRestartFile(), options.ignoreStateFile()))
     {
         ProcessPool::finalise();
@@ -76,7 +77,7 @@ int main(int args, char **argv)
     }
 
     // Iterate before launching the GUI?
-    if (options.nIterations() > 0)
+    if (options.inputFile() && options.nIterations() > 0)
     {
         // Prepare for run
         if (!dissolve.prepare())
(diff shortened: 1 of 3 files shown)


src/dissolve-gui.cpp
```diff
@@ -68,15 +68,16 @@ int main(int args, char **argv)
...
         ProcessPool::finalise();
         return 1;
     }
 
     // Iterate before launching the GUI?
-    if (options.nIterations() > 0)
+    if (options.inputFile() && options.nIterations() > 0)
     {
         // Prepare for run
         if (!dissolve.prepare())
```
trisyoungs, line 83: This could all be captured in one check of `if (options.inputFile())`, after which `openLocalFIle()` could be called and iterations (if requested) could be run.

trisyoungs (approved):
Nice. Only one small comment to consider.

trisyoungs:
Raising a `QMessageBox` in addition to the printed message would be a nice addition, especially as it's possible for the user to close the messages panel!

src/dissolve-gui.cpp
rprospero: One check is better than two.  I've made the change.

rprospero:
I've switched over to a warning box.

trisyoungs (approved):
Good to go!

PR #424, opened 2020-11-04
Fix MSVC build

trisyoungs:
This PR addresses the crashes encountered in the Windows version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,7 @@ if(BUILD_ANTLR_RUNTIME)
   if(BUILD_ANTLR_ZIPFILE)
     set(ANTLR_SOURCE_URL "${BUILD_ANTLR_ZIPFILE}")
   else(BUILD_ANTLR_ZIPFILE)
-    set(ANTLR_SOURCE_URL "https://www.antlr.org/download/antlr4-cpp-runtime-4.8-source.zip")
+    set(ANTLR_SOURCE_URL "https://www.antlr.org/download/antlr4-cpp-runtime-4.7.2-source.zip")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ if(BUILD_ANTLR_RUNTIME)
   if(BUILD_ANTLR_ZIPFILE)
     set(ANTLR_SOURCE_URL "${BUILD_ANTLR_ZIPFILE}")
   else(BUILD_ANTLR_ZIPFILE)
-    set(ANTLR_SOURCE_URL "https://www.antlr.org/download/antlr4-cpp-runtime-4.8-source.zip")
+    set(ANTLR_SOURCE_URL "https://www.antlr.org/download/antlr4-cpp-runtime-4.7.2-source.zip")
```
DanNixon, line 64: Was this required in the end?
I can't recall if this was part of the issue or something attempted to try and fix the issue.

trisyoungs: I didn't mention this before, but at some point during all this I realised that there was a mismatch of versions between the Java tool (from choco, 4.7.2) and the runtime source being downloaded during the build (v4.8). I'm not sure what was going on in the pipeline, but the step responsible for downloading the 4.8 jar from the ANTLR4 website simply refused to retrieve it (using `System.Net.WebClient`) without raising any error.  After an hour of trying to get around that, I just reverted to 4.7.2 to match the choco tool version.

DanNixon: Cool. my only concern was keeping things consistent across platforms, although I'd like to think that minor version differences wouldn't affect grammar syntax or parsing.

DanNixon (approved):

PR #426, opened 2020-11-06
Bugfix 06/11/20 - Fix CLI options

trisyoungs:
A quick bugfix addressing compilation issues on SCARF, related to deduction of lambda function arguments to CLI11 routines, possibly related to GCC v8.


rprospero (approved):
That is much cleaner

DanNixon (approved):

PR #428, opened 2020-11-06
Bugfix 06/11/20 Correctly handle empty NETA Rings

trisyoungs:
Bugfix PR to address an issue with NETA ring matching, where not specifying the complete set of atoms within a ring (as is the case for, e.g., benzene in the OPLSAA/Aromatics forcefield) would result in a failure to match the ring.

This has been patched, and the scoring of rings made more consistent.


rprospero (approved):
Looks good

PR #429, opened 2020-11-06
Update continuous pipeline

trisyoungs:
Quick fix to explicitly include triggering paths / files for the continuous build, rather than exclude an incomplete set.


DanNixon (approved):
:+1:

PR #433, opened 2020-11-06
Bugfix 06/11/20 Fix expanding GUI

### LINKED ISSUE #415: GUI extend infinitely off the screen
**Describe the bug**
Window of the GUI extends itself infinitely off the screen and cannot be used anymore. The display then hangs and must be killed in the terminal

image

**Screenshots**

**Operating System:**
 - OS:  Linux  (delete as appropriate)
 - OS version / flavour: Ubuntu 18.04

### LINKED ISSUE #425: GUI crashes when zooming in a graph
**Describe the bug**
While zooming in the energy graph, the GUI crashes with the following error message:

image

**Operating System:**
 - OS:   Linux
 - OS version / flavour: Ubuntu 18.04

trisyoungs:
Another bugfix PR, this time to deal with the expanding GUI issue. What may have been a complicated fix actually turned out to be rather trivial: an overload for QString::arg(double,double) was assumed, but did not exist. The second argument is in fact the field width, and so on some graphs where the y-axis is exponentially large, this causes the observed expansion as the QLabel displaying the current coordinate under the mouse becomes absurdly wide.

The crash on zooming the graph also seems to be fixed with these changes.

Closes #415.
Closes #425.


rprospero (approved):
Looks good.  Shame about that `arg` function, but there doesn't seem to be a good way out of it.  Unless c++ wants to add typeclasses...

PR #440, opened 2020-11-12
Remove unneeded parent pointers from ForcefieldAtomType

### LINKED ISSUE #347: Remove unused Forcefield *parent arg from ForcefieldAtomType ctors
This is another part of the Species class clean-up.

rprospero:
This closes #347


src/data/ff.cpp
```diff
@@ -59,14 +59,18 @@ EnumOptions<Forcefield::ShortRangeType> Forcefield::shortRangeTypes()
...
 // Add new atom type referencing existing parameters by name
 void Forcefield::addAtomType(int Z, int index, std::string_view name, std::string_view netaDefinition,
                              std::string_view description, double q, std::string_view parameterReference)
 {
-    atomTypes_.emplace_back(this, Z, index, name, netaDefinition, description, q, parameterReference);
+    OptionalReferenceWrapper<const ForcefieldParameters> parameterReference_ = shortRangeParameters(parameterReference);
+    if (!parameterReference_)
+        Messenger::error("Reference parameters named '{}' are not defined in the forcefield '{}'.\n", parameterReference,
+                         this->name());
+    atomTypes_.emplace_back(parameterReference_, Z, index, name, netaDefinition, description, q);
```
trisyoungs, line 73: This is a very sensible reorganisation of the pass/check of the parameter reference.  It also highlights a failing of the `addAtomType()` methods in that the only way they can currently raise an error is by throwing an exception. This, however, isn't ideal as a future feature of the code should be to allow reading in forcefield files at runtime, and checking them for errors in a more user-friendly manner than exceptions provide. Anyway, perhaps one to deal with in a future PR.

trisyoungs (approved):

PR #443, opened 2020-11-13
Bugfix 13/11/20 Recognise lowercase elements

trisyoungs:
A tiny bugfix PR to address an issue where element symbols written entirely in lowercase would not be recognised as valid elements.


rprospero (approved):
In theory, we could normalize the strings so that we could just go a straight comparison, but that's a pretty major architecture change, so this is probably the better answer for now.

PR #447, opened 2020-11-17
Use ANTLR4 for expression parsing (1/2)

trisyoungs:
This PR adds an ANTLR4-based expression parser to replace the existing Bison implementation. As a by-product, a basic unit-testing framework has also been put in place.

The ANTLR4 implementation exists alongside the old Bison grammar in this PR, with no attempt made to incorporate it into the rest of the codebase. This work is performed in a separate PR to follow once this has been approved and merged.

### DIFF
--- a/src/expression/ExpressionParser.g4
+++ b/src/expression/ExpressionParser.g4
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2020 Team Dissolve and contributors
+
+parser grammar ExpressionParser;
+
+options {
+	tokenVocab = ExpressionLexer;
+}
+
+// Parser file header
+@parser::header {/* Expression ANTLR Parser */}
+
+// Add custom includes after standard ANTLR includes in both *.h and *.cpp files
+@parser::postinclude { }
+
+// Appears in the private part of the parser in the h file.
+@parser::members { /* public parser declarations/members section */ }
+
+// Appears in the public part of the parser in the h file.
+@parser::declarations {/* private parser declarations section */}
+
+// Appears in line with the other class member definitions in the cpp file.
+@parser::definitions {/* parser definitions section */}
+
+/*
+ * Expression Grammar
+ */
+
+// Compound expression
+expression: expr EOF;
+
+// Expressions
+expr: OpenParenthesis expr CloseParenthesis                     #parentheses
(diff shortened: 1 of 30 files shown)


src/expression/ExpressionVisitor.cpp
```diff
@@ -0,0 +1,193 @@
...
+    currentContext()->addChild(node);
+
+    contextStack_.push_back(node);
+
+    auto result = visitChildren(ctx);
+
+    contextStack_.pop_back();
+
+    return result;
+}
```
rprospero, line 88: It's largely a mater of style, but I would invert the logic here.

```suggestion
antlrcpp::Any ExpressionVisitor::visitPower(ExpressionParser::PowerContext *ctx)
{
    return ExpressionBinaryOperatorNode(visit(ctx->left(), ctx->right());
}
```

This would, of course, involve labelling the left and right terms in the grammar file.  This has the advantage of letting the compiler handle the stack for us, plus everything remains immutable values.

src/expression/ExpressionParser.g4
```diff
@@ -0,0 +1,45 @@
...
+// Expressions
+expr: OpenParenthesis expr CloseParenthesis                     #parentheses
+| Subtract expr                                                 #unaryMinus
+| expr Power expr                                               #power
+| expr op=(Multiply | Divide) expr                              #multiplyDivide
+| expr op=(Add | Subtract) expr                                 #addSubtract
+| Name OpenParenthesis (expr (Comma expr)*)? CloseParenthesis   #function
+| Name                                                          #variable
+| value                                                         #number
+;
```
rprospero, line 41:
```suggestion
expr: OpenParenthesis expr CloseParenthesis                     #parentheses
| Subtract expr                                                 #unaryMinus
| left=expr Power right=expr                                               #power
| left=expr op=(Multiply | Divide) left=expr                              #multiplyDivide
| left=expr op=(Add | Subtract) right=expr                                 #addSubtract
| Name OpenParenthesis (expr (Comma expr)*)? CloseParenthesis   #function
| Name                                                          #variable
| value                                                         #number
;
```

Adding the names in makes it a lot easier to use the condensed parser I describe below.

rprospero (changes requested):
We can talk about this a bit more during the stand-up

rprospero (approved):

PR #448, opened 2020-11-17
Use ANTLR4 for expression parsing (2/2)

trisyoungs:
This PR represents the second part of the ANTLR4 expression parser work, and fully incorporates the new implementation into the rest of the code, resulting in no further dependence on Bison.

The majority of work focussed on replacing usage of `RefList<ExpressionVariable>` with `std::vector<std::shared_ptr<ExpressionVariable>>`, which touches several other sections of the code (most notably the `Procedure` class and its associated nodes).

**Do not merge until part 1 of this work (#447) has been approved and merged.**


src/expression/binary.cpp
```diff
@@ -0,0 +1,62 @@
...
+        case (OperatorPow):
+            result = pow(lhs.asDouble(), rhs.asDouble());
+            break;
+        case (OperatorMultiply):
+            if (ExpressionValue::bothIntegers(lhs, rhs))
+                result = lhs.asInteger() * rhs.asInteger();
+            else
+                result = lhs.asDouble() * rhs.asDouble();
+            break;
+    }
```
rprospero, line 59:
```suggestion
    typeof(std::plus) op;
    switch (operator_)
    {
        case (OperatorAdd):  op = std::plus; break;
        case (OperatorSubtract):  op = std::minus; break;
        case (OperatorDivide):op = std::divides; break
        case (OperatorPow):  op = pow; break;
        case (OperatorMultiply):  op = std::multiply;
    }
    
    if (ExpressionValue::bothIntegers(lhs, rhs))
        result = op(lhs.asInteger(), rhs.asInteger());
    else
        result = op(lhs.asDouble(), rhs.asDouble());
```

I /believe/ we can do this.  It'll also require including <functional> at the beginning.

src/expression/unary.cpp
```diff
@@ -0,0 +1,38 @@
...
+    auto rhs = (*optl);
+    switch (operator_)
+    {
+        case (OperatorNegate):
+            if (rhs.isInteger())
+                result = -rhs.asInteger();
+            else
+                result = -rhs.asDouble();
+            break;
+    }
```
rprospero, line 35: Do we need the case statement if there's only one possible operator?

rprospero (approved):
Just some stylistic suggestions, but otherwise looking pretty good.  Always nice to get rid of code.

src/expression/binary.cpp
trisyoungs: Nice idea, however, it would appear that these are templated functions, and so need the argument types in the declaration, which kinda defeats the purpose!

src/expression/unary.cpp
trisyoungs: Strictly no, but makes sense for future expansion with other unary operators.

PR #449, opened 2020-11-18
Data1d array

rprospero:
This changes the data storage in the Data1D class from the custom Array class to `std::vector`.  This has significant propagation throughout the code, but also enables quite a bit of algorithm usage which could eventually be parallelized.


src/genericitems/std_vector.h
```diff
@@ -0,0 +1,112 @@
...
+    // Read data through specified parser
+    bool read(LineParser &parser, CoreData &coreData)
+    {
+        if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+            return false;
+        int nItems = parser.argi(0);
+        data_.clear();
+        data_.resize(nItems);
+
+        for (auto n : data_)
```
trisyoungs, line 67: Shouldn't this be `for (auto &n : data_)` as we're assigning to `n`?

trisyoungs (commented):
Had a look through this as it stands - it's a monster piece of work!  All looks good though - it cleans up a lot of little things in the code. Still prefer the beauty and simplicity of `potCoeff = 0.0;` as opposed to `std::fill(potCoeff.begin(), potCoeff.end(), 0.0);` (for instance) but I'll get over it!

Made one important comment regarding the variable in a `for` loop in `genericitems/std_vector.h` - didn't check for (or perhaps notice) other similar potential issues in the rest of the files...

DanNixon (approved):
:+1:

PR #459, opened 2020-12-01
Move to using CTest as the test runner

rprospero:
This PR adds testing support to our CMake files.  All of the current system tests have been added into CTest.  This gives us the following advantages:

1) We no longer have to write and support our own test runner
2) Tests are run in parallel.  On my system, this has shown an almost linear speed-up with the number of cores
3) The test runner is cross-platform (our old setup only ran on Unix)
4) This will be needed when we later add support for unit tests
5) The test runner always has the correct binary (the path will be set at build time)

The following are the basic command to be run in the build directory.

* `ctest`  run the tests
* `ctest -j4` run the tests, performing four tests in parallel
* `ctest -L broadening` run only the broadening tests


.azure-pipelines/templates/system-tests.yml
```diff
@@ -18,21 +18,21 @@ jobs:
           artifactName: 'linux-tests-serial'
         displayName: 'Download Serial Test Artifacts'
       - bash: |
-          cd $(System.ArtifactsDirectory)/linux-tests-serial
-          chmod u+x ./go ./dissolve
-          ./go -a -s ./dissolve
+          cd $(System.ArtifactsDirectory)/linux-tests-serial/build
+          chmod +x bin/dissolve
+          ctest -j2 --output-on-failure
```
trisyoungs, line 23: Is there an option for `ctest` to deduce a suitable number of cores to run on, or does this have to be hardcoded?  Of course we know what resources the build runners have, so it isn't an issue.

.azure-pipelines/templates/system-tests.yml
```diff
@@ -52,18 +52,18 @@ jobs:
...
+      #     mkdir test-output
+      #     cp -rv $(System.ArtifactsDirectory)/linux-tests-parallel/*.out* test-output/
+      #   displayName: 'Copy Test Output'
+      #   condition: always()
+      # - task: PublishBuildArtifacts@1
+      #   inputs:
+      #     PathtoPublish: "test-output/"
+      #     ArtifactName: 'linux-test-output-parallel'
+      #   displayName: 'Publish Parallel Test Output Artifacts'
+      #   condition: always()
```
trisyoungs, line 69: Remove if no longer required!

.azure-pipelines/templates/system-tests.yml
```diff
@@ -18,21 +18,21 @@ jobs:
...
+      #     mkdir test-output
+      #     cp -rv $(System.ArtifactsDirectory)/linux-tests-serial/*.out test-output/
+      #   displayName: 'Copy Test Output'
+      #   condition: always()
+      # - task: PublishBuildArtifacts@1
+      #   inputs:
+      #     PathtoPublish: "test-output/"
+      #     ArtifactName: 'linux-test-output-serial'
+      #   displayName: 'Publish Serial Test Output Artifacts'
+      #   condition: always()
```
trisyoungs, line 35: Remove if no longer required!

tests/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+# Testing Code
+enable_testing()
+
+function(dissolve_system_test file_name count directory)
```
trisyoungs, line 4: Part of me wants to argue for the `directory` argument to be first in the list. To me, this has a greater "precedence" than the filename to run.

trisyoungs (approved):
Looking good, and extremely positive to have the tests running over all systems.

.azure-pipelines/templates/system-tests.yml
rprospero: Unfortunately, from everything I read, ctest can't figure out the number of available processors and you have to tell it how many parallel jobs to run.

.azure-pipelines/templates/system-tests.yml
rprospero: gone

.azure-pipelines/templates/system-tests.yml
rprospero: gone

tests/md/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+dissolve_system_test(benzene 2 md --restart reference.config)
```
trisyoungs, line 1: Just noticed this after looking through the build results - running for 2 steps as written here should result in a fail of the test, as the forces are not consistent with those supplied as the reference. Also, I notice that this test takes three minutes in the CI, when it has only taken seconds before. Is this just the CI machine being a factor of over 100 slower than my machine?

tests/CMakeLists.txt
rprospero: My original idea was that the test would figure out the directory on its own and that the name was the most important part.  However, as long as the parameter is in there, it should be first.  I've made the adjustment.

tests/md/CMakeLists.txt
rprospero: Something odd is going on with the md test.  I've been having trouble with it in one of my other branches.  I may need to investigate this further.

rprospero: For the record, all for MPI runs took a total of six seconds on my machine, so some of this may be the CI?

trisyoungs: Is there an easy way to capture the outputs of the tests and store them as a build artifact?

tests/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+# Testing Code
+enable_testing()
+
+function(dissolve_system_test directory file_name count)
+  if(PARALLEL)
+    foreach(nproc 1 2 3 4)
+      set(test_target_executable ../../build/bin/${target_name})
+      add_test(
+	NAME ${directory}-${file_name}-${nproc}
```
DanNixon, line 9: This formatting looks rather odd, you may need to update the cmake-format config.

rprospero: That's my own fault for trusting that cmake-language-server would give me the same formatting as cmake-format.  I've fix the whitespace.

CMakeLists.txt
```diff
@@ -32,22 +32,22 @@ conan_basic_setup(CMAKE_TARGETS)
...
 endif(PARALLEL)
 
 if(CHECKS)
-  add_definitions("-DCHECKS")
+  ADD_DEFINITIONS("-DCHECKS")
 endif(CHECKS)
 
 # Add local Modules dir to cmake search path
-list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules")
+LIST(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/Modules")
```
DanNixon: Why do we have this mixture of uppercase and lowercase CMake? Not that is makes any difference or needs to be fixed here, but when it is being changed why not settle on one format?

DanNixon (approved):

PR #463, opened 2020-12-03
Array2d overload

rprospero:
This PR overloads the [] operator to allow Array2D to have a slightly more intuitive accessor.

This builds on PR #458 and should not be merged before that PR.


trisyoungs (approved):
On balance I think these changes make a lot of sense. Contextually, the new `[{...}]` notation for multi-dimensional array access is more readable than the existing `at()` (and `constAt()`) accessors.

One question on efficiency - none of the changes made here impact what could be considered frequently-called, "critical" functions, but is there significant overhead imposed by passing a `tuple<int,int>` as opposed to plain `int`s? Not suggesting that extensive benchmarks need to be run - just thinking of possible downsides.

rprospero:
In theory, there doesn't need to be any overhead.  As it happens, clang will compile this down to identical code to the other.  Unfortunately, the GCC implementation of std::tuple has a slight issue that won't allow it to compile down identically and they cannot fix the issue without breaking the ABI.  However, even in this instance, the only overhead is that the tuple has to be passed on the stack and cannot be passed inside processor registers.  There's no other overhead over passing any other variable on the stack.

PR #464, opened 2020-12-03
Move Array3D onto a proper vector

rprospero:
This PR removes the raw array from array3D and uses a std::vector as the base instead.


src/math/sampleddouble.cpp
```diff
@@ -47,6 +47,7 @@ double SampledDouble::stDev() const { return (count_ < 2 ? 0.0 : sqrt(m2_ / (cou
 
 // Conversion (double)
```
trisyoungs, line 48: We can remove these "obvious" function description comments - must have been missed in the original sweep performed some time back.

trisyoungs (approved):
:+1: Particularly like the `iota`/`transform` combo!

PR #467, opened 2020-12-07
Bugfix 07/12/20 Register new GenericItems

trisyoungs:
This bugfix PR addresses an issue with the new `GenericItem` containers based on `std::vector<double>`, which were not registered in `src/main/dissolve.cpp` and so could not be read in from the restart file.


trisyoungs:
Closed as changes have been integrated into larger patch.

PR #470, opened 2020-12-08
Overload indexing Array3D

rprospero:
Just as we did with Array2D, I've overloaded the [] operator to allow accessing via tuples.  I've also gone through and, where possible, moved nested loops into a loop or transform operation.

Some of these remaining loops /could/ be converted to algorithms.  However, as a matter of style, I've only used algorithms for loops which could be run in parallel.  Loops which must be performed sequentially (e.g. because they depend on the start of a parser object) have been left as loops.


trisyoungs (approved):
All looks good!

PR #471, opened 2020-12-09
Move atom from DynamicArray to a vector shared_ptr

rprospero:
It'd be nice to have a vector of values and just use references, but that runs into issues with references being invalidated when the vector is resized.  Before merging, it's be worth considering if it's possible to set the number of atoms in the configuration at load time.  If it is possible, then this PR is /not/ the correct solution.


rprospero:
One little side toy that has sneaked into this PR is the ability to run MPI tests in parallel.  If the -j option is passed to ctest with the number of processors, an attempt will be made to intelligently run the tests in parallel.  For example, if you have four processors, then it should run four one-node tests concurrently or run two two-node tests in parallel, but mostly run the 3-node and 4-node tests individually.

src/classes/energykernel.cpp
```diff
@@ -576,14 +580,14 @@ double EnergyKernel::energy(std::shared_ptr<const Molecule> mol, ProcessPool::Di
 }
 
 // Return molecular correction energy related to intramolecular terms involving supplied atom
-double EnergyKernel::correct(const Atom *i)
+double EnergyKernel::correct(const std::shared_ptr<Atom> i)
 {
     // Loop over atoms in molecule
-    std::vector<Atom *> atoms = i->molecule()->atoms();
+    auto atoms = i->molecule()->atoms();
```
trisyoungs, line 586:
```suggestion
    auto &atoms = i->molecule()->atoms();
```

src/io/export/coordinates.cpp
```diff
@@ -119,7 +116,7 @@ bool CoordinateExportFileFormat::exportDLPOLY(LineParser &parser, Configuration
     // Export Atoms
     for (auto n = 0; n < cfg->nAtoms(); ++n)
```
trisyoungs, line 117: Minor stylistic suggestion - is it nicer to have the loop as `for (auto &i : cfg->atoms())` and use a counter for the index?

src/io/export/trajectory.cpp
```diff
@@ -66,7 +66,7 @@ bool TrajectoryExportFileFormat::exportXYZ(LineParser &parser, Configuration *cf
     // Write Atoms
     for (auto n = 0; n < cfg->nAtoms(); ++n)
     {
-        Atom *i = cfg->atom(n);
+        auto i = cfg->atom(n);
```
trisyoungs, line 69: Same here.

src/modules/rdf/functions.cpp
```diff
@@ -28,21 +28,11 @@ bool RDFModule::calculateGRTestSerial(Configuration *cfg, PartialSet &partialSet
...
-        for (jj = ii + 1; jj < cfg->nAtoms(); ++jj)
-        {
-            distance = box->minimumDistance(rI, atoms[jj]->r());
-            partialSet.fullHistogram(typeI, atoms[jj]->localTypeIndex()).bin(distance);
-        }
-    }
+    for_each_pair(cfg->atoms().begin(), cfg->atoms().end(), [box, &partialSet](auto i, auto ii, auto j, auto jj) {
+        if (ii != jj)
+            partialSet.fullHistogram(ii->localTypeIndex(), jj->localTypeIndex()).bin(box->minimumDistance(ii->r(), jj->r()));
+    });
```
trisyoungs, line 35: Ah, good old `for_each_pair`!

trisyoungs (approved):
Looking good. Another archaic custom class bites the dust! One minor bugfix, and some small style suggestions, but that's all.

PR #476, opened 2020-12-10
Bugfix 10/12/20 EPSRModuleWidget crash

### LINKED ISSUE #473: restart file cannot be openend
when loading a simulation with a restart file, the following error message occurs and the GUI cannot be opened:
image

trisyoungs:
This bugfix PR corrects an issue with `EPSRModuleWidget`, which always assumed that at least one data target was set.  It is entirely valid (if not particularly useful) input to have no targets defined, so this behaviour has been corrected.

Closes #473.


DanNixon (approved):
:+1:

rprospero (approved):

PR #478, opened 2020-12-11
Used standard containers in ChangeStore

rprospero:
Replacing the containers in the ChangeStore with standard containers will be part of whatever path forward we chose for the atoms.

### DIFF
--- a/src/classes/changedata.cpp
+++ b/src/classes/changedata.cpp
@@ -6,7 +6,7 @@
 #include "classes/atom.h"
 #include "classes/cell.h"
 
-ChangeData::ChangeData() : ListItem<ChangeData>()
+ChangeData::ChangeData()
 {
     atom_ = nullptr;
     cell_ = nullptr;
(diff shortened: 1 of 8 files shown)


src/classes/changedata.h
```diff
@@ -11,7 +11,7 @@ class Atom;
 class Cell;
 
 // Change Data
-class ChangeData : public ListItem<ChangeData>
+class ChangeData
 {
     public:
     ChangeData();
```
trisyoungs, line 17: The destructor here can be made `default` as its empty.

src/classes/changedata.cpp
```diff
@@ -6,7 +6,7 @@
 #include "classes/atom.h"
 #include "classes/cell.h"
 
-ChangeData::ChangeData() : ListItem<ChangeData>()
+ChangeData::ChangeData()
```
trisyoungs, line 9: Can use an initialiser list here.

trisyoungs (changes requested):
:+1:

The build failures here are caused by building with `-DCHECKS`, the poor-man's (i.e. my) equivalent of using assertions and `-DDEBUG`.

PR #479, opened 2020-12-14
Use ANTLR4 for TextPrimitve parsing

trisyoungs:
This PR converts the GUI's `TextPrimitive` grammer from Bison to ANTLR, allowing merge of the `antlr-expression-2` branch (PR #448) to completely remove Bison from our dependencies.

The `TextFragment`, `TextFormat`, `TextPrimitive`, and `TextPrimitiveList` classes are updated / modernised to some degree, remove uses of the `List` and `ListItem` classes for storage. Symbol data is scoped to its own namespace.

This PR should be merged **before** #448.  Afterwards, #448 should pass the CI build and be ready for merge as well.


src/gui/render/symbol.cpp
```diff
@@ -2,67 +2,95 @@
...
 {
-    for (auto n = 0; n < nSymbols; ++n)
-        if (name == symbols[n].name)
-            return &symbols[n];
+    for (const auto &sym : symbols())
+        if (std::get<3>(sym).find(text) != std::string::npos)
+            return std::get<0>(sym);
 
-    return nullptr;
+    return nSymbols;
```
rprospero, line 93:
```suggestion
    auto it = std::find_if(symbols().begin(), symbols().end(), [&text](auto & sym){ return std::get<3>(sym).find(text) != std::string::npos; });
    if (it == symbols().end()) return nSymbols;
    return std::get<0>(*it);
```

Using find_if again

rprospero (approved):
Only a couple of algorithm suggestions.  Otherwise, it looks good.

PR #481, opened 2020-12-15
Supply useful data from RDFModule

### LINKED ISSUE #398: Configuration weighting factors aren't always used
**Describe the bug**
Weighting factors can be given to the `Configuration` keyword for all modules, but no module except `RDFModule` actually pays any attention to them. An overhaul of the way modules depend on other modules data is underway in a separate branch - when this is completed, this will remove several problems related to this bug, but others will need to addressed specifically.

Related to #399

### LINKED ISSUE #400: Store species and configuration populations / weightings in class / restart file
Weights associated to Configurations targeted by modules are currently stored in restart file data and nowhere else, making access to them somewhat convoluted (e.g. from modules two degrees of separation away, such as is the case between `RDF` and `NeutronSQ`).

A better approach would be to use a new keyword to store configurations/weights as tuples, and dumping that data into the restart file if it gets modified (e.g. through some as-yet-unimplemented optimisation process). Related to generalities regarding module targets (#399).

A similar class for species weights (from, e.g., summing and blending configurations) would also make "higher-level" modules easier to write, as they wouldn't have to dig down into underlying modules to re-calculate the necessary information.

### LINKED ISSUE #451: providing an effective atomic density is not possible
image

trisyoungs:
This PR address several small bugs/issues in the code, and makes an effort to tidy up a few bits along the way. Most of these issues were highlighted by recent work that simplified the interaction between the correlation modules (#402).

### Overview
All the work centres around the calculation/storage of data by the `RDFModule`, and which is referenced by other modules when doing their own calculations. In particular:

- Effective densities from multiple configurations being specified in `RDFModule`.
- Species populations over multiple configurations.

### Resolution

Storage of trivially-calculable properties in the restart file is not performed, safeguarding the accuracy of those values when Configuration contents become dynamic. Functions to calculate and return necessary data have been added to `RDFModule`.

References to ConfigurationWeights values stored in processing data have been removed, as they were defunct. Proper weighting of configurations will be addressed in a future PR once that functionality is required.

### Related Issues

Closes #451.
Closes #398.
Closes #400.

### DIFF
--- a/src/modules/rdf/functions.cpp
+++ b/src/modules/rdf/functions.cpp
@@ -243,6 +243,27 @@ double RDFModule::effectiveDensity() const
     return rho0;
 }
 
+// Calculate and return used species populations based on target Configurations
+std::map<const Species *, double> RDFModule::speciesPopulations() const
+{
+    std::map<const Species *, double> populations;
+
+    for (auto *cfg : targetConfigurations())
+    {
+        // TODO Get weight for configuration
+        auto weight = 1.0;
+
+        ListIterator<SpeciesInfo> spInfoIterator(cfg->usedSpecies());
+        while (auto *spInfo = spInfoIterator.iterate())
+        {
+            populations.try_emplace(spInfo->species(), 0.0);
+            populations[spInfo->species()] += spInfo->population() * weight;
+        }
+    }
+
+    return populations;
+}
+
 // Calculate unweighted partials for the specified Configuration
 bool RDFModule::calculateGR(ProcessPool &procPool, Configuration *cfg, RDFModule::PartialsMethod method, const double rdfRange,
                             const double rdfBinWidth, bool &alreadyUpToDate)
--- a/src/modules/rdf/functions.cpp
+++ b/src/modules/rdf/functions.cpp
@@ -510,10 +510,6 @@ bool RDFModule::sumUnweightedGR(ProcessPool &procPool, Module *parentModule, con
     }
     summedUnweightedGR.setFingerprint(fingerprint);
 
-    // Store the overall density of our partials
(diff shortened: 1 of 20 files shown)


src/modules/rdf/functions.cpp
```diff
@@ -224,6 +224,46 @@ bool RDFModule::calculateGRCells(ProcessPool &procPool, Configuration *cfg, Part
...
+    {
+        // TODO Get weight for configuration
+        auto weight = 1.0;
+
+        totalWeight += weight;
+
+        rho0 += weight / cfg->atomicDensity();
+    }
+    rho0 /= totalWeight;
+    rho0 = 1.0 / rho0;
```
rprospero, line 241:
```suggestion
    auto [totalWeight, rho0] = std::accumulate(targetConfigurations().begin(), targetConfigurations().end(), std::make_tuple(0.0, 0.0), [](auto acc, auto *cfg) {
        // TODO Get weight for configuration
        auto weight = 1.0;
        auto [accWeight, accRho] = acc;
        return std::make_tuple(accWeight + weight, accRho + weight/cfg->atomicDensity());
    });

    rho0 /= totalWeight;
    rho0 = 1.0 / rho0;
```

I *think* this should work.  The tuple is a little unfortunate, but it saves us having to calculate the weight twice (when we start calculating the weight).

rprospero (approved):
I always love a PR that makes the code shorter.  I've suggested a possible accumulate algorithm that may help, but I haven't quite tested it.

src/modules/rdf/functions.cpp
trisyoungs: Happy to try it out - in the interests of getting the backlog of PRs in following the OSX failures, I'll address this in a future bugfix.

PR #500, opened 2020-12-28
Bugfix 23-12-20 Binary path for unit tests

trisyoungs:
Started as a quick bugfix to more reliably set paths to the generated binaries using `CMAKE_BINARY_DIR` rather than a relative path. Works for CI paths (`build/bin`) as well as more exotic paths used by IDEs (e.g. `cmake-build-debug/bin`).

Also addresses tests not being built in the CI, and adds better failure to parts of the system-tests pipeline.

**NOTE - Not all parallel system tests pass in this PR:**
 - energyforce2-full-2 (fixed in pending PR #504)
 - energyforce2-full-4 (fixed in pending PR #504)
 - exchangeable-watermeth-1 (fails - to be addressed)
 - exchangeable-watermeth-2 (fails - to be addressed)
 - exchangeable-watermeth-3 (fails - to be addressed)
 - exchangeable-watermeth-4 (fails - to be addressed)
 - md-benzene-1 (fixed in pending PR #501)
 - md-benzene-2 (fixed in pending PR #501)
 - md-benzene-3 (fixed in pending PR #501)
 - md-benzene-4 (fixed in pending PR #501)

### DIFF
--- a/.azure-pipelines/pr.yml
+++ b/.azure-pipelines/pr.yml
@@ -31,9 +31,6 @@ stages:
               extraflags: '-DCHECKS:bool=true'
               ppa: 'beineri/opt-qt-5.13.2-bionic'
               qtver: 513
-          - bash: |
-              cp build/bin/dissolve tests/
-            displayName: 'Copy Executable'
           - task: PublishBuildArtifacts@1
             inputs:
               PathtoPublish: "."
@@ -48,9 +45,6 @@ stages:
             fetchDepth: 1
           - template: templates/set-short-hash.yml
           - template: templates/build-linux-parallel.yml
-          - bash: |
-              cp build/bin/dissolve-mpi tests/
-            displayName: 'Copy Executable'
           - task: PublishBuildArtifacts@1
             inputs:
               PathtoPublish: "."
--- a/.azure-pipelines/pr.yml
+++ b/.azure-pipelines/pr.yml
@@ -33,7 +33,7 @@ stages:
               qtver: 513
           - task: PublishBuildArtifacts@1
             inputs:
-              PathtoPublish: "."
+              PathtoPublish: "build/"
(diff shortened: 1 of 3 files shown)


rprospero:
I had tried using `CMAKE_BINARY_DIR` during my initial work on putting in `ctest`.  The problem that I ran into is that it used the full path to the binary.  That might not seem like an issue, but when the artifact was copied between the build and testing phases, we moved into a new directory and `ctest` was looking for it in the old directory.  That's why I ultimately settled for a relative path.

If `CMAKE_BINARY_DIR` is working for us here, it's obviously the better route and we should go with it.  However, if the tests start failing, that might be a good place to look.

trisyoungs:
Yes, I remember the commit commentary well! Here, I'm just doing a lighter version of your work, copying over the `build` dir to the checked out source, so all paths (relative or otherwise) should be consistent. It's a little unfortunate that `cmake` doesn't use relative paths, as it makes this part of the CI a little fragile.

.azure-pipelines/pr.yml
```diff
@@ -28,15 +28,12 @@ stages:
           - template: templates/set-short-hash.yml
           - template: templates/build-linux-serial-gui.yml
             parameters:
-              extraflags: '-DCHECKS:bool=true'
+              extraflags: '-DCHECKS:bool=true -DBUILD_TESTS:bool=true'
```
DanNixon, line 31: Not for this PR but is there a reason for the `BUILD_TESTS` option? Could this be done via targets instead?

trisyoungs: Not sure. The `BUILD_TESTS` option does a couple of things, adding the system and unit test dirs to the configuration, and calling `ENABLE_TESTING()`.  Don't know if this can nicely be encapsulated in a single target...

DanNixon (approved):

PR #502, opened 2020-12-28
Namespace math-related classes

trisyoungs:
Quick PR to move math-related static functions encapsulated within classes into namespaces.

Closes #274.

**Requires rebase - do not merge until rebased after #501 and #500 are merged.**

PR #503, opened 2021-01-02
Improve error calculation for system tests

trisyoungs:
This PR focuses on implementing a more robust error metric for use in the `DataTest` module, with the intention to more reliably catch failures in the system tests. The new error metric is more sensitive to variations on data of the type frequently tested herein.

During implementation, all test thresholds were tightened as necessary, and all data checked visually for consistency with the reference sets.

Other bugfixes / improvements:
- Formatting of error output has been improved.
- Checks are now made for the calculated error (by any method) returning `nan` (and failing if so).
- Several instances of data saving from modules have been fixed.
- CRITICAL The summation var in the `std::inner_product` calls in `XRayWeights` were initialised to `int (`0`) rather than `double`, resulting in incorrect values.

**Requires rebase - do not merge until rebased after #502 is merged.**

### DIFF
--- a/src/math/error.cpp
+++ b/src/math/error.cpp
@@ -13,9 +13,10 @@ namespace Error
 // Return enum option info for AveragingScheme
 EnumOptions<ErrorType> errorTypes()
 {
-    static EnumOptionsList ErrorTypeOptions =
-        EnumOptionsList() << EnumOption(RMSEError, "RMSE") << EnumOption(MAAPEError, "MAAPE") << EnumOption(MAPEError, "MAPE")
-                          << EnumOption(PercentError, "Percent") << EnumOption(RFactorError, "RFactor");
+    static EnumOptionsList ErrorTypeOptions = EnumOptionsList()
+                                              << EnumOption(RMSEError, "RMSE") << EnumOption(MAAPEError, "MAAPE")
+                                              << EnumOption(MAPEError, "MAPE") << EnumOption(PercentError, "Percent")
+                                              << EnumOption(RFactorError, "RFactor") << EnumOption(EuclideanError, "Euclidean");
 
     static EnumOptions<ErrorType> options("ErrorType", ErrorTypeOptions, PercentError);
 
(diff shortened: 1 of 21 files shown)


src/modules/sq/process.cpp
```diff
@@ -228,7 +228,7 @@ bool SQModule::process(Dissolve &dissolve, ProcessPool &procPool)
                                             dissolve.processingModuleData().version("UnweightedGR", rdfModule->uniqueName()),
                                             includeBragg ? dissolve.processingModuleData().version("BraggReflections") : -1));
     // Save data if requested
-    if (saveData && configurationLocal_ && (!MPIRunMaster(procPool, unweightedsq.save())))
+    if (saveData && (!MPIRunMaster(procPool, unweightedsq.save())))
         return false;
```
rprospero, line 232:
```suggestion
    if (saveData && !MPIRunMaster(procPool, unweightedsq.save()))
        return false;
```

Remove extra parens.

src/math/error.cpp
```diff
@@ -292,4 +295,53 @@ double rFactor(const Data1D &A, const Data1D &B, bool quiet)
...
+
+        a = aY[n];
+
+        delta = a - interpolatedB.y(x);
+        sos += delta * delta;
+        a2 += a * a;
+
+        lastX = x;
+        ++nPointsConsidered;
+    }
```
rprospero, line 336: So I re-wrote this for-loop to use `std::inner_product`.  The nice part was that it could then be parallelised in the future.  The only downside was that it was completely incomprehensible.  Since I doubt that this is going to be our bottleneck, I'll leave it as it is.

The one addendum to that this does call out for the `zip` utility function.  The for loop could then look something more like

>    for(auto &[x, a] : zip(aX, aY))

It remove the unnecessary index and makes the connection between the values clearer.  We could either pull `zip` from a dependency (i.e. Boost header only libraries) or we could just add our own version in with `templates/algorithms.h`.  Either way, it could help clean up several `for` loops and algorithms in the code.

rprospero (approved):
Look pretty good. Just one minor formatting thing and one self-indulgent whinge on my part.

src/math/error.cpp
trisyoungs: Something like `zip` looks like it could be quite useful, as you say. Of course, being me I would settle for just reimplementing a small utility function such as this, rather than relying on Boost. Even if it only requires a single header dependency to be installed, it still seems like overkill for a small function like this.

PR #504, opened 2021-01-03
Bugfix 03-01-2021 Excluded atoms in energy/force calculations

trisyoungs:
This bugfix PR addresses an issue that has arisen during recent work, potentially during the move from `DynamicArray<Atom>` to `std::vector<std::shared_ptr<Atom>>`. Previously, some routines in `EnergyKernel` and `ForceKernel` compared the pointers of atoms in order to exclude pairs where atom `i` was "greater than" `j`. In the `DynamicArray` incarnation, objects were allocated in chunks and so were more likely (but not guaranteed) to allow this check to be valid. Using `std::shared_ptr`s, ordering by memory address for a sequence of created atoms is not guaranteed, especially across MPI processes which must all have the same ordering of atoms.

In the current PR all comparisons by pointer are removed and replaced with references to the `arrayIndex()` instead.

**Requires rebase - keep in draft until rebase following merge of #503.**


trisyoungs:
Incoporated into #501.

PR #511, opened 2021-01-06
Clean-up CMake

### LINKED ISSUE #491: Simplify library linkage
Currently, the main `CMakeLists.txt` list contains separate lists for the core module libraries, the `nogui` libs (if present), and the `gui` libs (if present). A more maintainable and localised way of incorporating the various libraries associate to a given module would be for the individual module `CMakeLists.txt` to append the relevant libs to global variables that are then used in the relevant link lines.

In a similar manner, forcefield libs could be incorporated in the same way.

trisyoungs:
This PR aims to simplify somewhat the build system, in particular in respect of including the various modules in the build. This previously involved a lot of repetition of common CMake, which have now been encapsulated within macros. In a similar manner, forcefields have undergone the same treatment. The benefit is that it is easier to maintain / add new modules and forcefields in the build system, and the lists of link libraries are automatically created for the top-level targets.

Code quality checks have been adjusted to actually fail on a cmake-format error, and to dump output in the same manner as clang-format.

For review, one may wish to focus on 27462b6 and b8f6efe, and take a cursory glance at the effects in f415ef1.

Closes #491.


DanNixon (approved):
:+1: 

I'm on the fence about dynamically finding source files. I believe that causes issues with some IDE integrations (but I wouldn't really know about that).

trisyoungs:
Me too.  I certainly wouldn't use this kind of approach for the entire build, but for the more frequently-extended categories of modules and forcefields it makes a little more sense, and removes most of the annoying integration issues.

PR #514, opened 2021-01-12
Zip iterator

rprospero:
This adds a zip iterator for looping over multiple lists simultaneously.  There's several example of its usage added (though not even close to every possible instance).  However, the general idea is to replace

```
for(int n=0; n<xs.size(); ++n) {
    auto x = xs[n];
    auto y = ys[n];
```

with

```
for(auto&& [x, y] : zip(xs, ys)) {
```


src/io/export/data1d.cpp
```diff
@@ -56,14 +57,13 @@ bool Data1DExportFileFormat::exportXY(LineParser &parser, const Data1D &data)
     if (data.valuesHaveErrors())
     {
         const auto &errors = data.errors();
-        // When we get Ranges, we can refactor this code
-        for (auto n = 0; n < x.size(); ++n)
-            if (!parser.writeLineF("{:16.10e}  {:16.10e}  {:16.10e}\n", x[n], values[n], errors[n]))
+        for (auto &&[n, value, error] : zip(x, values, errors))
+            if (!parser.writeLineF("{:16.10e}  {:16.10e}  {:16.10e}\n", n, value, error))
```
trisyoungs, line 61: From a style / readability perspective, I would rather grab the x-axis vector as `xAxis` rather than (my old) `x`, and then use `x` as the loop variable rather than `n`.  `n` conveys a certain `int`-like quality to me (probably thanks to Fortran...)

src/io/export/data1d.cpp
```diff
@@ -56,14 +57,13 @@ bool Data1DExportFileFormat::exportXY(LineParser &parser, const Data1D &data)
...
                 return false;
     }
     else
-        for (auto n = 0; n < x.size(); ++n)
-            if (!parser.writeLineF("{:16.10e}  {:16.10e}\n", x[n], values[n]))
+        for (auto &&[n, value] : zip(x, values))
+            if (!parser.writeLineF("{:16.10e}  {:16.10e}\n", n, value))
```
trisyoungs, line 66: And here.

src/math/data1d.cpp
```diff
@@ -512,13 +513,13 @@ bool Data1D::write(LineParser &parser)
...
-            if (!parser.writeLineF("{}  {}  {}\n", x_[n], values_[n], errors_[n]))
+        for (auto &&[n, value, error] : zip(x_, values_, errors_))
+            if (!parser.writeLineF("{}  {}  {}\n", n, value, error))
                 return false;
     }
     else
-        for (auto n = 0; n < x_.size(); ++n)
-            if (!parser.writeLineF("{}  {}\n", x_[n], values_[n]))
+        for (auto &&[n, value] : zip(x_, values_))
+            if (!parser.writeLineF("{}  {}\n", n, value))
```
trisyoungs, line 522: Similarly, `n` could be `x` here.

src/modules/epsr/process.cpp
```diff
@@ -572,11 +572,11 @@ bool EPSRModule::process(Dissolve &dissolve, ProcessPool &procPool)
     {
         if (procPool.isMaster())
         {
-            for (auto n = 0; n < estimatedSQ.size(); ++n)
+            for (auto sq : estimatedSQ)
```
trisyoungs, line 575:
```suggestion
            for (auto &sq : estimatedSQ)
```

src/templates/algorithms.h
```diff
@@ -94,3 +96,52 @@ template <class Lam> auto for_each_pair_early(int begin, int end, Lam lambda) ->
...
+    {
+        return ZipIterator(std::apply([](auto &... item) { return std::make_tuple(item.begin()...); }, sources));
+    }
+    auto end()
+    {
+        return ZipIterator(std::apply([](auto &... item) { return std::make_tuple(item.end()...); }, sources));
+    }
+
+    private:
+    std::tuple<Args &...> sources;
```
trisyoungs, line 146:
```suggestion
    std::tuple<Args &...> sources_;
```
For consistency with member visibility elsewhere.

src/modules/epsr/process.cpp
```diff
@@ -423,8 +423,8 @@ bool EPSRModule::process(Dissolve &dissolve, ProcessPool &procPool)
                 // Remove square of average normalisation, and apply average of squares
                 auto bbarOld = weights.boundCoherentSquareOfAverage(normalisedRef.xAxis());
                 auto bbarNew = weights.boundCoherentAverageOfSquares(normalisedRef.xAxis());
-                for (auto n = 0; n < bbarOld.size(); ++n)
-                    normalisedRef.value(n) *= bbarOld[n] / bbarNew[n];
+                for (auto &&[val, old, nw] : zip(normalisedRef.values(), bbarOld, bbarNew))
+                    val *= old / nw;
```
trisyoungs, line 427: Presumably the arguments in the tuple are all extracted from the source vectors by reference, so this kind of in-place modification will work as expected?

trisyoungs (changes requested):
This is good - makes a lot of sense seeing it used in context in the code. Some minor comments in the review, and I can't fully unpick the mechanism of the `operator!=` in `ZipIterator`, but I trust that it's doing the right thing!

src/modules/epsr/process.cpp
rprospero: It does allow for in place updates.  That's one of the things I made sure to check when I wrote the unit tests.

src/io/export/data1d.cpp
rprospero: `x` is obviously the better name there.

trisyoungs (approved):

PR #515, opened 2021-01-13
Bugfix 13-01-21 Parallel ifdef and tidying

trisyoungs:
Small PR to adress a possible issue causing an unpredictable failure when sanity checking data across processes (even in, or particularly for, serial builds) and a bit of floor sweeping.


rprospero (approved):
Good catch.  Thanks for not pointing out that the bugs were mine.

PR #516, opened 2021-01-13
Protein force fields

### LINKED ISSUE #493: Add protein forcefield

rprospero:
This would close #493, when it was completed.  As it stands, it's not near ready.  However, I wanted to put something up to make it possible to discuss the existing problems.


rprospero:
As a discussion of what's been done, the extra atom types were added to `atomtypes.cpp` file from the toplogy CHAARM file.  The element and name were pretty simple to include.  The description was taken from inline comments.  The three numerical parameters were set to the atomic mass and two zeros.

In protein.cpp, I've included the bond terms and angle terms from the parameter file, with the numeric parameters copies as is.

Questions:
1) Is any of this even on the right track?
2) The `OT` and `HT` types had no name.  I named them via analogy with `OW` and `HW`, but is there a better name for them?
3) The numerical parameters on the atom types are obviously wrong.  What sorts of values belong there and where I should go about locating those?
4) The bond and angle terms had their numerical values included as is.  However, I saw a note in the definition of the parameter file format that the bond terms are given in kCal and Angstroms.  What units should we be using?

trisyoungs:
First off, the implementation of this forcefield is using the original OPLS/AA as-implemented as a base, which I think is probably the best way forward. Assuming that only torsional terms were changed / added in OPLS/AA/M then the following *should* be true:

1) Atom type names in the CHARMM topology file are of the type `C135`, for instance, which should correspond to the existing type with id `135` in the OPLS/AA forcefield. While this does present an annoying renaming issue within the torsion terms (which are the only new intramolecular terms required) it does mean that no new atomtypes are required. The renaming itself is probably easiest by hacking some temporary code within Dissolve to read the new torsion terms (which use the new "numerical" atom type names), grab the type index from the name, and spit out the correct type name in its place.
2) The intramolecular bond and angle terms in OPLS/AA/M are identical to those in OPLS/AA, and so the `getBondTerm` and `getAngleTerm` functions in `OPLSAA2005BaseForcefield` can be used as-is.
3) The `getTorsionTerm` function will have to be overridden in order to provide the new values.
4) The biggest job is getting the automatic type detection correct, but the number of atom types to consider in the new forcefield is limited, and we can use the fact that they are only expected to occur in the defined residues as listed in the topology file to make life easier. Very happy to take this on as a joint piece of work!
5) Dissolve uses units of kJ and Angstrom, so anything in kCal will have to be converted to kJ (*4.184).

PR #520, opened 2021-01-23
Parameter Storage Modernisation

trisyoungs:
This PR modernises parts of the code relating to the storage of short-range interaction potentials parameters.

Summary of changes:
- Raw arrays of fixed size are removed in favour of `std::vector`s (removes `MAXSRPARAMETERS` define)
- Base Forcefield classes have been moved to `data/ff/`, and some CMake tidying performed.
- Simplify linking of `ForcefieldAtomType`s to common parameters, removing the `ForcefieldParameters` class and replacing it with a `std:vector` of `std::pair<std::string,std::vector<double>>`.
- Remove `InteractionParameters` class (which just contained a `double` and `double[]` for the charge and short-range parameters respectively) and use explicit members for storage in other classes.
- Atomtype parameters now always overwrite any existing values when assigning types from a forcefield.
- BUGFIX: Return total species charge based on assigned atom types by request.
- BUGFIX: Don't set charge on SpeciesAtom when assigning it a forcefield atomtype.

Notes:
- Removes `ForcefieldParameters` from #257.


src/data/ff/atomtype.h
```diff
@@ -3,25 +3,19 @@
...
 class ForcefieldAtomType
 {
     public:
     ForcefieldAtomType(Elements::Element Z = Elements::Unknown, int index = -1, std::string_view name = "",
                        std::string_view netaDefinition = "", std::string_view description = "", double q = 0.0,
-                       double data0 = 0.0, double data1 = 0.0, double data2 = 0.0, double data3 = 0.0);
-    ForcefieldAtomType(OptionalReferenceWrapper<const ForcefieldParameters> params, Elements::Element Z = Elements::Unknown,
-                       int index = -1, std::string_view name = "", std::string_view netaDefinition = "",
-                       std::string_view description = nullptr, double q = 0.0);
+                       const std::vector<double> parameters = {}, std::string_view equivalentName = "");
```
rprospero, line 18:
```suggestion
                       const std::vector<double> &parameters = {}, std::string_view equivalentName = "");
```

Passing by reference to prevent copying.  Default values for const reference parameters should be allowed.

src/data/ff/atomtype.cpp
```diff
@@ -1,33 +1,16 @@
...
-#include "data/ffatomtype.h"
-#include "data/ff.h"
-#include "data/ffparameters.h"
+#include "data/ff/atomtype.h"
+#include "data/ff/ff.h"
 
 ForcefieldAtomType::ForcefieldAtomType(Elements::Element Z, int index, std::string_view name, std::string_view netaDefinition,
-                                       std::string_view description, double q, double data0, double data1, double data2,
-                                       double data3)
+                                       std::string_view description, double q, std::vector<double> parameters,
```
rprospero, line 8:
```suggestion
                                       std::string_view description, double q, std::vector<double> &parameters,
```
Pass by reference

src/data/ff/ff.cpp
```diff
@@ -57,20 +56,20 @@ EnumOptions<Forcefield::ShortRangeType> Forcefield::shortRangeTypes()
 
 // Add new atom type with its own parameters
 void Forcefield::addAtomType(Elements::Element Z, int index, std::string_view name, std::string_view netaDefinition,
-                             std::string_view description, double q, double data0, double data1, double data2, double data3)
+                             std::string_view description, double q, std::vector<double> parameters)
 {
```
rprospero, line 60:
```suggestion
                             std::string_view description, double q, std::vector<double> &parameters)
{
```

Pass by reference.  Granted, a copy ultimately gets made, but at least we won't make two.

src/data/ff/ff.cpp
```diff
@@ -126,9 +125,9 @@ OptionalReferenceWrapper<const ForcefieldAtomType> Forcefield::determineAtomType
 }
 
 // Ass short-range parameters
-void Forcefield::addParameters(std::string_view name, double data0, double data1, double data2, double data3)
+void Forcefield::addParameters(std::string_view name, std::vector<double> parameters)
 {
```
rprospero, line 129:
```suggestion
void Forcefield::addParameters(std::string_view name, std::vector<double> *parameters)
{
```

src/data/ff/ff.cpp
```diff
@@ -150,14 +149,14 @@ bool Forcefield::createNETADefinitions()
 }
 
 // Return named short-range parameters (if they exist)
-const OptionalReferenceWrapper<const ForcefieldParameters> Forcefield::shortRangeParameters(std::string_view name) const
+std::optional<std::vector<double>> Forcefield::shortRangeParameters(std::string_view name) const
```
rprospero, line 152:
```suggestion
const OptionalReferenceWrapper<std::vector<double>> Forcefield::shortRangeParameters(std::string_view name) const
```

With all of the places were I've passed by reference instead of value, I'm not sure if this might not turn into a temporary somewhere and throw a fit because it can't be passed by reference.  This solves that issue, but then places the limit that the parameter vector must be const.  If we're changing that vector somewhere (that I haven't found yet), then we shouldn't do any of the reference changes that I've suggested.

src/gui/forcefieldtab_funcs.cpp
```diff
@@ -344,17 +344,19 @@ void ForcefieldTab::updateAtomTypesTableRow(int row, std::shared_ptr<AtomType> a
...
-    for (auto n = 0; n < MAXSRPARAMETERS; ++n)
+    auto col = 4;
+    for (auto x : atomType->shortRangeParameters())
     {
         if (createItems)
         {
             item = new QTableWidgetItem;
             item->setData(Qt::UserRole, QVariant::fromValue(atomType));
-            ui_.AtomTypesTable->setItem(row, n + 4, item);
+            ui_.AtomTypesTable->setItem(row, col, item);
```
rprospero, line 354:
```suggestion
            ui_.AtomTypesTable->setItem(row, col++, item);
```

src/gui/forcefieldtab_funcs.cpp
```diff
@@ -344,17 +344,19 @@ void ForcefieldTab::updateAtomTypesTableRow(int row, std::shared_ptr<AtomType> a
...
         }
         else
-            item = ui_.AtomTypesTable->item(row, n + 4);
-        item->setText(QString::number(atomType->parameters().parameter(n)));
+            item = ui_.AtomTypesTable->item(row, col);
```
rprospero, line 357:
```suggestion
            item = ui_.AtomTypesTable->item(row, col++);
```

src/gui/forcefieldtab_funcs.cpp
```diff
@@ -344,17 +344,19 @@ void ForcefieldTab::updateAtomTypesTableRow(int row, std::shared_ptr<AtomType> a
...
+        item->setText(QString::number(x));
+        ++col;
```
rprospero, line 359:
```suggestion
```

All three of these would need to be merged together.

src/gui/forcefieldtab_funcs.cpp
```diff
@@ -424,18 +426,20 @@ void ForcefieldTab::updatePairPotentialsTableRow(int row, PairPotential *pairPot
...
+    auto col = 5;
+    for (auto x : pairPotential->parameters())
     {
         if (createItems)
         {
             item = new QTableWidgetItem;
             item->setData(Qt::UserRole, VariantPointer<PairPotential>(pairPotential));
             item->setFlags(Qt::ItemIsSelectable | Qt::ItemIsEnabled);
-            ui_.PairPotentialsTable->setItem(row, n + 5, item);
+            ui_.PairPotentialsTable->setItem(row, col, item);
```
rprospero, line 437:
```suggestion
            ui_.PairPotentialsTable->setItem(row, col++, item);
```

src/gui/forcefieldtab_funcs.cpp
```diff
@@ -424,18 +426,20 @@ void ForcefieldTab::updatePairPotentialsTableRow(int row, PairPotential *pairPot
...
         }
         else
-            item = ui_.PairPotentialsTable->item(row, n + 5);
-        item->setText(QString::number(pairPotential->parameter(n)));
+            item = ui_.PairPotentialsTable->item(row, col);
+        item->setText(QString::number(x));
+        ++col;
```
rprospero, line 442:
```suggestion
            item = ui_.PairPotentialsTable->item(row, col++);
        item->setText(QString::number(x));
```

Same with these past two

rprospero (approved):
There's some minor changes in here.  The biggest one was passing several instances where we could pass the vectors by reference, instead of making copies.  Now, the vectors are small, so this may be a premature optimization.  This is especially important because passing by reference does force one of the functions to return a const value, which I've noted in its comment.

trisyoungs:
All good suggestions. Have incorporated them in a couple of squash commits.

PR #523, opened 2021-01-31
Use assertions

### LINKED ISSUE #275: Use assertions
Remove use on CHECKS define and prefer assertions.

trisyoungs:
This PR removes the custom error-checking from the code (activated with the `-DCHECKS` build define) which basically fulfilled the role of assertions. Now that STL classes are more widespread in the code, many of these checks are redundant and covered by the STL classes themselves. Where the STL is not yet implemented, code enclosed in an `#ifdef CHECKS` block has been replaced with suitable calls to `assert()`.

Closes #275.


DanNixon (approved):
:+1:

PR #527, opened 2021-02-03
Buff enums

trisyoungs:
This PR addresses an issue encountered where intramolecular parameters could not correctly be supplied for all the different functional forms as the number of arguments expected by the relevant keywords was limited to the wrong number.

Here, the `EnumOption` class is overhauled to make supplying argument limits a little more transparent, and various uses of it throughout the code are adjusted. A new enumeration for maximum arguments is added.

### DIFF
--- a/src/base/enumoption.cpp
+++ b/src/base/enumoption.cpp
@@ -45,9 +45,6 @@ std::string_view EnumOption::keyword() const { return keyword_; }
 // Return option description
 std::string_view EnumOption::description() const { return description_; }
 
-// Return whether the option has any associated arguments
-bool EnumOption::hasArguments() const { return (minArgs_ != 0); }
-
 // Return minimum number of arguments the option takes
 int EnumOption::minArgs() const { return minArgs_; }
 
--- a/src/base/enumoption.cpp
+++ b/src/base/enumoption.cpp
@@ -8,29 +8,16 @@
 // Static Singleton
 UnrecognisedEnumOption EnumOptionsBase::unrecognisedOption_;
 
-EnumOption::EnumOption()
+EnumOption::EnumOption() : enumeration_(0) {}
+EnumOption::EnumOption(const int enumeration, std::string_view keyword) : enumeration_(enumeration), keyword_(keyword) {}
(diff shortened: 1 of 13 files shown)


src/base/enumoption.h
```diff
@@ -21,9 +24,9 @@ class EnumOption
     // Argument Numbers
     enum ArgumentNumber
```
DanNixon, line 25: Ideally `enum class` should be used, this handles scope better than classic `enum`.

src/base/enumoption.cpp
```diff
@@ -8,29 +8,16 @@
 // Static Singleton
 UnrecognisedEnumOption EnumOptionsBase::unrecognisedOption_;
 
-EnumOption::EnumOption()
+EnumOption::EnumOption() : enumeration_(0) {}
+EnumOption::EnumOption(const int enumeration, std::string_view keyword) : enumeration_(enumeration), keyword_(keyword) {}
+EnumOption::EnumOption(const int enumeration, std::string_view keyword, int minArgs, std::optional<int> maxArgs)
```
DanNixon, line 13: Why is `minArgs` not optional here?
Could this and the above constructor be combined if `minArgs` and `maxArgs` were optional and defaulted to `nullopt`?

DanNixon (commented):
Looks good :+1:, although it would be nice to see unit tests accompany changes like this.

src/base/enumoption.cpp
trisyoungs: None as far as I can tell! Have merged those two ctors together.

src/base/enumoption.h
trisyoungs: I see the benefits, however declaring as `enum class` prevents implicit conversion of the members to `int`, which `EnumOption` depends on when constructing `EnumOptionsList`s. This could no doubt be circumvented, but might require templating the `EnumOption` class, and is out of scope of this PR. I'll add an issue to track it.

trisyoungs:
Unit test now present.

DanNixon (approved):

Issue #535, opened 2021-02-11
Move Cells array to standard container


trisyoungs:
`CellArray` utilises a raw array of `Cell`s at present, with raw pointers to the `Cell`s stored on individual `Atoms`. Changing this to a `std::vector` of `Cell`s is relatively trivial, but has repercussions for how the `Cell`s are referenced on the atoms.

PR #536, opened 2021-02-11
Remove CellNeighbour class

trisyoungs:
A small PR to modernise some aspects of the `Cell` management, undertaken while bug-hunting a specific issue (which in fact was just an error message that popped up now and again, but didn't have any effect or substance to it).

This PR stops short of removing the raw array of `Cell`s in `CellArray`, as this is a more wide-reachinc change (encapsulated in #535).


src/classes/cell.cpp
```diff
@@ -118,4 +108,4 @@ bool Cell::mimRequired(const Cell *otherCell) const
 }
 
 // Return list of all Cell neighbours
-const std::vector<CellNeighbour> &Cell::allCellNeighbours() const { return allCellNeighbours_; }
+const std::vector<std::pair<Cell *, bool>> &Cell::allCellNeighbours() const { return allCellNeighbours_; }
```
DanNixon, line 111: You might consider using type aliases (e.g. `using CellNeighbour = std::pair<Cell *, bool>;`) defined in a sensible header instead of specifying the underlying type in every place it is used.

DanNixon (approved):

src/classes/cell.cpp
trisyoungs: That would make things much neater - turns out, however, that that `bool` isn't actually used anywhere, so removing it cleans things up nicely.

PR #541, opened 2021-02-12
Tidy SpeciesIntra Base

trisyoungs:
Some housekeeping work focussing on the `SpeciesIntra` class.

### DIFF
--- a/src/classes/coredata.cpp
+++ b/src/classes/coredata.cpp
@@ -109,7 +109,7 @@ MasterIntra &CoreData::addMasterBond(std::string_view name)
             fmt::format("Refused to add a new master Bond named '{}' since one with the same name already exists.\n", name)));
 
     // OK to add new master Bond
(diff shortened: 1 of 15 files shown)


rprospero (approved):
Certainly a nice tidy up.  The only though I'd had was that, if `i_`, `j_`, `k_`, an `l_` were all made part of `SpeciesIntra`, then you could put the assign functions there as well.  There were a couple of advantage to that, but I don't think they outweigh the dangers of having some variables that we know that we'll never use (e.g. what is `k_` on a `SpeciesBond`).

trisyoungs:
Along those lines there's a lot that could be condensed / tidied by, for instance, storing the target `SpeciesAtom`s in a `std::vector`. That would be preferable to having the unused vars (i.e. `k_` and/or `l_`) there. The generalised `assign()` method would have to be aware of the type of interaction, so it knows how to (re-)attach the term correctly.

PR #543, opened 2021-02-15
Move MasterIntra terms out of lists and into vectors.

rprospero:
This PR handles three different issues.  The main issues is taking the MasterIntra terms (i.e. MasterBond, MasterAngle, MasterTorsion, and MasterImproper) and removing the ListItem dependency.  The previous use of pointers is now replaced with reference (or optional references where necessary).

As part of this setup, I've had to go through and update the `const` correctness of the tree and list updaters.

I've also discovered that the copy operator for the SpeciesIntra class did not copy the `form_` member.  This has been corrected.


src/classes/coredata.cpp
```diff
@@ -101,182 +101,219 @@ int CoreData::atomTypesVersion() const { return atomTypesVersion_; }
...
+// Return whether named master Bond parameters exist
+OptionalReferenceWrapper<MasterIntra> CoreData::hasMasterBond(std::string_view name)
+{
+    // Remove leading '@' if necessary
+    std::string_view trimmedName = name[0] == '@' ? &name[1] : name;
+
+    for (auto &b : masterBonds_)
+        if (DissolveSys::sameString(trimmedName, b.name()))
+            return b;
+    return {};
```
trisyoungs, line 150: Here it would be better to rename the `hasMasterBond` (my original name, I know!) to `getMasterBond` as it is a getter, and have a separate `hasMasterBond` function that returns just a `bool`. This would be consistent with function names/return types used in the majority of other places in the code.

src/classes/coredata.cpp
```diff
@@ -101,182 +101,219 @@ int CoreData::atomTypesVersion() const { return atomTypesVersion_; }
...
-const List<MasterIntra> &CoreData::masterAngles() const { return masterAngles_; }
+const std::vector<MasterIntra> &CoreData::masterAngles() const { return masterAngles_; }
 
 // Return nth master Angle parameters
-MasterIntra *CoreData::masterAngle(int n) { return masterAngles_[n]; }
+MasterIntra &CoreData::masterAngle(int n) { return masterAngles_[n]; }
 
 // Return whether named master Angle parameters exist
-MasterIntra *CoreData::hasMasterAngle(std::string_view name) const
+OptionalReferenceWrapper<MasterIntra> CoreData::hasMasterAngle(std::string_view name)
```
trisyoungs, line 180: Same here.

src/classes/coredata.cpp
```diff
@@ -101,182 +101,219 @@ int CoreData::atomTypesVersion() const { return atomTypesVersion_; }
...
     // OK to add new master Bond
-    MasterIntra *b = masterBonds_.add();
-    b->setName(name);
-    b->setType(SpeciesIntra::BondInteraction);
+    masterBonds_.emplace_back();
+    auto &b = masterBonds_.back();
+    b.setName(name);
+    b.setType(SpeciesIntra::BondInteraction);
 
     return b;
```
trisyoungs, line 117: Tempting to add a constructor with default `name` and `type` args to `MasterIntra` so that this (and the other related occurrences) can be reduced to `masterBonds_.emplace_back(name, SpeciesIntra::BondInteraction);` followed by `return masterBonds_.back();`.

src/classes/coredata.h
```diff
@@ -62,57 +63,61 @@ class CoreData
...
-    MasterIntra *addMasterBond(std::string_view name);
+    MasterIntra &addMasterBond(std::string_view name);
     // Return number of master Bond parameters in list
     int nMasterBonds() const;
     // Return list of master Bond parameters
-    const List<MasterIntra> &masterBonds() const;
+    const std::vector<MasterIntra> &masterBonds() const;
     // Return nth master Bond
-    MasterIntra *masterBond(int n);
+    MasterIntra &masterBond(int n);
```
trisyoungs, line 82: I don't think this is used anywhere now, so can be removed along with the related calls for angle, torsion, and improper.

src/classes/masterintra.cpp
```diff
@@ -5,12 +5,14 @@
...
-MasterIntra::MasterIntra() : SpeciesIntra(), ListItem<MasterIntra>()
+MasterIntra::MasterIntra() : SpeciesIntra()
 {
     parent_ = nullptr;
     masterParameters_ = nullptr;
 }
 
+MasterIntra::MasterIntra(const MasterIntra &source) : SpeciesIntra(source), name_(source.name_), type_(source.type_) {}
+
 MasterIntra::~MasterIntra() {}
```
trisyoungs, line 16: Make this a `default` dtor.

src/classes/speciesintra.cpp
```diff
@@ -28,6 +28,7 @@ SpeciesIntra &SpeciesIntra::operator=(const SpeciesIntra &source)
     attached_[0] = source.attached_[0];
     attached_[1] = source.attached_[1];
     inCycle_ = source.inCycle_;
+    form_ = source.form_;
```
trisyoungs, line 31: Good catch.  Not quite sure what the effect of this would have been!

src/gui/importspecieswizard_funcs.cpp
```diff
@@ -391,23 +391,20 @@ void ImportSpeciesWizard::updateMasterTermsPage()
...
+        if (dissolveReference_->coreData().findMasterTerm(intra.name()))
         {
             conflicts = true;
             break;
         }
-    ListIterator<MasterIntra> torsionIterator(temporaryCoreData_.masterTorsions());
-    while (MasterIntra *intra = torsionIterator.iterate())
-        if (dissolveReference_->coreData().findMasterTerm(intra->name()))
+    for (auto &intra : temporaryCoreData_.masterTorsions())
+        if (dissolveReference_->coreData().findMasterTerm(intra.name()))
```
trisyoungs, line 407: We're missing a check for the improper terms here (my fault).

src/gui/keywordwidgets/modulegroups_funcs.cpp
```diff
@@ -39,10 +39,10 @@ ModuleGroupsKeywordWidget::ModuleGroupsKeywordWidget(QWidget *parent, KeywordBas
  */
 
 // Selection table update function
-void ModuleGroupsKeywordWidget::updateSelectionRow(int row, Module *module, bool create)
+void ModuleGroupsKeywordWidget::updateSelectionRow(int row, const Module *module, bool create)
 {
     // Grab the target groups
-    auto &groups = keyword_->data();
```
trisyoungs, line 45: Why not `auto` here?

trisyoungs (changes requested):
Some minor function naming suggestions and other bits, but all looks good.

src/classes/coredata.cpp
rprospero: That makes the code much cleaner.

src/classes/coredata.h
rprospero: It might be worth looking for a decent dead-code elimination tool

src/classes/masterintra.cpp
rprospero: Done

src/gui/importspecieswizard_funcs.cpp
rprospero: Good catch

src/gui/keywordwidgets/modulegroups_funcs.cpp
rprospero: That should obviously be `auto`.  I'm wondering if maybe I had briefly messed with the return type of `data` and I needed to use an explicit type to properly cast out of a reference wrapper?

src/classes/coredata.cpp
rprospero: I've renamed all of the `hasMaster` functions to `getMaster`.  I have *not* yet added in the new, `bool`-returning `hasMaster` functions.  My reasoning is that the only places in the code that we'd be using the new `hasMaster` is within the `CoreData` class itself.  That would tend to imply that the new methods should be `private`, but it felt kind of odd to add a private wrapper around a classes own method, especially since the `getMaster` method can be use in place with now changes.

If we think that `hasMaster` will become more important, I'm happy to add it in, but I've left it out for now.

PR #550, opened 2021-02-19
UFF

trisyoungs:
This PR corrects and significantly updates the existing implementation of the Universal Force Field.  All aspects are coded as accurately as is possible, given the lack of information on the forcefield itself.


src/classes/speciestorsion.cpp
```diff
@@ -378,6 +379,21 @@ double SpeciesTorsion::energy(double angleInDegrees, int form, const std::vector
...
+         *
+         * Parameters:
+         * 0 : Force constant, k
+         * 1...N : Coefficients C0 - CN
+         */
+        auto U = params[0] * params[1];
+        for (auto n = 2; n < params.size(); ++n)
+            U += params[0] * params[n] * cos((n - 1) * phi);
+
+        return U;
```
rprospero, line 395:
```suggestion
        std::vector<int> ns(params.size);
        std::iota(ns.begin(), ns.end(), 1);
        return params[0] * std::transform_reduce(params.begin(), params.end(), ns.begin(), params[1], std::plus<>(), [phi](auto &param, auto &n) {return param * (1.0 + cos(n * phi));});
```

With same caveats as above.

src/classes/speciestorsion.cpp
```diff
@@ -524,6 +540,23 @@ double SpeciesTorsion::force(double angleInDegrees, int form, const std::vector<
...
+         *
+         * Parameters:
+         * 0 : Force constant, k
+         * 1...N : Coefficients C0 - CN
+         */
+        auto dU_dphi = 0.0;
+        for (auto n = 2; n < params.size(); ++n)
+            dU_dphi -= params[0] * params[n] * (n - 1) * sin((n - 1) * phi);
+
+        return -dU_dphi * dphi_dcosphi;
```
rprospero, line 558: Besides the same comments from above, I've noticed that this algorithm never uses parameter 1.  Is that intentional?

rprospero (changes requested):
I mostly have some suggested improvements, but I do think we need to check the `SpeciesTorion::FourierForm` before we merge.

src/data/ff/uff/uffatomtype.h
```diff
@@ -25,7 +24,6 @@ class UFFAtomType : public ForcefieldAtomType
     private:
     // Generator parameters
     double r_, theta_, x_, D_, zeta_, effectiveCharge_, chi_, V_, U_;
```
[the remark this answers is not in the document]
trisyoungs: Done.

src/classes/speciestorsion.cpp
```diff
@@ -341,11 +342,11 @@ double SpeciesTorsion::energy(double angleInDegrees, int form, const std::vector
...
          */
-        auto result = 0.0;
+        auto U = 0.0;
         for (auto n = 0; n < params.size(); ++n)
-            result += params[n] * (1.0 + cos((n + 1) * phi));
+            U += params[n] * (1.0 + cos((n + 1) * phi));
 
-        return result;
+        return U;
     }
```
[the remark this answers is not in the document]
trisyoungs: Here I am inclined to keep the readability of the original version, primarily for my own sanity when I go back and inevitably have to check a functional form for correctness.  However, these are amongst the most demanding intramolecular energy calculations, and so a speed-up may be worthwhile, especially when considering particularly large and globulous molecules.

I think this is one worth properly benchmarking in a separate PR. I'll add an issue.

src/classes/speciestorsion.cpp
trisyoungs: The code is correct, but definitely worth asking! The `params[1]` is present as a constant with no multiplicative `sin()` term, and so is not present in the derivative w.r.t. angle.

PR #554, opened 2021-02-22
Tidy Modules

trisyoungs:
A short, boring PR to use default destructors in all modules, and mark all functions reimplemented from the `Module` base class as `override`.


DanNixon (approved):
:+1:

PR #555, opened 2021-02-25
Range splitter

rprospero:
Use a general range splitter during MPI instead of using a stride step.  This will make it easier to switch to using algorithms, since they always look at every element.

### DIFF
--- a/src/modules/rdf/functions.cpp
+++ b/src/modules/rdf/functions.cpp
@@ -326,9 +326,10 @@ bool RDFModule::calculateGR(ProcessPool &procPool, Configuration *cfg, RDFModule
 
     // Loop over molecules...
     std::shared_ptr<Atom> i, j;
-    for (auto m = start; m < cfg->nMolecules(); m += stride)
+    auto [begin, end] = cut_range(cfg->molecules().begin(), cfg->molecules().end(), stride, start);
+    for (auto it = begin; it < end; ++it)
     {
-        std::shared_ptr<Molecule> mol = cfg->molecule(m);
+        std::shared_ptr<Molecule> mol = *it;
         auto &atoms = mol->atoms();
 
         for (auto ii = atoms.begin(); ii < std::prev(atoms.end()); ++ii)
--- a/src/modules/rdf/functions.cpp
+++ b/src/modules/rdf/functions.cpp
@@ -77,8 +77,8 @@ bool RDFModule::calculateGRSimple(ProcessPool &procPool, Configuration *cfg, Par
     double rbin = 1.0 / binWidth;
 
     // Loop context is to use all processes in Pool as one group
-    auto start = procPool.interleavedLoopStart(ProcessPool::PoolStrategy);
-    auto stride = procPool.interleavedLoopStride(ProcessPool::PoolStrategy);
+    auto offset = procPool.interleavedLoopStart(ProcessPool::PoolStrategy);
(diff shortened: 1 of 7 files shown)


src/templates/algorithms.h
```diff
@@ -7,6 +7,15 @@
 #include <optional>
 #include <tuple>
 
+// Cut a range into a smaller segment for MPI
+template <typename T> auto cut_range(const T begin, const T end, const int count, const int step)
```
trisyoungs, line 11: Should `count` represent the number of "chunks" (processors) and `step` the current index of the chunk/process? If so, names could be more descriptive.

Also, is this more appropriately named `chop_range` than a `cut_range`?

src/templates/algorithms.h
```diff
@@ -7,6 +7,15 @@
...
+{
+    auto diff = end - begin;
+    T start = begin + std::ldiv((const long int)step * diff, (const long int)count).quot;
+    T stop = begin + std::ldiv((const long int)(step + 1) * diff, (const long int)count).quot;
+    return std::make_tuple(start, stop);
```
trisyoungs, line 16: Shouldn't the `step` multiply against `count` rather than `diff`?  And how is the remainder handled when the number of elements in the array doesn't divide evenly over processes?

trisyoungs (changes requested):
Definitely heading in the right direction, but need to take a close look at a couple of bits.

src/templates/algorithms.h
rprospero: Multiplying against `diff` is the correct value and is how the remainder is handled.  For example, if you have a five element array and a count of 3.

step 0 =>
start = (0 * 5) // 3 = 0 // 3 = 0
stop = 1*5 // 3 = 5 // 3 = 1

step 1 =>
start = (1 * 5) // 3 = 1 // 3 = 1
stop = 2*5 // 3 = 10 // 3 = 3

step 2 =>
start = (2 * 5) // 3 = 10 // 3 = 3
stop = 3*5 // 3 = 15 // 3 = 5

src/modules/rdf/functions.cpp
```diff
@@ -89,15 +89,14 @@ bool RDFModule::calculateGRSimple(ProcessPool &procPool, Configuration *cfg, Par
...
-        for (i = start; i < maxr[typeI]; i += stride)
-        {
-            centre = ri[i];
-            for (j = i + 1; j < maxr[typeI]; ++j)
-                bins[j] = box->minimumDistance(centre, ri[j]) * rbin;
-            for (j = i + 1; j < maxr[typeI]; ++j)
-                if (bins[j] < nPoints)
-                    ++histogram[bins[j]];
-        }
+        for_each_pair(ri, ri + maxr[typeI], stride, start,
```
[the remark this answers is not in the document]
rprospero: I believe that the code should follow the same semantics as the original code.  Each `typeI` is followed in the outer loop.  The inner loops finds the distances for each pair in `ri`.  So it will calculate the upper-right triangular element for the entire range.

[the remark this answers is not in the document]
trisyoungs: OK. Sorry - rushed to go through this before home schooling started!

src/templates/algorithms.h
trisyoungs: Again, apologies - not firing on all cylinders today.

trisyoungs (approved):

PR #559, opened 2021-02-27
EnumOptions Modernisation

trisyoungs:
This PR simplifies and modernises `EnumOptions` and related classes, principally to permit the use of `enum class` definitions as well as straight `enum`s.

- `EnumOptionsBase` (required as a generic interface to GUI keyword widgets) has been greatly simplified.
- The `EnumOptionsList` class is removed completely, with `std::vector`s of `EnumOption`s created via brace-initalisation.
- Pointless and ill-used default enumeration argument to `EnumOption` has been removed.
- Class functions now just return copies of `EnumOptions` objects, rather than references to local static objects.
- Unnecessary enum entries (e.g. those intended to capture the size of the enum list such as `nCoulombTruncationSchemes`, `nSiteKeywords`) are removed.

Other optimisations may be possible, and that have been missed. Comments welcome!


src/base/enumoptions.h
```diff
@@ -18,6 +18,20 @@ template <class E> class EnumOptions : public EnumOptionsBase
         if (options_.size() > 0)
             currentOptionIndex_ = 0;
     }
+    EnumOptions<E> &operator=(E value)
+    {
+        // Find the index of the enumeration
+        currentOptionIndex_ = std::nullopt;
```
DanNixon, line 24: Could you use `std::find_if` here?

DanNixon (approved):
:+1:

src/base/enumoptions.h
trisyoungs: Coulda, shoulda, now have(a). Also led to doing a few other modernisations / tidying, hence the request for re-review (on the three new commits alone).

DanNixon (approved):

PR #560, opened 2021-02-27
Simplify WindowFunction class

trisyoungs:
A PR to simplify the `WindowFunction` class, which made provision for extendability that was never utilised.

**Do not merge until #559 is merged, and this PR is rebased.**


StephenSmith25 (approved):
👍

PR #570, opened 2021-03-04
Change cell array to vector of unique pointers instead of c array of raw pointers

### LINKED ISSUE #535: Move Cells array to standard container

StephenSmith25:
This PR updates cell array to use a std::vector of unique pointers to the cells, rather than a C array of raw pointers. Also removes the `nCells_` member variable, as this should be retrievable from the vector, unless I've missed something!

Fixes #535

### DIFF
--- a/src/classes/cellarray.cpp
+++ b/src/classes/cellarray.cpp
@@ -141,10 +141,7 @@ bool CellArray::generate(const Box *box, double cellSize, double pairPotentialRa
             fracCentre.z = fractionalCellSize_.z * 0.5;
             for (z = 0; z < divisions_.z; ++z)
             {
-                auto &cell = cells_.emplace_back(std::make_unique<Cell>());
-                cell->setIndex(count);
-                cell->setGridReference(x, y, z);
-                cell->setCentre(box_->fracToReal(fracCentre));
+                cells_.emplace_back(std::make_unique<Cell>(count, Vec3<int>(x, y, z), box_->fracToReal(fracCentre)));
                 fracCentre.z += fractionalCellSize_.z;
                 ++count;
             }
--- a/src/classes/cellarray.cpp
+++ b/src/classes/cellarray.cpp
@@ -5,11 +5,7 @@
 #include "classes/box.h"
 #include "classes/cell.h"
 
-CellArray::CellArray()
-{
-    cells_ = nullptr;
-    nCells_ = 0;
-}
+CellArray::CellArray() : box_(nullptr) {}
 
 CellArray::~CellArray() {}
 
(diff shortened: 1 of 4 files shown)


src/classes/cellarray.cpp
```diff
@@ -138,9 +128,9 @@ bool CellArray::generate(const Box *box, double cellSize, double pairPotentialRa
 
     // Construct Cell arrays
     clear();
-    nCells_ = divisions_.x * divisions_.y * divisions_.z;
-    Messenger::print("Constructing array of {} cells...\n", nCells_);
-    cells_ = new Cell[nCells_];
+    int nCells = divisions_.x * divisions_.y * divisions_.z;
```
trisyoungs, line 131:
```suggestion
    auto nCells = divisions_.x * divisions_.y * divisions_.z;
```

src/classes/cellarray.cpp
```diff
@@ -253,10 +244,10 @@ bool CellArray::generate(const Box *box, double cellSize, double pairPotentialRa
     Messenger::print("Constructing neighbour lists for individual Cells...\n");
     std::vector<Cell *> nearNeighbours, mimNeighbours;
     Vec3<int> gridRef, delta;
-    for (n = 0; n < nCells_; ++n)
+    for (auto n = 0u; n < cells_.size(); ++n)
```
trisyoungs, line 247: This is entirely correct, of course, but Dissolve adopts the Google-style of (effectively) "no unsigned ints" - https://google.github.io/styleguide/cppguide.html#Integer_Types

src/classes/cellarray.h
```diff
@@ -34,9 +34,8 @@ class CellArray
     // Cell axes
     Matrix3 axes_;
     // Total number of Cells in Box
```
trisyoungs, line 36:
```suggestion
```

src/classes/cellarray.cpp
```diff
@@ -151,9 +141,10 @@ bool CellArray::generate(const Box *box, double cellSize, double pairPotentialRa
             fracCentre.z = fractionalCellSize_.z * 0.5;
             for (z = 0; z < divisions_.z; ++z)
             {
-                cells_[count].setIndex(count);
-                cells_[count].setGridReference(x, y, z);
-                cells_[count].setCentre(box_->fracToReal(fracCentre));
+                auto &cell = cells_.emplace_back(std::make_unique<Cell>());
+                cell->setIndex(count);
+                cell->setGridReference(x, y, z);
+                cell->setCentre(box_->fracToReal(fracCentre));
```
trisyoungs, line 147: I believe that the index, grid reference, and centre are only ever set once / here, so for brevity one could undo my laziness and extend `Cell::Cell()` to take those arguments.

trisyoungs (changes requested):
Two small style comments, and one further suggestion, but otherwise spot on.

trisyoungs (approved):
:+1:

PR #571, opened 2021-03-04
Generic Modernisation 1

trisyoungs:
This PR is the first of a sequence modernisnig various aspects of the `GenericList` and `GenericItem` classes.

The present PR focuses on adding fully `const` return of data values from the list and removing the `GenericItemHelper` class in favour of templated functions within the `GenericList` class itself. The latter promotes code readability when retrieving data, and simplifies the function call.

Many files are touched by this PR as a result of the `GenericItemHelper` class removal - for review purposes, focussing on the changes to the `src/genericitems/list.h` file will provide the majority of interesting changes!


src/modules/bragg/functions.cpp
```diff
@@ -24,28 +23,22 @@ bool BraggModule::calculateBraggTerms(ProcessPool &procPool, Configuration *cfg,
...
-    Array2D<double> &braggAtomVectorYSin =
-        GenericListHelper<Array2D<double>>::realise(cfg->moduleData(), "BraggAtomVectorYSin");
-    Array2D<double> &braggAtomVectorZSin =
-        GenericListHelper<Array2D<double>>::realise(cfg->moduleData(), "BraggAtomVectorZSin");
-    auto &braggMaximumHKL = GenericListHelper<Vec3<int>>::realise(cfg->moduleData(), "BraggMaximumHKL");
+    auto &braggKVectors = cfg->moduleData().realise<Array<KVector>>("BraggKVectors");
+    auto &braggReflections =
+        cfg->moduleData().realise<Array<BraggReflection>>("BraggReflections", "", GenericItem::InRestartFileFlag);
+    Array2D<double> &braggAtomVectorXCos = cfg->moduleData().realise<Array2D<double>>("BraggAtomVectorXCos");
+    Array2D<double> &braggAtomVectorYCos = cfg->moduleData().realise<Array2D<double>>("BraggAtomVectorYCos");
```
StephenSmith25, line 36: In keeping with other examples in the files, I guess these could also be auto declared 
```suggestion
    auto &braggAtomVectorYCos = cfg->moduleData().realise<Array2D<double>>("BraggAtomVectorYCos");
```

src/genericitems/list.h
```diff
@@ -28,30 +29,163 @@ class GenericList
...
+            throw std::runtime_error(
+                fmt::format("GenericListHelper::value({}) failed, because the target item is of the wrong type.", name));
+
+        if (found != nullptr)
+            (*found) = true;
+        return castItem->data();
+    }
+    // Retrieve named item from specified list as template-guided type, assuming that it is going to be modified
+    template <class T>
+    T &retrieve(std::string_view name, std::string_view prefix = "", T defaultValue = T(), bool *found = nullptr)
```
rprospero, line 106: Am I correct that `retrieve` and `value` do the same thing, except that `value` is `const` while `retrieve` is mutable?  If so, is there any reason that we didn't give them the same name and use const overloading?

src/genericitems/list.h
```diff
@@ -28,30 +29,163 @@ class GenericList
...
+            if (created != nullptr)
+                (*created) = true;
+            return add<T>(name, prefix, flags);
+        }
+
+        // Cast to correct type
+        auto *castItem = dynamic_cast<GenericItemContainer<T> *>(item);
+        if (!castItem)
+            throw std::runtime_error(fmt::format(
+                "GenericListHelper::realise({}) failed, because the item couldn't be cast to the desired type.", name));
```
StephenSmith25, line 155: Very minor, but do these throws need to be updated given the refactor/moving about of the code. For instance, something like
```suggestion
                "GenericList.realise({}) failed, because the item couldn't be cast to the desired type.", name));
```

StephenSmith25 (commented):
Fantastic, lots of good modernisation, just some minor comments.

rprospero (approved):
I had one question about the naming, which may just be more of my understanding of the GenericItem setup than an actual naming issue.  Even if I'm right about it, it's probably not worth changing at this point.

src/genericitems/list.h
trisyoungs: Yes. I made the distinction a long time ago (before `auto` was even a concept in my head) to ensure separation between just needing a value, and needing to modify the data. In the latter case, the internal version of the data object is also bumped automatically to reflect that it has been changed.

StephenSmith25 (approved):

PR #575, opened 2021-03-05
Remove state I/O for tabs, views, and charts.

### LINKED ISSUE #574: Don't write detailed Viewer info to the state file
On closing the GUI, a state file is written which contains every detail on every `BaseViewer`-derived widget that currently exists. Included are the specification of the view (axes, rotation etc.) and all `Renderable` and `RenderableGroup` definitions.

Writing all of this data is unnecessary, and causes some problems with ongoing modernisation of related parts of the code. Only a minimal set of data should be output - this may include View information, but should certainly not include data related to `Renderables`.

trisyoungs:
This PR is immensely boring, and removes writing of GUI "state" information from `BaseViewer`, `Chart`, and `MainTab` classes and their derivatives, since the vast majority of data written was either unimportant, awkward to handle, or simply represented opportunities for bugs / crashes to occur. The information in the state file was used to reconstruct most of the view information and data targets in the `Viewer` classes in order to reproduce the "user-modified" state when the GUI was next started on the related input file, but this only served to override the default construction state of the widgets, so no core functionality is lost.

A side effect of this is that the `WorkspaceTab` is now effectively only a temporary construct, as no window / widget information is retained on GUI exit. However, this aspect of the GUI was hastily implemented in the first place, and would benefit from a re-think (if, indeed, it is a worthwhile addition to do so).

It may set the size record for a "deletions only" PR!

Closes #574.


StephenSmith25 (approved):
Looks good, always nice to see code removal! You've just picked up a conflict in one of the files.

PR #576, opened 2021-03-07
Move Renderable and RenderableGroup to std::vector

trisyoungs:
This PR moves the `Renderable` and `RenderableGroup` classes on to `std::vector`, with the former stored as a `std:vector` of `std::shared_ptr`s. This also lays groundwork for other changes to the rendering system, which requires some modification once the `ObjectStore` class is removed.

`TreeWidgetUpdater` has been politely hacked with additional functions so as to not break the GUI, but this should be considered a temporary fix.

It also:
- Adds missing plot of improper energy to `EnergyModule`.
- Removes `Renderable` and `RenderableGroup` from depending on `List` and `ListItem` (#257).

### Related PR Hierarchy

Listed in order of review / merge ordering (rebasing required at each stage):

- ~#575 Remove state I/O for tabs, views, and charts~
- `THIS` Move `Renderable` and `RenderableGroup` to `std:vector`

PR #581, opened 2021-03-09
Use template-guided renderable creation

trisyoungs:
This PR removes the old `RenderableFactory` class, and uses template-guided instantiation of `Renderable`s. This neatens the creation of such objects, particularly when a copy of the pointer to the renderable must be maintained by the usage site.

Two overloaded `createRenderable()` functions are now present in `BaseViewer` - one which takes a single template argument (the renderable class type) and the name of the data to search for (at some point), and one which takes two template arguments (the renderable and data class types) and a pointer to the data to be rendered. This latter function will likely change to take a reference in a near-future commit, in line with ongoing modernisation work in multiple branches.

This is part of a sequence of PRs updating the general storage and location of data within the code.

### Related PR Hierarchy

- `THIS` Template renderable creation.
- #582 No local configuration data
- #586 New GenericList
- #599 Data Name Tidying


rprospero (approved):
I always love offloading more work from humans onto the compiler.

PR #582, opened 2021-03-09
No local configuration data

trisyoungs:
This PR removes storage for module data (via `GenericList`) on `Configuration`s. All such data is now stored in the main (and only) `Dissolve::processingModuleData_`, prefixed by the configuration name. The advantages of doing this are:

- All relevant data is contained in a single location, facilitating searching (particularly when `ObjectStore` is removed).
- `Configuration`s are now pure elements of the simulation (making them more like `Species`), rather than containing data produced by the simulation.

The benefit of storing data locally on `Configuration`s is that it is easily separated from other data. However, suitable prefixing of the data names in the main `GenericList` handles this well enough. If more separation is required at a later stage (e.g. for access speed when searching for data) the `GenericList` class can be overhauled to "bucket" data into relevant groups (e.g. by prefix).

Notes:
- Module widget constructors now take the processing data list as an argument to enable relevant data to be extracted for display where necessary.
- The "restart" system test has been basically hacked to work. A near-future PR will break it again, so replacing it properly is left until then.

This is part of a sequence of PRs updating the general storage and location of data within the code.

### Related PR Hierarchy

- ~#581 Use template-guided renderable creation~
- `THIS` No local configuration data
- #586 New GenericList
- #599 Data Name Tidying
- #606 Remove ObjectStore 2

### DIFF
--- a/src/classes/configuration.cpp
+++ b/src/classes/configuration.cpp
@@ -179,22 +179,6 @@ void Configuration::setTemperature(double t) { temperature_ = t; }
 double Configuration::temperature() const { return temperature_; }
 
 /*
- * Modules
- */
-
-// Associate Module to the Configuration
-bool Configuration::ownModule(Module *module) { return moduleLayer_.own(module); }
-
-// Return number of Modules associated to this Configuration
-int Configuration::nModules() const { return moduleLayer_.nModules(); }
-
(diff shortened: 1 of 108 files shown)


StephenSmith25 (approved):
All looks good 👍

PR #583, opened 2021-03-09
Xml

### LINKED ISSUE #539: Support loading Forcefields from OpenMM XML files
Currently, all Forcefields are hard-coded into Dissolve. It would be helpful if Forcefields could be loaded from an XML file.  Since the OpenMM file format is already standardized, that would be a useful place to start.

**Additional details**
Due to the type mismatch between the atom types in the XML files and the atom types in Dissolve, it may need to be a GUI only option for now.

rprospero:
This is a draft on the XML importer.  There's still some little bits that needs cleaning up, but I wanted it up and visible while I'm out of commission for a week at ACCU so that any big changes that need to be made can be noticed.

This closes #539 

Running
=====

To use the XML importer
1. Load `dissolve-gui`
2. Start a new simulation
3. Add atom types for Hydrogen, Carbon, and Oxygen
4. Under the species menu, choose `Import Forcefield...`
5. Load an OpenMM XML file (e.g. `tests/ff/methanol.xml`)
6. Click `Continue`
7. Check the values in the tables.  Specifically, the `Atom Types` table automatically matches each type from the XML file with the first pre-existing atom type of the same element.  If there is no pre-existing type, then it marks it as missing and displays the type in red.  The type box can be edited with the name of an existing atom type.
8. Click `Continue`

Tasks
===

- [x] Add unit test that missing atom types are red
- [x] Properly accept dialog on completion
- [x] Properly clear dialog after use
- [x] Validate atom types to ensure that every type has a match
- [x] Ensure that user's chosen atom type has the correct element
- [x] Use a combobox for putting in atom type

Major Changes
==========

Under the `src` directory, there is now a `models` subdirectory that contains Qt Models.  These were put into a separate directory and library from the rest of the GUI code so that the models can be iterated and recompiled quickly without needing to relink all of the GUI component libraries.  This was important because the unit tests link against the models, but not the components, so being able to link them separately saved quite a bit of time.

There are now six unit tests that check the XML importer.  These unit tests are only built if both testing and GUI is enabled.  This allows tests to continue to be built on platforms that do not support the GUI while still allowing linking against the Qt library in the tests.


src/classes/species_io.cpp
```diff
@@ -351,7 +351,14 @@ bool Species::read(LineParser &parser, CoreData &coreData)
...
             case (Species::ForcefieldKeyword):
-                forcefield_ = ForcefieldLibrary::forcefield(parser.argsv(1));
+	        switch (parser.nArgs()) {
+		    case 2:
+		      forcefield_ = ForcefieldLibrary::forcefield(parser.argsv(1));
+		      break;
+		    default:
+		      // forcefield_ = ForcefieldLibrary::forcefield(parser.argsv(1), parser.argsv(2));
+		      break;
+		}
```
rprospero, line 361: Remove these changes.  They are remnants of the original CLI attempt at the XML import

src/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ add_subdirectory(procedure)
 add_subdirectory(neta)
 
 if(GUI)
+  add_subdirectory(models)
```
trisyoungs, line 15: What's the reason for having the (obviously GUI-related) `models` directory located outside the `gui` dir?

src/classes/species_io.cpp
```diff
@@ -351,7 +351,15 @@ bool Species::read(LineParser &parser, CoreData &coreData)
...
-                forcefield_ = ForcefieldLibrary::forcefield(parser.argsv(1));
+                switch (parser.nArgs())
+                {
+                    case 2:
+                        forcefield_ = ForcefieldLibrary::forcefield(parser.argsv(1));
+                        break;
+                    default:
+                        // forcefield_ = ForcefieldLibrary::forcefield(parser.argsv(1), parser.argsv(2));
+                        break;
+                }
```
trisyoungs, line 362: If this is a placeholder for future loading of an xml forcefield from the command-line?  Could (should) it be removed for now?

src/data/ff/ff.h
```diff
@@ -29,6 +29,8 @@ class Forcefield
 {
     public:
     Forcefield() = default;
+    Forcefield(std::vector<ForcefieldAtomType> &, std::vector<ForcefieldBondTerm> &, std::vector<ForcefieldAngleTerm> &,
+               std::vector<ForcefieldTorsionTerm> &, std::vector<ForcefieldImproperTerm> &);
```
trisyoungs, line 33: I appreciate that the argument types are self-explanatory, but naming the vars (e.g. `bonds`, `angles`, etc.) is nicer from a stylistic sense IMO, and consistent with the rest of the code.

src/data/ff/xml/base.h
```diff
@@ -0,0 +1,42 @@
...
+#pragma once
+
+#include "data/ff/ff.h"
+#include "templates/optionalref.h"
+#include <string_view>
+
+// Forward Declarations
+/* none */
+
+// OPLS-AA (2005) Base Forcefield
```
trisyoungs, line 13: Need to change the comment.

src/data/ff/xml/base.cpp
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2021 Team Dissolve and contributors
+
+#include "data/ff/xml/base.h"
+#include <iostream>
+
+Forcefield_XML::Forcefield_XML(std::vector<ForcefieldAtomType> atoms, std::vector<ForcefieldBondTerm> bonds,
+                               std::vector<ForcefieldAngleTerm> angles, std::vector<ForcefieldTorsionTerm> torsions,
+                               std::vector<ForcefieldImproperTerm> impropers)
+    : Forcefield(atoms, bonds, angles, torsions, impropers)
```
trisyoungs, line 10: If `Forcefield` accepts references to `std::vector`s then surely the `Forcefield_XML` ctor should do the same?

src/gui/importforcefieldwizard_funcs.cpp
```diff
@@ -0,0 +1,106 @@
...
+        ui_.atomTable->resizeColumnsToContents();
+        ui_.bondTable->resizeColumnsToContents();
+        ui_.angleTable->resizeColumnsToContents();
+        ui_.torsionTable->resizeColumnsToContents();
+        ui_.improperTable->resizeColumnsToContents();
+    }
+    else
+    {
+        atoms_.clear();
+    }
```
trisyoungs, line 103:
```suggestion
        atoms_.clear();
```

trisyoungs (changes requested):
Huge amount of work here - definitely warrants the `32`!  I haven't checked out the branch and compiled it for myself so I don't know what the UX is like, but from reading the code I have a query about how we are handling the atom types here.  At present, there seems to be a dependence on having an existing (simulation) atom type of the same element in order to properly process the forcefield. I may have missed the point here, or forgotten a discussion that happened two weeks ago, but I appreciate that there is no way of automatically or cleverly assigning the atom types that we read in from the file on to the atoms in a molecule, as of course there is no NETA information in the XML.  This isn't really a problem for the current PR to solve, though, as it's more of a general issue.

There are a couple of possibilities for doing this semi-automatically in the present case - one is to use the rudimentary bond information in the XML file (in the Residue key) to construct a basic NETA description for each atom based only on simple topology. However, all of that is beyond the scope of this PR.

src/CMakeLists.txt
rprospero: There were three parts to the idea

1) I wanted to place a distinction that the models aren't directly GUI code, but rather a system state that the GUI can query.  It's possible to have multiple front-ends to a mode, including web based ones and even command line ones.  They don't have to be GUI code.  Now, in practice, we only use them for the GUI, so this is a meaningless distinction.
2) The models could be compiled using only the Qt::Core and not needing the Qt widgets, so the hope was that you could compile the model unit tests on machines that couldn't support a full GUI.  Except that you needed QT::Widgets the first time a model requested that something is displayed in a different colour, so that fell apart pretty quickly.
3) When I was iterating on my unit tests, I could recompile libmodels without needing to relink the massive libcomponents in the GUI folder, which made iterating go faster.  Except I think that this broke when I got the unit tests building on Azure, so this advantage is lost.

In other words, there's no good reason for this now.  I'll go ahead and move it into the GUI folder.

src/classes/species_io.cpp
rprospero: I shouldn't have left that in there.  It's been removed.

src/data/ff/ff.h
rprospero: You're right that it's better to match the rest of the code.

trisyoungs:
From an initial trial of the GUI in situ, this is excellent because of the extra forcefield possibilities it opens up, however I wasn't able to assign forcefield parameters from the supplied XML methanol forcefield to the related molecule. Perhaps I misunderstood the current workflow: I created `H`, `C`, and `O` atom types, imported the xml forcefield, then attempted to assign it. The complaint was that the bond between `H` and `C` could not be found, so it looks like either an issue with the transfer of type names from the xml on to the existing ones, or I am not using it correctly! I still think that a workflow where the xml forcefield is loaded independently of existing types is preferable, and is more consistent with the existing implementation of forcefields.

Some other observations:
- Once the XML forcefield has been imported, the list of forcefields in the Add Forcefield Terms wizard contains only the imported xml forcefield - all others have been deleted.
- The second page of the import xml ff wizard which shows all the imported parameters is nice, but I wonder if a more concise report of the available terms would be more friendly to the user. It's one of those cases of "do they need to know"?
- The issue of how to eventually apply the imported atom types to the molecule still remains, but this is not an issue with the xml importer. This is a general workflow that is missing in Dissolve, and would cover proper manual assignment of atom types to molecules, as well as merging / splitting of atom types. This is at least one separate issue in itself.
- The atom types read in from the xml file can/should assume the `class` as their names. Then, the intramolecular term definitions can be directly applied to the molecule, *if* the atom types have been manually assigned to the atoms (previous point).

trisyoungs:
Looking much better.  Some initial observations / comments based on a quick run-through of the process:

1. Auto-populate the "Name" QLineEdit with something based on the xml file name?
2. Append the number of terms read in to the table root items (e.g. "Bonds (4)") to give some immediate feedback to the user that something successful happened.
3. The "Atoms" tree should really be called "AtomTypes" - on that note, where are the interaction (e.g. Lennard-Jones) parameters?
4. Use "resizeColumnsToContents()" on the table.
5. Units - for the bonds we appear to be in `nm` rather than Angstroms, and the force constant units look way off (too large by about a factor 100)...
6. Units 2 - for the angles are the equilibrium values in radians?  The force constants look OK here.
7. We shouldn't be immediately adding atom types in to the main Dissolve lists (in `CoreData`) - rather, those definitions / values should be stored in the Forcefield we create, and then in turn it is up to the user to apply this forcefield to a given species.

rprospero:
> 1. Auto-populate the "Name" QLineEdit with something based on the xml file name?

I now pull the file name from the base name of the xml file, if the user hasn't already given it a name.

> 2. Append the number of terms read in to the table root items (e.g. "Bonds (4)") to give some immediate feedback to the user that something successful happened.

Done

> 3. The "Atoms" tree should really be called "AtomTypes" - on that note, where are the interaction (e.g. Lennard-Jones) parameters?

I have not added the interaction parameters.  I'll get started on that.

> 4. Use `resizeColumnsToContents()` on the tree.

I was doing that already, but it seemed to be resizing to the columns of the collapsed trees.  I'm now briefly opening all the trees, resizing the columns, and collapsing the trees again.  This seems to be providing the appropriate behavior.

> 5. Units - for the bonds we appear to be in nm rather than Angstroms, and the force constant units look way off (too large by about a factor 100)...

I've switched the bonds to Angstroms and made the corresponding unit change to the force constant.

> 6. Units 2 - for the angles are the equilibrium values in radians?  The force constants look OK here.

The angles are the equilibrium values in radians and the force constant is per radian squared.  Do we need to switch these to degrees?

> 7. We shouldn't be immediately adding atom types in to the main Dissolve lists (in `CoreData`) - rather, those definitions / values should be stored in the Forcefield we create, and then in turn it is up to the user to apply this forcefield to a given species.

The part that has been catching me on this is that the ForcefieldAtomType constructor takes an index.  In my reading of the code, I thought that this was the index of the corresponding AtomType in the main vector, so there needs to be an AtomType.  Now, I can set that index to -1 (the value for the default constructor).  Alternately, I might be wrong about its purpose.  If it only needs to be the index in the vector of ForcefieldAtomTypes, then I can populate it manually.

trisyoungs:
> The angles are the equilibrium values in radians and the force constant is per radian squared. Do we need to switch these to degrees?

Force constants OK as per radian squared,  Equilibrium angles need the "human" form of degrees (Dissolve assumes degrees, and I've never met a user-facing computational code that doesn't use degrees as input).

> The part that has been catching me on this is that the ForcefieldAtomType constructor takes an index. In my reading of the code, I thought that this was the index of the corresponding AtomType in the main vector, so there needs to be an AtomType. Now, I can set that index to -1 (the value for the default constructor). Alternately, I might be wrong about its purpose. If it only needs to be the index in the vector of ForcefieldAtomTypes, then I can populate it manually.

Ah, OK - so the indices in the `ForcefieldAtomType`s can be whatever you like.  They are only there to provide a numerical ID if you want to use references when writing NETA descriptions.  They bear no relation to the indices on the `AtomType`s used in the main simulation. `ForcefieldAtomType`s in `Forcefield` only serve to hold the necessary data to create an `AtomType` proper, *if* that type is required when assigning to a species.

trisyoungs:
> So an important note on this. The commit I just pushed _should_ have moved us to GCC 8 on Azure. If we only want to support GCC 7, then I'll have to rip out the `filesystem` library.

Well, we'll need GCC 9 for the STL Parallel features, so leaving GCC 7 (and 8) behind is fine by me

trisyoungs:
Had another run through of the XML import and this is all looking pretty good. There is still the lack of ability to assign atom types automatically or manually, but this is not the fault of this PR and we can work that up once this is merged.

Couple of issues that still remain:
1. Angle equilibrium values are still in radians (need to be degrees). Force constants are fine.
2. If I import the forcefield and then open the Add Forcefield Terms wizard for the first time, the list of forcefields only contains the imported XML one.
3. If I open the Add Forcefield Terms wizard *before* importing the XML forcefield, then import the XML forcefield, then open the Add Forcefield Terms wizard again, the built-in forcefields are there, but the imported XML one is not.

Some cosmetics:
1. Small naming convention change to the generated forcefield name - could be prefixed with "OpenMM/" to give it the same convention as the built-in ones.
2. Font size and icons on buttons (e.g. cross on the Cancel button) could be more in-line with, e.g., the Add Forcefield Terms wizard.

trisyoungs (approved):

PR #584, opened 2021-03-11
Remove list item inheritance from species info

StephenSmith25:
This PR removes the list item inheritance from the species info class. 

Do not merge before #573 as this will likely need a rebase.

PR #585, opened 2021-03-11
Remove list item inheritance from species site

StephenSmith25:
This PR removes the list item inheritance from the species site class. 

Will probably need rebasing if #584 is merged first.

### DIFF
--- a/src/classes/species_site.cpp
+++ b/src/classes/species_site.cpp
@@ -8,33 +8,35 @@
 // Add a new SpeciesSite to this Species
 SpeciesSite *Species::addSite(std::string_view name)
 {
-    auto *site = sites_.add();
-    site->setParent(this);
-    site->setName(uniqueSiteName(name));
+    auto &site = sites_.emplace_back();
+    site.setParent(this);
+    site.setName(uniqueSiteName(name));
 
-    return site;
+    return &site;
 }
 
 // Remove specified SpeciesSite
 void Species::removeSite(SpeciesSite *site)
 {
+    auto it = std::remove_if(sites_.begin(), sites_.end(), [&](const auto &p) { return &p == site; });
     if (site == nullptr)
         Messenger::error("NULL_POINTER - NULL SpeciesSite passed to Species::removeSite().\n");
-    else if (sites_.contains(site))
+    else if (it != sites_.end())
     {
         Messenger::print("Removed SpeciesSite '{}' from Species '{}'.\n", site->name(), name_);
-        sites_.remove(site);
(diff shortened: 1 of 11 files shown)


src/classes/species_site.cpp
```diff
@@ -51,13 +53,21 @@ std::string Species::uniqueSiteName(std::string_view base, const SpeciesSite *ex
 }
 
 // Search for SpeciesSite by name
-SpeciesSite *Species::findSite(std::string_view name, const SpeciesSite *exclude) const
+const SpeciesSite *Species::findSite(std::string_view name, const SpeciesSite *exclude) const
```
trisyoungs, line 56: In keeping with similar modernisations, this should return an `OptionalReferenceWrapper<const SpeciesSite>` rather than a `const` ptr.

src/classes/species_site.cpp
```diff
@@ -51,13 +53,21 @@ std::string Species::uniqueSiteName(std::string_view base, const SpeciesSite *ex
...
+        return ((&p != exclude) && (DissolveSys::sameString(name, p.name())));
+    });
+    if (it != sites_.end())
+    {
+        return &(*it);
+    }
+    else
+    {
+        return nullptr;
+    }
```
trisyoungs, line 68: Can remove the extraneous braces here.

src/gui/helpers/listwidgetupdater.h
```diff
@@ -367,6 +367,65 @@ template <class T, class I> class ListWidgetUpdater
         }
     }
 
+    ListWidgetUpdater(QListWidget *listWidget, const std::vector<I> &data, Qt::ItemFlags flags = Qt::NoItemFlags,
```
trisyoungs, line 370: I'm beginning to really hate this class... :)

src/gui/speciestab_sites.cpp
```diff
@@ -35,8 +35,11 @@ void SpeciesTab::setCurrentSiteFromViewer()
         return;
 
     // Sanity check that the displayed site actually exists in our species
-    if (!species_->sites().contains(displayedSite))
+    if (std::find_if(species_->sites().begin(), species_->sites().end(),
+                     [displayedSite](const auto &p) { return &p == displayedSite; }) == species_->sites().end())
+    {
         return;
+    }
```
trisyoungs, line 42: And here. We sit in the "no braces if it's just a single line" camp.

trisyoungs (approved):
:+1:  One small change and a couple of style tweaks, but I've pre-approved.

PR #586, opened 2021-03-11
New GenericList

### LINKED ISSUE #404: Modernise GenericList / GenericListHelper
The present implementation of `GenericListHelper` provides several function for the creation and / or retrieval of data from a `GenericList`. Part of this functionality includes returning whether the named object is found / created, and which is currently fed back to the calling function via a `bool` accessed via a provided pointer.

### Proposed Work
- [x] Return an`OptionalReferenceWrapper` from the `value()` and `retrieve()` functions, removing the need to pass a `bool` (`std::nullopt` can be returned on failure)
- [x] Either pass the `bool` to the `realise()` function as a reference, or return a tuple containing the reference to the object as well as its creation status
- [x] Remove `GenericListHelper` entirely, and reimplement its functionality as templated member functions within `GenericList`. This would reduce the code required for use, as the target `GenericList` is the first parameter provided to all of `GenericListHelper`s functions.

### LINKED ISSUE #453: Remove GenericItem classes
Currently, there is a GenericItem class template that handles many aspects needed for multiple data types (e.g. broadcasting, RTTI, serialisation).  However, it has multiple problems to it

1) It's something of a god class that handles multiple independent bits of functionality whose only commonality is that they're needed everywhere
2) It assumes that instances that are containers are full of instances that are also members of GenericItem.  This can be an issue for base types or classes that cannot be extended
3) All functionality must be defined for all classes, even if only parts of it are needed.
4) There is significant boilerplate.

It might be better to reorganize the class as a series of templated functions.  Template specialisation would allow the function to be defined for each type, even if it is not a class and without requiring an inheritance hierarchy.  It would also allow only defining implementations for the classes that need them.

trisyoungs:
This PR removes the clumsy inheritance hierarchy holding up the functionality of `GenericList`, and replaces it with a more elegant inheritance-free solution.

### Implementation Details

The existing `GenericList`/`GenericItem` implementation uses multiple layers of inheritance in order to give a generic base-class that could be stored in a homogeneous list, but whose derived siblings could contain any data type, POD or otherwise. For a custom class to be store in the `GenericList` it has to inherit `GenericItemBase` which required several virtual functions to be implemented in order to get the name of the class, and to provide (de)serialisation and MPI-related broadcast functions that could be used "blind" by the `GenericList` when required. For a POD-type class to be stored in a `GenericList`, specialisations of the relevant classes had to be written.

The proposed new implementation is based around a `std::map` at the suggestion of @rprospero. The map keys are the names of the items, and the map values are a `std::tuple` containing the data item in the form of `std::any`, the class name (for convenience), and the internal version and flags for the item. Since `std::visit` cannot be used with `std::any` (since deducing the type contained in the latter may only be performed at run-time) templated functions ro provide the necessary serialisation etc. cannot directly be used. To circumvent this a series of templated `std::function`s are registered on construction of `GenericList` in order to perform the various functions (after internally performing `std::any_cast<T>` on the passed `std::any`).

Where a suitable (de)serialisation function was implemented within a custom class (typically forced by inheriting the old `GenericItemBase`) this is retained and called in the relevant templated function.

### Notes

- Old `T::equality()` members, along with the corresponding `GenericList` function, have been removed. Their original purpose was to ensure data consistency between processes, particularly when data unique to individual processes had to be shared back to all other processes. However, while early plans for Dissolve envisaged process groups working on specific isolated tasks, this was never implemented, and so the equality check is somewhat redundant as all processes currently follow the same workflow path in the simulation. Data equivalence is also covered by the various system tests as each process performs its own test against reference data.
- Extending this, most MPI broadcast() functions can be removed as well, leaving just a handful of data broadcast functions. However, due to the scope of this work, this is left for a follow-up PR.

#### Suggestions
- [x] Use singletons / factories
#### Issue #404 
- [x] Get rid of `bool *created` in `GenericList::retrieve()` somehow?
- [x] ~Return `std::optional<T>` from `GenericList::value()`, and ditch the `bool *created`~ Implemented as separate function `valueIf()` to avoid confusion.
#### Tidy
- [x] Rename old `GenericItemBase`-enforced `read()` and `write()` functions as `serialise()` and `deserialise()`.
- [x] Clean up (new) `serialise()` and `deserialise()` function arguments - e.g. `Data1D` read() does not need `CoreData`.
- [x] ~Move `SampledDouble` serialisation/deserialisation code to singletons~ Rejected
- [x] Decision - see https://github.com/disorderedmaterials/dissolve-docs/pull/5

### Related Issues

Closes #404.
Closes #453.

### Related PR Hierarchy

- ~#581 Use template-guided renderable creation~
- ~#582 No local configuration data~
- `THIS` New GenericList
- #599 Data Name Tidying
- #606 Remove ObjectStore 2


StephenSmith25 (commented):
In general I think it seems like a reasonable solution. In my opinion, things like the producers should probably be abstracted out from the Generic List class at some point (maybe after you have a minimal working example) something like a factory class which handles registering and creation of the types would probably be a good idea. You could then have a singleton instance of this factory, that’s used for all the lists to create items. But I haven't considered the implementation any further than that.

I know you've kept away from boost, but I'm wondering if it would be worth looking at boost.fusion, https://theboostcpplibraries.com/boost.fusion, its a library that supports heterogeneous containers. Maybe it doesn't do everything you need (as I'm not really familiar with the GenericList yet) but I think it may be worth a glance.

trisyoungs:
The factory is a good idea - there is a clear separation between the creation of the items and the intent of the `GenericList`.

The Boost fusion stuff is cool and, if I understand it correctly, could allow this to be rewritten using templated functions instead of all this `std::any` business.  Overhead isn't really an issue here, but of course the Boost way would be faster.

src/genericitems/list.cpp
```diff
@@ -2,293 +2,186 @@
...
-        nextItem = item->next();
-        if (!item->hasFlag(GenericItem::ProtectedFlag))
-            items_.remove(item);
-        item = nextItem;
-    }
+    for (auto it = items_.begin(); it != items_.end();)
+        if (!(std::get<ItemData::Flags>(it->second) & GenericList::ProtectedFlag))
+            items_.erase(it);
+        else
+            ++it;
```
rprospero, line 25:
```suggestion
    items_.erase(std::remove_if(items_.begin(), items_.end(), [=](auto& [key, value]){ return !(std::get<ItemData::Flags>(value) & GenericList::ProtectedFlag);}, items_.end());
```

Might need to take a closer look at what I've included in with the lambda, but I think I got that right.

src/genericitems/list.cpp
```diff
@@ -2,293 +2,186 @@
...
-    {
-        if (DissolveSys::endsWith(item->name(), suffix))
-        {
-            nextItem = item->next();
-            items_.remove(item);
-            item = nextItem;
-        }
+    for (auto it = items_.begin(); it != items_.end();)
+        if (DissolveSys::endsWith(it->first, suffix))
+            items_.erase(it);
```
rprospero, line 96:
```suggestion
    items_.erase(std::remove_if(items_.begin(), items_.end(), [&suffix](auto& [key, value]){ return DissolveSys::endsWith(key, suffix);}, items_.end());

```

src/genericitems/list.cpp
```diff
@@ -2,293 +2,186 @@
...
-                checkedItems.append(item);
-            }
+// Serialise all objects via the specified LineParser
+bool GenericList::serialiseAll(LineParser &parser, const CoreData &coreData, std::string_view headerPrefix) const
+{
+    for (auto &item : items_)
+    {
+        // If it is not flagged to be saved in the restart file, skip it
+        if (!(std::get<ItemData::Flags>(item.second) & GenericList::InRestartFileFlag))
+            continue;
```
rprospero, line 130:
```suggestion
    for (auto &[key, value] : items_)
    {
        // If it is not flagged to be saved in the restart file, skip it
        if (!(std::get<ItemData::Flags>(value) & GenericList::InRestartFileFlag))
            continue;
```

It's a style thing, but I happen to like using the destructuring binds to give names to the individual parts

src/genericitems/list.cpp
```diff
@@ -2,293 +2,186 @@
...
-        }
-        else
-            while (procPool.decision(n))
-            {
-                // Receive the name of the item we're targetting
-                if (!procPool.broadcast(itemName, n))
-                    return false;
+        if (!parser.writeLineF("{}  {}  {}  {}  {}\n", headerPrefix, item.first, std::get<ItemData::ClassName>(item.second),
+                               std::get<ItemData::Version>(item.second), std::get<ItemData::Flags>(item.second)))
+            return false;
```
rprospero, line 134:
```suggestion
        if (!parser.writeLineF("{}  {}  {}  {}  {}\n", headerPrefix, key, std::get<ItemData::ClassName>(value),
                               std::get<ItemData::Version>(value), std::get<ItemData::Flags>(value)))
            return false;
```

Names again

src/genericitems/list.cpp
```diff
@@ -2,293 +2,186 @@
...
 
-                Messenger::printVerbose("Checking equality of item '{}'...\n", itemName);
+        // Find a suitable serialiser and call it
+        auto &data = std::get<ItemData::AnyObject>(item.second);
```
rprospero, line 137:
```suggestion
        auto &data = std::get<ItemData::AnyObject>(value);
```

src/genericitems/list.cpp
```diff
@@ -2,293 +2,186 @@
...
+        auto it = serialisers_.find(data.type());
+        if (it == serialisers_.end())
+            throw(std::runtime_error(
+                fmt::format("Item '{}' cannot be serialised as no suitable broadcaster is available (for class '{}').\n",
+                            item.first, std::get<ItemData::ClassName>(item.second))));
```
rprospero, line 142:
```suggestion
                            key, std::get<ItemData::ClassName>(value))));
```

src/genericitems/list.cpp
```diff
@@ -2,293 +2,186 @@
...
+    {
+        Messenger::printVerbose("Broadcasting item '{}'...\n", item.first);
+        auto &data = std::get<ItemData::AnyObject>(item.second);
+        auto it = broadcasters_.find(data.type());
+        if (it == broadcasters_.end())
+            throw(std::runtime_error(
+                fmt::format("Item '{}' cannot be broadcast as no suitable broadcaster is available (for '{}').\n", item.first,
+                            data.type().name())));
+        if (!(it->second)(data, procPool, root, coreData))
+            return Messenger::error(fmt::format("Broadcast of item '{}' failed.\n", item.first));
```
rprospero, line 184:
```suggestion
    for (auto &[key, value] : items_)
    {
        Messenger::printVerbose("Broadcasting item '{}'...\n", key);
        auto &data = std::get<ItemData::AnyObject>(value);
        auto it = broadcasters_.find(data.type());
        if (it == broadcasters_.end())
            throw(std::runtime_error(
                fmt::format("Item '{}' cannot be broadcast as no suitable broadcaster is available (for '{}').\n", key,
                            data.type().name())));
        if (!(it->second)(data, procPool, root, coreData))
            return Messenger::error(fmt::format("Broadcast of item '{}' failed.\n", key));
```

src/genericitems/producers.cpp
```diff
@@ -0,0 +1,32 @@
...
+    registerProducer<std::vector<double>>("std::vector<double>");
+
+    // Custom Classes
+    //    registerProducer<>("");
+    registerProducer<Data1D>("Data1D");
+    registerProducer<Data2D>("Data2D");
+    registerProducer<Data3D>("Data3D");
+    registerProducer<PartialSet>("PartialSet");
+    registerProducer<Vec3<int>>("Vec3<int>");
+    registerProducer<Vec3<double>>("Vec3<double>");
```
rprospero, line 31: There's a beautiful bit with consteval and fragments coming in c++23 that will handle this boilerplate for us.  I'm sure that we'll get c++23 support on SCARF in just a couple more popes.

rprospero (commented):
Just taken a quick look through on this and it looks great.  I'll have to do a more thorough look later on.  I did make a couple of comments on places where we could better use some algorithms and my personal preference for destructuring binds, though I'm not going to fight for that one.

I think that there may be a better way to handle the broadcasters and serialisers through `std::visit` which would save the need to register, but I'll have to give that some more thought.

src/genericitems/list.cpp
trisyoungs: Much nicer!

src/genericitems/list.cpp
trisyoungs: For some reason I was under the impression that `std::map` didn't have an `erase()` overload that took begin and end iterators.  It's been a tough week!

src/genericitems/producers.cpp
trisyoungs: Bring on FRANCISCVS Tertium!

trisyoungs:
@rprospero if you can get `std::visit` to play ball with `std::any`, then that would be great.  I failed on what I thought were quite fundamental levels - I tried with a type-matching visitor of the type exemplified on https://en.cppreference.com/w/cpp/utility/variant/visit to avoid the need of a `std::variant`, but failed quickly as passing a `std::any` raises a `valueless_by_exception` error.  Sure, you can check the type at run time with `type()` but there's no way to translate that to a (compile-time) templated function call or cast it before passing to `std::visit`.  If you can unpick this, then I'm definitely listening!

rprospero:
That’s my mistake. You’re right that std::any and std::visit are fundamentally incompatible.  My original thought had been a giant std::variant (with a short and friendly alias).  That should work with std::visit, but would require every type that’s stored in the generic list to be manually included in the variant.  I’d suggested std::any to get around that, forgetting that I just broke std::visit.

I’m not sure that adding all the types to the variant and visit isn’t just as much boilerplate as registering all the types with visit.  I guess it might give more compile time type safety?

trisyoungs:
I had considered the catch-all `std::variant` as the alternative solution, but I don't really like the fact that the variant takes on the size of the largest container, so storing a `bool` whose memory footprint is that of a `ALargeClass` doesn't sit well with me.  Sure, the overall amount of memory wastage is probably small right now, but it will only grow.

trisyoungs:
> Just taken a quick look through on this and it looks great. I'll have to do a more thorough look later on. I did make a couple of comments on places where we could better use some algorithms and my personal preference for destructuring binds, though I'm not going to fight for that one.

Alas, when using iterators you can't use a structured bind.

src/genericitems/producers.h
```diff
@@ -0,0 +1,63 @@
...
+    std::unordered_map<std::type_index, std::string> classNames_;
+    std::unordered_map<std::type_index, ProducerFunction> producers_;
+
+    private:
+    // Register producer for specific class
+    template <class T> void registerProducer(std::string className)
+    {
+        producers_[typeid(T)] = []() { return T(); };
+        classNames_[typeid(T)] = std::move(className);
+    }
```
rprospero, line 38:
```suggestion
    template <class T> void registerProducer()
    {
        producers_[typeid(T)] = []() { return T(); };
        classNames_[typeid(T)] = std::move(typeid(T).name());
    }
```

This would remove the need to give the type name in both the template and the string for every type.  It does remove the ability to register a producer under the "wrong" name, but I didn't find an instance of using that.

src/genericitems/producers.cpp
```diff
@@ -0,0 +1,107 @@
...
+    registerProducer<double>("double");
+    registerProducer<int>("int");
+
+    // stdlib
+    registerProducer<std::string>("std::string");
+    registerProducer<std::streampos>("streampos");
+    registerProducer<std::vector<double>>("std::vector<double>");
+
+    // Custom Classes
+    registerProducer<Array<double>>("Array<double>");
```
rprospero, line 30:
```suggestion
    registerProducer<Array<double>>();
```

This change goes with the change I suggested to `registerProducer` in producers.cpp.  The same change would obviously be applied to all the following lines.

src/genericitems/producers.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+// Return class name for specified type
+std::string GenericItemProducer::className(const std::type_info &objectType) const
+{
+    auto it = classNames_.find(objectType);
+    if (it == classNames_.end())
+        throw(std::runtime_error(fmt::format("Class name has not been registered for type '{}'.\n", objectType.name())));
+
+    return it->second;
+}
```
rprospero, line 88:
```suggestion
std::string GenericItemProducer::className(const std::type_info &objectType) const
{ return objectType.name(); }
```

This follows from my suggested change to the registration function.  I guess it could be a problem if we're using that exception for flow control elsewhere in the program, but I'm really hoping that we're not.

src/genericitems/serialisers.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+GenericItemSerialiser::GenericItemSerialiser()
+{
+    // PODs
+    registerSerialiser<bool>([](const std::any &a, LineParser &parser, const CoreData &coreData) {
+        return parser.writeLineF("{}\n", DissolveSys::btoa(std::any_cast<bool>(a)));
+    });
+    registerSerialiser<double>([](const std::any &a, LineParser &parser, const CoreData &coreData) {
+        return parser.writeLineF("{}\n", std::any_cast<double>(a));
+    });
```
rprospero, line 24: Further down, I'm going to bring up again the idea of doing the serialisation slightly recursively.    As an example of why I think that this is important, I've noticed that, for a vector of doubles, the values are all formatted `{:16.9e}`, but there's no such formatting for a raw double.  Is this inconsistency intentional?  Using the recursive method would prevent it from occuring.

src/genericitems/serialisers.cpp
```diff
@@ -0,0 +1,199 @@
...
+    registerSerialiser<std::vector<double>>([](const std::any &a, LineParser &parser, const CoreData &coreData) {
+        const auto &v = std::any_cast<const std::vector<double> &>(a);
+        if (!parser.writeLineF("{}\n", v.size()))
+            return false;
+        for (auto &n : v)
+            if (!parser.writeLineF("{:16.9e}\n", n))
+                return false;
+        return true;
+    });
```
rprospero, line 44:
```suggestion
    registerSerialiser<std::vector<double>>([](const std::any &a, LineParser &parser, const CoreData &coreData) {
        const auto &v = std::any_cast<const std::vector<double> &>(a);
        if (!parser.writeLineF("{}\n", v.size()))
            return false;
        for (auto &n : v)
            if (!serialiseObject(n, parser, coreData))
                return false;
        return true;
    });
```

Just as with the deserialisers, I think it could be useful to go slightly recursive here.  I've mentioned before that we already have an inconsistency, plus this would be more useful on some of the later classes (example to follow).

src/genericitems/serialisers.cpp
```diff
@@ -0,0 +1,199 @@
...
+        const auto &v = std::any_cast<const Array2D<std::vector<double>> &>(a);
+        if (!parser.writeLineF("{}  {}  {}\n", v.nRows(), v.nColumns(), DissolveSys::btoa(v.halved())))
+            return false;
+        for (auto &data : v)
+        {
+            if (!parser.writeLineF("{}\n", data.size()))
+                return false;
+            for (auto &n : data)
+                if (!parser.writeLineF("{:16.9e}\n", n))
+                    return false;
```
rprospero, line 106:
```suggestion
            if (!serialiseObject(data, parser, coreData))
                return false;
```

A bigger example of the advantages of the recursive route.

src/genericitems/serialisers.cpp
```diff
@@ -0,0 +1,199 @@
...
+        if (!parser.writeLineF("{}  {}  {}\n", v.nRows(), v.nColumns(), DissolveSys::btoa(v.halved())))
+            return false;
+        for (auto &n : v)
+            if (!n.serialise(parser))
+                return false;
+        return true;
+    });
+    registerSerialiser<AtomTypeList>([](const std::any &a, LineParser &parser, const CoreData &coreData) {
+        return std::any_cast<const AtomTypeList &>(a).serialise(parser);
+    });
```
rprospero, line 121:
```suggestion
    registerSerialiser<AtomTypeList>(simpleSerialise<AtomLtypeList>);
```

Like my suggestion with the deserialiser, this would require making another function

```
template <class T>
bool simpleSerialise(const std::any &a, lineParser &parser, const CoreData &coreData) {
    return std::any_cast<const T &>(a).serialise(parser);
}
```

rprospero (changes requested):
I've made some suggestions that might clean up a bunch of the boilerplate around registering the various helper functions.

src/genericitems/deserialisers.cpp
```diff
@@ -0,0 +1,216 @@
...
+        std::any_cast<double &>(a) = parser.argd(0);
+        return true;
+    });
+    registerDeserialiser<int>([](std::any &a, LineParser &parser, const CoreData &coreData) {
+        if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+            return false;
+        std::any_cast<int &>(a) = parser.argi(0);
+        return true;
+    });
+
```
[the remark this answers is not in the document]
trisyoungs: It's a good suggestion.  I know we talked a long time ago about replacing the custom `LineParser` with a more general lib though, so this probably gets consumed under that as well.

unit/test_genericlist.cpp
```diff
@@ -0,0 +1,49 @@
...
+TEST(GenericListTest, GenericList)
+{
+    GenericList list;
+
+    // POD set / get
+    const auto d = 1.23456;
+    list.realise<double>("DOUBLE") = d;
+    EXPECT_DOUBLE_EQ(d, list.value<double>("DOUBLE"));
+    EXPECT_DOUBLE_EQ(d, list.retrieve<double>("DOUBLE"));
+
```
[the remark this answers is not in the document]
trisyoungs: :+1:

src/genericitems/deserialisers.cpp
```diff
@@ -0,0 +1,247 @@
...
+            data.resize(nItems);
+            for (auto &n : data)
+            {
+                if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+                    return false;
+                n = parser.argd(0);
+            }
+        }
+        return true;
+    });
```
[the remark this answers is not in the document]
trisyoungs: :+1:

src/genericitems/deserialisers.cpp
```diff
@@ -0,0 +1,247 @@
...
+            if (!n.deserialise(parser))
+                return false;
+        return true;
+    });
+    registerDeserialiser<AtomTypeList>([](std::any &a, LineParser &parser, const CoreData &coreData) {
+        return std::any_cast<AtomTypeList &>(a).deserialise(parser, coreData);
+    });
+    registerDeserialiser<Data1D>([](std::any &a, LineParser &parser, const CoreData &coreData) {
+        return std::any_cast<Data1D &>(a).deserialise(parser);
+    });
```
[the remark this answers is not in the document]
trisyoungs: Good suggestion - worth a try!

src/genericitems/producers.h
trisyoungs: Unfortunately, since type names derived from `type_info` are implementation-dependent and often mangled, this would break restart-file compatibility across platforms. So, for consistency the manual naming approach has to remain.

src/genericitems/serialisers.cpp
trisyoungs: No reason for the inconsistency at all, and another benefit to your original suggestion.

src/genericitems/producers.h
rprospero: Good call.  I haven't considered the cross platform issues.

src/genericitems/producers.cpp
rprospero: Obviously not possible the the cross platform issues that we discussed.

src/genericitems/producers.cpp
rprospero: Again, not safely cross-platform.

src/genericitems/deserialisers.cpp
```diff
@@ -0,0 +1,247 @@
...
+        v.clear();
+        v.resize(parser.argi(0));
+        for (auto &n : v)
+        {
+            if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+                return false;
+            n = parser.argd(0);
+        }
+        return true;
+    });
```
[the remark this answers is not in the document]
trisyoungs: Looking at how this change works under the hood, this would involve at the very least retrieving the relevant function from the `deserialiers_` `std::map` - for a single `double` that isn't an issue, but performance may start to suffer if we write containers (of containers) of `double`s this way. Same goes for the serialisers below. I'd probably suggest that we keep the explicit read/write of PODs as they are (with consistent formatting, of course) and utilise the recursion only for `class` objects (e.g. `Array<Data1D>`).

src/genericitems/deserialisers.cpp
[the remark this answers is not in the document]
trisyoungs: This works very nicely.  Had to create two static template functions for the deserialisers, as some don't take the `CoreData` arg, but that's only a minor inconvenience.

src/genericitems/deserialisers.cpp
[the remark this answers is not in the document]
rprospero: That feels like the proper compromise.

src/genericitems/serialisers.cpp
```diff
@@ -0,0 +1,156 @@
...
+/*
+ * Instance
+ */
+
+// Return the serialiser instance
+const GenericItemSerialiser &GenericItemSerialiser::instance()
+{
+    static GenericItemSerialiser _instance;
+
+    return _instance;
```
StephenSmith25, line 132: Nitpicky but I think google coding standards would avoid a leading underscore in a variable name.

StephenSmith25 (approved):
All looks good to me, just some minor things I spotted.

src/modules/epsr/process.cpp
```diff
@@ -224,19 +224,15 @@ bool EPSRModule::process(Dissolve &dissolve, ProcessPool &procPool)
 
     // Loop over target data
     auto rFacTot = 0.0;
-    bool found, created;
+    bool created;
```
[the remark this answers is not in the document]
trisyoungs: Indeed it is not.  Found a few others as well...

src/procedure/nodes/collect2d.cpp
```diff
@@ -113,8 +113,8 @@ bool Collect2DProcedureNode::prepare(Configuration *cfg, std::string_view prefix
     // Construct our data name, and search for it in the supplied list
     std::string dataName = fmt::format("{}_{}_Bins", name(), cfg->niceName());
     bool created;
```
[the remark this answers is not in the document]
trisyoungs: :+1:

src/genericitems/serialisers.cpp
trisyoungs: To be honest, I'm not sure why I used them in the first place! Now removed.

PR #592, opened 2021-03-15
Remove ordered pointer list

StephenSmith25:
Removes the ordered pointer list from the code - replacing the one instance with a std::set of pointers, which equally will be sorted and will not duplicate entries.


src/classes/regionaldistributor.h
```diff
@@ -21,7 +21,6 @@ class RegionalDistributor
     public:
     RegionalDistributor(const std::deque<std::shared_ptr<Molecule>> &moleculeArray, const CellArray &cellArray,
                         ProcessPool &procPool, ProcessPool::DivisionStrategy strategy);
-    ~RegionalDistributor();
```
trisyoungs, line 24: Better to mark this as `default`.

trisyoungs (approved):
Tiny style suggestion, otherwise ready to go!

Issue #597, opened 2021-03-16
Single-item delegate for forcefield parameter editing


trisyoungs:
Instead of using individual items in the tablewidget for intramolecular (and LJ) term parameters, create a delegate that displays all parameters (and their variable names?) as a formatted string, with editing performed through a separate widget.  That way, only a single table item is required. This should also make it easier to handle cases where there are a variable number of parameters for a function - there is currently no way in the GUI to add numbers by hand for this kind of form,

trisyoungs:
This could / should also be combined with the functional form item in order to provide a consistent editing experience.

PR #598, opened 2021-03-16
Bugfix 16/03/21 Table items in SpeciesTab

trisyoungs:
A quick bugfix PR to address a crash when updating SpeciesTab - the problem was caused by the assumption of pre-existing table items for interaction parameters, but these were never guaranteed to be there since we no longer assume a fixe
d maximum number of parameters for each interaction.


rprospero (approved):
Another reason that I need to get onto the gui refactor

trisyoungs:
@rprospero I was thinking the same. This kind of failure / annoyance seems to be coming more frequently. There's also #597 which is kinda related.

PR #599, opened 2021-03-17
Data Name Tidying

trisyoungs:
This PR tidies up the "naming" of the `DataND` classes ahead of their removal from `ObjectStore`.

### Background

There are currently multiple ways to name or tag a `DataND` object:

1. Through a base-class member `name_`, inherited from `PlottableData`
2. Through the `tag_` member of `ObjectInfo`, accessed via an `ObjectStore`
3. Through the association of a name within a `GenericList`
4. Through a display name given when creating a `Renderable` based off of the `DataND`

Obviously, this is horribly confusing. 3) and 4) separates the name from the underlying `DataND` object and should remain as-is, while 2) will be removed once #586 is merged. This leaves 1), which is used inconsistently and variously as:

a. Storing the names of reference data (akin to a matching object `tag_`) within a `DataNDStore`.
b. Storing filenames used when exporting data.
c. To identify the object with a searchable tag

This PR makes the following changes:
- Removes the inherited `name_` member from `PlottableData`
- Adds a `tag_` member to each of the `DataND` to allow tagging with relevant data (e.g. `ObjectInfo`-like names)
- Generate filenames dynamically rather than relying on the stored names
- Reworks the `DataNDStore` classes accordingly, using an `std::list` rather than an `std::vector` to sidestep the need for immediate modernisation of other classes (namely those related to keyword storage).

Related to other modernisation projects, it removes inheritance from `ListItem` from the `DataND` classesi (part of #257), as well as uses of `List` and `RefDataList` from `DataNDStore`.

### Related PR Hierarchy

- #581 Use template-guided renderable creation
- ~#582 No local configuration data~
- ~#586 New GenericList~
- `THIS` Data Name Tidying
- #606 Remove ObjectStore 2

### DIFF
--- a/src/classes/data1dstore.cpp
+++ b/src/classes/data1dstore.cpp
@@ -5,10 +5,6 @@
 #include "base/lineparser.h"
 #include "io/import/data1d.h"
 
-Data1DStore::Data1DStore() {}
-
-Data1DStore::~Data1DStore() {}
-
 /*
  * Data
  */
--- a/src/classes/data1dstore.cpp
+++ b/src/classes/data1dstore.cpp
@@ -14,15 +14,15 @@ bool Data1DStore::addData(std::string_view dataName, LineParser &parser, int sta
                           const CoreData &coreData)
 {
     // Create new data
-    auto &data = data_.emplace_back();
-    data.first.setTag(dataName);
+    auto &[data, format] = data_.emplace_back();
+    data.setTag(dataName);
 
     // Read the file / format
-    if (!data.second.read(parser, startArg, endKeyword, coreData))
+    if (!format.read(parser, startArg, endKeyword, coreData))
         return false;
 
     // Load the data
-    return data.second.importData(data.first, parser.processPool());
+    return format.importData(data.first, parser.processPool());
 }
 
 // Check to see if the named data is present in the store
--- a/src/classes/data1dstore.cpp
+++ b/src/classes/data1dstore.cpp
@@ -14,47 +14,32 @@ bool Data1DStore::addData(std::string_view dataName, LineParser &parser, int sta
                           const CoreData &coreData)
 {
     // Create new data
-    Data1D *data = data_.add();
-    data->setName(dataName);
-
-    // Add reference to data
-    RefDataItem<Data1D, Data1DImportFileFormat> *ref = dataReferences_.append(data);
+    auto &data = data_.emplace_back();
+    data.first.setTag(dataName);
 
     // Read the file / format
-    if (!ref->data().read(parser, startArg, endKeyword, coreData))
+    if (!data.second.read(parser, startArg, endKeyword, coreData))
         return false;
 
     // Load the data
-    return ref->data().importData(*data, parser.processPool());
+    return data.second.importData(data.first, parser.processPool());
 }
 
 // Check to see if the named data is present in the store
 bool Data1DStore::containsData(std::string_view name) const
 {
-    ListIterator<Data1D> dataIterator(data_);
-    while (Data1D *data = dataIterator.iterate())
-        if (DissolveSys::sameString(name, data->name()))
-            return true;
-
-    return false;
+    return std::find_if(data_.begin(), data_.end(), [&name](auto &data) { return data.first.tag() == name; }) != data_.end();
 }
 
 // Return named data
-const Data1D &Data1DStore::data(std::string_view name) const
+OptionalReferenceWrapper<const Data1D> Data1DStore::data(std::string_view name) const
 {
-    ListIterator<Data1D> dataIterator(data_);
-    while (Data1D *xyData = dataIterator.iterate())
-        if (DissolveSys::sameString(name, xyData->name()))
-            return (*xyData);
+    auto it = std::find_if(data_.begin(), data_.end(), [&name](auto &data) { return data.first.tag() == name; });
+    if (it == data_.end())
+        return {};
 
-    static Data1D dummy;
-    Messenger::warn("Data named '{}' was requested from Data1DStore, but it does not exist. Returning an empty Data1D...\n",
-                    name);
-    return dummy;
+    return it->first;
 }
 
-// Return list of all data
-const List<Data1D> &Data1DStore::data() const { return data_; }
-
-// Return list of all data references
-const RefDataList<Data1D, Data1DImportFileFormat> &Data1DStore::dataReferences() const { return dataReferences_; }
+// Return vector of all data
+const std::list<std::pair<Data1D, Data1DImportFileFormat>> &Data1DStore::data() const { return data_; }
--- a/src/classes/data1dstore.cpp
+++ b/src/classes/data1dstore.cpp
@@ -22,7 +22,7 @@ bool Data1DStore::addData(std::string_view dataName, LineParser &parser, int sta
         return false;
 
     // Load the data
-    return format.importData(data.first, parser.processPool());
+    return format.importData(data, parser.processPool());
 }
 
 // Check to see if the named data is present in the store
--- a/src/classes/data1dstore.h
+++ b/src/classes/data1dstore.h
@@ -5,11 +5,8 @@
 
 #include "io/import/data1d.h"
 #include "math/data1d.h"
-#include "templates/list.h"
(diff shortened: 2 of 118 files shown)


src/classes/data1dstore.h
```diff
@@ -5,27 +5,22 @@
 
 #include "io/import/data1d.h"
 #include "math/data1d.h"
-#include "templates/list.h"
-#include "templates/refdatalist.h"
-
-// Forward Declarations
-/* none */
+#include "templates/optionalref.h"
+#include <list>
```
rprospero, line 9: We do need to put an ticket in to keep track of all of the std::list instances and remember to try and eventually get the converted over to vectors.

src/classes/data1dstore.cpp
```diff
@@ -18,47 +14,32 @@ bool Data1DStore::addData(std::string_view dataName, LineParser &parser, int sta
                           const CoreData &coreData)
 {
     // Create new data
-    Data1D *data = data_.add();
-    data->setName(dataName);
-
-    // Add reference to data
-    RefDataItem<Data1D, Data1DImportFileFormat> *ref = dataReferences_.append(data);
+    auto &data = data_.emplace_back();
+    data.first.setTag(dataName);
```
rprospero, line 18:
```suggestion
    auto &[data, format] = data_.emplace_back();
    data.setTag(dataName);
```
Use structured binding

src/classes/data1dstore.cpp
```diff
@@ -18,47 +14,32 @@ bool Data1DStore::addData(std::string_view dataName, LineParser &parser, int sta
...
 
     // Read the file / format
-    if (!ref->data().read(parser, startArg, endKeyword, coreData))
+    if (!data.second.read(parser, startArg, endKeyword, coreData))
```
rprospero, line 21:
```suggestion
    if (!format.read(parser, startArg, endKeyword, coreData))
```

Structured binding

src/classes/data1dstore.cpp
```diff
@@ -18,47 +14,32 @@ bool Data1DStore::addData(std::string_view dataName, LineParser &parser, int sta
...
         return false;
 
     // Load the data
-    return ref->data().importData(*data, parser.processPool());
+    return data.second.importData(data.first, parser.processPool());
```
rprospero, line 25:
```suggestion
    return format.importData(data.first, parser.processPool());
```

Structured binding

src/classes/data2dstore.cpp
```diff
@@ -18,47 +14,32 @@ bool Data2DStore::addData(std::string_view dataName, LineParser &parser, int sta
...
+    data.first.setTag(dataName);
 
     // Read the file / format
-    if (!ref->data().read(parser, startArg, endKeyword, coreData))
+    if (!data.second.read(parser, startArg, endKeyword, coreData))
         return false;
 
     // Load the data
-    return ref->data().importData(*data, parser.processPool());
+    return data.second.importData(data.first, parser.processPool());
```
rprospero, line 25: Worth doing the structured bindings again

src/classes/data3dstore.cpp
```diff
@@ -18,47 +14,32 @@ bool Data3DStore::addData(std::string_view dataName, LineParser &parser, int sta
...
+    data.first.setTag(dataName);
 
     // Read the file / format
-    if (!ref->data().read(parser, startArg, endKeyword, coreData))
+    if (!data.second.read(parser, startArg, endKeyword, coreData))
         return false;
 
     // Load the data
-    return ref->data().importData(*data, parser.processPool());
+    return data.second.importData(data.first, parser.processPool());
```
rprospero, line 25: Structured bindings

src/modules/rdf/functions.cpp
```diff
@@ -703,9 +702,9 @@ bool RDFModule::testReferencePartials(const Data1DStore &testData, double testTh
 
         // Check first argument to check it has the corect prefix
         if (!DissolveSys::sameString(prefix, parser.argsv(0)))
-            return Messenger::error("Unrecognised test data name '{}'.\n", data->name());
+            return Messenger::error("Unrecognised test data name '{}'.\n", data.tag());
 
-        if (!testReferencePartial(partials, testThreshold, *data, parser.argsv(1), parser.hasArg(2) ? parser.argsv(2) : nullptr,
+        if (!testReferencePartial(partials, testThreshold, data, parser.argsv(1), parser.hasArg(2) ? parser.argsv(2) : nullptr,
                                   parser.hasArg(3) ? parser.argsv(3) : nullptr))
```
rprospero, line 708:
```suggestion
        if (!testReferencePartial(partials, testThreshold, data, parser.argsv(1), parser.hasArg(2) ? parser.argsv(2) : "",
                                  parser.hasArg(3) ? parser.argsv(3) : ""))
```

src/modules/rdf/functions.cpp
```diff
@@ -739,11 +737,11 @@ bool RDFModule::testReferencePartials(const Data1DStore &testData, double testTh
...
             setA = false;
         else
-            return Messenger::error("Unrecognised test data name '{}'.\n", data->name());
+            return Messenger::error("Unrecognised test data name '{}'.\n", data.tag());
         const PartialSet &targetSet = (setA ? partialsA : partialsB);
 
-        if (!testReferencePartial(targetSet, testThreshold, *data, parser.argsv(1),
-                                  parser.hasArg(2) ? parser.argsv(2) : nullptr, parser.hasArg(3) ? parser.argsv(3) : nullptr))
+        if (!testReferencePartial(targetSet, testThreshold, data, parser.argsv(1), parser.hasArg(2) ? parser.argsv(2) : nullptr,
+                                  parser.hasArg(3) ? parser.argsv(3) : nullptr))
```
rprospero, line 744:
```suggestion
        if (!testReferencePartial(targetSet, testThreshold, data, parser.argsv(1), parser.hasArg(2) ? parser.argsv(2) : "",
                                  parser.hasArg(3) ? parser.argsv(3) : ""))
```

src/classes/partialset.cpp
```diff
@@ -260,33 +260,37 @@ Data1D PartialSet::unboundTotal(bool applyConcentrationWeights) const
...
-
-    Messenger::printVerbose("Writing total file '{}'...\n", total_.name());
-    Data1DExportFileFormat exportFormat(total_.name());
+    // Set titles for partials
+    std::string title;
+    for_each_pair_early(atomTypes_.begin(), atomTypes_.end(),
+                        [&](int typeI, const AtomTypeData &at1, int typeJ, const AtomTypeData &at2) -> EarlyReturn<bool> {
+                            // Open file and check that we're OK to proceed writing to it
+                            std::string filename{
+                                fmt::format("{}-{}-{}-{}.{}", prefix, tag, at1.atomTypeName(), at2.atomTypeName(), suffix)};
```
rprospero, line 273: Is there usually a prefix?  If the prefix is present in the generic case, then I'm fine with this.  If it's only present in unusual cases, then it might be worth the extra logic to not start the file name with a dash when prefix is ""

rprospero (approved):
Looks good.  Just a couple of minor suggestions

src/classes/data1dstore.h
trisyoungs: Done (#619).  The reason for using `std::list` here was to avoid rewriting the `FileAndFormat` classes, whose ctor implementations don't like being default constructed.

rprospero: I fully appreciate the justification.  I literally just added a new instance of `List` because I'm working on the DataManagerDialog and I didn't want to update too much at one.

src/classes/data3dstore.cpp
trisyoungs: :+1:

src/classes/data2dstore.cpp
trisyoungs: :+1:

src/classes/partialset.cpp
trisyoungs: There is, in fact, always supposed to be a prefix provided to this function - otherwise, as you say, the output filenames start with a `-`- which isn't good for anybody!

I've added an assertion to prevent this kind of unwanted usage.

PR #609, opened 2021-03-22
Bugfix 22/03/21 Species Atoms/Bond Handling

trisyoungs:
A bugfix PR to address a crash observed when deleting an atom (in the EditSPeciesDialog) that was in the middle of the std::list.

Removal of objects from the `std::list` does not invalidate references to the remaining objects, but odd behaviour (crashes) were observed when deleting anything but the last atom in the list.  This was traced to the use of the erase/remove idiom when removing the `SpeciesAtom` from the list - replacing this with find_if/erase causes no such problems, and the container functions as expected.

### DIFF
--- a/src/classes/species_atomic.cpp
+++ b/src/classes/species_atomic.cpp
@@ -38,6 +38,14 @@ void Species::removeAtom(SpeciesAtom *i)
 // Return the number of Atoms in the Species
 int Species::nAtoms() const { return atoms_.size(); }
 
+// Renumber atoms so they are sequential in the list
+void Species::renumberAtoms()
+{
+    auto count = 0;
+    for (auto &i : atoms_)
+        i.setIndex(count++);
+}
+
 // Return the first Atom in the Species
 const SpeciesAtom &Species::firstAtom() const { return atoms_.front(); }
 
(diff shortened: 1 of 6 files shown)


src/classes/species_atomic.cpp
```diff
@@ -24,32 +24,59 @@ void Species::removeAtom(SpeciesAtom *i)
...
+        }
+        else
+            return false;
+    });
+    if (it != bonds_.end())
+        bonds_.erase(it, bonds_.end());
 
     // Now remove the atom
-    atoms_.erase(std::remove_if(atoms_.begin(), atoms_.end(), [&](const auto &p) { return i == &p; }), atoms_.end());
+    auto atomIt = std::find_if(atoms_.begin(), atoms_.end(), [&](const auto &p) { return i == &p; });
```
StephenSmith25, line 46: Interesting that his makes a difference, I will note this for the future! Also after looking at the `std::list` docs I see it has its own `remove_if` method, I wonder if that would make a difference.

StephenSmith25 (approved):
Looks good 👍

src/classes/species_atomic.cpp
trisyoungs: Indeed, the `remove_if` method works as expected - the fact that it's there I suppose hints at the fact that erase/remove doesn't do quite what is expected of the container!

I elected to go with the two-line version as written since `find_if` stops at the first matching item, whereas the `remove_if` method traverses the whole list.  Not a critical performance gain, but wen you know you have unique items in the list, you might as well take that in to account!

Issue #607, opened 2021-03-22
GUI crahses after adding forcefield terms wizard


tfheaden:
**Describe the bug**
GUI crahses after adding forcefield terms wizard (or on occasion has just hung). Works fine for water and ethane, but crashes or hang for anything more complicated (e.g. propane, run as a test). Goes through the wizard fine and seems to correctly work out the intramolecular terms, but crashes when click Finish (or similar - can't remember exact word.

**Operating System:**
 - OS:  Linux  (delete as appropriate)
 - OS version / flavour: Ubuntu 18.04 [e.g. Windows 10, Catalina, Ubuntu 18.04]

trisyoungs:
This looks to be fixed in the latest versions.  @tfheaden can you confirm so the issue can be closed?

PR #610, opened 2021-03-22
Fix iterator invalidation on Species vectors

rprospero:
This is an emergency PR to fix the issues with pointer and references being invalidated when elements are added and remove to the SpeciesBond, SpeciesAngle, etc. lists.  The move to std::list should prevent the pointer and reference invalidation.  To keep things self-contained, this PR does include the kinds of changes from PR #609, but those kinds of changes may be needed late.

### DIFF
--- a/src/classes/species.h
+++ b/src/classes/species.h
@@ -121,13 +121,13 @@ class Species
      */
     private:
     // Array of bonds between atoms in the Species
-    std::list<SpeciesBond> bonds_;
+    std::vector<SpeciesBond> bonds_;
     // Array of angles between atoms in the Species
-    std::list<SpeciesAngle> angles_;
+    std::vector<SpeciesAngle> angles_;
     // Array of torsions between atoms in the Species
-    std::list<SpeciesTorsion> torsions_;
+    std::vector<SpeciesTorsion> torsions_;
     // Array of impropers between atoms in the Species
-    std::list<SpeciesImproper> impropers_;
+    std::vector<SpeciesImproper> impropers_;
     // Whether the attached atoms lists have been created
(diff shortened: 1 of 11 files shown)


trisyoungs (approved):
:+1:

trisyoungs (changes requested):
Hold on - I retract my previous approval!  It isn't the vectors of `SpeciesBond` etc. that are the issue - it's the vectors of `MasterIntra` in the `CoreData` class...

rprospero:
I've just pushed in a patch that should handle the CoreData MasterIntras as
well.  This did involve making some new updater templates.

src/classes/species.h
```diff
@@ -121,13 +121,13 @@ class Species
...
+    std::list<SpeciesBond> bonds_;
     // Array of angles between atoms in the Species
-    std::vector<SpeciesAngle> angles_;
+    std::list<SpeciesAngle> angles_;
     // Array of torsions between atoms in the Species
-    std::vector<SpeciesTorsion> torsions_;
+    std::list<SpeciesTorsion> torsions_;
     // Array of impropers between atoms in the Species
-    std::vector<SpeciesImproper> impropers_;
+    std::list<SpeciesImproper> impropers_;
```
trisyoungs, line 130: These can be reverted back to `std::vector` in order to squeeze the most efficiency out of calculations elsewhere.

trisyoungs (approved):
The `SpeciesBond`, `SpeciesAngle`, `SpeciesTorsion`, and `SpeciesImproper` lists can safely be reverted back to `std::vector`.

Tested this patch and looks good (can close #607) when merging.

PR #616, opened 2021-03-29
NETA Tidying / Modernisation

trisyoungs:
This PR tidies up several aspects of the NETA code, modernising a few elements, and adding a unit test.


rprospero (approved):
I like the switch to the enum classes

PR #617, opened 2021-03-29
Update CMake and add PSTL and TBB to conan

StephenSmith25:
Adds the parallel STL (PSTL) and TBB to conan. The parallel STL is only included for OSX/clang (where Libc++ is yet to implement its own PSTL). This also adds a basic parallel definition file to define the parallel policies. 

This also updates the cmake minimum version to 3.15. In doing so it permits a more "target" based approach.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,4 @@
+cmake_minimum_required(VERSION 3.15)
 project(Dissolve)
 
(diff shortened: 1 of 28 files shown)


trisyoungs (approved):
Looks good. CMake 3.15 is pushing SCARF to its limit (only up to version 3.16 on there) but we can manage!

PR #618, opened 2021-03-29
NETA Geometry

trisyoungs:
This PR adds the capability to match atoms in the NETA context based on their actual bond geometry.

Essentially, it moves the (existing, unused) functionality from the `Forcefield` class and localises it to the `SpeciesAtom` itself, and updates the grammar to recognise a new `option` rule to sit alongside the `modifier` and `flag` rules.  For clarity, these three are distinct because:

- A `modifier` keyword requires a comparison operator and an integer value.
- A `flag` keyword sets an internal option to `true`/`false`, toggling a default option, and takes no arguments.
- An `option` keyword requires a comparison operator (specifically '=' or '!=') followed by a string value (typically one to be converted to an enumeration).

An obvious comment that one can make purely on this descriptive text is "Why is the equivalence operator not `==`?". Well, the whole point of NETA is that it's supposed to be readable from a human perspective, and typically those humans do not know of programming language conventions. Thus, the operator remains as "culturally-accepted" `=`.


StephenSmith25 (approved):
Looks good

PR #626, opened 2021-04-03
ProcedureNode Fix & Tidy

### LINKED ISSUE #20: ProcessND node data access
ProcessND nodes store pointers to their processed data (since its store in a processing data list), and so access to that data through the associated functions will not work until the node has been run, affecting usability within the GUI after loading a restart file, for instance.

### LINKED ISSUE #591: CalculateDAngle module cannot be restarted
**Describe the bug**
Running the system test `calculate_dangle` for the full 95 steps passes correctly, but restarting the test (e.g. running 94 steps followed by the final 1) results in failure.

The reason lays in data being apparently zeroed after being read in from the restart file - e.g. `DAngle(X-H...O)//Process1D//Bulk//RDF(BC)` has no data points present after the `CalculateDAngle` module has been run.

This may be indicative of a wider issue.

trisyoungs:
This PR performs some modernisation and tidying of the various `ProcedureNode` classes, and harmonises their actions a little in order to clarify their output.

### Background

At present, the `ProcedureNode`s are used for one of two purposes, either: 1) generating the contents of a `Configuration`, or; 2) performing some kind of analysis on a `Configuration`. In the latter case, there is a slight mismatch between the raw data generated by some nodes (e.g. `CollectND`), and the processing of that data by other nodes (e.g. `ProcessND`). Each node has (potentially) a set-up stage during which data structures are initialised, an execution stage during which the actual meat of the calculation is done, and a final stage in which summary output is printed, or important processing is performed.

The principal aim of this PR is to make the interplay of finalise stages between nodes more consistent, yielding the correct behaviour and results. Before these fixes, data produced and processed by pairs of `Collect1D` and `Process1D` nodes were working on data from different cycles of the procedure, and running for a single iteration yielded no output of the latter (since the data of the former was out-of-step).

### Resolution

Code previously in the overridden `ProcedureNode::execute()` function has been moved `ProcedureNode::finalise()` for classes that are deemed to be "final" in the procedure, ensuring that data is up-to-date between them.

This PR does not attempt to modernise the container (custom `List`) used for storage of `ProcedureNode`s - while `std::vector<std::shared_ptr>>` seems like an obvious choice, this raises non-trivial problems in the related keyword classes (`NodeKeywordBase`, `NodeKeyword` etc.) and is left for a separate PR.

### Related Issues

Closes #20.
Closes #591.


StephenSmith25 (approved):
Looks good, tidies things up nicely.

PR #627, opened 2021-04-04
Accumulation 1/3 - Data Classes

trisyoungs:
This PR adds `SampledData1D` and `PartialSetAccumulator` classes, and updates `SampledVector`, ahead of introducing the `AccumulateModule`.

### DIFF
--- a/src/classes/partialsetaccumulator.cpp
+++ b/src/classes/partialsetaccumulator.cpp
@@ -1,0 +1,197 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2021 Team Dissolve and contributors
+
+#include "classes/partialsetaccumulator.h"
+#include "base/lineparser.h"
+#include "base/sysfunc.h"
+#include "io/export/data1d.h"
+#include "templates/algorithms.h"
+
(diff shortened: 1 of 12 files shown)


src/classes/partialsetaccumulator.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+    for (auto &partial : array)
+    {
+        if (!parser.writeLine(partial.tag()))
+            return false;
+        if (!partial.serialiseValues(parser))
+            return false;
+    }
+
+    return true;
```
rprospero, line 141:
```suggestion
    return std::all_of(array.begin(), array.end(), [&parser](auto& partial){ return parser.writeLine(partial.tag()) && partial.serialiseValues(parser);});
```

Possible algorithm switch

rprospero (approved):
One minor suggestion, but looks pretty clean.

src/classes/partialsetaccumulator.cpp
trisyoungs: As a caution, we'll stick with the explicit loop as `std::any_of` potentially isn't guaranteed to respect ordering of items in the container.

PR #632, opened 2021-04-13
Bugfix 13/04/21 String Return From Keywords

trisyoungs:
Quick bugfix PR to address corrupt strings in the printed output.

Since all `EnumOptions` objects containing keyword names & descriptions are now returned as full objects (rather than references to static objects) the data they contain is not long-lived. Hence, when printing retrieved `std::string_view`s of keyword names or descriptions, corruption of the text occurred as the `std::string_view`s no longer point to valid `std::string`s.

This PR switches to returning full `std::string`s instead of `std::string_view`s from keyword-related classes.

### DIFF
--- a/src/base/enumoption.h
+++ b/src/base/enumoption.h
@@ -45,9 +45,9 @@ template <class E> class EnumOption
     // Return option enumeration (i.e. from enum value)
     E enumeration() const { return enumeration_; }
     // Return option keyword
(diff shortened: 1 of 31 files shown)


StephenSmith25 (approved):
Looks good - pre-approving as just some builds to get through (clang format is complaining at the moment).

PR #639, opened 2021-04-14
More ListItem Removal

trisyoungs:
This PR comprises a little bit of tidying following on from the recent `remove-objectstore-2` merge, and removes some further uses of `ListItem` (#257).

In summary:
- Small modernisations to the `HistogramND` classes.
- Removal of `ListItem` dependency from `HistogramND` classes.
- Remove single use of `ListVec3` class (in `CellArray`), replacing with `std::vector<Vec3<int>>`.


StephenSmith25 (approved):
👍

PR #646, opened 2021-04-16
Data Manager Reference Point Table Model

rprospero:
This is a simple conversion of the Reference Point Table in the Data Manager Dialog to Model/View.  This is both showing the process of converting an existing table to Model/View with testing and also setting the groudwork for converting the whole dialog to a more testable configuration.


src/gui/datamanagerdialog.h
```diff
@@ -20,14 +22,16 @@ class DataManagerDialog : public QDialog
...
-    DataManagerDialog(QWidget *parent, Dissolve &dissolve, List<ReferencePoint> &referencePoints);
+    DataManagerDialog(QWidget *parent, Dissolve &dissolve, std::vector<ReferencePoint> &referencePoints);
     ~DataManagerDialog();
 
     private:
     // Reference to Dissolve
     Dissolve &dissolve_;
     // List of current ReferencePoints
-    List<ReferencePoint> &referencePoints_;
+    std::vector<ReferencePoint> &referencePoints_;
```
trisyoungs, line 32: I wonder if this even needs to be a separate class - a `std::pair`  of `std::string` would do the job...

src/gui/gui_io.cpp
```diff
@@ -16,10 +16,9 @@ bool DissolveWindow::saveState()
...
 
     // Write reference points
-    ListIterator<ReferencePoint> referencePointIterator(referencePoints_);
-    while (ReferencePoint *refPoint = referencePointIterator.iterate())
+    for (auto &refPoint : referencePoints_)
     {
-        if (!stateParser.writeLineF("ReferencePoint  '{}'  '{}'\n", refPoint->suffix(), refPoint->restartFile()))
+        if (!stateParser.writeLineF("ReferencePoint  '{}'  '{}'\n", refPoint.suffix(), refPoint.restartFile()))
             return false;
     }
```
trisyoungs, line 23: Can remove the extraneous braces here.

unit/gui/test_datamanager.cpp
```diff
@@ -0,0 +1,47 @@
...
+    dissolve.clear();
+    dissolve.loadInput("restart/benzene.txt");
+
+    DataManagerReferencePointModel model(dissolve, points);
+
+    // Expect an error if the file does not exist
+    EXPECT_TRUE(model.addFile("error", "restart/not.exist.restart"));
+
+    // Expect this file to load properly
+    ASSERT_FALSE(model.addFile("benzene", "restart/benzene.txt.restart.test"));
```
trisyoungs, line 37: Shouldn't this be `ASSERT_TRUE`?

trisyoungs (approved):
Looking good.  Only minor comments, so pre-approving for merge.

unit/gui/test_datamanager.cpp
rprospero: This does need to be `ASSERT_FALSE`, since it should *not* return an optional error message.  Similarly, the previous line did need to be `EXPECT_TRUE`, because it does return an error message.

src/gui/datamanagerdialog.h
rprospero: I had the same thought.  Didn't act on it before, but I'm trying it now.

PR #650, opened 2021-04-19
Add GitHub action for benchmarking

StephenSmith25:
This PR adds a GitHub action to store, plot and visualise the benchmark results.

I tested this locally at : https://stephensmith25.github.io/dissolve/dev/bench/

For this to work we'll need to setup a gh-pages branch, mine was a minimal example: 

https://github.com/StephenSmith25/dissolve/tree/gh-pages

I'm not sure if we already have something setup?

**The only thing you need to look at is the benchmark_linux.yml action**

This is built upon the benchmarking branch - so don't merge before thats in.

### DIFF
--- a/.github/workflows/benchmark_linux.yml
+++ b/.github/workflows/benchmark_linux.yml
@@ -38,3 +38,6 @@ jobs:
               output-file-path: benchmark_result.json
               github-token: ${{ secrets.GITHUB_TOKEN }}
               auto-push: true
+              comment-on-alert: true
+              alert-threshold: '200%'
+              alert-comment-cc-users: '@disorderedmaterials/dissolve-devs'
--- a/.github/workflows/benchmark_linux.yml
+++ b/.github/workflows/benchmark_linux.yml
@@ -1,0 +1,40 @@
+name: benchmarks_ubuntu
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark_linux.yml
```diff
@@ -0,0 +1,40 @@
...
+            ninja
+        - name: Run benchmark
+          run: ./build/bin/benchmarks --benchmark_format=json | tee benchmark_result.json
+        - name: Store benchmark result
+          uses: rhysd/github-action-benchmark@v1
+          with:
+              tool: 'googlecpp'
+              output-file-path: benchmark_result.json
+              github-token: ${{ secrets.GITHUB_TOKEN }}
+              auto-push: true
```
trisyoungs, line 40: It feels like we should add the `comment_on_alert` flag as well as set the `alert-comment-cc-users` (this could be dissolve-devs for simplicity).

trisyoungs (approved):
Looks fine to me.  I'm wondering what we should set for the `alert-threshold`, which defaults to 200%. This seems a bit generous, but I suppose we could stick with that for now until we've got a few datapoints under our belt, and we understand what the variance on the benchmark results is likely to be.

PR #653, opened 2021-04-21
Store forces in std::vector

trisyoungs:
Here we tidy up storage of force arrays in multiple places in the code. Previously each component (x, y, and z) of the forces for all atoms were stored in separate `Array<double>` which is convenient for MPI broadcast but annoying for almost everything else.

This PR moves force storage to `std::vector<Vec3<double>>`, and added new broadcast, produce, serialise, and deserialise functions accordingly.

### DIFF
--- a/src/base/processpool.cpp
+++ b/src/base/processpool.cpp
@@ -4,6 +4,7 @@
 #include "base/processpool.h"
 #include "base/messenger.h"
 #include "base/sysfunc.h"
+#include "templates/algorithms.h"
 
 // Static Members
 int ProcessPool::nWorldProcesses_ = 1;
@@ -1207,6 +1208,36 @@ bool ProcessPool::allSum(double *source, int count, ProcessPool::CommunicatorTyp
     return true;
 }
 
+// Reduce (sum) vector of Vec3<double> data to all processes
+bool ProcessPool::allSum(std::vector<Vec3<double>> &source, ProcessPool::CommunicatorType commType)
+{
+#ifdef PARALLEL
+    timer_.start();
+    if ((commType == ProcessPool::GroupLeadersCommunicator) && (!groupLeader()))
+        return true;
+
+    // Construct a contiguous POD buffer that we can send via MPI
+    std::vector<double> buffer(source.size());
+    for (auto n = 0; n < 3; ++n)
+    {
+        // Copy elements (x, y, or z) to buffer
+        for (auto &&[buf, val] : zip(buffer, source))
+            buf = val[n];
+
+        // Reduce
+        if (MPI_Allreduce(source, buffer.data(), buffer.size(), MPI_DOUBLE, MPI_SUM, communicator(commType)) != MPI_SUCCESS)
+            return false;
+
+        // Put reduced data aback into source
+        for (auto &&[buf, val] : zip(buffer, source))
+            val[n] = buf;
+    }
+
+    timer_.accumulate();
+#endif
+    return true;
+}
+
 // Reduce (sum) int data to all processes
 bool ProcessPool::allSum(int *source, int count, ProcessPool::CommunicatorType commType)
 {
@@ -1245,7 +1276,7 @@ bool ProcessPool::allSum(long int *source, int count, ProcessPool::CommunicatorT
     return true;
 }
 
-// Reduce (sum) double data over processes relevant to specifeid strategy
+// Reduce (sum) double data over processes relevant to specified strategy
 bool ProcessPool::allSum(double *source, int count, ProcessPool::DivisionStrategy strategy)
 {
 #ifdef PARALLEL
--- a/src/base/processpool.cpp
+++ b/src/base/processpool.cpp
@@ -1216,21 +1216,24 @@ bool ProcessPool::allSum(std::vector<Vec3<double>> &source, ProcessPool::Communi
     if ((commType == ProcessPool::GroupLeadersCommunicator) && (!groupLeader()))
         return true;
 
-    // Construct a contiguous POD buffer that we can send via MPI
-    std::vector<double> buffer(source.size());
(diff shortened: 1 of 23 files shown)


src/base/processpool.cpp
```diff
@@ -1207,6 +1208,36 @@ bool ProcessPool::allSum(double *source, int count, ProcessPool::CommunicatorTyp
...
+    timer_.start();
+    if ((commType == ProcessPool::GroupLeadersCommunicator) && (!groupLeader()))
+        return true;
+
+    // Construct a contiguous POD buffer that we can send via MPI
+    std::vector<double> buffer(source.size());
+    for (auto n = 0; n < 3; ++n)
+    {
+        // Copy elements (x, y, or z) to buffer
+        for (auto &&[buf, val] : zip(buffer, source))
```
StephenSmith25, line 1224: Can we use the copy constructor / assignment operator
`buffer = source[n];`

src/modules/md/process.cpp
```diff
@@ -106,28 +107,29 @@ bool MDModule::process(Dissolve &dissolve, ProcessPool &procPool)
...
         // Read in or assign random velocities
         // Realise the velocity array from the moduleData
-        auto [v, status] = dissolve.processingModuleData().realiseIf<Array<Vec3<double>>>(
+        auto [velocities, status] = dissolve.processingModuleData().realiseIf<std::vector<Vec3<double>>>(
             fmt::format("{}//Velocities", cfg->niceName()), uniqueName(), GenericItem::NoFlags);
         if (status == GenericItem::ItemStatus::Created)
         {
             randomVelocities = true;
-            v.initialise(cfg->nAtoms());
+            velocities.resize(cfg->nAtoms());
```
StephenSmith25, line 132: Is initialise the same as resize? Or maybe it doesn't matter in this case. But I think initialise cleared everything and resized the container, but resize would just do the latter.

StephenSmith25 (approved):
Looks good, just two small comments

src/base/processpool.cpp
trisyoungs: Unfortunately not here - the source array is basically an N*3 array arising from it containing N copies of `Vec3<double>`. So the loop here is extracting first all the x elements (`val[0]`) into a simple vector, then reducing, then extracting all the y elements (`val[1]`) etc.

One could possibly make some assumptions about memory organisation here and do something lower level, but I prefer to keep things dumb and straightforward where speed isn't necessarily an issue!

src/modules/md/process.cpp
trisyoungs: Good spot.  Missed the value argument to `resize` here.

src/modules/md/functions.cpp
```diff
@@ -28,15 +26,15 @@ int MDModule::capForces(Configuration *cfg, double maxForce, Array<double> &fx,
...
-    fTemp = fz.maxAbs();
-    if (fTemp > fMax)
-        fMax = fTemp;
+    auto fMax = 0.0;
+    for (auto &fxyz : f)
+    {
+        auto m = fxyz.absMax();
+        if (m > fMax)
+            fMax = m;
+    }
```
rprospero, line 37:
```suggestion
    auto fMax = *std::max_element(f.begin(), f.end(), [](auto & left, auto& right){ return left.absMax() < right.absMax(););
```

src/modules/md/process.cpp
```diff
@@ -138,41 +139,40 @@ bool MDModule::process(Dissolve &dissolve, ProcessPool &procPool)
...
         // Calculate instantaneous temperature
         // J = kg m2 s-2  -->   10 J = g Ang2 ps-2
         // If ke is in units of [g mol-1 Angstroms2 ps-2] then must use kb in units of 10 J mol-1 K-1 (= 0.8314462)
         const auto kb = 0.8314462;
         ke = 0.0;
-        for (n = 0; n < cfg->nAtoms(); ++n)
-            ke += 0.5 * mass[n] * v[n].dp(v[n]);
-        tInstant = ke * 2.0 / (3.0 * cfg->nAtoms() * kb);
+        for (auto &&[m, v] : zip(mass, velocities))
+            ke += 0.5 * m * v.dp(v);
```
rprospero, line 170:
```suggestion
        ke = std::transform_reduce(momentum.begin(), momentum.end(), velocity.begin(), 0.0, std::plus<>, 
                                                    [](const auto &m, const auto &v) { return 0.5 * m * v.dp(v); });
```

rprospero (approved):
That really does clear things up.  I have some algorithm suggestions, but I'm not going to fight over any of them.

src/modules/md/functions.cpp
trisyoungs: Good suggestion - just needs a minor tweak to retrieve the `absMax()` element of the returned `Vec3<double>`.

src/modules/geomopt/functions.cpp
```diff
@@ -6,70 +6,45 @@
...
 {
-    for (auto n = 0; n < sp->nAtoms(); ++n)
-    {
-        auto r = sp->atom(n).r();
-        xRef_[n] = r.x;
-        yRef_[n] = r.y;
-        zRef_[n] = r.z;
-    }
+    for (auto &&[ref, i] : zip(rRef_, sp->atoms()))
+        ref = i.r();
```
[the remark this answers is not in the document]
trisyoungs: Probably noticed a theme here. I do like your `zip` algorithm!

src/modules/md/process.cpp
trisyoungs: Odd - I get compiler errors on my local machine along the lines that `std::transform_reduce` doesn't exist.  In the interests of safety and not losing time finding out why, I'll revert this particular change for now.

PR #655, opened 2021-04-22
Add energy benchmarks

StephenSmith25:
This PR adds some initial benchmarks to the energy calculations

Mainly wondering if I've missed anything. Naming could possible be improved, at the moment I'm trying to name the benchmarks as `BM_BENCHMARKMODULE_***`, which means benchmarks from a particular module can be run using:

`./bin/benchmarks --benchmark_filter=BM_CalculateEnergy ` for instance.


StephenSmith25:
Agreed. I've had a go at abstracting out both the problem type and the problem size. Its fairly arbitrary that small is a population of 1k and medium 5k and large 10k, but happy to change them.

trisyoungs (approved):
I think the sizes are OK - for me, they're reasonable reflections of simulations on "modest" computing resources (i.e. not supercomputers). We can consider extending the set to include a "huge" (1e6 atom) simulation, but this is probably not tractable to run on the GH Actions CI.

PR #656, opened 2021-04-22
Data Manager Simulation Table Testing

rprospero:
This converts the Simulation table on the Data Manager dialog over to the Model/View setup.  It also adds a unit test to confirm that the data from the restart file is properly loaded.


src/gui/datamanagerdialog_funcs.cpp
```diff
@@ -25,54 +31,13 @@ DataManagerDialog::~DataManagerDialog() {}
...
-            if (item->isSelected())
-                ui_.SimulationDataTable->setCurrentItem(nullptr);
-
-            ui_.SimulationDataTable->setRowHidden(n, true);
-        }
-    }
+    if (filterText.isEmpty())
+        return;
+    simProxy_.setFilterRegExp(QRegExp(filterText, Qt::CaseInsensitive, QRegExp::Wildcard));
+    simProxy_.setFilterKeyColumn(0);
```
trisyoungs, line 40: Should the passing of an empty filter need to "turn off" the filtering explicitly here?

src/gui/models/dataManagerSimulationModel.h
```diff
@@ -0,0 +1,34 @@
...
+#include <QModelIndex>
+#include <optional>
+#include <vector>
+
+class DataManagerSimulationModel : public QAbstractTableModel
+{
+    Q_OBJECT
+
+    private:
+    // std::vector<Simulation> &referencePoints_;
```
trisyoungs, line 17:
```suggestion
```

trisyoungs (approved):
:+1:

Very happy with this new-look GUI code, and where it''s taking us. Only a couple of small comments to check before merging.

PR #659, opened 2021-04-27
Parallel energy routines

StephenSmith25:
Creates parallel energy routines. This PR does three things:

- Changes energy API to take references rather than shared ptr (for performance and generality)
- Creates generic parallel_reduce algorithms, which can be parallel if we've built with tbb, or otherwise the parallel flag is ingored and a serial reduce will be performed.
- Changes energy routines to use the parallel reduce algorithm.

My results:
[performance_results.pdf](https://github.com/disorderedmaterials/dissolve/files/6383783/performance_results.pdf)


src/templates/algorithms.h
```diff
@@ -169,3 +169,31 @@ template <typename... Args> class zip
     private:
     std::tuple<Args &...> sources_;
 };
+
+namespace dissolve
```
rprospero, line 173: I love this `dissolve` namespace

src/templates/algorithms.h
```diff
@@ -169,3 +169,31 @@ template <typename... Args> class zip
...
+          std::enable_if_t<dissolve::internal::is_execution_policy<ParallelPolicy>::value, bool> = true>
+T transform_reduce(ParallelPolicy policy, Iter begin, Iter end, T initialVal, BinaryOp binaryOp, UnaryOp unaryOp)
+{
+    return std::transform_reduce(policy, begin, end, initialVal, binaryOp, unaryOp);
+}
+
+// Enabled if parallelpolicy is not a real execution policy, i.e. we haven't compiled with multithreading but attempted to set a
+// parallel policy
+template <typename ParallelPolicy, class Iter, typename T, class UnaryOp, class BinaryOp,
+          std::enable_if_t<std::is_same_v<ParallelPolicy, FakeParallelPolicy>, bool> = true>
```
rprospero, line 194: I want to make sure that I understand the logic here.  The idea is that we have one special case that occurs if we give an actual execution policy and another `FakeParallelPolicy` special case where we just doing everything single threaded.  However, it's also possible to put in a type for `ParallelPolicy` which fits neither of these criteria, which results in the code not compiling.

The fact that the code doesn't compile is a *good* thing, because it means that we can add more special cases in the future.  If we simply made the second condition the inverse of the first (`= false`), then it would be harder to add more policies going forward.

Is that the gist of it?

src/classes/energykernel.cpp
```diff
@@ -547,22 +537,22 @@ double EnergyKernel::energy(std::shared_ptr<const Molecule> mol, ProcessPool::Di
...
-        scale = 1.0 - i->scaling(j);
-        if (scale <= 1.0e-3)
-            return acc;
-        r = box_->minimumDistance(rI, j->r());
-        return acc + pairPotentialEnergy(i, j, r) * scale;
-    });
+    auto &atoms = i.molecule()->atoms();
+    const auto &rI = i.r();
+
+    double correctionEnergy = dissolve::transform_reduce(ParallelPolicies::par, atoms.begin(), atoms.end(), 0.0,
```
rprospero, line 546: I'm glad that you pushed for `transform_reduce` here.  It's so much cleaner than accumulate

rprospero (approved):
I love everything about this PR.

src/classes/energykernel.cpp
```diff
@@ -48,16 +44,14 @@ double EnergyKernel::energyWithMim(const std::shared_ptr<Atom> i, const std::sha
  */
 
 // Return PairPotential energy between atoms (provided as pointers)
```
trisyoungs, line 46: Need a quick search/replace to change the function comments to reflect the change in arg types.

src/classes/energykernel.cpp
```diff
@@ -48,16 +44,14 @@ double EnergyKernel::energyWithMim(const std::shared_ptr<Atom> i, const std::sha
...
-double EnergyKernel::energy(const std::shared_ptr<Atom> i, const std::shared_ptr<Atom> j, bool applyMim, bool excludeIgeJ)
+double EnergyKernel::energy(const Atom &i, const Atom &j, bool applyMim, bool excludeIgeJ)
 {
-    assert(i && j);
-
     // If Atoms are the same, we refuse to calculate
-    if (i == j)
+    if (&i == &j)
         return 0.0;
```
trisyoungs, line 51: I know this is only a simple check, but it's in a routine that get's hammered quite a lot.  If our loop constructions going over atom pairs automatically exclude this pair then this could be replaced with an assertion.  Might be best left to a follow-up PR, as it probably requires a new implementation of `for_each_pair` that excludes the diagonal i == j elements (`for_each_pair_off`?).

src/classes/energykernel.cpp
```diff
@@ -581,16 +571,21 @@ double EnergyKernel::energy(const CellArray &cellArray, bool interMolecular, Pro
...
-    for (auto cellId = begin; cellId < end; ++cellId)
-    {
-        cell = cellArray.cell(cellId);
+    dissolve::counting_iterator<int> countingIterator(begin, end);
 
-        // This cell with itself
-        totalEnergy += energy(cell, cell, false, true, interMolecular, subStrategy, performSum);
+    totalEnergy = dissolve::transform_reduce(
+        ParallelPolicies::par, countingIterator.begin(), countingIterator.end(), 0.0, std::plus<double>(), [&](int i) {
+            double localEnergy = 0.0;
```
trisyoungs, line 578:
```suggestion
            auto localEnergy = 0.0;
```

src/templates/algorithms.h
trisyoungs: I was going to write a comment here about casing of the namespace, but decided to shut up instead!  For such low-level  functionality, consistency with e.g. `std` makes sense.

trisyoungs (approved):
Lots to digest in there, but this is great!

src/templates/algorithms.h
StephenSmith25: Correct, my goal was to be able to write the following for both the single and multithreaded case.

`
transform_reduce(ParallelPolicy::par, ...)
`

When we configure CMake with multithreading the ParallelPolicy will be the correct STL variant, but when we have no multithreading it will be substituted for our fake policy and call the single threaded base function. Essentially, it's providing a hint that we want this algorithm to be parallel if tbb is available.

src/classes/energykernel.cpp
StephenSmith25: I agree, an additional algorithm (or template the original) to ignore diagonal terms would be good. I suppose we could make it optionally multithreaded  as well, like the `dissolve::transform_reduce`

PR #660, opened 2021-04-27
UFF4MOF Testing

### LINKED ISSUE #561: Implement UFF-MOF Forcefield
Extend, derive from, or otherwise enhance the present UFF implementation to be capable of simulating MOFs.

**Additional details**
Paper can be found here: https://pubs.acs.org/doi/abs/10.1021/ct400952t

trisyoungs:
This PR implements the UFF4MOF forcefield (Addicoat 2014).  The base UFF forcefield on which it is based has been modified in order to make it suitable for subclassing and building on, since the current UFF4MOF forcefield only adds parameters to the existing set.

Closes #561.


DanNixon (approved):
I do wonder if there is a nicer way to define atom types that avoid the "wall of values" that they currently are, which is not a huge issue, but something like a thin wrapper would 1. make it obvious what each numerical value is without having to check the function signature, 2. increase type safety. Not a significant issue for right now though.

trisyoungs:
Valid point. I get a bit spoiled by my IDE labelling each parameter in-place. Not sure how you'd remove such ambiguity without having individual wrapper classes for each param tho?

PR #664, opened 2021-05-04
Parallel force calculations

StephenSmith25:
This PR adds parallel force calculations and benchmarks. Performance look good, just comparing three of the macro benchmarks:

On my macbook pro quad core.
 |  Benchmark  |  Single threaded  |  Multithreaded | 
 |  ------------ |  ----------------- |  ------------- | 
 | BM_CalculateForces_TotalInterAtomic |  1524ms |  341ms
 | BM_CalculateForces_TotalIntraMolecular  |   42.4ms | 9.27ms
 | BM_CalculateForces_TotalForces |  1605 ms |  355ms

This PR also restructed the force kernel:

- References instead of shared ptr
- Extract force vector from class, now we can create a thread local force vector and combine at the end of parallel computations
- Remove storage of torision and angle parameters -> makes it easier to parallelize
- Most methods are now const -> implying some degree of thread safety

### DIFF
--- a/src/classes/forcekernel.cpp
+++ b/src/classes/forcekernel.cpp
@@ -23,9 +23,8 @@ ForceKernel::ForceKernel(ProcessPool &procPool, const Box *box, const PotentialM
  */
 
 // Calculate PairPotential forces between Atoms provided (no minimum image calculation)
-void ForceKernel::forcesWithoutMim(const Atom &i, const Atom &j, std::optional<double> scaleOpt, ForceVector &f) const
+void ForceKernel::forcesWithoutMim(const Atom &i, const Atom &j, ForceVector &f, double scale) const
 {
-    auto scale = scaleOpt.value_or(1.00);
     auto force = j.r() - i.r();
     auto distanceSq = force.magnitudeSq();
     if (distanceSq > cutoffDistanceSquared_)
@@ -38,9 +37,8 @@ void ForceKernel::forcesWithoutMim(const Atom &i, const Atom &j, std::optional<d
 }
 
 // Calculate PairPotential forces between Atoms provided (minimum image calculation)
-void ForceKernel::forcesWithMim(const Atom &i, const Atom &j, std::optional<double> scaleOpt, ForceVector &f) const
+void ForceKernel::forcesWithMim(const Atom &i, const Atom &j, ForceVector &f, double scale) const
 {
-    auto scale = scaleOpt.value_or(1.00);
     auto force = box_->minimumVector(i, j);
     auto distanceSq = force.magnitudeSq();
     if (distanceSq > cutoffDistanceSquared_)
@@ -69,9 +67,9 @@ void ForceKernel::forces(const Atom &i, const Atom &j, bool applyMim, bool exclu
         return;
 
     if (applyMim)
-        forcesWithMim(i, j, std::nullopt, f);
+        forcesWithMim(i, j, f);
     else
-        forcesWithoutMim(i, j, std::nullopt, f);
+        forcesWithoutMim(i, j, f);
 }
 
 // Calculate forces between atoms in supplied cells
@@ -107,12 +105,12 @@ void ForceKernel::forces(Cell *centralCell, Cell *otherCell, bool applyMim, bool
 
                 // Check for atoms in the same Molecule
                 if (molI != jj->molecule())
-                    forcesWithMim(*ii, *jj, std::nullopt, f);
+                    forcesWithMim(*ii, *jj, f);
                 else
                 {
                     double scale = ii->scaling(jj);
                     if (scale > 1.0e-3)
-                        forcesWithMim(*ii, *jj, scale, f);
+                        forcesWithMim(*ii, *jj, f, scale);
                 }
             }
         }
@@ -135,12 +133,12 @@ void ForceKernel::forces(Cell *centralCell, Cell *otherCell, bool applyMim, bool
 
                 // Check for atoms in the same molecule
                 if (molI != jj->molecule())
-                    forcesWithoutMim(*ii, *jj, std::nullopt, f);
+                    forcesWithoutMim(*ii, *jj, f);
                 else
                 {
                     double scale = ii->scaling(jj);
                     if (scale > 1.0e-3)
-                        forcesWithoutMim(*ii, *jj, scale, f);
+                        forcesWithoutMim(*ii, *jj, f, scale);
                 }
             }
         }
@@ -185,12 +183,12 @@ void ForceKernel::forces(const Atom &i, Cell *cell, int flags, ProcessPool::Divi
 
                 // Check for atoms in the same species
                 if (moleculeI != jj->molecule())
-                    forcesWithMim(i, *jj, std::nullopt, f);
+                    forcesWithMim(i, *jj, f);
                 else
                 {
                     double scale = i.scaling(jj);
                     if (scale > 1.0e-3)
-                        forcesWithMim(i, *jj, std::nullopt, f);
+                        forcesWithMim(i, *jj, f);
                 }
             }
         else if (flags & KernelFlags::ExcludeIGEJFlag)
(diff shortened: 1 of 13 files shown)


src/classes/forcekernel.h
```diff
@@ -36,98 +38,107 @@ class ForceKernel
...
-
-    private:
-    // Calculate angle force parameters from supplied vectors, storing result in local class variables
-    void calculateAngleParameters(Vec3<double> vecji, Vec3<double> vecjk);
-    // Calculate torsion force parameters from supplied vectors, storing result in local class variables
-    void calculateTorsionParameters(const Vec3<double> vecji, const Vec3<double> vecjk, const Vec3<double> veckl);
-    // Sum torsion forces for atom 'i' in 'i-j-k-l' into the specified vector index
-    void sumTorsionForceI(double du_dphi, int index);
+    // Add torsion forces for atom 'i' in 'i-j-k-l' into the specified vector index
+    void addTorsionForceI(double du_dphi, int index, TorsionParameters &torisionParameters, ForceVector &f) const;
```
trisyoungs, line 97: Type here `torisionParameters` -> `torsionParameters`.

src/classes/forcekernel.h
```diff
@@ -36,98 +38,107 @@ class ForceKernel
...
     // Sum torsion forces for atom 'j' in 'i-j-k-l' into the specified vector index
-    void sumTorsionForceJ(double du_dphi, int index);
+    void addTorsionForceJ(double du_dphi, int index, TorsionParameters &torisionParameters, ForceVector &f) const;
     // Sum torsion forces for atom 'k' in 'i-j-k-l' into the specified vector index
-    void sumTorsionForceK(double du_dphi, int index);
+    void addTorsionForceK(double du_dphi, int index, TorsionParameters &torisionParameters, ForceVector &f) const;
     // Sum torsion forces for atom 'l' in 'i-j-k-l' into the specified vector index
-    void sumTorsionForceL(double du_dphi, int index);
+    void addTorsionForceL(double du_dphi, int index, TorsionParameters &torisionParameters, ForceVector &f) const;
```
trisyoungs, line 103: And here.

src/classes/forcekernel.h
```diff
@@ -36,98 +38,107 @@ class ForceKernel
...
     /*
      * Internal Force Calculation
      */
     private:
     // Calculate inter-particle forces between Atoms provided (no minimum image calculation)
-    void forcesWithoutMim(const std::shared_ptr<Atom> i, const std::shared_ptr<Atom> j, double scale = 1.0);
+    void forcesWithoutMim(const Atom &i, const Atom &j, std::optional<double>, ForceVector &f) const;
     // Calculate inter-particle forces between Atoms provided (minimum image calculation)
-    void forcesWithMim(const std::shared_ptr<Atom> i, const std::shared_ptr<Atom> j, double scale = 1.0);
+    void forcesWithMim(const Atom &i, const Atom &j, std::optional<double>, ForceVector &f) const;
```
trisyoungs, line 49: Can we reinstate the variable names (`scale`) on the `optional` parameters. I get easily confused!

src/classes/potentialmap.cpp
```diff
@@ -127,6 +127,16 @@ double PotentialMap::force(const std::shared_ptr<Atom> i, const std::shared_ptr<
...
+// Return force between Atoms at distance specified
+double PotentialMap::force(const Atom &i, const Atom &j, double r) const
+{
+    // Check to see whether Coulomb terms should be calculated from atomic charges, rather than them being included in the
+    // interpolated potential
+    auto *pp = potentialMatrix_[{i.masterTypeIndex(), j.masterTypeIndex()}];
+    return pp->includeCoulomb()
+               ? pp->force(r)
+               : pp->force(r) + pp->analyticCoulombForce(i.speciesAtom()->charge() * j.speciesAtom()->charge(), r);
+}
```
trisyoungs, line 139: If this completely replaces the `const std::shared_ptr<Atom>` version then that can be removed.

src/classes/potentialmap.cpp
```diff
@@ -140,6 +150,15 @@ double PotentialMap::force(const SpeciesAtom *i, const SpeciesAtom *j, double r)
...
 
+// Return force between SpeciesAtoms at distance specified
+double PotentialMap::force(const SpeciesAtom &i, const SpeciesAtom &j, double r) const
+{
+    // Check to see whether Coulomb terms should be calculated from atomic charges, rather than them being included in the
+    // interpolated potential
+    auto *pp = potentialMatrix_[{i.atomType()->index(), j.atomType()->index()}];
+    return pp->includeCoulomb() ? pp->force(r) : pp->force(r) + pp->analyticCoulombForce(i.charge() * j.charge(), r);
+}
+
```
trisyoungs, line 161: Same here.

src/modules/forces/process.cpp
```diff
@@ -229,9 +228,14 @@ bool ForcesModule::process(Dissolve &dissolve, ProcessPool &procPool)
                         veckl = box->minimumVector(l, k);
 
                         // Calculate torsion force parameters
-                        ForceKernel::calculateTorsionParameters(vecji, vecjk, veckl, phi, dxpj_dij, dxpj_dkj, dxpk_dkj,
-                                                                dxpk_dlk, dcos_dxpj, dcos_dxpk);
-                        du_dphi = torsion.force(phi * DEGRAD);
+                        auto torisionParameters = ForceKernel::calculateTorsionParameters(vecji, vecjk, veckl);
```
trisyoungs, line 231: `torision`.

src/modules/forces/process.cpp
```diff
@@ -229,9 +228,14 @@ bool ForcesModule::process(Dissolve &dissolve, ProcessPool &procPool)
...
+                        auto dxpj_dij = torisionParameters.dxpj_dij_;
+                        auto dxpj_dkj = torisionParameters.dxpj_dkj_;
+                        auto dxpk_dkj = torisionParameters.dxpk_dkj_;
+                        auto dxpk_dlk = torisionParameters.dxpk_dlk_;
+                        auto dcos_dxpj = torisionParameters.dcos_dxpj_;
+                        auto dcos_dxpk = torisionParameters.dcos_dxpk_;
+                        du_dphi = torsion.force(torisionParameters.phi_ * DEGRAD);
```
trisyoungs, line 238: Question - is there any (performance) reason to prefer not copying / referencing these parameters out of the `torsionParameters` object, and just using the object, apart from readability? A shorter name for the object (e.g. `tp`) would help alleviate that.

src/classes/forcekernel.cpp
```diff
@@ -271,161 +261,161 @@ void ForceKernel::forces(const std::shared_ptr<Atom> i, Cell *cell, int flags, P
...
 }
 
 /*
  * Intramolecular Terms
  */
 
-// Calculate angle force parameters from supplied vectors, storing result in local class variables
-void ForceKernel::calculateAngleParameters(Vec3<double> vecji, Vec3<double> vecjk)
+// Add torsion forces for atom 'i' in 'i-j-k-l' into the specified vector index and input vector
+void ForceKernel::addTorsionForceI(double du_dphi, int index, ForceKernel::TorsionParameters &torisionParameters,
```
trisyoungs, line 361: Mr `torision` again!

src/classes/forcekernel.h
trisyoungs: Also, it makes more sense to me to have the `optional` arg as the last parameter, since we're always specifying the `ForceVector`.

trisyoungs (changes requested):
Looking really good. Only minor comments.

src/modules/forces/process.cpp
StephenSmith25: No good reason, both should give the same performance. I got a bit lazy and for some reason was utterly hopeless at writing the word `torsion`.. I've changed it to `tp` and removed these local variables.

trisyoungs (approved):

PR #668, opened 2021-05-08
Bugfix 08/05/21 Data Export

trisyoungs:
Bugfix PR to allow analysis modules to save out data to a specific filename, rather than an automatically generated one which was potentially quite similar / identical across analysis modules.
Also fixes format of saved accumulation data files, which missed 50% of the intended data!


DanNixon (approved):
:+1: 

How much effort would it be to add some testing of the accumulation file to existing tests?

trisyoungs:
Not hard, but awkward since it's a "non-standard" output file in Dissolve's language (it isn't a recognised format in the `Data1D` I/O class. In that sense, there are no checks to ensure that data is read / written consistently by those classes, so I'll add a general issue to cover this.

PR #657, opened 2021-04-24
Remove Array 1

trisyoungs:
This is an exceedingly uninteresting PR which removes a handful of uses of the custom `Array` class and moves them to `std::vector`, within the "bigger picture" view of going all STL/parallel.

Here we also remove some "dead code" classes that were either unused (steepest descent and simplex minimisers) or pointless wrappers around simple data (`CoordinateSet`).

Further removal of `Array` use is left for a second commit in order to keep reviewers from falling asleep on their desks.

PR #670, opened 2021-05-08
Remove Array 2

trisyoungs:
## Not ready for review until #657 is merged

This PR is part of general work to remove the custom `Array<T>` class.

Here, site-related classes, `ColourScale`, and `Primitive` classes are addressed.


rprospero (approved):
Just one suggestion, but it's not mandatory

PR #671, opened 2021-05-08
Remove Array 3 (Isotopes)

trisyoungs:
## Not ready for review until #657 and #670 are merged.

This PR is part of general work to remove the custom `Array<T>` class.

Here we address the `Isotope` storage class, removing uses of `Array<T>`, undoing some archaic practices, and storing an enumeration to an `Isotope` rather than a pointer throughout the code. The isotope data is also placed in a namespace, in alignment with other similar classes.

### DIFF
--- a/src/gui/delegates/isotopecombo_funcs.cpp
+++ b/src/gui/delegates/isotopecombo_funcs.cpp
@@ -6,6 +6,8 @@
 #include "templates/list.h"
 #include "templates/variantpointer.h"
 
+Q_DECLARE_METATYPE(Sears91::Isotope);
+
 IsotopeComboDelegate::IsotopeComboDelegate(QObject *parent) : QItemDelegate(parent) {}
 
 IsotopeComboDelegate::~IsotopeComboDelegate() {}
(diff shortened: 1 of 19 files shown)


src/gui/delegates/isotopecombo_funcs.cpp
```diff
@@ -16,17 +18,13 @@ QWidget *IsotopeComboDelegate::createEditor(QWidget *parent, const QStyleOptionV
...
-            editor->addItem(textForIsotope(tope));
-    }
-    else
-        Messenger::error("IsotopeComboDelegate::createEditor() - Did not find an Isotope* in the associated QModelIndex.\n");
+    // Get the data from the model index
+    auto isotope = index.data(Qt::UserRole).value<Sears91::Isotope>();
+    auto isotopes = Sears91::isotopes(Sears91::Z(isotope));
+
+    // Populate combo with all possible Isotopes for this element
+    for (auto tope : isotopes)
```
rprospero, line 26: Should this be a reference or possibly a const reference?

rprospero (approved):
A couple minor notes, but much cleaner than what we had before.

src/gui/delegates/isotopecombo_funcs.cpp
trisyoungs: Could be a reference, but the object is just an enumerated isotope, so a copy of basically a POD seems enough to me.

PR #679, opened 2021-05-10
Small bug fix in parallel computations + Clean up parallel GR implementation

StephenSmith25:
This PR fixes a small bug in the parallel computations when distributing the loop combinations between MPI processors. It also cleans up the GR cells implementation by removing unnecessary/boilerplate code using the developments made in the other PRs.

### DIFF
--- a/src/modules/rdf/functions.cpp
+++ b/src/modules/rdf/functions.cpp
@@ -171,9 +171,9 @@ bool RDFModule::calculateGRCells(ProcessPool &procPool, Configuration *cfg, Part
 
     auto combinableHistograms = dissolve::CombinableValue<Array2D<Histogram1D>>([&partialSet]() {
         Array2D<Histogram1D> histograms;
-        histograms.initialise(partialSet.nAtomTypes(), partialSet.nAtomTypes());
+        histograms.initialise(partialSet.nAtomTypes(), partialSet.nAtomTypes(), true);
         for (auto i = 0; i < partialSet.nAtomTypes(); ++i)
-            for (auto j = 0; j < partialSet.nAtomTypes(); ++j)
+            for (auto j = i; j < partialSet.nAtomTypes(); ++j)
                 histograms[{i, j}] = partialSet.fullHistogram(i, j);
         return histograms;
     });
--- a/src/templates/array2d.h
+++ b/src/templates/array2d.h
@@ -243,11 +243,13 @@ template <class A> class Array2D
     Array2D<A> operator+(const Array2D<A> &other) const
     {
         assert(nColumns_ == other.nColumns_ && nRows_ == other.nRows_);
-        Array2D<A> ret(nRows_, nColumns_);
-        // Could optimize with transform and zip iterator over this and other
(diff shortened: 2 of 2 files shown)


src/modules/rdf/functions.cpp
```diff
@@ -152,141 +158,65 @@ bool RDFModule::calculateGRSimple(ProcessPool &procPool, Configuration *cfg, Par
...
     auto nChunks = procPool.interleavedLoopStride(ProcessPool::PoolStrategy);
+    auto [cStart, cEnd] = chop_range(0, comb.getNumCombinations(), nChunks, offset);
+
+    auto combinableHistograms = dissolve::CombinableValue<Array2D<Histogram1D>>([&partialSet]() {
+        Array2D<Histogram1D> histograms;
+        histograms.initialise(partialSet.nAtomTypes(), partialSet.nAtomTypes());
+        for (auto i = 0; i < partialSet.nAtomTypes(); ++i)
+            for (auto j = 0; j < partialSet.nAtomTypes(); ++j)
+                histograms[{i, j}] = partialSet.fullHistogram(i, j);
+        return histograms;
```
trisyoungs, line 178: This could be the halved array (`histograms.initialise(partialSet.nAtomTypes(), partialSet.nAtomTypes(), true)`) if you so wished, since the partial for (i,j) is exactly equal to the partial for (j,i).  Nothing really to gain in speed since this is only the final combiner, but you could save a chunk of memory for systems with a lot of different atom types.

src/modules/rdf/functions.cpp
```diff
@@ -152,141 +158,65 @@ bool RDFModule::calculateGRSimple(ProcessPool &procPool, Configuration *cfg, Par
...
-    auto combinableHistograms = dissolve::combinable<RDFModuleHelpers::PartialHistograms>(
-        [&partialSet]() { return RDFModuleHelpers::PartialHistograms(partialSet); });
+        auto &atomsI = cellI->atoms();
+        auto &atomsJ = cellJ->atoms();
 
-    auto lambda = [&combinableHistograms, start, cfg, &comb, rdfRange](const auto &range) {
-        for (auto i = range.begin(); i < range.end(); ++i)
+        // Perform minimum image calculation on all atom pairs -
+        // quicker than working out if we need to given the absence of a 2D look-up array
+        for (auto &i : atomsI)
```
trisyoungs, line 200: The more I see old comments of mine like this, the more I want to rewrite `CellArray` to be more useful than it currently is. There is a function `CellArray::minimumImageRequired()` which would tell you if you need to do the necessary calculation between atoms in cells I and J, but it is pretty ugly.  An alternative would be to use the `Cell::mimRequired()` function.  One simple check may give a decent performance boost now that the main function has been streamlined.

trisyoungs (approved):
Now looking real neat and tidy - certainly easier to understand from casual reading in this form!  Had a couple of comments to make on the minimum image checking of the algorithm, but any significant investigation on this can be left for a separate issue / PR for the future.

PR #681, opened 2021-05-11
Apply clang tidy

StephenSmith25:
This PR runs clang tidy with the following options

- Override usage
- Modernize
- Performance

It also fixes an issue where we were capturing the address of a stack variable.


src/classes/atom.cpp
```diff
@@ -2,8 +2,10 @@
 // Copyright (c) 2021 Team Dissolve and contributors
 
 #include "classes/atom.h"
+
```
trisyoungs, line 5: Random addition?

src/data/ff/angleterm.cpp
```diff
@@ -6,7 +6,7 @@
 #include "data/ff/ff.h"
 
 ForcefieldAngleTerm::ForcefieldAngleTerm(std::string_view typeI, std::string_view typeJ, std::string_view typeK,
-                                         SpeciesAngle::AngleFunction form, const std::vector<double> parameters)
+                                         SpeciesAngle::AngleFunction form, const std::vector<double> &parameters)
```
trisyoungs, line 9: I wouldn't have thought having a default argument (an empty reference) to this would be possible? Obviously I'm missing something subtle...

trisyoungs (approved):
I've spun through the larger commits quite quickly, but this all looks good.  One point for discussion in the TSC on Friday perhaps would be whether to add more stringent `clang-tidy` testing in the CI, to enforce the style that has just been applied here.

src/classes/atom.cpp
StephenSmith25: Looks like it, I wonder why its done that 🤔

src/data/ff/angleterm.cpp
StephenSmith25: This is a bit strange and likely doesn't save us anything - Just clang really doesn't like to see copies of things such as vectors. But I think what happens is our const reference binds to the temporary vector constructed by the default argument (which is legal as the temporary will live as long as the function call). We subsequently take a copy of the vector within the function so the lifetime issues are avoided.

src/classes/atom.cpp
trisyoungs: Keep us on our toes!  When I first saw this I wondered if you were using a newer version of `clang-format` that decided to separate the "closely-related" include for the current file from the rest, but this didn't seem to be the case for any other file...

src/data/ff/angleterm.cpp
trisyoungs: That makes sense.

PR #682, opened 2021-05-12
Improve fallback implementation of counting iterator

StephenSmith25:
This PR improves the fallback implementation of the counting iterator. Previously it just created a vector and used that for iteration. If we have lots of atoms and therefore lots of combinations, the vector could have millions of elements.

### DIFF
--- a/src/modules/forces/functions.cpp
+++ b/src/modules/forces/functions.cpp
@@ -43,7 +43,6 @@ void ForcesModule::interAtomicForces(ProcessPool &procPool, Configuration *cfg, 
     auto [begin, end] = chop_range(0, cellArray.nCells(), stride, start);
 
     // algorithm parameters
-    dissolve::counting_iterator<int> countingIterator(begin, end);
     auto unaryOp = [&combinableForces, &kernel, &cellArray, strategy](const int id) {
         auto *cellI = cellArray.cell(id);
         auto &fLocal = combinableForces.local();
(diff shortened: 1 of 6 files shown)


src/templates/tbb-fallbacks.h
```diff
@@ -2,26 +2,71 @@
...
+    // Operators : misc
+    reference operator*() { return counter_; }
+    value_type operator[](difference_type i) const { return *(*this + i); };
+    difference_type operator-(const counting_iterator &it) const { return counter_ - it.counter_; };
+    counting_iterator &operator+=(difference_type forward)
+    {
+        counter_ += forward;
+        return *this;
+    }
+    counting_iterator &operator-=(difference_type backward) { return *this += -backward; };
```
trisyoungs, line 32: This is more curiosity coming from my lack of experience with TBB, but why are the function bodies of these two operators constructed differently?

trisyoungs (approved):
:+1:

One comment on the `+=` and `-=` operators, but I'm probably thinking too much about how this works in-place.

src/templates/tbb-fallbacks.h
StephenSmith25: The second one is implemented in terms of the first one, gotta save that one line of code

src/modules/forces/functions.cpp
```diff
@@ -149,8 +149,7 @@ void ForcesModule::intraMolecularForces(ProcessPool &procPool, Configuration *cf
 
     // Loop over supplied atom indices
     const auto &atoms = cfg->atoms();
-    auto [begin, end] = chop_range(0, (int)targetIndices.size(), stride, start);
-    dissolve::counting_iterator<int> countingIterator(begin, end);
+    auto [begin, end] = chop_range(0, static_cast<int>(targetIndices.size()), stride, start);
```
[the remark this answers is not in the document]
StephenSmith25: I think it would be worth looking into, with our increasing use of standard containers we are going to have to deal with more and more unsigned integers.

src/templates/tbb-fallbacks.h
trisyoungs: Doh! Yep, read far too much into that.

PR #684, opened 2021-05-13
Const Species

trisyoungs:
This PR enforces storage of only `const Species` pointers in the `SpeciesKeyword` class and elsewhere, meaning that `Species` are read-only within the main workflow modules, as should be the case.

We also remove the `SpeciesInfo` class (replaced with a `std::pair<const Species *,int>` and unused `UsedSpeciesComboDelegate` gui class.

Some functions have been renamed for clarity and consistency with the rest of the codebase (e.g. `SpeciesAtom::hasBond()` returned a reference to a bond, so has been renamed `getBond`).


src/classes/species.h
```diff
@@ -144,12 +144,13 @@ class Species
     std::vector<SpeciesBond> &bonds();
     const std::vector<SpeciesBond> &bonds() const;
     // Return whether SpeciesBond between SpeciesAtoms exists
-    bool hasBond(SpeciesAtom *i, SpeciesAtom *j) const;
+    bool getBond(SpeciesAtom *i, SpeciesAtom *j) const;
```
StephenSmith25, line 147: It feels to me the method returning a boolean should still be called `hasBond()`, as this doesn't really get anything.

StephenSmith25 (approved):
Looks great - the more const things the better!

PR #685, opened 2021-05-14
Speciestab model

### LINKED ISSUE #600: Move SpeciesTab tables onto Model/View pattern
The tables in SpeciesTab have been a source of unexpected bugs.  We want to move them onto the Model/View pattern both to improve reliability and to enable testing.

rprospero:
This PR moves all of the tables *except the Isotopologues* into a model/view system and adds unit tests.  There should be not behaviour changes from this, but it's worth a quick review of the behaviour.

This should close #600


src/gui/models/speciesAngleModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+int SpeciesAngleModel::rowCount(const QModelIndex &parent) const
+{
+    Q_UNUSED(parent);
+    return angles_.size();
+}
+
+int SpeciesAngleModel::columnCount(const QModelIndex &parent) const
+{
+    Q_UNUSED(parent);
+    return 8;
```
trisyoungs, line 18: There is of course a continuing "issue" with all of the tables on the `SpeciesTab` in that the number of parameters (and hence total number of columns) is not known unless we (dynamically) assess current functional forms of all the terms in the vector.  This is encapsulated by #597. This shouldn't be addressed as part of the current PR, but we may want to get on to it as a prioritised next step.

src/gui/models/speciesAngleModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    switch (index.column())
+    {
+        case 0:
+        case 1:
+        case 2:
+            return angle.index(index.column()) + 1;
+        case 3:
+            return angle.masterParameters() ? ("@" + std::string(angle.masterParameters()->name())).c_str()
+                                            : SpeciesAngle::angleFunctions().keywordFromInt(angle.form()).c_str();
```
trisyoungs, line 42: Here and elsewhere, should we avoid `c_str()` and just use `QVariant::ftomStdString()`?

src/gui/models/speciesAtomModel.h
```diff
@@ -0,0 +1,41 @@
...
+class SpeciesAtomModel : public QAbstractTableModel
+{
+    Q_OBJECT
+
+    private:
+    std::list<SpeciesAtom> &atoms_;
+    Dissolve &dissolve_;
+
+    signals:
+    void updateIsotopologuesTab();
```
trisyoungs, line 23: This is a very specific name for a signal that could trigger a variety of events - perhaps a more general name, e.g. `dataChanged()`, might be more appropriate?

src/gui/speciestab.h
```diff
@@ -79,31 +82,15 @@ class SpeciesTab : public QWidget, public ListItem<SpeciesTab>, public MainTab
...
-    void on_AtomTable_itemSelectionChanged();
-    void on_BondTable_itemChanged(QTableWidgetItem *w);
-    void on_AngleTable_itemChanged(QTableWidgetItem *w);
-    void on_TorsionTable_itemChanged(QTableWidgetItem *w);
-    void on_ImproperTable_itemChanged(QTableWidgetItem *w);
-
-    public slots:
-    // Update Geometry tab
+    // public slots:
+    // // Update Geometry tab
```
trisyoungs, line 92: Double indent

src/gui/speciestab_funcs.cpp
```diff
@@ -17,7 +17,9 @@
 
 SpeciesTab::SpeciesTab(DissolveWindow *dissolveWindow, Dissolve &dissolve, MainTabsWidget *parent, const QString title,
                        Species *species)
-    : ListItem<SpeciesTab>(), MainTab(dissolveWindow, dissolve, parent, QString("Species: %1").arg(title), this)
+    : ListItem<SpeciesTab>(), MainTab(dissolveWindow, dissolve, parent, QString("Species: %1").arg(title), this),
+      atoms_(species->atoms(), dissolve), bonds_(species->bonds(), dissolve), angles_(species->angles(), dissolve),
+      torsions_(species->torsions(), dissolve), impropers_(species->impropers(), dissolve)
```
trisyoungs, line 22: Getting a `reorder` compiler warning - `angles_` and `bonds_` need to be swapped.

src/gui/speciestab_geometry.cpp
```diff
@@ -649,31 +83,6 @@ void SpeciesTab::updateGeometryTab()
 {
     Locker refreshLocker(refreshLock_);
```
trisyoungs, line 84: Redundant now.

trisyoungs (changes requested):
Overall, looking really good. My only concerns that should be addressed is the editing of the items in the various tables:

1) Some time ago I made a change to the edit triggers on the tables, and after playing with the `SpeciesTab` a bit I really don't like it at all! At present we have all edit triggers enabled on items, including just clicking on one (that isn't currently selected). This immediately raises an editor delegate, when it feels entirely alien to do so.  If we could remove that trigger from all the tables, that would undo my bad decision and make the UX much nicer.

2) When a delegate is raised for an item on the table, the initial value is some default and not the current data in the table. For instance, numeric parameters all come up as `0` (or blank) and the functional form combos default to `None`. This is because the (custom) delegates used on those tables (see, e.g. `ExponentialSpinDelegate`) request their data via the `EditRole`, which isn't currently implemented as a return value.

src/gui/models/speciesAngleModel.cpp
rprospero: I've gone ahead and assigned #597 to myself, as it's a natural follow on from this PR.

src/gui/models/speciesAngleModel.cpp
rprospero: I've moved all of the models to explicitly returning a `QString`, instead of having it implicitly cast from `c_str`.  The intent is clearer this way.

src/gui/models/speciesAtomModel.h
rprospero: I've switched this to `atomTypeChanged`.  This is because this is only trigged when the atom type changes, since this will propagate to the isotopologues table.  The `dataChanged` name was too general because there are many changes (e.g. changing the charge) that will not trigger the signal.  We could have used the pre-built `dataChanged` signal, but that would trigger a bunch of events on changes that weren't going to alter the table.

trisyoungs (approved):

PR #692, opened 2021-05-18
Show all parameters in a single cell

### LINKED ISSUE #597: Single-item delegate for forcefield parameter editing

rprospero:
Instead of having four parameters displayed for each row, we display a single cell that has a comma separated list of the parameter values.  Eventually we'll want the parameters to be labelled, but that's a larger issue than the immediate interface one.

This closes #597.

### DIFF
--- a/src/gui/models/speciesModelUtils.cpp
+++ b/src/gui/models/speciesModelUtils.cpp
@@ -1,0 +1,25 @@
+#include "gui/models/speciesModelUtils.h"
+#include <QStringList>
+#include <algorithm>
+#include <vector>
+
+QString joinParameters(const SpeciesIntra &source)
+{
+    std::vector<QString> vec(source.parameters().size());
+    std::transform(source.parameters().begin(), source.parameters().end(), vec.begin(),
+		   [](const double value) { return QString::number(value); });
+    QStringList terms(vec.begin(), vec.end());
+    return terms.join(",");
+}
+
+bool splitParameters(const QString &params, SpeciesIntra &destination)
+{
+    if (destination.masterParameters())
+	return false;
+    auto terms = params.split(",");
+    if (terms.size() != destination.parameters().size())
+	return false;
+    for (int i = 0; i < terms.size(); ++i)
+	destination.setParameter(i, terms[i].toDouble());
+    return true;
+}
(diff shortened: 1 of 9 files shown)


src/gui/models/speciesModelUtils.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <vector>
+
+QString joinParameters(const SpeciesIntra &source)
+{
+    std::vector<QString> vec(source.parameters().size());
+    std::transform(source.parameters().begin(), source.parameters().end(), vec.begin(),
+                   [](const double value) { return QString::number(value); });
+    QStringList terms(vec.begin(), vec.end());
+    return terms.join(",");
+}
```
trisyoungs, line 13: This is the part that causes compilation errors for me as I _only_ have Qt 5.12.7, and the `QStringList` ctor you're using here wasn't introduced until 5.14. Who'd have though nobody needed it until then!

Since I'm pretty stuck with 5.12.7 and there are many other places in the code where a `", "`-separated list is required, I might suggest that this is a utility function to be added in a more general manner, perhaps in `DissolveSys`. Such a function could be something like:

```
std::string join_strings(Iter begin, Iter end, std::string_view glue, Lambda lam)
{
    return std::accumulate(std::next(begin), end, lam(*begin), [&lam](const auto &acc, const auto *item) { return fmt::format("{}{}{}", acc, glue, lam(*item)); };);
}
```
Invocation would then be:
```
auto joined = join_strings(source.parameters.begin(), source.parameters.end(), ", ", [](const auto value) { return fmt::format("{}", value); });
```

This should give enough flexibility for future use cases (e.g. "name=value" lists).

trisyoungs (changes requested):
After hacking around the bit that doesn't work for me (see sole comment on the code) this all looks fine, although I have to say that I much prefer a space following the comma in the list to break things up.

The same model can also be applied to the ForcefieldTab, perhaps in a subsequent PR.

src/gui/models/speciesModelUtils.cpp
rprospero: Ironically, that's what I had initially written, but I switched it over to the QStringList because I didn't like needing to check for the empty container. I had also assumed that it had been around since Qt 1, since it's so obvious.  I'll change it back.

trisyoungs (approved):
:+1: Like the use of `fmt::format` to give a general join function for probably all of the datatypes we'd ever need to be gluing together.

PR #694, opened 2021-05-20
Store MasterIntra terms in vector instead of list

StephenSmith25:
This PR stores the MasterIntra terms in a vector. To keep stability in the referenced MasterIntra, the terms are now stored in shared_ptrs. 

Rationale for shared_ptrs is that both coreData and the GUI share ownership over the MasterTerms.

This PR also updates the MasterTerms table so that it is in the model view format.


trisyoungs:
While `MasterIntra` really should be on a `std::vector`, there is a hitch - the `SpeciesIntra` class optionally stores a pointer to a `MasterIntra` from which numerical parameters are taken. Alas, this means that using a `std::vector` isn't going to work all of the time, if we consider that they could be added / removed at will by the user.

One possibility would be to have a two-way relationship between `MasterIntra` and `SpeciesIntra` (much like `SpeciesAtom` and the intramolecular classes like `SpeciesBond` etc.). Another more easily-realised solution would be to reference `MasterIntra` by (unique) name (or hash key from a `std::map`?) rather than pointer, and simply copy over the numerical parameters as and when required.

trisyoungs (commented):
This looks like a good solution, but it got me thinking about the relationship between `MasterIntra` and `SpeciesIntra` - I'm wondering now if it is more complicated than we need.

Essentially, I'm questioning whether the `SpeciesIntra` needs to be aware of the source `MasterIntra` objects / vectors at all. As and when we require a `SpeciesIntra` to utilise a `MasterIntra`, we can pass a `const` reference to the `SpeciesIntra`, which then immediately copies the functional form and parameters, and stores the name of the term in a `std::optional<std::string>`. From this point, the following user-driven events can happen in the GUI:

1) The term is "detached" from the `MasterIntra` - here all we need to do is nullify the `std::optional`.
2) The master term is renamed - we would need a function in `Species` along the lines of `masterTermRenamed(SpeciesIntra::InteractionType type, std::string oldName, std::string newName)` which would search through the local terms and rename any occurrences accordingly (no re-setting of parameters is necessary).
3) Parameters in a master term are changed - lazily, one could just have a `Species::updateMasterTerms(const CoreData &coreData)` function to search through local terms and re-pass references to terms to enforce another copy of the form / parameters.
4) The master term is removed - this could also be handled by 3), nullifying any `std::optional`s containing non-existent names.

Thoughts?

src/gui/delegates/intraformcombo.hui
```diff
@@ -4,28 +4,25 @@
...
 #include <QtWidgets/QItemDelegate>
 
-// Forward Declarations
-class ComboListItems;
-
 // Delegate class, providing list of functional forms / master terms for an intramolecular definition
-class IntraFormComboDelegate : public QItemDelegate
+template <typename T> class IntraFormComboDelegate : public QItemDelegate
 {
-    Q_OBJECT
```
rprospero, line 18: Does the MOC have a problem with the Q_OBJECT inside a template?  I'm not opposed to removing the line here, but I'm used to needing it and I'm most just confused why it's not needed here.

rprospero (approved):
I'm quite happy with this code.  I have a question for my own personal education, but I'm otherwise ready for it to merge in.  This also seems to have taken care of a lot of the code that I was going to need to do on the Forcefield Tab.

src/gui/delegates/intraformcombo.hui
StephenSmith25: So a Q object can't be templated it seems, Maybe the MOC doesn't know what to do with it- I'm feeling we don't need it here as we aren't defining any signals or slots (those are defined in the class we inherit from).

Issue #686, opened 2021-05-14
Wizard for Configuration Generation


trisyoungs:
Configuration generation options are currently implemented as individual options ("Simple Random Mix", "Relative Random Mix" etc.) and adding more (e.g. based on adsorbates in frameworks) now begins to create a long list of functions with similar or largely identical functionality.

A wizard to guide the user through the process of constructing a configuration would allow a fair amount of code re-use and permit common tasks (setting a set of relative ratios, for instance) to be generalised across different system types (e.g. random mix vs random mix inside a framework).

PR #695, opened 2021-05-23
Wizard Dialogs

trisyoungs:
On attempting to set up some new simulations, it quickly became clear that the bridge between the XML (OpenMM) import and automated forcefield assignment was a priority. This PR does some groundwork on a fairly neglected aspect of the code (the so-called "wizards"), modernising and tidying the related classes, and simplifying the usage of wizard-based dialogs.

An AtomType model for the GUI has been added, incorporating a solution towards allowing optional extensions to that model (specifically icons in the present case). Opinions / alternative suggestions welcome! I have also elected to add a specialised `splitStringToDoubles` function to `DissolveSys` (it is not templated), returning a vector of `double`s. I felt this was more in line with where we are ultimately heading (probably a splitter returning a vector of pairs of parameter name and value) and wanted to deal with the case where the number of function parameters required only has a minimum limit (see, e.g., the `CosN` torsion terms).

Finally, the `AddConfigurationDialog` and `AddConfigurationWizard` classes are completely removed, since they were incomplete and most likely replaced with new functionality (#686).


src/gui/addforcefieldtermsdialog_funcs.cpp
```diff
@@ -1,29 +1,654 @@
...
+    if (!ok)
+        return;
+
+    QList<QTreeWidgetItem *> selectedItems = ui_.MasterTermsTree->selectedItems();
+    QList<QTreeWidgetItem *>::iterator i;
+    for (i = selectedItems.begin(); i != selectedItems.end(); ++i)
+    {
+        MasterIntra *intra = VariantPointer<MasterIntra>((*i)->data(0, Qt::UserRole));
+        intra->setName(fmt::format("{}{}", qPrintable(prefix), intra->name()));
+    }
```
rprospero, line 632:
```suggestion
    for (i : selectedItems)
    {
        MasterIntra *intra = VariantPointer<MasterIntra>(i->data(0, Qt::UserRole));
        intra->setName(fmt::format("{}{}", qPrintable(prefix), intra->name()));
    }
```

src/gui/addforcefieldtermsdialog_funcs.cpp
```diff
@@ -1,29 +1,654 @@
...
+    if (!ok)
+        return;
+
+    QList<QTreeWidgetItem *> selectedItems = ui_.MasterTermsTree->selectedItems();
+    QList<QTreeWidgetItem *>::iterator i;
+    for (i = selectedItems.begin(); i != selectedItems.end(); ++i)
+    {
+        MasterIntra *intra = VariantPointer<MasterIntra>((*i)->data(0, Qt::UserRole));
+        intra->setName(fmt::format("{}{}", intra->name(), qPrintable(suffix)));
+    }
```
rprospero, line 651:
```suggestion
    for (auto i : selectedItems)
    {
        MasterIntra *intra = VariantPointer<MasterIntra>(i->data(0, Qt::UserRole));
        intra->setName(fmt::format("{}{}", intra->name(), qPrintable(suffix)));
    }
```

src/base/sysfunc.h
```diff
@@ -58,6 +59,8 @@ class DissolveSys
     static bool isNumber(std::string_view text, bool &isFloatingPoint);
     // Replace all occurrences of search string with replace string
     static std::string replace(const std::string_view source, const std::string_view search, const std::string_view replace);
+    // Split a string over a delimiter, returning a vector of converted double values
+    static std::vector<double> splitStringToDoubles(std::string str, std::string_view delim = ", ");
```
rprospero, line 63: After my review, perhaps I missed something, but I didn't see where we used it.  I can appreciate having the function if it's commonly used, but if it's only in one place, we might just want to use the already existing splitter.

rprospero (commented):
Some possible design suggestions on this.  Also, it feels like there's two follow on tickets from here.  First, we should probably get the other trees onto models and continue getting rid of the TreeUpdater.  Second, we should adapt the Forcefield Import Wizard to use the wizard framework.  However, I think that both of these are probably outside the scope of this PR.

src/gui/importspeciesdialog_funcs.cpp
```diff
@@ -1,23 +1,427 @@
...
+    switch (currentIndex)
+    {
+        case (ImportSpeciesDialog::SelectFilePage):
+            // Check that the input/species file exists, and can be read in successfully
+            temporaryDissolve_.loadInput(qPrintable(ui_.InputFileEdit->text()));
+            if (temporaryDissolve_.nSpecies() == 0)
+                return Messenger::error("No species loaded from input file.");
+
+            // Update widget contents
+            speciesModel_.setData(temporaryDissolve_.species());
```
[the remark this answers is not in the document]
trisyoungs: A unit test for importing a species would be good (I now realise have been a bit lax here) but I'm not sure whether it should be a role for the `SpeciesModel` to perform. Such a unit test could be just as easily made without that functionality, and keeping the `SpeciesModel` as more of a read-only interface to the data is sensible, as it protects the most sensitive data in the whole simulation.

As for the use of a temporary Dissolve, at present this is still required as the `ImportSpeciesDialog` allows you to open an existing Dissolve input file and extract any `Species` definitions. This is probably an argument for a parser/lexer setup, rather than the current "initliase while reading" approach (I'm sure there's a technical term for that!

src/gui/models/atomTypeModel.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+/*
+ * QAbstractItemModel overrides
+ */
+
+int AtomTypeModel::rowCount(const QModelIndex &parent) const
+{
+    Q_UNUSED(parent);
+    return atomTypes_ ? atomTypes_->get().size() : 0;
+}
```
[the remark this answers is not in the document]
rprospero: How did I wind up highlighting `rowCount`?  I'm glad that you realized that I was talking about `rawData`

src/base/sysfunc.h
rprospero: Makes sense

src/gui/importspeciesdialog_funcs.cpp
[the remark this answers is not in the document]
rprospero: This was mostly my own bias sneaking in.  I tend to prefer the design of the interface being a thin wrapper over a complex core.  Adding `loadInput` to `SpeciesModel` made for a thinner wrapper and a more complex core, but the benefits that noteworthy.

PR #700, opened 2021-05-25
Energy Optimisation

trisyoungs:
This PR adjusts the parallelism target of the `EnergyKernel::energy(Molecule &mol, ...)` function. Previously, the parallelism was over atoms in the target molecule, but since a very common case is for the majority of molecules to be small (e.g. water) there is a limit to the benefits that this approach can yield. This in particular affects the `MolShake` module which depends on molecule-centric energy calculation.

Here, the outer loop over atoms is sequential, with the inner loops over cells treated in parallel. On the `examples/water/` system this gives a speedup of around 30 - 35% on my 4-core i7 laptop relative to the original version which, interestingly, is about as performant as the non-multithreaded version.

On a larger machine (48 core) the effects are significantly worse - in fact, the multithreaded version is 1.7 - 2x slower than the serial version before optimisation. With this PR, the multithreaded version is faster than the serial version, but only by perhaps 15 - 17%.

The MC challenge continues!

### DIFF
--- a/src/classes/energykernel.cpp
+++ b/src/classes/energykernel.cpp
@@ -504,29 +504,28 @@ double EnergyKernel::energy(const Atom &i, ProcessPool::DivisionStrategy strateg
 // Return PairPotential energy of Molecule with world
 double EnergyKernel::energy(const Molecule &mol, ProcessPool::DivisionStrategy strategy, bool performSum)
 {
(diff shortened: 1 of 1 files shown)


StephenSmith25 (approved):
Looks great! Had overlooked the size of the molecule atoms vectors in this one!

PR #702, opened 2021-05-27
Remove List Dependency from Tabs

rprospero:
This PR remove all uses of the custom List container from the MainTabsWidget.  The eliminates four more instance of subclasses of ListItem


trisyoungs (approved):
Sticking on a code formatting issue, but otherwise looks good.

I was going to make a comment about returning references to the `std::shared_ptr<MainTab>`s to save the copy, but given the context it isn't important.

PR #703, opened 2021-05-27
Molecule Energy Optimisation 2

trisyoungs:
This PR performs some further (and much better) optimisation on the `EnergyKernel::energy(Molecule &mol, ...)` function, improving the performance of the core `MolShake` module.

Timing information is as follows, run on a 48-core machine. Speed increases vs the serial version are nowhere near 48x, but this is perhaps unsurprising given the localised nature of the routine making it difficult to scale.

```
------------------------------------------------------------------------------
Serial @ 9093bf7 (first optimisation)        Time             CPU   Iterations
------------------------------------------------------------------------------
smallMolecule, Population::small>        0.260 ms        0.260 ms         2491
mediumMolecule, Population::small>        2.12 ms         2.12 ms          335
mediumMolecule, Population::medium>       1.80 ms         1.80 ms          394

------------------------------------------------------------------------------
Threaded @ 9093bf7 (first optimisation)      Time             CPU   Iterations
------------------------------------------------------------------------------
smallMolecule, Population::small>        0.452 ms        0.452 ms         1485
mediumMolecule, Population::small>        2.82 ms         2.82 ms          226
mediumMolecule, Population::medium>       1.77 ms         1.77 ms          350

------------------------------------------------------------------------------
Serial @ Current                             Time             CPU   Iterations
------------------------------------------------------------------------------
smallMolecule, Population::small>        0.199 ms        0.199 ms         3533
mediumMolecule, Population::small>        1.10 ms         1.10 ms          638
mediumMolecule, Population::medium>      0.900 ms        0.900 ms          753

------------------------------------------------------------------------------
Threaded @ Current                           Time             CPU   Iterations
------------------------------------------------------------------------------
smallMolecule, Population::small>        0.047 ms        0.047 ms        11028
mediumMolecule, Population::small>       0.212 ms        0.212 ms         3090
mediumMolecule, Population::medium>      0.237 ms        0.237 ms         2766
```

### DIFF
--- a/src/classes/energykernel.cpp
+++ b/src/classes/energykernel.cpp
@@ -505,43 +505,55 @@ double EnergyKernel::energy(const Atom &i, ProcessPool::DivisionStrategy strateg
 // Return PairPotential energy of Molecule with world
 double EnergyKernel::energy(const Molecule &mol, ProcessPool::DivisionStrategy strategy, bool performSum)
 {
-    const Cell *lastCell = nullptr;
-    std::vector<Cell *> allCells;
+    std::vector<const Cell *> allCells;
 
-    auto totalEnergy = std::accumulate(mol.atoms().begin(), mol.atoms().end(), 0.0, [&](const auto &acc, const auto &i) {
-        auto *cellI = i->cell();
-        if (cellI != lastCell)
-        {
-            allCells.resize(cellI->allCellNeighbours().size() + 1, nullptr);
-            std::copy(cellI->allCellNeighbours().begin(), cellI->allCellNeighbours().end(), allCells.begin());
-            allCells[cellI->allCellNeighbours().size()] = cellI;
-        }
+    // Create a map of atoms in cells so we can treat all atoms with the same set of neighbours at once
+    std::map<Cell *, std::vector<const Atom *>> locationMap;
+    for (auto &i : mol.atoms())
+        locationMap[i->cell()].push_back(i.get());
 
-        // All cell neighbours (all treated with minimum image)
-        auto localEnergy = dissolve::transform_reduce(
-            ParallelPolicies::par, allCells.begin(), allCells.end(), 0.0, std::plus<double>(),
-            [&i, cellI, this, strategy](const auto *cell) {
-                const auto &ii = *i.get();
-                auto mimRequired = cell->mimRequired(cellI);
-                return std::accumulate(
-                    cell->atoms().begin(), cell->atoms().end(), 0.0, [&ii, mimRequired, this](const auto acc, const auto &j) {
-                        auto &jj = *j.get();
+    auto totalEnergy =
+        std::accumulate(locationMap.begin(), locationMap.end(), 0.0, [&](const auto totalAcc, const auto &location) {
+            const auto *centralCell = location.first;
+            const auto &centralCellAtoms = location.second;
 
-                        // Calculate rSquared distance between atoms, and check it against the stored cutoff distance
-                        auto rSq = mimRequired ? box_->minimumDistanceSquared(ii.r(), jj.r()) : (ii.r() - jj.r()).magnitudeSq();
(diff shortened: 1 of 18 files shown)


src/classes/energykernel.cpp
```diff
@@ -504,28 +505,55 @@ double EnergyKernel::energy(const Atom &i, ProcessPool::DivisionStrategy strateg
...
+            // All cell neighbours (all treated with minimum image)
+            auto localEnergy = dissolve::transform_reduce(
+                ParallelPolicies::par, allCells.begin(), allCells.end(), 0.0, std::plus<double>(),
+                [&centralCellAtoms, centralCell, this](const auto *cell) {
+                    auto mimRequired = cell->mimRequired(centralCell);
+                    return std::accumulate(
+                        centralCellAtoms.begin(), centralCellAtoms.end(), 0.0,
+                        [cell, centralCell, mimRequired, this](const auto acc, const auto &i) {
+                            auto &ii = *i;
+                            return acc + std::accumulate(cell->atoms().begin(), cell->atoms().end(), 0.0,
```
rprospero, line 534: So we have an `accumulate` inside and `accumulate` inside a `transform_reduce` inside an `accumulate`.  It's a little confusing, if I'm completely honest.  Unfortunately, I can't see a good solution that doesn't greatly expand the scope of this PR.  Ideas that would be outside the scope:

1) Add a `PairIterator` class, like our `ZipIterator`, except gives every pair from the two containers, instead of the matched pairs.  The two inner accumulates, then becomes a single accumulate over the `PairIterator`

2) Create a `double_accumulate` class template function that iterates one function over two containers.  I've tried it before and it's surprisingly tricky to go in a completely general way.

3) Split the inner lambda into a private member function.  The nice part is that you can understand the method in isolation.  The issue is that you can't capture local variables if it's just a method.  We can get around this, though, using the reader pattern and a private inner struct on the type.

Any of these options are going to just make this PR more confusing, though, so they should be tagged as an issue for a follwup.

src/classes/energykernel.cpp
```diff
@@ -504,28 +505,55 @@ double EnergyKernel::energy(const Atom &i, ProcessPool::DivisionStrategy strateg
...
-            ParallelPolicies::par, cellI->mimCellNeighbours().begin(), cellI->mimCellNeighbours().end(), 0.0,
-            std::plus<double>(), [&i, this, strategy](const auto *cell) {
-                // Cell neighbours requiring minimum image
-                return energy(*i, cell, KernelFlags::ApplyMinimumImageFlag | KernelFlags::ExcludeIntraIGEJFlag, strategy,
-                              false);
-            });
-
-        return acc + localEnergy;
-    });
+    std::vector<const Cell *> allCells;
```
rprospero, line 508: Does this need to be a top level value for the function?  As I'm reading this, it gets wipes on each iteration through the out most loop.

rprospero (approved):
Overall, the PR looks good.  Ironically, the bit that I'm currently working on had to do with the fact the `shared_ptr` does get to be overkill sometimes.  I have a couple of concerns about the complexity, but only one honest suggestion that could improve the PR immediately.

src/classes/energykernel.cpp
trisyoungs: Agreed - as this work went on I could see this particular bit of code becoming a little bit obtuse. A custom class to tidy this kind of "accumulate hierarchy" would be nice, but we should be careful not to prevent vectorisation potential of this version.

src/classes/energykernel.cpp
trisyoungs: Probably not.  It remained outside of the first loop body simply to avoid having to reallocate an array every time. That's a micro-optimisation, I appreciate, and the compiler no doubt knows better. In fact, how this is done will almost immediately be irrelevant, since #701 provides exactly the array of cells (neighbours plus self) required for the calculation.

PR #706, opened 2021-06-02
Storing SpeciesSite

trisyoungs:
This PR moves `SpeciesSite` storage from being a `std::vector<SpeciesSite>` to a `std::vector<std::unique_ptr<SpeciesSite>>` - many keywords rely on the ability to refer to `SpeciesSite`s by `const` pointer, which cannot be maintained in the previous form when manipulating (adding/removing) sites.

A handful of uses of `RefList<SpeciesAtom>` are replaced with `std::vector<const SpeciesAtom *>` as part of the work, and a model/proxy filter is added for `SpeciesSite` in order to allow reimplementation of `SpeciesSiteVectorKeywordWidget`. The model is retrospectively applied to the `SpeciesTab` as well.


src/gui/speciestab_sites.cpp
```diff
@@ -27,26 +24,25 @@ SpeciesSite *SpeciesTab::currentSite()
...
-                     [displayedSite](const auto &p) { return &p == displayedSite; }) == species_->sites().end())
+    // Sanity check that the displayed site actually exists in our species, and get its index in the sites list
+    auto index = 0;
+    for (auto &site : species_->sites())
+    {
+        if (site.get() == displayedSite)
+            break;
+        ++index;
+    }
+    if (index >= species_->sites().size())
```
rprospero, line 40:
```suggestion
    // Sanity check that the displayed site actually exists in our species, and get its index in the sites list
    auto it = std::find_if(species_->sites().begin(), species_->sites.end(),
        [displayedSite](const auto &site){ return site.get() == displayedSitel }));
    auto index = it - species_->sites().begin();
    if (it == species->sites().end())
```

This pulls us back to the finding logic while still retrieving the index value.

rprospero (approved):
One minor suggestion, but otherwise looks good

Issue #590, opened 2021-03-12
GenericList item tidying


trisyoungs:
Work in progress will modernise `GenericList` and remove the need for `GenericItem`s.  Following that, there are several other modifications that can be made to reduce the number of legacy classes registered in the new approach.

- [x] Store data as the recently-introduced `SampledVector` class instead of `Array<SampledDouble>` with 
- [x] Store `std::vector<double>` instead of `Array<double>`.
- [x] Store `std::vector<BraggReflection>` rather than `Array<BraggReflection>`.
- [x] Store `std::vector<KVector>` rather than `Array<KVector>`.
- [x] Store `std::vector<Vec3<double>>` rather than `Array<Vec3<double>>`
- [x] Add specialised class to replace `ArrayND<Data1D>`, e.g. `Data1DArray`, `Data1DArray2D` - this will allow for efficient serialisation (related to #579).

PR #708, opened 2021-06-03
Fixup AvgMol Module

### LINKED ISSUE #112: CalculateAvgMol needs proper system test
CalculateAvgMol module is not covered by a system test, since there is currently no way to load in test coordinates.

Implement this with a keyword maintaining an Array< Vec3<double> >.

### LINKED ISSUE #132: CalculateAvgMol needs to be set up after reading restart file
CalculateAvgMol needs to have its setUp() stage run after the restart file is read, as it is this that updates the stored Species information.

Move necessary code to initialise Species with stored coordinates to a separate function and call from readSession()? Need to store pointer to x, y, and z arrays stored in module data, as the Dissolve object is not available in readSession().

trisyoungs:
The focus of this PR is the removal of further uses of the custom `Array` class in favour of replacing with `std::vector`, and removes the `SampledVector` item from #590.

Summary of changes:
- To aid with system testing a simple mechanism for providing reference vectors of values has been implemented in the `ValueStore` and `ValueStoreKeyword`, following the spirit of e.g. `Data1DStore` and `Data1DStoreKeyword`.
- The ability for `ValueStore` to read data directly from the input file, rather than relying on external files, has been added. Rather than supplying a filename, an `@` symbol indicates that data is to be read from the current input stream *following the end of the `FileAndFormat` block*. Thus, the data to read in must immediately follow the end of the `FileAndFormat` block. See https://github.com/disorderedmaterials/dissolve/blob/fixup-avgmol/tests/calculate_avgmol/bendy.txt for an example of this.
- The system test for the AvgMol module has been partially updated. A subsequent PR will finish this along with making some changes to the way system tests are registered.

Closes #112.
Closes #132.


src/classes/valuestore.h
```diff
@@ -0,0 +1,35 @@
...
+    public:
+    ValueStore() = default;
+    ~ValueStore() = default;
+
+    /*
+     * Data
+     */
+    private:
+    // Value of contained data
+    std::list<std::tuple<std::string, std::vector<double>, ValueImportFileFormat>> data_;
```
rprospero, line 23:
```suggestion
    std::vector<std::unique_ptr<std::tuple<std::string, std::vector<double>, ValueImportFileFormat>>> data_;
```

I'm guessing that you went with `std::list` because it doesn't have the pointer invalidation issues that `std::vector` has.  I've been finding that `std::vector<std::unique_ptr<>>` tends to get around that issue.  The only part of the interface that would actually have to change is the `data` member function and most of those call sites could be updated fairly trivially.

It may be a pointless micro-optimisation, so I'm not going to fight on it if you have a reason for sticking with `std::list`

rprospero (commented):
Besides one comment about an underlying data-type, everything else looks good.

src/classes/valuestore.h
trisyoungs: Ah, now I recall why I used `std::list` (here and for `DataNDStore`s too) - it's a nasty tangle of a problem.  The stored `FileAndFormat` has a `KeywordList` that can't be copied since it stores a bunch of objects derived from `KeywordBase`, which itself is abstract. So using `std::vector` is a no-no since there can be no default-constructed (empty) object.

A `std::vector` of `std::unique_ptr` would work as well as a `std::list`, but given the context of the usage (i.e. in routines that are only used sparingly for testing, and not in production simulations) I would rather leave it for now, and stick another pin in the "Fix KeywordList" voodoo doll.

PR #709, opened 2021-06-03
System Test Definitions

trisyoungs:
This PR makes some changes to the way system tests are defined / detected, hopefully making them more transparent and a little bit more flexible. It stems from a desire to start adding tests that restart from a point mid-way through a process, but for which the input file is the same as the normal "full" test. Previously, these had to be defined using two distinct input files in the same / different directory.

In the present PR:
- Specification of the source directory of the test is removed from the definition function, since this can be determined automatically with `${CMAKE_CURRENT_LIST_DIR}`
- Tests are named after the first argument (prepended with the automatically-determined directory name) rather than the input file.
- The input file is given explicitly (including its extension).
- The `tests/restart` directory contained a restart test for the `CalculateRDF` module, so this has been moved to `tests/calculate_rdf` for consistency.


rprospero (approved):
Much cleaner than my original version.

PR #710, opened 2021-06-03
Refactor Forcefield Tab

rprospero:
This refactors the last bits of the Forcefield Tab.  As a part of this process, the PairPotentials lose their ListItem dependency, as part of #257.

Much of the work was already taken care of with the conversion of the MasterIntra terms.  The AtomTypes similarly allowed the reuse of the old model.  Thus, the only new model is the PairPotential model.

The decision to use a vector of `unique_ptr` and keep much of the pointer overlay instead of using a simple array of `PairPotential` and references was two-fold.  First, the references has a bad habit of being invalidated as elements were added to the vector.  `unique_ptr` with raw pointer observers was chosen because the vector is the canonical owner of these elements and the other functions already behaved as observers.  Additionaly, this avoids the `shared_ptr` overhead previous seen in #700


src/gui/forcefieldtab.h
```diff
@@ -30,6 +32,9 @@ class ForcefieldTab : public QWidget, public MainTab
     private:
     // Main form declaration
     Ui::ForcefieldTab ui_;
+    // Model of Atom Types
```
trisyoungs, line 35:
```suggestion
    // Models
```

src/gui/models/pairPotentialModel.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+const PairPotential *PairPotentialModel::rawData(const QModelIndex index) const { return pairs_[index.row()].get(); }
+
+PairPotential *PairPotentialModel::rawData(const QModelIndex index) { return pairs_[index.row()].get(); }
+
+QVariant PairPotentialModel::data(const QModelIndex &index, int role) const
+{
+    // return QString::number(pairs_.size());
+
```
trisyoungs, line 36:
```suggestion
```

src/main/pairpotentials.cpp
```diff
@@ -32,48 +32,53 @@ void Dissolve::setPairPotentialsIncludeCoulomb(bool b) { pairPotentialsIncludeCo
...
 
 // Return number of defined PairPotentials
-int Dissolve::nPairPotentials() const { return pairPotentials_.nItems(); }
+int Dissolve::nPairPotentials() const { return pairPotentials_.size(); }
 
 // Add new pair potential to list
 PairPotential *Dissolve::addPairPotential(std::shared_ptr<AtomType> at1, std::shared_ptr<AtomType> at2)
 {
-    PairPotential *pp = pairPotentials_.add();
+    auto &pp = pairPotentials_.emplace_back(new PairPotential());
```
trisyoungs, line 48:
```suggestion
    auto &pp = pairPotentials_.emplace_back(std::make_unique<PairPotential>());
```
Is this the line that was causing the issue you were having?  We do exactly this for `Species`, so not sure why it can't be the same here...

trisyoungs (approved):
All looks good to me. Just one small deletion, one picky comment change, and one question.

src/main/pairpotentials.cpp
rprospero: The issue I was having was that you can't use
```
auto &pp = pairPotentials_.emplace_back();
```

This gives a `nullptr` in `pp`, while this would work with `shared_ptr`

PR #711, opened 2021-06-03
Move Geometry and SystemTemplate out of ListItem

rprospero:
Just removing two more ListItem Dependencies as part of #257

### DIFF
--- a/src/keywords/geometrylist.cpp
+++ b/src/keywords/geometrylist.cpp
@@ -5,8 +5,8 @@
 #include "base/lineparser.h"
 #include "classes/coredata.h"
 
-GeometryListKeyword::GeometryListKeyword::GeometryListKeyword(std::vector<Geometry> &data, Geometry::GeometryType type)
-    : KeywordData<std::vector<Geometry> &>(KeywordBase::GeometryListData, data), type_(type)
+GeometryListKeyword::GeometryListKeyword::GeometryListKeyword(std::vector<Geometry> data, Geometry::GeometryType type)
+    : KeywordData<std::vector<Geometry>>(KeywordBase::GeometryListData, data), type_(type)
 {
 }
 
--- a/src/keywords/geometrylist.cpp
+++ b/src/keywords/geometrylist.cpp
@@ -63,7 +63,7 @@ bool GeometryListKeyword::write(LineParser &parser, std::string_view keywordName
 {
     std::string index;
 
-    for (auto ref : data_)
+    for (auto &ref : data_)
     {
         index.clear();
         for (auto n = 0; n < maxArguments() - 1; ++n)
--- a/src/keywords/geometrylist.cpp
+++ b/src/keywords/geometrylist.cpp
@@ -5,8 +5,8 @@
 #include "base/lineparser.h"
 #include "classes/coredata.h"
 
-GeometryListKeyword::GeometryListKeyword::GeometryListKeyword(List<Geometry> &data, Geometry::GeometryType type)
-    : KeywordData<List<Geometry> &>(KeywordBase::GeometryListData, data), type_(type)
+GeometryListKeyword::GeometryListKeyword::GeometryListKeyword(std::vector<Geometry> &data, Geometry::GeometryType type)
+    : KeywordData<std::vector<Geometry> &>(KeywordBase::GeometryListData, data), type_(type)
 {
 }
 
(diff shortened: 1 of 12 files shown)


src/keywords/geometrylist.cpp
```diff
@@ -5,8 +5,8 @@
 #include "base/lineparser.h"
 #include "classes/coredata.h"
 
-GeometryListKeyword::GeometryListKeyword::GeometryListKeyword(List<Geometry> &data, Geometry::GeometryType type)
-    : KeywordData<List<Geometry> &>(KeywordBase::GeometryListData, data), type_(type)
+GeometryListKeyword::GeometryListKeyword::GeometryListKeyword(std::vector<Geometry> &data, Geometry::GeometryType type)
+    : KeywordData<std::vector<Geometry> &>(KeywordBase::GeometryListData, data), type_(type)
```
trisyoungs, line 9:
```suggestion
GeometryListKeyword::GeometryListKeyword::GeometryListKeyword(std::vector<Geometry> data, Geometry::GeometryType type)
    : KeywordData<std::vector<Geometry>>(KeywordBase::GeometryListData, data), type_(type)
```
In line with the trend happening for other keywords, rather than storing references to target objects as the keyword data (as is the case here) I have been moving old and writing new keywords to store the object itself in the keyword.  I used to think that storing a reference to the target list / vector was a cool idea, but it creates a an unnecessary separation between the keyword and the data it is supposed to be managing.

src/keywords/geometrylist.cpp
```diff
@@ -63,13 +63,12 @@ bool GeometryListKeyword::write(LineParser &parser, std::string_view keywordName
 {
     std::string index;
 
-    ListIterator<Geometry> GeoIterator(data_);
-    while (Geometry *ref = GeoIterator.iterate())
+    for (auto ref : data_)
```
trisyoungs, line 66:
```suggestion
    for (auto &ref : data_)
```

src/keywords/geometrylist.h
```diff
@@ -5,16 +5,15 @@
...
 #include "base/geometry.h"
 #include "keywords/data.h"
-#include "templates/list.h"
 
 // Forward Declarations
 /* none */
 
 // Keyword with Geometry Data
-class GeometryListKeyword : public KeywordData<List<Geometry> &>
+class GeometryListKeyword : public KeywordData<std::vector<Geometry> &>
```
trisyoungs, line 13:
```suggestion
class GeometryListKeyword : public KeywordData<std::vector<Geometry>>
```

src/modules/checks/checks.h
```diff
@@ -59,7 +59,7 @@ class ChecksModule : public Module
      */
     private:
     // Distances between atoms
-    List<Geometry> distances_;
+    std::vector<Geometry> distances_;
     // Angles between atoms
-    List<Geometry> angles_;
+    std::vector<Geometry> angles_;
```
trisyoungs, line 64: Making the change from stored reference to store object, these two local variables would disappear.

src/modules/checks/process.cpp
```diff
@@ -45,13 +45,13 @@ bool ChecksModule::process(Dissolve &dissolve, ProcessPool &procPool)
          */
 
         // Loop over distances to check
-        for (auto *d = distances_.first(); d != nullptr; d = d->next())
+        for (const auto &d : distances_)
```
trisyoungs, line 48:
```suggestion
        auto &distances = keywords_.retrieve<std::vector<Geometry>>("Distance");
        for (const auto &d : distances)
```

src/modules/checks/process.cpp
```diff
@@ -66,13 +66,13 @@ bool ChecksModule::process(Dissolve &dissolve, ProcessPool &procPool)
          */
 
         // Loop over angles to check
-        for (auto *a = angles_.first(); a != nullptr; a = a->next())
+        for (const auto &a : angles_)
```
trisyoungs, line 69:
```suggestion
        auto &angles = keywords_.retrieve<std::vector<Geometry>>("Angle");
        for (const auto &a : angles)
```

trisyoungs (changes requested):
More good `ListItem` culling.  Just one request for changing the data type `GeometryListKeyword` stores, to bring it in line with the others.

trisyoungs (approved):

PR #713, opened 2021-06-05
Remove Array 5

trisyoungs:
This PR finalises removal of the custom `Array` class. It also removes `ListItem` from `ProcessGroup` (#257).


src/gui/render/primitiveassembly.cpp
```diff
@@ -108,6 +108,6 @@ void PrimitiveAssembly::createCylinderBond(Primitive *bondPrimitive, Vec3<double
 // Send to OpenGL (i.e. render)
 void PrimitiveAssembly::sendToGL(double pixelScaling)
 {
-    for (auto n = 0; n < assembly_.nItems(); ++n)
-        assembly_[n]->sendToGL(pixelScaling);
+    for (auto *pi : assembly_)
+        pi->sendToGL(pixelScaling);
```
rprospero, line 112:
```suggestion
    for (auto *prim : assembly_)
        prim->sendToGL(pixelScaling);
```

Naming the variable π was throwing me for a second.  Perhaps we could use a slightly more indicative name?

rprospero (commented):
One minor complaint about a variable name, but everything looks great otherwise.

PR #714, opened 2021-06-07
Better import and export of PartialSet

### LINKED ISSUE #579: Be more sensible about writing PartialSet data to the restart file
At present, a very lazy method is used to serialise `PartialSet`s, with each individual `Data1D` component written out in full, whereas all of them share the same x-axis.  This could easily be changed to store the common x-axis followed by the individual y-arrays, saving a ton of space in the restart file.

rprospero:
This includes a custom serialisation and parsing code for the PartialSet class.  This moves the Full, Bound, and Unbound terms into the same table, leading to a third as many lines and no repetition of the x-coordinate.  This should close #579.


src/classes/partialset.cpp
```diff
@@ -509,12 +517,37 @@ bool PartialSet::deserialise(LineParser &parser, const CoreData &coreData)
...
+            {
+                if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+                    return false;
+                int argIndex = 0;
+                part.xAxis().push_back(parser.argd(argIndex));
+                bound.xAxis().push_back(parser.argd(argIndex));
+                unbound.xAxis().push_back(parser.argd(argIndex++));
+                argIndex = readDataPoint(argIndex, parser, part);
+                argIndex = readDataPoint(argIndex, parser, bound);
+                argIndex = readDataPoint(argIndex, parser, unbound);
```
trisyoungs, line 549: Part of me wants to convert the x-axis value first and re-use it for clarity:
```suggestion
                auto x = parser.argd(argIndex++);
                part.xAxis().push_back(x);
                bound.xAxis().push_back(x);
                unbound.xAxis().push_back(x);
                argIndex = readDataPoint(argIndex, parser, part);
                argIndex = readDataPoint(argIndex, parser, bound);
                argIndex = readDataPoint(argIndex, parser, unbound);
```

src/classes/partialset.cpp
```diff
@@ -509,12 +517,37 @@ bool PartialSet::deserialise(LineParser &parser, const CoreData &coreData)
...
-            if (!unboundPartials_[{typeI, typeJ}].deserialise(parser))
+            part.setTag(parser.argsv(0));
+            bound.setTag(parser.argsv(1));
+            unbound.setTag(parser.argsv(2));
+
+            Messenger::warn("Tags: {} {} {}", parser.argsv(0), parser.argsv(1), parser.argsv(2));
+
+            if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
                 return false;
+            auto nPoints = parser.argi(0);
```
trisyoungs, line 534: Given we know the size, we could `reserve` the vectors in advance if we wanted to.

trisyoungs (changes requested):
Looks good.  Couple of small suggestions to consider

src/classes/partialset.cpp
```diff
@@ -509,12 +517,37 @@ bool PartialSet::deserialise(LineParser &parser, const CoreData &coreData)
...
+            auto &unbound = unboundPartials_[{typeI, typeJ}];
+
+            if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
                 return false;
-            if (!unboundPartials_[{typeI, typeJ}].deserialise(parser))
+            part.setTag(parser.argsv(0));
+            bound.setTag(parser.argsv(1));
+            unbound.setTag(parser.argsv(2));
+
+            Messenger::warn("Tags: {} {} {}", parser.argsv(0), parser.argsv(1), parser.argsv(2));
```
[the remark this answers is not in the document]
rprospero: Good catch.  That was some debugging code that snuck in there.  I'm removing it entirely.

src/classes/partialset.cpp
rprospero: I've added a block to reserve the space for all the terms.

src/classes/partialset.cpp
rprospero: That is clearer.

trisyoungs (approved):

PR #717, opened 2021-06-09
Remove ListItem dependency from SiteStack

rprospero:
Just a small PR to put SiteStacks in a proper vector.  I went with `unique_ptr` against because I had some issues with references getting invalidated on my first attempt.  This is another part of #257.


trisyoungs (approved):
Another one off the list (literally!).

There are two included files in `sitestack.h` which can be removed:
```
#include "templates/array.h"
#include "templates/listitem.h"
```

Apart from that, all good to go.

PR #719, opened 2021-06-10
Remove ListItem base from Configuration

rprospero:
The main `List<Configuration>` was replaced with a `std::vector<std::unique_ptr<Configuration>>` while all of the `RefList<Configuration>` could then be converted to `std:vector<Configuration*>`.  This is another part of #257.

### DIFF
--- a/src/classes/coredata.cpp
+++ b/src/classes/coredata.cpp
@@ -399,9 +399,9 @@ Configuration *CoreData::addConfiguration()
 // Remove specified Configuration
 void CoreData::removeConfiguration(Configuration *cfg)
 {
-    auto it = std::find_if(configurations_.begin(), configurations_.end(), [cfg](const auto &c) { return cfg == c.get(); });
-    if (it != configurations_.end())
-        configurations_.erase(it);
+    configurations_.erase(
+        std::remove_if(configurations_.begin(), configurations_.end(), [cfg](const auto &c) { return cfg == c.get(); }),
+        configurations_.end());
 }
 
 // Return number of Configurations in list
--- a/src/classes/coredata.cpp
+++ b/src/classes/coredata.cpp
@@ -388,27 +388,32 @@ Species *CoreData::findSpecies(std::string_view name) const
 // Add new Configuration
 Configuration *CoreData::addConfiguration()
 {
-    Configuration *newConfiguration = configurations_.add();
+    auto &newConfiguration = configurations_.emplace_back(std::make_unique<Configuration>());
 
     // Create a suitable unique name
     newConfiguration->setName(uniqueConfigurationName("NewConfiguration"));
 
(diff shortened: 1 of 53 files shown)


src/modules/analyse/process.cpp
```diff
@@ -9,15 +9,12 @@
 bool AnalyseModule::process(Dissolve &dissolve, ProcessPool &procPool)
 {
     // Check for zero Configuration targets
-    if (targetConfigurations_.nItems() == 0)
+    if (targetConfigurations_.size() == 0)
```
trisyoungs, line 12:
```suggestion
    if (targetConfigurations_.empty())
```

src/modules/calculate_axisangle/process.cpp
```diff
@@ -15,7 +15,7 @@ bool CalculateAxisAngleModule::setUp(Dissolve &dissolve, ProcessPool &procPool)
 bool CalculateAxisAngleModule::process(Dissolve &dissolve, ProcessPool &procPool)
 {
     // Check for zero Configuration targets
-    if (targetConfigurations_.nItems() == 0)
+    if (targetConfigurations_.size() == 0)
```
trisyoungs, line 18:
```suggestion
    if (targetConfigurations_.empty())
```

src/modules/calculate_dangle/process.cpp
```diff
@@ -15,7 +15,7 @@ bool CalculateDAngleModule::setUp(Dissolve &dissolve, ProcessPool &procPool) { r
 bool CalculateDAngleModule::process(Dissolve &dissolve, ProcessPool &procPool)
 {
     // Check for zero Configuration targets
-    if (targetConfigurations_.nItems() == 0)
+    if (targetConfigurations_.size() == 0)
```
trisyoungs, line 18:
```suggestion
    if (targetConfigurations_.empty())
```

trisyoungs (approved):
A few missed opportunities to use `empty()` over `size() == 0`, but otherwise all good.

src/classes/coredata.cpp
```diff
@@ -388,27 +388,32 @@ Species *CoreData::findSpecies(std::string_view name) const
...
 }
 
 // Remove specified Configuration
-void CoreData::removeConfiguration(Configuration *cfg) { configurations_.remove(cfg); }
+void CoreData::removeConfiguration(Configuration *cfg)
+{
+    auto it = std::find_if(configurations_.begin(), configurations_.end(), [cfg](const auto &c) { return cfg == c.get(); });
+    if (it != configurations_.end())
+        configurations_.erase(it);
+}
```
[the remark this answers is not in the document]
rprospero: I had a reason, but, looking back on it, it wasn't a good one.  I've changed back to remove_if.

PR #718, opened 2021-06-10
Tidy Cells

trisyoungs:
This PR moves the storage of neighbours from the `Cell` class itself to `CellArray`, reducing memory requirements and data redundancy.

The main classes affected are `EnergyKernel` and `ForceKernel`, but since those classes already stored a reference to the correct `CellArray`, refactoring was not so bad. Removing the actual cell arrays from cells is a more involved job, and is left for a separate PR.

Other points to note:
- The `EnergyKernel` has been tidied in order to clarify the role of functions, make private some members, and prepare a little for the removal of `ProcessPool` from being stored in the object.
- There are still some speed increases in the `Molecule&`-based energy calculation, but not the 50% originally quoted. Turns out that it's better to calculate the energy *correctly* rather than quickly.
- For the `Atom&`-based energy calculation, there is a small reduction in performance using the new cell vector. However, the old routine was not threaded, so this penalty is offset somewhat by the now-parallel nature of the routine.

```
------------------------------------------------------------------------------------------------
MoleculeEnergy (EnergyKernel::energy(const Molecule &mol,...)) Time             CPU   Iterations
------------------------------------------------------------------------------------------------
<ProblemType::smallMolecule, Population::small>    OLD     0.193 ms        0.193 ms         3636
<ProblemType::smallMolecule, Population::small>    NEW     0.185 ms        0.185 ms         3788

<ProblemType::mediumMolecule, Population::small>   OLD      1.08 ms         1.08 ms          643
<ProblemType::mediumMolecule, Population::small>   NEW      1.01 ms         1.01 ms          696

<ProblemType::mediumMolecule, Population::medium>  OLD     0.893 ms        0.893 ms          749
<ProblemType::mediumMolecule, Population::medium>  NEW     0.859 ms        0.859 ms          827

-------------------------------------------------------------------------------------------------------------------------------
Atom / world energy (EnergyKernel::energy(const Atom &,...)    Time             CPU   Iterations
-------------------------------------------------------------------------------------------------------------------------------
<ProblemType::smallMolecule, Population::small>    OLD     55425 ns        55425 ns        12568
<ProblemType::smallMolecule, Population::small>    NEW     61518 ns        61518 ns        11394

<ProblemType::mediumMolecule, Population::small>   OLD     42291 ns        42289 ns        16143
<ProblemType::mediumMolecule, Population::small>   NEW     45792 ns        45791 ns        14953
```

PR #721, opened 2021-06-11
Kernel Tidy

trisyoungs:
## Not ready for review until #718 is merged.

This PR removes the storage of neighbour cell vectors on individual cells. There were some unanticipated knock-on effects from doing so, leading to a more significant portion of work refactoring several classes. Along the way, the opportunity to tidy a few things up was taken.

Notes:
- Intramolecular energy and force routines checked whether to use minimum image vectors by calling the `Cell::mimRequired(Cell*)` function, which is now defunct. Rather than reimplement these checks using the new vectors in `CellArray`, it was decided to simply always calculate the minimum image vectors. In practice this turns out to be as performant (perhaps marginally more so) than checking for and applying minimum image only when necessary.
- Constructors to energy and force kernels have been generalised.
- Intramolecular energy functions have been updated.


rprospero (approved):
Looks really good.  The cleaner Force and Energy terms will help with getting code moved over to the GPU

PR #742, opened 2021-07-05
Qt6

rprospero:
This switches us to Qt 6.

Qt should now install itself through conan.  Unfortunately, this currently prevents installing the SVG libraries, so the SVG files have been converted to PNG for the time being.  The SVG files remain so that the eventual return to SVG has the necessary files


rprospero:
Superseded  by #749

PR #746, opened 2021-07-07
Parallel Fourier Transform

rprospero:
This enables a parallel calculation of the Fourier transform.  There is also a benchmark included to track the speed of the Fourier transform.  On my four core machine, the benchmark run approximately twice as face with the new, parallel code.

### DIFF
--- a/src/math/ft.cpp
+++ b/src/math/ft.cpp
@@ -3,6 +3,9 @@
 
 #include "math/ft.h"
 #include "math/data1d.h"
+#include <algorithm>
+#include <numeric>
+#include <vector>
 
 namespace Fourier
 {
@@ -43,10 +46,16 @@ bool sineFT(Data1D &data, double normFactor, double wMin, double wStep, double w
     // Create working arrays
     std::vector<double> newX, newY;
 
+    auto omega = wMin;
+    while (omega <= wMax)
+    {
+        newX.push_back(omega);
+        omega += wStep;
+    }
+
     // Perform Fourier sine transform, apply general and omega-dependent broadening, as well as window function
     double ft, deltaX;
-    double omega = wMin;
-    while (omega <= wMax)
+    for (const auto omega : newX)
     {
         ft = 0.0;
         if (omega > 0.0)
@@ -85,10 +94,7 @@ bool sineFT(Data1D &data, double normFactor, double wMin, double wStep, double w
         }
 
         // Add point
-        newX.push_back(omega);
         newY.push_back(ft);
-
-        omega += wStep;
     }
 
     // Apply normalisation factor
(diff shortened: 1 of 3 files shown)


src/math/ft.cpp
```diff
@@ -36,63 +42,53 @@ bool sineFT(Data1D &data, double normFactor, double wMin, double wStep, double w
...
-        {
-            for (m = 0; m < nX - 1; ++m)
-            {
-                deltaX = x[m + 1] - x[m];
+    // auto omega = wMin;
+    // for (auto &xval : newX)
+    // {
+    //     xval = omega;
+    //     omega += wStep;
+    // }
```
trisyoungs, line 56:
```suggestion
```

src/math/ft.cpp
```diff
@@ -36,63 +42,53 @@ bool sineFT(Data1D &data, double normFactor, double wMin, double wStep, double w
...
-
-    // Apply normalisation factor
-    std::transform(newY.begin(), newY.end(), newY.begin(), [normFactor](auto value) { return value * normFactor; });
+    // Perform Fourier sine transform, apply general and omega-dependent broadening, as well as window function
+    dissolve::transform(ParallelPolicies::par, newX.begin(), newX.end(), newY.begin(),
+                        [normFactor, &x, &y, &windowFunction, &broadening](const auto omega) {
+                            double ft = 0.0;
+                            const auto nX = x.size();
+                            std::vector<double> deltas(nX), windows(nX - 1), broadens(nX - 1), result(nX - 1);
+                            std::adjacent_difference(x.begin(), x.end(), deltas.begin());
```
trisyoungs, line 66: Since we're doing pretty simple math in this loop, I wonder if the instantiation of these `std::vector`s is a bottleneck we should remove.  They could be created at the correct size outside the `dissolve::transform` loop (the `std::adjacent_difference` call could be put there too.  Now, if we do this then we'll obviously create a segfault as all threads will be trying to write to some of these vectors at the same time, *however* in my experience of multithreading the various energy loops it is a significant performance hit to "multithread inside a multithread"ed loop. While this makes perfect sense to do so in principle, the overhead of continually creating a thread pool for a task is relatively large compared to the calculation being done. As such, I would be interested to see what the performance gain is if you change the `dissolve::transform` calls below this line into straight `std::transform`s - i.e. a multithreaded outer loop with (potentially) vectorised inner loops. My suspicion is that you will improve performance beyond the factor 2 you see already on your 4-core machine,

trisyoungs (commented):
Completely understand your logic behind this approach, but I wonder if we can make it faster - see single large comment on `src/math/ft.cpp`.

src/math/ft.cpp
rprospero: As it happens, I original had `std:transform` instead of `dissolve::transform`. Oddly enough, I got about 10% better speeds with the nested parallelism, which surprised me as well.

However, you've brought to my attention that there's a way of doing this with only one heap allocation in each thread.  Hopefully that should get us another speed boost.

trisyoungs (commented):
I like the newer version - looks a lot cleaner to me.  How does it perform?

rprospero:
Honestly, it only performs a little faster than the ugly version.  From some cursory poking around, it looks like the window function and broadening function eat up enough time that other optimizations don't significantly effect the runtime.

trisyoungs (approved):

PR #748, opened 2021-07-12
Transmute Species

trisyoungs:
This PR adds in two further `ProcedureNode`s - `PickProximity` to select molecules based on their proximity to other molecules (i.e. by coordination number) and a basic `Transmute` which converts a target picked selection of molecules (of any type) into new molecules of a target species.

The `Transmute` node will be extended in future PRs to introduce randomisation / rotation / orientation of the new molecules, but this is left until other related aspects of the code are implemented, in particular a forthcoming PR updating the `Box` classes and node in order to permit the necessary groundwork before transmutation.

### DIFF
--- a/src/gui/helpers/comboboxcontroller.h
+++ b/src/gui/helpers/comboboxcontroller.h
@@ -15,12 +15,38 @@
  * are stored as data in the QComboBox to allow tests of equivalence.
  */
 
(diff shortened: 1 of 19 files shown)


rprospero (approved):
At some point, we might want to switch to the Model/View setup for the comboboxes like we did for the tables.  It might save us some template code, but it might also be more trouble than its worth.

PR #750, opened 2021-07-14
Box Updates 1

trisyoungs:
This is a short PR to tidy up commenting and variable names in the `Box` classes, and add a more comprehensive benchmark ahead of changes to the core operations.

PR #751, opened 2021-07-15
Box updates 2

### LINKED ISSUE #736: Monoclinic / Triclinic boxes wrongly detected
**Describe the bug**
Monoclinic boxes are detected correctly only if the beta != 90, whereas any single angle not equal to 90degrees qualifies as monoclinic.

**Expected behaviour**
Any single angle not equal to 90degrees should use specific `MonoclinicBox` classes - for maximal optimisation this needs to be implemented as separate classes, depending on which angle is not equal to 90degrees.

trisyoungs:
This PR follows on from #750 and updates the set of `Box` classes, and covers two principal areas:

### Optimisation / Simplification

Initial attempts at simplification / optimisation reduced the number of required function reimplementations in each derived class to two - conversion of supplied `Vec3<double>` position to real or fractional coordinates. This allowed the `Box` base class to contain all of the necessary minimum image, fold etc. functions and simply call the overloaded `toFractional` or `toReal` functions as necessary. However, even with inlining / use of `constexpr` this turned out to be a significant performance hit.

Reverting to a point half-way between the old and trialled extremes seems to be a happy medium. A total of six reimplemented functions are required, covering the coordinate conversion and minimum image routines.  This gives the following rough results:

- Minimum image / vector / distance functions show speed increases of between 20 - 50% across the derived classes.
- Fold functions are roughly the same performance as before, but do not require reimplementation per-class
- Performance of random coordinate generation is slightly degraded by around 5 - 10%, but does not require reimplementation per-class.

### Monoclinic Boxes

Previously only a single `MonoclinicBox` class existed, and which assumed the non-right angle to be `beta`, however in practice the non-right angle may be any of the three available. As such, the original class is split into three separate classes, allowing optimisation of the matrix multiplications based on the known non-right angle.

### Related Issues
Closes #736.

### DIFF
--- a/benchmark/box/box_benchmark.cpp
+++ b/benchmark/box/box_benchmark.cpp
@@ -19,10 +19,10 @@ template <typename BoxType> BoxType createTestBox()
         OrthorhombicBox box(lengths);
         return box;
     }
-    else if constexpr (std::is_same_v<BoxType, MonoclinicBox>)
+    else if constexpr (std::is_same_v<BoxType, MonoclinicAlphaBox>)
     {
         Vec3<double> lengths(1.00, 1.00, 1.00);
-        MonoclinicBox box(lengths, 45);
+        MonoclinicAlphaBox box(lengths, 45);
         return box;
     }
     else if constexpr (std::is_same_v<BoxType, TriclinicBox>)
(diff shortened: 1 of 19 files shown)


rprospero (approved):
I'm surprised that the virtual function calls created that much overhead, but I can't argue with data.

This does remind me that, at some point, I need to benchmark my alternate idea for the Vec3.  The ideas kind of steals a bit from Quaternions, but basically you store the normalized vector along with its magnitude in a separate float.  The advantage is that a lot of operations become faster (normalizing, rescaling, calculating magnitude).  The down side is that it uses 33% more memory for the vectors and that translating vectors becomes slower (but operations on translated vectors often work out about the same).

trisyoungs:
I didn't understand it either, but the benchmarking consistently came out much worse for the simple version.

The `Vec3<double>` change sounds inherently sensible for some applications, but less for others. In particular I'm thinking of `Atom` coordinates which constitute the major use of the class, and are always taken as-is from the class. Doing the modification you suggest would require the scale factor to be applied every time we use the coordinates (which is a lot!) or (worse) converting from the implicit unit cell coordinates into real space with a matrix multiplication. We need to be careful that we don't generate a performance hit here!

PR #752, opened 2021-07-16
Module Targets 1

trisyoungs:
Short, dull PR to make the most of our `joinStrings` algorithm in several places in the code, to rename `ConfigurationRefList` into `ConfigurationVector` to reflect the actual underlying datatype, and to store an actual `std::vector<Configuration *>` rather than a reference to an object located elsewhere.

This lays the groundwork for harmonising `Module` targets, which will be the subject of a future PR.


DanNixon (approved):
:+1:

PR #753, opened 2021-07-20
Box Scaling

### LINKED ISSUE #737: Scale specific Box parameter when adding species
Add the ability to scale only a specific Box axis when adding in new species. This is required for systems such as templated silicas where the A and B cell lengths are fixed by the pore spacing, but the C axis can be freely modified.

trisyoungs:
This PR adds the ability to restrict scaling of box axes to as few as one when changing box volume in the `AddSpecies` node.

The specific requirement of this is to allow a monoclinic cell with enforced *A* and *B* axis lengths to be kept fixed, while the *C* axis is freely scaled to accommodate the necessary volume changes.

Closes #737.

### DIFF
--- a/src/classes/configuration_contents.cpp
+++ b/src/classes/configuration_contents.cpp
@@ -201,29 +201,38 @@ std::shared_ptr<Atom> Configuration::atom(int n)
     return atoms_[n];
 }
 
-// Scale contents of the box by the specified factor
-void Configuration::scaleContents(double factor)
+// Scale contents of the box by the specified factors along each axis
+void Configuration::scaleContents(Vec3<double> scaleFactors)
 {
-    Vec3<double> oldCog, newCog, newPos;
+    Vec3<double> r, cog;
     for (auto &mol : molecules_)
     {
         // If the related species has a periodic box, scale atom positions rather than COG position
         if (mol->species()->box()->type() != Box::BoxType::NonPeriodic)
         {
             for (auto &i : mol->atoms())
-                i->setCoordinates(i->r() * factor);
+            {
+                r = i->r();
+                box()->toFractional(r);
+                r.multiply(scaleFactors);
+                box()->toReal(r);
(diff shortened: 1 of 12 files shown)


src/classes/configuration_contents.cpp
```diff
@@ -201,29 +201,38 @@ std::shared_ptr<Atom> Configuration::atom(int n)
...
         if (mol->species()->box()->type() != Box::BoxType::NonPeriodic)
         {
             for (auto &i : mol->atoms())
-                i->setCoordinates(i->r() * factor);
+            {
+                r = i->r();
+                box()->toFractional(r);
+                r.multiply(scaleFactors);
+                box()->toReal(r);
+                i->setCoordinates(r);
```
rprospero, line 219: This is the part that I was concerned about on the voice chat.  Imagine that you had a water molecule in the box where the hydrogen and oxygen were all part of the same XY plane.  If you now rescale the Y-axis by one million, the bond angle is going to be essentially 180°, which is obviously not correct.  Wouldn't this then effect all of the energy and torsion calculations, or am I missing something?

rprospero (approved):
Everything looks good, as long as I'm wrong about the math/science on my question about the angle.

src/classes/configuration_contents.cpp
trisyoungs: The routine you highlight here only scales individual atom coordinates if, and only if, the source `Species` is a periodic model such as a crystal (i.e. it has a box definition). You are absolutely right that if you apply non-uniform scaling here you'll completely mess up the crystal (or the molecules as in your example). This is not prohibited by the function, but attempting to do so would be a good example of a "very bad idea" on the part of the user.

The typical use case of this routine is scaling the box uniformly when applying a scale factor, but it was generalised to permit "odd" scalings necessary for some systems when `Add`ing molecules during configuration generation.

PR #757, opened 2021-07-25
Remove GUI Start Screen

### LINKED ISSUE #727: Update start screen
Make it less ugly and more modern.

trisyoungs:
This PR removes the clunky and largely un-necessary start screen, preferring to enter straight into a main, empty view with skinned background. Messages output has been moved to its own tab rather than exist at the bottom of the main tabs taking up valuable space. Several other behaviours have been improved along the way, notably the handling of non-existent or broken input files from the GUI.

Closes #727.
image
image


src/gui/messagestab_funcs.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "gui/messagestab.h"
+#include <QScrollBar>
+
+MessagesTab::MessagesTab(DissolveWindow *dissolveWindow, Dissolve &dissolve, MainTabsWidget *parent, const QString title)
+    : MainTab(dissolveWindow, dissolve, parent, title, this)
+{
+    ui_.setupUi(this);
+
+    ui_.MessagesEdit->setFont(QFont("Cousine", 10));
```
rprospero, line 12: Instead of using Cousine, could we use the `defaultFamily` member to get the system font.  This would cause the font to match more naturally with the other system applications.

If we do go this route, it would probably make sense as a separate PR that goes through and removes the custom font code.

rprospero (approved):
One small possible extension forward, but otherwise looks good.  I like the new look.

src/gui/messagestab_funcs.cpp
trisyoungs: Agreed. Encapsulated in #760 which will be addressed in a subsequent PR.

PR #758, opened 2021-07-26
Remove ListItem Depedency from ObjectFactory

rprospero:
This removes the ListItem dependency from the ObjectFactory and ObjectChunk classes.  This is part of #257.  This PR also sets a constant chunkSize, since a) the dynamic chunk size was never used and b) this allowed using static, stack declared arrays instead of putting them on the heap.


trisyoungs (approved):
All looks good to me.

The `ObjectFactory` class itself is now only used in one place (`PrimitiveAssembly`)  - it might be worth considering refactoring that class to not use `ObjectFactory` at all, simplifying the rendering a bit ahead of moving to Qt3D. Is there a "chunked" storage class in the STL?

rprospero:
That's a very good point.  `std::deque` is a chunked container that performs the same effect.  I may try and remove the factor entirely.

trisyoungs (approved):

PR #761, opened 2021-07-28
Remove ListItem dependency from Isotopologues and IsotopeData

rprospero:
This is part of #257.  IsotopeData just needed to have ListItem removed as a parent class, but there were some remaining List stragglers for Isotopologues.

### DIFF
--- a/src/classes/isotopedata.h
+++ b/src/classes/isotopedata.h
@@ -4,10 +4,9 @@
 #pragma once
 
 #include "data/isotopes.h"
-#include "templates/listitem.h"
 
 // sotopeData Definition
-class IsotopeData : public ListItem<IsotopeData>
+class IsotopeData
 {
     public:
     IsotopeData(Sears91::Isotope isotope = Sears91::Isotope::Unknown, double population = 0.0, double fraction = 0.0);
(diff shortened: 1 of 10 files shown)


src/classes/isotopedata.h
```diff
@@ -4,10 +4,9 @@
 #pragma once
 
 #include "data/isotopes.h"
-#include "templates/listitem.h"
 
 // sotopeData Definition
```
trisyoungs, line 8:
```suggestion
// IsotopeData Definition
```
Not your typo, but nice to fix while we're here!

src/classes/species_isotopologues.cpp
```diff
@@ -75,14 +80,18 @@ const Isotopologue *Species::findIsotopologue(std::string_view name, const Isoto
...
-    for (auto *iso = isotopologues_.first(); iso != nullptr; iso = iso->next())
-        if (iso == exclude)
+    for (auto &iso : isotopologues_)
+        if (iso.get() == exclude)
             continue;
         else if (DissolveSys::sameString(name, iso->name()))
-            return iso;
+            return iso.get();
 
     return nullptr;
```
trisyoungs, line 89: Tempting to combine the logic here into something like this?
```
auto it = std::find_if(isotopologues_.begin(), isotopologues_.end(), [exclude,name](auto &i) { return i.get() != exclude && DissolveSys::sameString(name, iso->name()); });
if (it != isotopologues_.end())
    return it->get();
return nullptr;
```

trisyoungs (approved):
Looks good.  Just one minor algorithm suggestion.

PR #765, opened 2021-08-04
Remove ListItem dependency from Modules

rprospero:
This is part of #257.

### DIFF
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -209,7 +209,7 @@ class Dissolve
      */
     private:
     // List of defined processing layers
-    List<ModuleLayer> processingLayers_;
+    std::vector<std::unique_ptr<ModuleLayer>> processingLayers_;
     // Data associated with processing Modules
     GenericList processingModuleData_;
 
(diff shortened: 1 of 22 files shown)


src/main/dissolve.h
```diff
@@ -202,14 +202,14 @@ class Dissolve
     // Generate unique Module name with base name provided
     std::string uniqueModuleName(std::string_view name, Module *excludeThis = nullptr);
     // Delete specified Module instance
-    bool deleteModuleInstance(Module *instance);
+    bool deleteModuleInstance(std::unique_ptr<Module> instance);
```
trisyoungs, line 205: Isn't `std::unique_ptr` non-copyable?  So this should be a reference or just a raw pointer?

trisyoungs (approved):
Looks good to me. Paves the way for further modernisation of `Module` creation in particular (i.e. get rid of the list of master modules and move to a "producer"-style approach).

src/main/dissolve.h
rprospero: It isn't copyable, but it can be moved with `std::move`.  Then, when the end of the function is reached, the instance leaves scope and the unique_ptr deletes its contents.  Originally, `deleteModuleInstance` would call `delete instance`, but this didn't work when `instance` was an observer with a `unique_ptr`.  The options I saw were to let `deleteModuleInstance` take the `unique_ptr` or to have `deleteModuleInstance` not delete the Module Instance, which felt unituitive.

When I mentioned using move semantics in anger during the stand-up, this is what I was talking about.

trisyoungs: Sorry - should have paid more attention to the function definition!

PR #767, opened 2021-08-04
Cylindrical Regions

trisyoungs:
This PR continues to rework the picking/region selectors to avoid repetition of code.  As such, the `PickCylinder` node is removed here and replaced with `CylindricalRegion` and `PickRegion` nodes.  This makes any defined region usable in a "pick", and only requires the necessary geometry to be written once.

The `CylindricalRegion` class brings up an excellent case for modernisation of the keywords, since the `isVoxelValid()` override requires keyword data to be retrieved and used in the calculation every time. This cannot be precalculated easily owing to the `const`-ness enforced when referencing `ProcedureNode`s from other `ProcedureNode`s, but can be made more efficient by not requiring searches of the available `keywords_` every time.


src/classes/region.h
```diff
@@ -0,0 +1,71 @@
...
+
+        // Initialise 3D map and determine valid voxels
+        voxelMap_.initialise(nVoxels_.x, nVoxels_.y, nVoxels_.z);
+        for (auto x = 0; x < nVoxels_.x; ++x)
+            for (auto y = 0; y < nVoxels_.y; ++y)
+                for (auto z = 0; z < nVoxels_.z; ++z)
+                    voxelMap_[{x, y, z}] = {
+                        Vec3<int>(x, y, z),
+                        voxelCheckFunction(cfg, box_->getReal({(x + 0.5) * voxelSizeFrac_.x, (y + 0.5) * voxelSizeFrac_.y,
+                                                               (z + 0.5) * voxelSizeFrac_.z}))};
```
rprospero, line 53:
```suggestion
        for (auto &[x, y, z] : voxelMap_.indices())
            voxelMap_[{x, y, z}] = {
                Vec3<int>(x, y, z),
                voxelCheckFunction(cfg, box_->getReal({(x + 0.5) * voxelSizeFrac_.x, (y + 0.5) * voxelSizeFrac_.y,
                                                       (z + 0.5) * voxelSizeFrac_.z}))};
```

You can't merge the above patch in because the `indices` method doesn't exist.  However, it seems like an obvious extension that would be useful in multiple parts of the code base, besides also making this particular bit of code cleaner.

src/classes/region.h
```diff
@@ -0,0 +1,71 @@
...
+
+        // Create linear vector of all available voxels
+        auto nFreeVoxels = std::count_if(voxelMap_.begin(), voxelMap_.end(), [](const auto &voxel) { return voxel.second; });
+        freeVoxels_.clear();
+        freeVoxels_.resize(nFreeVoxels);
+        std::copy_if(voxelMap_.begin(), voxelMap_.end(), freeVoxels_.begin(), [](const auto &voxel) { return voxel.second; });
```
rprospero, line 59:
```suggestion
        auto nFreeVoxels = std::count_if(voxelMap_.begin(), voxelMap_.end(), std::get<1>);
        freeVoxels_.clear();
        freeVoxels_.resize(nFreeVoxels);
        std::copy_if(voxelMap_.begin(), voxelMap_.end(), freeVoxels_.begin(), std::get<1>);
```

No need to define a new lambda.

rprospero (approved):
One small suggestion and then one huge, cumbersome suggestion that involves a lot of work and should probably be its own issue.

src/classes/region.h
trisyoungs: Good suggestion (captured in #771).

src/classes/region.h
trisyoungs: This can't work because `std::get<>` needs the `std::pair` object as a parameter, unless I'm missing some C++ voodoo in which the current item can be implicitly passed?

rprospero: `std::get<1>` should get the object as a parameter just as the lambda function is passed the object as a parameter, unless I'm missing something?

trisyoungs: Sounds sensible, but doesn't compile. Not in C++17 perhaps?

rprospero: I learned something new today:  taking function pointers to functions in the `std` namespace is explicitly defined in the standard as being undefined behaviour.  I've rolled back to before I made the suggestion.

Issue #760, opened 2021-07-28
Use system fonts


trisyoungs:
Don't package and use custom fonts in the GUI - use system fonts for a consistent UI experience.

- [ ] Don't register custom `Cousine` font (`src/gui/gui_funcs.cpp`)
- [ ] Remove `Cousine` font from Qt resource file, and actual files from repo.
- [ ] Apply suitable fixed-width font in MessagesTab (`src/gui/messagestab_funcs.cpp`)

PR #769, opened 2021-08-06
Convert to system fonts

EthanDevlinSTFC:
Modified code to use system fonts. For issue #760


src/gui/messagestab_funcs.cpp
```diff
@@ -9,7 +9,7 @@ MessagesTab::MessagesTab(DissolveWindow *dissolveWindow, Dissolve &dissolve, Mai
 {
     ui_.setupUi(this);
 
-    ui_.MessagesEdit->setFont(QFont("Cousine", 10));
+    ui_.MessagesEdit->setFont(QFont("Helvetica", 10));
```
rprospero, line 12: It might be worth using `QGuiApplication::font()` instead of specifically calling on Helvetica. For example, my own personal machine does not have Helvetica installed.

rprospero (changes requested):

PR #775, opened 2021-08-10
Singularity Container

trisyoungs:
This PR introduces building of Singularity containers for Dissolve. The containers contain both the command-line and GUI versions, with the latter being the default run target, and the former accessed by passing `--app dissolve`.

Two flavours of container are defined in `ci/singularity/`, each with a specific issue for discussion:

### Ubuntu 20.04

This simply brings in the binary artifacts from the build stage of the CI, but which has a missing runtime dependency in the form of `libicui18n.so.56` which doesn't appear to be in any accessible Ubuntu repo. I assume that this dependency comes from the Qt6 packages brought in by `aqtinstall`, and which presumably breaks the AppImages as well.

### OpenSuSE Leap 15.2

This builds Dissolve from scratch (again) but all required libraries are native since they are all in official OpenSuSE repositories. The GUI opens without error, but as soon as a `QOpenGLWidget` is constructed by loading / starting a new simulation we get lots a GLX error:
```
Falling back to using screens root_visual.
qt.glx: qglx_findConfig: Failed to finding matching FBConfig for QSurfaceFormat(version 2.0, options QFlags<QSurfaceFormat::FormatOption>(), depthBufferSize -1, redBufferSize 1, greenBufferSize 1, blueBufferSize 1, alphaBufferSize -1, stencilBufferSize -1, samples -1, swapBehavior QSurfaceFormat::SingleBuffer, swapInterval 1, colorSpace QColorSpace(), profile  QSurfaceFormat::NoProfile)
Could not initialize GLX
```

### DIFF
--- a/.azure-pipelines/pr.yml
+++ b/.azure-pipelines/pr.yml
@@ -29,10 +29,10 @@ stages:
           - checkout: self
             fetchDepth: 1
           - template: templates/set-short-hash.yml
-          - template: templates/build-linux.yml
-            parameters:
-              extraflags: "-DBUILD_SYSTEM_TESTS:bool=true -DBUILD_UNIT_TESTS:bool=true -DCMAKE_BUILD_TYPE=Debug"
-          - template: templates/package-linux.yml
+            #- template: templates/build-linux.yml
+            #parameters:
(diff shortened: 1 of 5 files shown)


rprospero:
This has been obviated by #807.

PR #778, opened 2021-08-16
Open recent implementation

EthanDevlinSTFC:
Fully implemented open recent menu


src/gui/gui.ui
```diff
@@ -835,6 +828,11 @@
     <string>To Master Terms</string>
    </property>
   </action>
+  <action name="actiond">
+   <property name="text">
+    <string>d</string>
+   </property>
+  </action>
```
trisyoungs, line 835: Randomly-named action?

src/gui/gui.h
```diff
@@ -97,6 +97,23 @@ class DissolveWindow : public QMainWindow
     // List of ReferencePoints currently loaded
     std::vector<ReferencePoint> referencePoints_;
 
+    private:
+    QList<QAction *> recentFileActionList;
+    const int recentFileNo;
```
trisyoungs, line 102: For us, private member variables always have a suffixed underscore:
```suggestion
    QList<QAction *> recentFileActionList_;
    const int recentFileNo_;
```

src/gui/gui.h
```diff
@@ -97,6 +97,23 @@ class DissolveWindow : public QMainWindow
...
+    /*
+     * OpenRecent Functions
+     */
+    public slots:
+    // Manage opening of recent files
+    void openRecent();
+    // Fill open recent with placeholders
+    void createRecentMenu();
+    // Add current file to recent files
+    void adjustForCurrentFile(const QString &filePath);
```
trisyoungs, line 113: This function could be renamed to more clearly reflect what it's acting on - `addCurrentToRecentFiles` for instance.

src/gui/gui_funcs.cpp
```diff
@@ -18,9 +19,10 @@
 #include <QLCDNumber>
 #include <QMdiSubWindow>
 #include <QMessageBox>
+#include <QSettings>
 
 DissolveWindow::DissolveWindow(Dissolve &dissolve)
-    : QMainWindow(nullptr), dissolve_(dissolve), threadController_(this, dissolve)
+    : QMainWindow(nullptr), dissolve_(dissolve), threadController_(this, dissolve), recentFileNo(4)
```
trisyoungs, line 25: Four is a bit stingy!  Perhaps 10 as the default?

src/gui/gui_funcs.cpp
```diff
@@ -214,13 +216,89 @@ bool DissolveWindow::openLocalFile(std::string_view inputFile, std::string_view
...
+        itEnd = recentFilePaths.size();
+    }
+    else
+    {
+        itEnd = recentFileNo;
+    }
+    // Fill recent menu
+    for (auto i = 0u; i < itEnd; ++i)
+    {
+        QString strippedName = QFileInfo(recentFilePaths.at(i)).fileName();
```
trisyoungs, line 292: I have a lot of input files called the same thing, so I can't distinguish between them based on just the filename. You could add the full path after the filename in the menu - e.g. `input.txt (/home/tris/work/stuff/input.txt)` for example.

src/gui/gui_funcs.cpp
```diff
@@ -214,13 +216,89 @@ bool DissolveWindow::openLocalFile(std::string_view inputFile, std::string_view
...
+        recentFileActionList.at(i)->setText(strippedName);
+        recentFileActionList.at(i)->setData(recentFilePaths.at(i));
+        recentFileActionList.at(i)->setVisible(true);
+    }
+
+    for (auto i = itEnd; i < recentFileNo; ++i)
+        recentFileActionList.at(i)->setVisible(false);
+}
```
trisyoungs, line 300:
```suggestion
    // Fill recent menu
    for (auto i = 0u; i < recentFileNo; ++i)
    {
        if (i < recentFilePaths.size())
        {
            QString strippedName = QFileInfo(recentFilePaths.at(i)).fileName();
            recentFileActionList.at(i)->setText(strippedName);
            recentFileActionList.at(i)->setData(recentFilePaths.at(i));
            recentFileActionList.at(i)->setVisible(true);
        }
        else
            recentFileActionList.at(i)->setVisible(false);
    }
}
```

trisyoungs (changes requested):
Overall very good - organisation of code, use of `auto` etc.  Just a few small suggestions/changes to address.

src/gui/gui.ui
EthanDevlinSTFC: Stray code, issue resolved

trisyoungs:
One further observation from testing the code - the font for the recent files actions is the default 11 pt, whereas all of the other menus are set to 10 pt.  Annoyingly, The submenu doesn't seem to inherit the font from its parent, so you'll have to set it manually (you can copy the font of the parent item and set that on your new menu).

src/gui/gui.h
trisyoungs: Actually - the code doesn't really make use of the private member variable containing the current file because you're passing that as an argument, so even my suggested name is a touch misleading!  Perhaps simply `addRecentFile` would be enough?

src/gui/menu_file.cpp
```diff
@@ -80,8 +80,8 @@ void DissolveWindow::on_FileOpenLocalAction_triggered(bool checked)
         return;
 
     // Request a new file to open
-    QString inputFile =
-        QFileDialog::getOpenFileName(this, "Choose input file to open", QDir().absolutePath(), "Dissolve input files (*.txt)");
+    QString inputFile = QFileDialog::getOpenFileName(this, "Choose input file to open", QDir().absolutePath(),
+                                                     "Dissolve input files (All files (*.*);;Text Files (*.txt))");
```
rprospero, line 84:
```suggestion
                                                     "Dissolve input files (*.txt);;All Files (*.*)");
```

I would suggest defaulting to the filtered view and letting the users choose an unfiltered one, instead of the other way around.

rprospero (changes requested):
A couple of small suggestions

src/gui/menu_file.cpp
```diff
@@ -80,8 +80,8 @@ void DissolveWindow::on_FileOpenLocalAction_triggered(bool checked)
         return;
 
     // Request a new file to open
-    QString inputFile =
-        QFileDialog::getOpenFileName(this, "Choose input file to open", QDir().absolutePath(), "Dissolve input files (*.txt)");
+    QString inputFile = QFileDialog::getOpenFileName(this, "Choose input file to open", QDir().absolutePath(),
+                                                     "Dissolve input files (*.txt;;All Extensions (*.*);;All Files (*))");
```
trisyoungs, line 84: This is starting to get a bit messy now IMO. I don't think I've ever seen an option to choose between all files with/without extensions in a file dialog.  Question - if you remove the filter argument completely from the call to `getOpenFileName` do we just get "all files"?

EthanDevlinSTFC: Removing the argument seems to disable all files. 
The issue with messiness is the use of file name over file type. I could modify the save functions to ensure all files are saved as "name".txt, or we can try to change how the function operates? 
I'm not sure how else we'd manage files without an explicit extension.
The simplest change would be to just have (*) be the only parameter and let the program handle invalid input files as it does already. Though, this would make for a messy file interface.

src/gui/menu_file.cpp
```diff
@@ -81,7 +81,7 @@ void DissolveWindow::on_FileOpenLocalAction_triggered(bool checked)
 
     // Request a new file to open
     QString inputFile = QFileDialog::getOpenFileName(this, "Choose input file to open", QDir().absolutePath(),
-                                                     "Dissolve input files (*.txt;;All Extensions (*.*);;All Files (*))");
+                                                     "Dissolve input files (Text files (*.txt);;All Files (*))");
```
trisyoungs, line 84: There's some odd bracketing going on here!  The second part of the line should read:
```
"Dissolve Input Files (*.txt);;All Files (*)");
```

trisyoungs (approved):
Two small adjustments to make to keep the project manager happy, then this can be merged in.

PR #781, opened 2021-08-19
Array index iteration

EthanDevlinSTFC:
*Not for commit*
This pr is not for committing, unit testing/ code changes pending on approval of algorithms.

Changes implemented allow for iterating over array index's and conversion of nested loops to requested format.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,5 +36,4 @@ changes
 *.restart.prev
 *.state
 
-.vscode/launch.json
-.vscode/settings.json
+.vscode/
--- a/src/dissolve-gui.cpp
+++ b/src/dissolve-gui.cpp
@@ -87,7 +87,8 @@ int main(int args, char **argv)
             }
         }
     }
-
+    // Create recent files menu
+    dissolveWindow.createRecentMenu();
     // Update the main window and exec the app
     dissolveWindow.fullUpdate();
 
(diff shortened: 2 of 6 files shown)


src/templates/array2d.h
```diff
@@ -187,6 +187,25 @@ template <class A> class Array2D
...
+            y++;
+            if (y == nColumns_)
+            {
+                x++;
+                y = 0;
+            }
+        }
+        return indexes;
+    }
+
```
rprospero, line 208: This creates a full array of indices every time we perform the iteration, which will consume around the same order of memory as the Array2D itself.  Instead, what we probably want to do is create two utility classes.  The first class `ArrayIndex` would contain the dimensions of the array and the raw index in the underlying vector.  You could overload the standard increment  to merely operate on the raw index.  The dereferencing operator would then convert that index into a tuple.

The second class, `ArrayRange` would have two member functions, `begin()` and `end()`, which would each return an `ArrayIndex`.

If you're especially eager, overloading more addition, subtraction, and comparison operators would enable random access iterators, which we'll need for parallel loops.

rprospero (changes requested):
The updated code in data2d and data3d looks very clean and supports this abstraction, but I think we'll need a different implementation if we don't want to hammer the heap.  I've included more details below.

PR #786, opened 2021-08-25
openRecent CLI Fixes

### LINKED ISSUE #785: Immediate crash when opening file from CLI
**Describe the bug**
Running `dissolve-gui` from the command line with an input file given as an argument causes a crash before the GUI even starts. This is related to the recent files menu added in #778.

**To Reproduce**
Steps to reproduce the behaviour (if necessary):
1. Run `build/bin/dissolve-gui examples/benzene/input.txt`

**Expected behaviour**
No crash!

EthanDevlinSTFC:
Moved function call order to prevent usage of menu before creation.
Found and fixed other CLI related bugs and issues
closes #785

### DIFF
--- a/src/gui/gui_funcs.cpp
+++ b/src/gui/gui_funcs.cpp
@@ -289,7 +289,7 @@ void DissolveWindow::updateRecentActionList()
             fileInfo = QFileInfo(recentFilePaths.at(i));
             strippedName = fileInfo.fileName();
             filePath = fileInfo.absoluteDir().absolutePath();
-            recentFileActionList_.at(i)->setText(strippedName + "    (" +filePath+")");
+            recentFileActionList_.at(i)->setText(strippedName + "    (" + filePath + ")");
             recentFileActionList_.at(i)->setData(recentFilePaths.at(i));
             recentFileActionList_.at(i)->setVisible(true);
         }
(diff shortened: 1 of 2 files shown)


src/gui/gui_funcs.cpp
```diff
@@ -148,6 +148,10 @@ bool DissolveWindow::openLocalFile(std::string_view inputFile, std::string_view
     auto loadResult = false;
     if (inputFileInfo.exists())
     {
+        // Add file to recent menu
+        QString filePath = inputFileInfo.absoluteDir().absolutePath();
+        std::string fileName = dissolve_.inputFilename().data();
+        addRecentFile(filePath + QString::fromStdString("/" + fileName));
```
trisyoungs, line 154: This would appear to be the root of the issue with the filename not appearing in the menu item - at this stage `dissolve_.inputFilename()` isn't returning anything useful, but you have the whole path + filename in the `QFileInfo` object.   Also, don't you need to be passing the whole absolute path to your `addRecentFile()` function? I f so, you may as well use `QFileInfo::absoluteFilePath()` (see https://doc.qt.io/qt-5/qfileinfo.html#absoluteFilePath).

trisyoungs (changes requested):
Threaded tests still fail - however this is still not your fault, and is now fixed in `develop`.

One big thing - generating the menu entry has a blank space where the filename should be (the path is correct). See my comment below.

One small thing: there's a compiler warning generated because of the initialisation order of objects in `DissolveWindow`.  On line 25 of `gui_funcs.cpp` you should to swap your recentFileNo and threadController initialisations over:
```
QMainWindow(nullptr), dissolve_(dissolve), recentFileNo_(10), threadController_(this, dissolve)
```

PR #789, opened 2021-08-26
Reorganise Unit Tests

trisyoungs:
This PR reorganises the unit tests into categories, and implements a CMake function to register individual tests and reduce code duplication, bringing them in line with the system tests.

A basic unit test covering CIF file read/parse is also added.


DanNixon (approved):
:+1:

PR #790, opened 2021-09-01
Array index iteration 

EthanDevlinSTFC:
New version of index iteration incorporating class and pointers to traverse array index.
I believe this achieves the objectives set out in the previous feedback.
Seeking feedback on the acceptability of the solution prior to full integration.


rprospero:
This looks like a good start here.  I'll admit that my own mental model had been to store the `nX`, `nY`, and `nZ` properties in the class, which would make the dereferencing operator instantaneous, but would make the arithmatic operators slower.  Honestly, I think that your design is the better one.

PR #792, opened 2021-09-02
Array index iteration implementation

### LINKED ISSUE #771: Add index iteration for Array2D and Array3D
As mentioned in #767, the 2D and 3D array classes have ways of iterating through their values, but not through their multi-dimensional indices.  This leads to a pattern in the code of

    for (int x=0; x < arr.x; x++)
        for (int y=0; y < arr.y; y++)
            for (int z=0; z < arr.z; z++)
                arr[{x, y, z}] = f(x, y, z);

What we would like to be able to write is:

    for (auto &[x, y, z] : arr.indices())
        arr[{x, y, z}] = f(x, y, z)

Besides moving away from arrow-shaped code, constructing the iterator for random-access support would enable parallelising these loops in the future.

EthanDevlinSTFC:
Full solution and implementation of raw index based iteration of multi-dimensional arrays.
closes #771


src/math/data3d.cpp
```diff
@@ -346,18 +348,13 @@ bool Data3D::deserialise(LineParser &parser)
     // Read errors / valuse
     if (hasError_)
     {
-        for (auto x = 0; x < x_.size(); ++x)
+        ArrayIndex3D index(values_);
+        while (index++.hasNext())
```
trisyoungs, line 352: I find this notation a bit "opaque" in terms of its logic.  Does the `ArrayIndex3D` class permit the use of constructions like `for (auto id : index)` as well?

src/math/data3d.cpp
```diff
@@ -346,18 +348,13 @@ bool Data3D::deserialise(LineParser &parser)
...
-                    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
-                        return false;
-                    values_[{x, y, z}] = parser.argd(0);
-                    errors_[{x, y, z}] = parser.argd(1);
-                }
-            }
+            if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+                return false;
+            values_[*index] = parser.argd(0);
+            errors_[*index] = parser.argd(1);
```
trisyoungs, line 357: Dereferencing to get the tuple is a bit raw IMO.  Could perhaps define a conversion operator allowing implicit conversion of the index to a `std::tuple<int,int,int>`?

src/templates/arrayIndex2d.h
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2021 Team Dissolve and contributors
+
+#include "base/messenger.h"
+#include "templates/array2d.h"
+#include "templates/list.h"
+#include "templates/vector3.h"
+#include <numeric>
+#include <tuple>
+#include <vector>
```
trisyoungs, line 10: Most of these includes are redundant:
```suggestion
#include <tuple>
```

src/templates/arrayIndex2d.h
```diff
@@ -0,0 +1,75 @@
...
+        return *this;
+    }
+
+    std::tuple<int, int> operator*()
+    {
+        int rawIndex = ptr_ - &arr_[{0, 0}];
+        int x = rawIndex / nY_;
+        int y = rawIndex % nY_;
+        std::tuple<int, int> arrayIndex = std::make_tuple(x, y);
+        return arrayIndex;
```
trisyoungs, line 61:
```suggestion
        return {rawIndex / nY_, rawIndex % nY_};
```

src/templates/arrayIndex2d.h
```diff
@@ -0,0 +1,75 @@
...
+    std::tuple<int, int> begin() { return std::make_tuple(0, 0); }
+
+    std::tuple<int, int> end() { return std::make_tuple(nX_, nY_); }
+
+    bool hasNext() { return ptr_ <= &arr_[{nX_, nY_}]; }
+
+    private:
+    int nX_;
+    int nY_;
+    const double *ptr_;
```
trisyoungs, line 73: Assuming that this is a `double` will break the class for any array that contains other objects.  Does the `ptr_` member really need to be a pointer at all? It's only used to determine the raw index, so could just be a straight `int`.  Extending this, does the reference to the array even need to be stored?  All we need to do is generate indices over some range - the underlying array itself is kinda irrelevant.

src/templates/arrayIndex3d.h
```diff
@@ -0,0 +1,79 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2021 Team Dissolve and contributors
+
+#include "base/messenger.h"
+#include "templates/array3d.h"
+#include "templates/list.h"
+#include "templates/vector3.h"
+#include <numeric>
+#include <tuple>
+#include <vector>
```
trisyoungs, line 10:
```suggestion
#include <tuple>
```

src/templates/arrayIndex3d.h
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    std::tuple<int, int, int> operator*()
+    {
+        int rawIndex = ptr_ - &arr_[{0, 0, 0}];
+        int x = rawIndex / (nY_ * nZ_);
+        int y = rawIndex % nY_ / nZ_;
+        int z = rawIndex % nY_ % nZ_;
+        std::tuple<int, int, int> arrayIndex = std::make_tuple(x, y, z);
+        return arrayIndex;
```
trisyoungs, line 64:
```suggestion
        return {rawIndex / (nY_ * nZ_), rawIndex % nY_ / nZ_, rawIndex % nY_ % nZ_};
```

src/templates/arrayIndex3d.h
```diff
@@ -0,0 +1,79 @@
...
+
+    std::tuple<int, int, int> end() { return std::make_tuple(nX_, nY_, nZ_); }
+
+    bool hasNext() { return (ptr_ <= &arr_[{nX_, nY_, nZ_}]); }
+
+    private:
+    int nX_;
+    int nY_;
+    int nZ_;
+    const double *ptr_;
```
trisyoungs, line 77: Same here.

trisyoungs (changes requested):
This is all good, however I think there are some nice simplifications that can be made. It would be good to get @rprospero 's opinion on some of my comments here.

Also, this is crying out for a unit test!

src/math/data2d.cpp
```diff
@@ -355,27 +356,23 @@ bool Data2D::deserialise(LineParser &parser)
     // Read errors / valuse
     if (hasError_)
     {
-        for (auto x = 0; x < x_.size(); ++x)
+        ArrayIndex2D index(values_);
+        while (index++.hasNext())
```
rprospero, line 360: Using the `while` loop and our own `hasNext` method is a bit fragile.  It would be better if we could treat this like a standard for loop.  It should look something more like

```suggestion
        for (auto index : ArrayIndex2D(values_))
```

This also offers the coder the option of 

```suggestion
        for (auto &[x, y] : ArrayIndex2D(values_))
```

for when the individual coordinates are more significant (which they are not, in this instance).

src/math/data2d.cpp
```diff
@@ -355,27 +356,23 @@ bool Data2D::deserialise(LineParser &parser)
...
-            {
-                if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
-                    return false;
-                values_[{x, y}] = parser.argd(0);
-                errors_[{x, y}] = parser.argd(1);
-            }
+            if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+                return false;
+            values_[*index] = parser.argd(0);
+            errors_[*index] = parser.argd(1);
```
rprospero, line 365: With my changes suggested above, we get a little cleaner code here
```suggestion
            values_[index] = parser.argd(0);
            errors_[index] = parser.argd(1);
```

src/math/data2d.cpp
```diff
@@ -355,27 +356,23 @@ bool Data2D::deserialise(LineParser &parser)
...
         {
-            for (auto y = 0; y < y_.size(); ++y)
-            {
-                if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
-                    return false;
-                values_[{x, y}] = parser.argd(0);
-            }
+            if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+                return false;
+            values_[*index] = parser.argd(0);
```
rprospero, line 375: Same comments as above

src/math/data2d.cpp
```diff
@@ -406,20 +403,20 @@ bool Data2D::serialise(LineParser &parser) const
...
     {
-        for (auto x = 0; x < x_.size(); ++x)
+        ArrayIndex2D index(values_);
+        while (index++.hasNext())
         {
-            for (auto y = 0; y < y_.size(); ++y)
-                if (!parser.writeLineF("{:e}  {:e}\n", values_[{x, y}], errors_[{x, y}]))
-                    return false;
+            if (!parser.writeLineF("{:e}  {:e}\n", values_[*index], errors_[*index]))
+                return false;
```
rprospero, line 410: Same as above

src/math/data2d.cpp
```diff
@@ -406,20 +403,20 @@ bool Data2D::serialise(LineParser &parser) const
...
     {
-        for (auto x = 0; x < x_.size(); ++x)
+        ArrayIndex2D index(values_);
+        while (index++.hasNext())
         {
-            for (auto y = 0; y < y_.size(); ++y)
-                if (!parser.writeLineF("{:e}\n", values_[{x, y}]))
-                    return false;
+            if (!parser.writeLineF("{:e}  {:e}\n", values_[*index], errors_[*index]))
+                return false;
```
rprospero, line 419: Same as above

src/math/data3d.cpp
rprospero: Hopefully, my suggestion will fix that opacity.

src/math/data3d.cpp
rprospero: Implementing my suggested course should remove the need for the dereference.

src/templates/arrayIndex2d.h
```diff
@@ -0,0 +1,75 @@
...
+
+template <class A> class ArrayIndex2D
+{
+    public:
+    ArrayIndex2D(A &inArray) : arr_(inArray)
+    {
+        nX_ = inArray.nRows();
+        nY_ = inArray.nColumns();
+        ptr_ = &inArray[{0, 0}];
+    }
```
rprospero, line 20: You'll probably want another constructor so that you can create indices at arbitrary parts of the array, and not just at {0, 0}

src/templates/arrayIndex2d.h
rprospero: Agreed.  We only ever care about the index, so it doesn't need to be a pointer and we don't need the refrence to the raw array.  This also means that the class can be used for iteration where there isn't necessarily an underlying Array2D.

src/templates/arrayIndex2d.h
```diff
@@ -0,0 +1,75 @@
...
+        int rawIndex = ptr_ - &arr_[{0, 0}];
+        int x = rawIndex / nY_;
+        int y = rawIndex % nY_;
+        std::tuple<int, int> arrayIndex = std::make_tuple(x, y);
+        return arrayIndex;
+    }
+
+    std::tuple<int, int> begin() { return std::make_tuple(0, 0); }
+
+    std::tuple<int, int> end() { return std::make_tuple(nX_, nY_); }
```
rprospero, line 66: For the range-based `for` loops and algorithms to work properly, these really should return `ArrayIndex` instances.  `begin` should have an index of zero and `end` should have an index equal to the size of the array.  You'll need to correct this if you're going to use my suggestions.

src/templates/arrayIndex2d.h
```diff
@@ -0,0 +1,75 @@
...
+
+    bool hasNext() { return ptr_ <= &arr_[{nX_, nY_}]; }
+
```
rprospero, line 69: If we have proper bounding iterators to compare against (i.e. `begin` and `end`), then we shouldn't need this class.  I'm also a little concerned about the logic on this.  For a 3x3 array, wouldn't `{nX_, nY_}` be `{3, 3}`, but our final array element would be `{2, 2}`?

rprospero (changes requested):
I think we're getting closer to where we want to be

rprospero:
I noticed this morning that the unit tests won't actually run.  You'll need to add

```
dissolve_unit_test(arrayIteration.cpp)
```

to the end of `unit/algorithms/CMakeLists.txt`

PR #795, opened 2021-09-03
Tab Management

trisyoungs:
This PR moves storage of the various tabs from `std::vector<std::shared_ptr>` to `std::vector<QPointer>`, since the former imposed additional memory management _on top of_ that already present in Qt (via `QObject`). This caused odd behaviour and repeatable crashes in places. Some other small issues discovered while performing this work are also fixed.

Moving to `QPointer`and the crash fixes will be incorporated into v0.8.3.

### DIFF
--- a/src/dissolve-gui.cpp
+++ b/src/dissolve-gui.cpp
@@ -96,9 +96,6 @@ int main(int args, char **argv)
 
     auto result = app.exec();
 
-    // Clear all data
-    dissolve.clear();
-
     // End parallel communication
     ProcessPool::finalise();
 
--- a/src/gui/gettabnamedialog.h
+++ b/src/gui/gettabnamedialog.h
@@ -17,14 +17,14 @@ class GetTabNameDialog : public QDialog
     Q_OBJECT
 
     public:
-    GetTabNameDialog(QWidget *parent, const std::vector<std::shared_ptr<MainTab>> &currentTabs);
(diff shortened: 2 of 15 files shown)


rprospero (commented):
I think that I'm understanding the problem here.  We have shared pointers to Qt elements.  Qt, on its own, frees certain elements when they go out of scope.  Using the `shared_ptr` then also tries to free the element when it leaves scope, leading to a double free situation.

The only thing that bugs me is that we've replaced the vectors of shared pointers with vectors of raw pointers.  I know that QPointer, when the object it's pointing to is deleted, will automatically adjust itself to `nullptr`.  The raw pointer will not.  Is there any reason that these couldn't all be vectors of QPointer?  That would ensure that the pointers are updated if the data is lost and also indicate where the memory management is being handled.

trisyoungs:
We're talking about the `std::vector<MainTab *> allTabs_` vector here, right?  I forgot to mention this in the PR text. Initially I had gone with the obvious choice of a `std::vector<QPointer<MainTab>>` but this was not an option since `MainTab` is not a `QObject` - rather, it is the derived tab classes.  Thus, since `QPointer` must reference a `QObject`, maintaining a vector of base-class pointers can't be done using that wrapper.

I suspect that, in the medium term, it may well be possible to just ditch the `allTabs_` array altogether, since it is there mostly for convenience.

PR #797, opened 2021-09-06
Flat atom packed

rprospero:
This is a WIP on moving all of the atoms into a single vector.  I've only posted this at the moment for a lazy way to diff against develop


rprospero:
This has been superseded by #801

PR #804, opened 2021-10-04
Update qt install

rprospero:
It seems that the qt install of aqt has changed and is breaking the build.  This PR should hopefully fix the problem.


rprospero:
This should solve issue #805.

PR #807, opened 2021-10-11
Bundle and container support through nix flakes

rprospero:
This is a plan to use the nix infrastructure for building dissolve on Linux.  It *should* provide the following benefits.

- Deterministic builds.  No one can break the build but us.
- Singularity containers.  Dissolve provides utilities for automatically creating containers from packages.
- Simple build configurability.  Switching out GCC versions, or replacing it with Clang, it a one line change.
- Linux packaging.  The nix bundler should provide universal linux binaries that can be run on any system with all dependencies included.
- Developer onboarding.  A single `nix develop` and new developers have a full featured development environment.

Things that I need to take care of before removing the `draft` tag on this pr:

- Support building with multi-threading support
- Support building with GUI support
- Update cache from CI

Things that are outside the scope of this PR

- Mac builds.  We'll leave in the mac infrastructure for right now and only focus on linux packaging.  This is all that really matter for the container generation and is also one of the bigger sources of packaging issue.  It will also be *non-trivial* to get Qt 6 working on the nix version
- Windows builds.  Again, we'll leave the existing infrastructure in place.  The documentation says that windows builds are possible, but that's about all that it says on the issue.  Qt 6 will likely be an even larger issue here.
- ARM builds.  Nix support gives us these for free, but, without an ARM system to test it on, this will be very difficult to test.  Not to be a broken record, but Qt 6 will also be a bit of an issue here.


rprospero:
@trisyoungs I believe that I now have a proper docker container with OpenGL support.  I can load it up within the docker container and still create molecules via the 3D drawing interface.  It should try and connect to the video card and fall back on software rendering if that fails.  Having the software rendering backend adds a lot of overhead to the docker image, but I feel its worth it for the fairly common case of 3d hardware issues.

Testing steps:
1. Download the [image](https://dev.azure.com/DisorderedMaterials/_apis/resources/Containers/22825676/linux-GUI-docker?itemPath=linux-GUI-docker%2Fresult).
2. load the image with `docker load`
3. `docker run --rm -e DISPLAY=unix$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix dissolve-gui`

trisyoungs:
@rprospero - Success! Performance is of course compromised (viewing a Configuration in full 3D is slow, but very usable for smaller systems) but a small price to pay.

rprospero:
@trisyoungs Does view the configuration run any faster if you add a `--privileged` flag to the `docker run` call?  That *might* give you access to the GPU.

trisyoungs:
> @trisyoungs Does view the configuration run any faster if you add a `--privileged` flag to the `docker run` call? That _might_ give you access to the GPU.

@rprospero Alas, no.

changeversion
```diff
@@ -106,9 +106,12 @@ then
   if [ "$CHECK" = "false" ]
   then
     sed -i -e "s/#define DISSOLVEVERSION \"[0-9\.]\+\(.*\)\"/#define DISSOLVEVERSION \"$MAJOR.$MINOR.$PATCH\"\1/g" src/main/version.cpp
+    sed -i -e "s/version = \".*\"/version = \"$MAJOR.$MINOR.$PATCH\"/" flake.nix
   fi
   echo -n "         version.cpp (Program Version) : "
   grep '#define DISSOLVEVERSION' src/main/version.cpp | sed -e 's/#define DISSOLVEVERSION "\([0-9a-z\.]\+\).*"/\1/g'
+  echo -n "         flake.nix (Program Version) : "
+  grep " version =" flake.nix | sed -e "s/\W*version = \"\(.*\)\";/\1/"
```
trisyoungs, line 114:
```suggestion
  fi
  echo -n "         version.cpp (Program Version) : "
  grep '#define DISSOLVEVERSION' src/main/version.cpp | sed -e 's/#define DISSOLVEVERSION "\([0-9a-z\.]\+\).*"/\1/g'
  
  
  # Flake (flake.nix)
  if [ "$CHECK" = "false" ]
  then
    sed -i -e "s/version = \".*\"/version = \"$MAJOR.$MINOR.$PATCH\"/" flake.nix
  fi
  echo -n "           flake.nix (Program Version) : "
  grep " version =" flake.nix | sed -e "s/\W*version = \"\(.*\)\";/\1/"
```

trisyoungs: Just a picky layout thing with the (admittedly clunky) `changeversion` script.

tests/epsr/H2Ox-weighted-total.gr.broad
```diff
@@ -0,0 +1,310 @@
+1.2500000000e-02  -3.5088629018e-01
```
trisyoungs, line 1: Looks like a data file got added somehow...

trisyoungs (approved):
Just one small formatting thing with the `changeversion` script, and a file that probably should be removed.

PR #834, opened 2021-11-01
NETA Strings

trisyoungs:
This PR adds basic creation of NETA strings from a specified `SpeciesAtom` - this allows connectivity descriptions to be generated automatically and used for similarity matching, forcefield creation etc.

Currently, only basic connectivity is considered - including ring information is left for a future commit when such functionality is required.

### DIFF
--- a/src/neta/neta.cpp
+++ b/src/neta/neta.cpp
@@ -21,6 +21,8 @@ NETADefinition::NETADefinition(std::string_view definition) : rootNode_(nullptr)
         create(definition);
 }
 
+NETADefinition::NETADefinition(const SpeciesAtom *i, int maxDepth) : rootNode_(nullptr), valid_(false) { create(i, maxDepth); }
+
 /*
  * Data
  */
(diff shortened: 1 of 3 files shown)


src/neta/neta.cpp
```diff
@@ -91,6 +93,50 @@ bool NETADefinition::create(std::string_view definition, const Forcefield *assoc
...
+    if (i->Z() != Elements::H)
+        neta += fmt::format(",nh={}", nH);
+    return neta;
+}
+
+// Create from specified atom and its connectivity
+bool NETADefinition::create(const SpeciesAtom *i, int maxDepth)
+{
+    std::vector<const SpeciesAtom *> path;
+    definitionString_ = netaString(i, 0, maxDepth, path);
```
rprospero, line 136:
```suggestion
    definitionString_ = netaString(i, 0, maxDepth, {});
```

Just clearing out an extra temporary.

rprospero (approved):
I have one small code suggestion which I *believe* will work.  Otherwise, it's ready to go in.

PR #837, opened 2021-11-02
Remove list dependency from ChartBlock and ChartHotSpot

rprospero:
This is a quickie to remove the `ListItem` dependency from `ChartBlock` and `ChartHotSpot`.  There's some other tidying that could be done, but I didn't want to get too deep into it, since we're planning to eventually remove all this code anyway.  This removes two more parts of #257.


trisyoungs (approved):
:+1:

PR #847, opened 2021-11-08
Cause CI to fail when build fails

### LINKED ISSUE #844: CI passes even if compilation fails
**Describe the bug**
Following recent modifications to the CI pipelines, it appears that they are letting through uncompilable code. Noticed this when I check out `develop` just now, which contains commits from #792 which breaks the build on my local machine.

You can see in the transcript for the build that there a errors generated, but the stage still passes:
https://dev.azure.com/DisorderedMaterials/Dissolve/_build/results?buildId=4068&view=logs&jobId=7e36aa70-1544-5ed8-e9c7-9b511b32e9bc&j=7e36aa70-1544-5ed8-e9c7-9b511b32e9bc&t=aaa4af23-b6fe-50ec-9c49-e86ae5b50441

rprospero:
Currently, the build on the CI has its output piped to a cache.  The problem is that the pipe then clobbers the exit code of the build, so the CI continues on even when the build fails.  The current hypothesis is to run the build twice.  The first run performs the actual build and unit tests and the CI will fail if this fails.  The second run is then piped out to the cache.  Since nix locally caches the builds, there's no extra time added to run the build twice.

This will close #844.

### DIFF
--- a/.azure-pipelines/templates/build-linux.yml
+++ b/.azure-pipelines/templates/build-linux.yml
@@ -31,6 +31,7 @@ steps:
   - bash: |
       set -ex
       source ~/.nix-profile/etc/profile.d/nix.sh
+      nix build -L .#${{ parameters.package }}
       nix build -L .#${{ parameters.package }} | jq -r '.[].outputs | to_entries[].value' | cachix push dissolve-nix
     env:
       CACHIX_AUTH_TOKEN: $(CACHIX_AUTH_TOKEN)


rprospero:
I confirmed that the CI now fails when the compile or unit tests fail.  I've also fixed the issues with the array iterators that were preventing the build.

trisyoungs (approved):
Looks like a reasonable solution. :+1:

PR #848, opened 2021-11-08
Move SpeciesAtom off std::list

trisyoungs:
This PR moves `SpeciesAtom` storage from `std::list` to `std::vector`. References between `SpeciesAtom`s and `SpeciesBond` objects are maintained via the new move constructor/assignment operator.

Some functions in the `Species` class itself have been modified to only accept atom indices rather than `SpeciesAtom` pointers.


src/classes/species_atomic.cpp
```diff
@@ -5,25 +5,48 @@
 #include "classes/species.h"
 #include "data/atomicmasses.h"
 
-// Add a new atom to the Species
-SpeciesAtom &Species::addAtom(Elements::Element Z, Vec3<double> r, double q)
+// Recursively select atoms along any path from the specified one, ignoring the bond(s) provided
+void Species::selectFromAtomRecursive(std::vector<SpeciesAtom *> selection, SpeciesAtom *i, SpeciesBond &exclude,
```
rprospero, line 9: I believe that I understand the logic here, but I just wanted to confirm that `selection` was not supposed to be a reference.

rprospero (approved):
I had one question where I wanted to double check that we didn't want a reference (and I think we don't) but it otherwise looks good.

src/classes/species_atomic.cpp
trisyoungs: Oh, good catch! The algorithm probably works more-or-less in the absence of a reference, but it should most certainly be one.

PR #849, opened 2021-11-09
Singularity no fork

rprospero:
This is another fork of the singularity branch that exists in the main dissolve repo so that it can access secrets and I can confirm that the release code works.


trisyoungs (commented):
This all looks OK to me, provided we can overcome or explain away the inability of my system to run the GUI sif file.

One minor niggle - in the `Testing` stages we refer to a plain `Build` job, but in reality this is a cached build with the intent of running the tests, so I would suggest renaming to `Rebuild / Test` (or something similar) for clarity.

DanNixon:
The GUI container seems to work fine for me via Xwayland. @trisyoungs I'd be curious to see if Xwayland is running or started when you try and launch the container.

trisyoungs:
OK, so I'm on Gnome (I know, I know...) 3.34 which allegedly has "dynamic" support for Wayland apps, running a Wayland layer when necessary. However, it seems that I can't even run the login screen on Wayland unless I want an unusable black void - I had apparently already changed a setting for the desktop manager to disable Wayland. I suspect this might be a combination of relatively old Gnome coupled with the proprietary Nvidia drivers, so I may not solve this problem soon...

DanNixon:
What kernel are you running? I know there was odd issues with Nvidia and Wayland on IIRC 5.10 and below. Certainly the latest release has it fixed.

trisyoungs:
5.3.18.  Practically stone-age.

DanNixon:
Very retro, although I now recall it was a new issue in one of the newer releases, so should not be the issue here. If you are on a particularly old version of Gnome I wonder if it just does not support Wayland properly, does anything definitely work under Wayland? (Firefox is likely the easiest to check, `about:support` and look for "Window Protocol")

trisyoungs:
Well that's a hard "x11" right there. I am a minor version behind in the OpenSuSE world, but even the latest release sticks with 5.3.18 as it is tied to their enterprise kernel. Might try an upgrade to see if I get any joy, but in the meantime I guess we discount my failure to run the Singularity images and proceed.

Issue #827, opened 2021-10-24
Epic / Keyword Modernisation


trisyoungs:
### Focus
Modernise keyword-related classes, and remove them from `List` as part of #257. This issue keeps track of a draft PR stack / work in progress, briefly stating the significant changes in each PR.

### 1. Simplify Data Storage
Branch: [`new-keywords-1`](https://github.com/disorderedmaterials/dissolve/tree/new-keywords-1)
- [x] Temporarily move to `std::vector` of pointers over `List` in `KeywordList`
- [x] Move all keywords to store data within keyword, rather than references
- [x] Update `KeywordList` retrieval functions to permit changes.

### 2. Local Variables
Branch: [`new-keywords-2`](https://github.com/disorderedmaterials/dissolve/tree/new-keywords-2)
- [x] Remove inconsistent `arguments_` storage for keywords
- [x] Rework `KeywordList` to store base class ref and derived class `typeid`
- [x] Move POD keywords to be local class members, and add them to managing class
- [x] Move class-based keywords to be local class members, and add them to managing class
- [x] Move STL-based keywords to be local class members, and add them to managing class
- [x] Remove `LinkToKeyword` class since this is no longer necessary?
- [x] Make `KeywordBase::set_` private (?) and remove derived-class `setData`/`setValue` members in favour of non-const return of `data_`.
- [x] Remove need for `KeywordXXX` -> `KeywordData` -> `KeywordBase` class hierarchy
- [x] Remove `asInt()`, `asDouble()` etc. functions (no longer necessary once move to local members)
- [x] Better setting of `optionMask` on `KeywordInfo` - separate function in `KeywordList`?

### 3. Remove Type, Move To Widget Producers
Branch: [`new-keywords-3`](https://github.com/disorderedmaterials/dissolve/tree/new-keywords-3)
- [x] Remove `KeywordData::type_` enum
- [x] Clearly separate keywords from widgets via producer-style class
- [x] Take derived class keyword pointers in widget constructors.

### 4. Tidy
Branch: [`new-keywords-4`](https://github.com/disorderedmaterials/dissolve/tree/new-keywords-4)
- [x] Rename `read` and `write` to `serialise` and `deserialise` for consistency.
- [x] ~Rename module `init.cpp` for `control.cpp` to be more descriptive and consistent~
- [x] Rename `KeywordList` to `KeywordStore`.
[...]

PR #854, opened 2021-11-12
Keywords 1 & 2

trisyoungs:
This PR addresses the handling of keywords, greatly improving data locality in `Module`s and `Node`s and converting keyword classes into "managing wrappers".

To summarise the major changes:
1. No data is stored on the keyword - instead, references to member variables are stored instead. This obfuscates the need for `KeywordData<T>`, which has been removed.
2. Keywords are created via a templated `add()` function in `KeywordList`, passing the member variable reference and any other relevant parameters to the keyword constructor via a parameter pack.
3. Setting and getting of keyword data is now greatly reduced as the data are local to where they are most often needed, but templated setter/getter functions have been implemented with a minimal set of classes in order to provide this functionality.

A follow-up PR will remove the `KeywordBase::KeywordDataType` enum and work on simplifying keyword creation and usage in the GUI.

Note: This PR includes the branch `new-keywords-1` (see epic tracking this work #827 for more details on the content).


src/modules/skeleton/init.cpp
```diff
@@ -7,5 +7,5 @@
 // Perform any necessary initialisation for the Module
 void SkeletonModule::initialise()
 {
-    // 	keywords_.add(new BoolKeyword(true), "ExampleKeyword", "Example keyword description", "<args>");
+    // 	keywords_.add<BoolKeyword>("ExampleKeyword", "Example keyword description", localVariable_);
```
rprospero, line 10: Can we just make this an empty function?

src/modules/test/init.cpp
```diff
@@ -7,5 +7,5 @@
 // Perform any necessary initialisation for the Module
 void TestModule::initialise()
 {
-    // 	keywords_.add(new BoolKeyword(true), "ExampleKeyword", "Example keyword description", "<args>");
+    // 	keywords_.add<BoolKeyword>("ExampleKeyword", "Example keyword description", localVariable_);
```
rprospero, line 10: Empty?

rprospero (approved):
This feels a lot cleaner than the old system.  There's two very minor suggestions.  Ignore my comment about using a template base class.  I realised later on that the trivial implementation would cause problems with storing the pointers.  There's still a way we can do it, but it's outside the scope of this PR.

src/modules/skeleton/init.cpp
trisyoungs: Yes we can. I have earmarked tidying of the module directories in a future issue - there are too many files in there with too little content than is sensible.

PR #863, opened 2021-11-15
Bugfix 15/11/21 Large Molecule Manipulations

trisyoungs:
This PR addresses issues with the manipulation of the coordinates of large molecules, specifically being related to the calculation of the centre-of-geometry.

### DIFF
--- a/src/classes/box.cpp
+++ b/src/classes/box.cpp
@@ -198,7 +198,11 @@ Vec3<double> Box::getReal(Vec3<double> r) const
 }
 
 /*
+<<<<<<< HEAD
  * Geometry Calculation
+=======
+ * Geometry
+>>>>>>> efb17328b... Add manipulator functions to Molecule, and calculate COG correctly for large (all) molecules.
  */
 
 // Return angle (in degrees) between coordinates
--- a/src/classes/box.h
+++ b/src/classes/box.h
@@ -136,7 +136,11 @@ class Box
     virtual double minimumDistanceSquared(const Vec3<double> &r1, const Vec3<double> &r2) const = 0;
 
     /*
+<<<<<<< HEAD
      * Geometry Calculation
+=======
+     * Geometry
+>>>>>>> efb17328b... Add manipulator functions to Molecule, and calculate COG correctly for large (all) molecules.
      */
     public:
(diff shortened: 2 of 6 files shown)


src/classes/box.cpp
```diff
@@ -198,7 +198,11 @@ Vec3<double> Box::getReal(Vec3<double> r) const
 }
 
 /*
+<<<<<<< HEAD
  * Geometry Calculation
+=======
+ * Geometry
+>>>>>>> efb17328b... Add manipulator functions to Molecule, and calculate COG correctly for large (all) molecules.
```
rprospero, line 205: This feels like a git merge that didn't got through properly

src/classes/box.h
```diff
@@ -136,7 +136,11 @@ class Box
     virtual double minimumDistanceSquared(const Vec3<double> &r1, const Vec3<double> &r2) const = 0;
 
     /*
+<<<<<<< HEAD
      * Geometry Calculation
+=======
+     * Geometry
+>>>>>>> efb17328b... Add manipulator functions to Molecule, and calculate COG correctly for large (all) molecules.
```
rprospero, line 143: Git merge again

src/classes/molecule.cpp
```diff
@@ -65,6 +67,81 @@ void Molecule::updateAtoms(std::vector<Atom> &source)
...
+    }
+}
+
+// General manipulation function working on reassembled molecule
+void Molecule::traverseLocal(const Box *box, ManipulationFunction action)
+{
+    std::vector<bool> flags(atoms_.size(), false);
+    action(atoms_[0], atoms_[0]->r());
+    recurseLocal(flags, box, 0, action);
+}
```
rprospero, line 126: Do we need the non-const version?  It seems to do exactly the same as the const one.

rprospero (approved):
A couple of minor points, but otherwise looks good.

src/classes/molecule.cpp
trisyoungs: I'm afraid so - sometimes the lambda is read-only on the `Atom`s and called from a local `const` function, whereas other times it's used for actual manipulations.

PR #876, opened 2021-11-18
Remove list dependency from Procedure Node

rprospero:
This should remove the last dependency on the old List container


rprospero:
This PR now completely removes the the List template.  It also removes the `SelectForcefieldDialog`, since it depended on the List class and didn't seem to be used anywhere.

src/gui/models/procedureNodeModel.cpp
```diff
@@ -3,35 +3,35 @@
 
 #include "gui/models/procedureNodeModel.h"
 
-Q_DECLARE_METATYPE(const ProcedureNode *);
+Q_DECLARE_METATYPE(ConstNodeRef);
```
trisyoungs, line 6: No longer necessary?
```suggestion
```

src/keywords/dynamicsitenodes.cpp
```diff
@@ -44,7 +45,7 @@ bool DynamicSiteNodesKeyword::read(LineParser &parser, int startArg, const CoreD
         return Messenger::error("Parent ProcedureNode not set, so can't read DynamicSiteNode data.\n");
 
     // Create a new DynamicSite and add it to our data RefList
-    auto *dynamicSite = new DynamicSiteProcedureNode(parentNode_);
+    auto dynamicSite = std::dynamic_pointer_cast<DynamicSiteProcedureNode>(parentNode_->shared_from_this());
```
trisyoungs, line 48: This fixes the error in the CI for the `calculate_dangle` test.
```suggestion
    auto dynamicSite = std::make_shared<DynamicSiteProcedureNode>(
        std::dynamic_pointer_cast<SelectProcedureNode>(parentNode_->shared_from_this()));
```

trisyoungs (changes requested):
Requested changes only because there's a fix for the failing CI in my comments. Otherwise looks OK.

I note that there's around a 10% performance penalty when running analysis-heavy stuff (tested with the `calculate_dangle` system test) because of the move to `std::ahared_ptr`s here. Not a job for right now, but we might be able to reclaim that in the hot loops by getting and using the raw pointer reference temporarily, rather than suffer the cost of indirection each time.

trisyoungs:
Before my PRs cause this PR any more grief, I suggest that you merge in once the conflicts are fixed _before_ #879 - I'll then deal with the merge conflicts that result from all my moving of files about.

PR #887, opened 2021-11-26
Optimise species loading

trisyoungs:
This PR addresses incredible slowness of input file loading, discovered whilst working on a simulation with a Species containing approximately 27k atoms.  The root causes were:

1. Continual reallocation of vectors
2. Unnecessary repeated updates of associated objects when setting `AtomType`s on atoms.

The resolutions are as follows:
1. Input files now contain (optional) keywords allowing vectors to be reserved.
2. An internal `AtomTypeMix` is no longer maintained by the `Species` - instead this is calculated and returned when required.

### DIFF
--- a/src/classes/isotopologue.cpp
+++ b/src/classes/isotopologue.cpp
@@ -32,7 +32,7 @@ std::string_view Isotopologue::name() const { return name_; }
 void Isotopologue::update()
 {
     // Prune any types in our list that are not used in the parent species
-    const auto &atomTypes = parent_->atomTypes();
+    const auto atomTypes = parent_->atomTypes();
     isotopes_.erase(std::remove_if(isotopes_.begin(), isotopes_.end(),
                                    [&atomTypes](auto value) { return !atomTypes.contains(std::get<0>(value)); }),
                     isotopes_.end());
--- a/src/classes/isotopologue.cpp
+++ b/src/classes/isotopologue.cpp
@@ -32,7 +32,7 @@ std::string_view Isotopologue::name() const { return name_; }
 void Isotopologue::update()
 {
     // Prune any types in our list that are not used in the parent species
-    const auto atomTypes = parent_->atomTypes();
+    const auto &atomTypes = parent_->atomTypes();
     isotopes_.erase(std::remove_if(isotopes_.begin(), isotopes_.end(),
                                    [&atomTypes](auto value) { return !atomTypes.contains(std::get<0>(value)); }),
                     isotopes_.end());
--- a/src/classes/neutronweights.cpp
+++ b/src/classes/neutronweights.cpp
@@ -156,7 +156,7 @@ void NeutronWeights::calculateWeightingMatrices()
 
         // Using the underlying Species, construct a flag matrix which states the AtomType interactions we have present
         const auto *sp = topes.species();
-        const auto speciesAtomTypes = sp->atomTypes();
+        const auto &speciesAtomTypes = sp->atomTypes();
         intraFlag = false;
         dissolve::for_each_pair(
             ParallelPolicies::seq, atomTypes_.begin(), atomTypes_.end(),
--- a/src/classes/neutronweights.cpp
+++ b/src/classes/neutronweights.cpp
@@ -156,7 +156,7 @@ void NeutronWeights::calculateWeightingMatrices()
 
         // Using the underlying Species, construct a flag matrix which states the AtomType interactions we have present
         const auto *sp = topes.species();
-        const auto &speciesAtomTypes = sp->atomTypes();
+        const auto speciesAtomTypes = sp->atomTypes();
         intraFlag = false;
         dissolve::for_each_pair(
             ParallelPolicies::seq, atomTypes_.begin(), atomTypes_.end(),
--- a/src/classes/species_intra.cpp
+++ b/src/classes/species_intra.cpp
@@ -68,9 +68,8 @@ const std::vector<SpeciesBond> &Species::bonds() const { return bonds_; }
 // Return whether SpeciesBond between specified SpeciesAtoms exists
 bool Species::hasBond(const SpeciesAtom *i, const SpeciesAtom *j) const
 {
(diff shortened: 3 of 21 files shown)


src/classes/isotopologue.cpp
```diff
@@ -32,7 +32,7 @@ std::string_view Isotopologue::name() const { return name_; }
 void Isotopologue::update()
 {
     // Prune any types in our list that are not used in the parent species
-    const auto &atomTypes = parent_->atomTypes();
+    const auto atomTypes = parent_->atomTypes();
```
rprospero, line 35: Did we need to allocate a copy of the vector here?  It's const, so we don't have to worry about making upstream changes.

src/classes/neutronweights.cpp
```diff
@@ -156,7 +156,7 @@ void NeutronWeights::calculateWeightingMatrices()
 
         // Using the underlying Species, construct a flag matrix which states the AtomType interactions we have present
         const auto *sp = topes.species();
-        const auto &speciesAtomTypes = sp->atomTypes();
```
rprospero, line 159: Same question

src/classes/species_intra.cpp
```diff
@@ -68,9 +68,8 @@ const std::vector<SpeciesBond> &Species::bonds() const { return bonds_; }
 // Return whether SpeciesBond between specified SpeciesAtoms exists
 bool Species::hasBond(const SpeciesAtom *i, const SpeciesAtom *j) const
 {
-    auto it = std::find_if(bonds_.cbegin(), bonds_.cend(), [i, j](const auto &bond) { return bond.matches(i, j); });
-
-    return it != bonds_.cend();
+    return std::find_if(bonds_.cbegin(), bonds_.cend(), [i, j](const auto &bond) { return bond.matches(i, j); }) !=
+           bonds_.cend();
```
rprospero, line 72:
```suggestion
    return std::any_of(bonds_.cbegin(), bonds_.cend(), [i, j](const auto &bond) { return bond.matches(i, j); });
```

src/classes/species_intra.cpp
```diff
@@ -237,9 +236,8 @@ const std::vector<SpeciesAngle> &Species::angles() const { return angles_; }
 // Return whether SpeciesAngle between SpeciesAtoms exists
 bool Species::hasAngle(SpeciesAtom *i, SpeciesAtom *j, SpeciesAtom *k) const
 {
-    auto it = std::find_if(angles_.cbegin(), angles_.cend(), [i, j, k](const auto &angle) { return angle.matches(i, j, k); });
-
-    return it != angles_.cend();
+    return std::find_if(angles_.cbegin(), angles_.cend(), [i, j, k](const auto &angle) { return angle.matches(i, j, k); }) !=
+           angles_.cend();
```
rprospero, line 240:
```suggestion
    return std::any_of(angles_.cbegin(), angles_.cend(), [i, j, k](const auto &angle) { return angle.matches(i, j, k); });
```

src/classes/species_intra.cpp
```diff
@@ -300,8 +298,8 @@ const std::vector<SpeciesTorsion> &Species::torsions() const { return torsions_;
 // Return whether SpeciesTorsion between SpeciesAtoms exists
 bool Species::hasTorsion(SpeciesAtom *i, SpeciesAtom *j, SpeciesAtom *k, SpeciesAtom *l) const
 {
-    auto it = std::find_if(torsions_.cbegin(), torsions_.cend(), [&](const auto &t) { return t.matches(i, j, k, l); });
-    return it != torsions_.end();
+    return std::find_if(torsions_.cbegin(), torsions_.cend(), [&](const auto &t) { return t.matches(i, j, k, l); }) !=
+           torsions_.cend();
```
rprospero, line 302:
```suggestion
    return std::any_of(torsions_.cbegin(), torsions_.cend(), [&](const auto &t) { return t.matches(i, j, k, l); });
```

src/classes/species_intra.cpp
```diff
@@ -370,12 +368,8 @@ const std::vector<SpeciesImproper> &Species::impropers() const { return improper
...
 bool Species::hasImproper(SpeciesAtom *i, SpeciesAtom *j, SpeciesAtom *k, SpeciesAtom *l) const
 {
-    auto it = std::find_if(impropers_.begin(), impropers_.end(),
-                           [i, j, k, l](auto &improper) { return improper.matches(i, j, k, l); });
-    if (it == impropers_.end())
-        return false;
-
-    return true;
+    return std::find_if(impropers_.cbegin(), impropers_.cend(),
+                        [i, j, k, l](auto &improper) { return improper.matches(i, j, k, l); }) != impropers_.cend();
```
rprospero, line 372:
```suggestion
    return std::any_of(impropers_.cbegin(), impropers_.cend(),
                        [i, j, k, l](auto &improper) { return improper.matches(i, j, k, l); });
```

src/classes/speciesangle.cpp
```diff
@@ -139,13 +140,7 @@ int SpeciesAngle::index(int n) const
...
 bool SpeciesAngle::matches(const SpeciesAtom *i, const SpeciesAtom *j, const SpeciesAtom *k) const
 {
-    if (j_ != j)
-        return false;
-    if ((i_ == i) && (k_ == k))
-        return true;
-    if ((i_ == k) && (k_ == i))
-        return true;
-    return false;
+    return (j_ == j) && ((i_ == i && k_ == k) || (i_ == k && k_ == i));
```
rprospero, line 143: I like this line.

rprospero (approved):
There was one part that I was a little confused by, plus some minor suggestions, but nothing that would stop me from wanting this merged in.

src/classes/isotopologue.cpp
trisyoungs: No, you're absolutely right - a const reference it shall be.

src/classes/neutronweights.cpp
trisyoungs: :+1:

src/classes/speciesangle.cpp
trisyoungs: Might be the most concise line of code I've ever written.

Issue #873, opened 2021-11-16
Species and Configuration viewers should respect dark mode foreground and background colours


trisyoungs:
**Describe the bug**
In dark mode, the `SpeciesViewer` and `ConfigurationViewer` widgets still use a white background with black foreground (e.g. axes), whereas they should inherit the current Qt colours.

PR #891, opened 2021-11-30
Issue 873

rprospero:
This PR adds support for Dark Move in the 3D views, closing issue #873 .  The current work in progress is visible below:

20211130_13h47m42s_grim

The text and background pull their colours from the palette.  The axes are causing a bit of trouble, though.  The code for controlling the colour through VBOs doesn't seem to work, even if I hard code the colour in.

### DIFF
--- a/src/gui/render/axes.cpp
+++ b/src/gui/render/axes.cpp
@@ -77,7 +77,7 @@ Axes::Axes(View &parent, FontInstance &fontInstance) : parentView_(parent), font
     // GL
     for (auto n = 0; n < 3; ++n)
     {
-        axisPrimitives_[n].initialise(GL_LINES, false);
+        axisPrimitives_[n].initialise(GL_LINES, true);
         axisPrimitives_[n].setNoInstances();
         gridLineMajorPrimitives_[n].initialise(GL_LINES, false);
         gridLineMajorPrimitives_[n].setNoInstances();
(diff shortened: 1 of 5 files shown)


trisyoungs:
The `Primitive` class doesn't always store/use colour data in the VBO - the `Axes` class is an example of this (see [here](https://github.com/disorderedmaterials/dissolve/blob/6b62cd332e19bbc181d2b0cd8115cbc5b52d3581/src/gui/render/axes.cpp#L80), so when rendering it it will take on the current setting of `glColor*()`.

rprospero:
I found where I was making the mistake with the axes.  I believe that we now have a patch that full respects dark mode.

Screenshot_20211201_125620

trisyoungs:
Unrelated comment - is this last picture only using one quarter of the display widget? I've seen that before on some OSX versions.  Or is this just a zoom in the view that makes it look that way?

src/gui/render/axes.cpp
```diff
@@ -77,7 +77,7 @@ Axes::Axes(View &parent, FontInstance &fontInstance) : parentView_(parent), font
     // GL
     for (auto n = 0; n < 3; ++n)
     {
-        axisPrimitives_[n].initialise(GL_LINES, false);
+        axisPrimitives_[n].initialise(GL_LINES, true);
```
trisyoungs, line 80: Part of me wants to argue against using a VBO with a colour component since the axes are always a single colour. However, storing the foreground colour on the font instance and not hardcoding it in the rendering function is a much nicer solution!

trisyoungs (approved):
:+1:

PR #902, opened 2021-12-04
Optional Double Keyword

### LINKED ISSUE #901: Magic numbers in keywords
There are several keyword options in modules that either use an unexplained magic number (e.g. `-1` in `CutoffDistance` to indicate to use the pair potential range) or which have "hidden" magic numbers (e.g. `0.0` for `ReferenceFTQMax` in `NeutronSQ` to indicate no upper cutoff limit in the Fourier transform).

This should be resolved to make it clear to the user what the magical values mean.

trisyoungs:
Here we introduce a companion keyword to `DoubleKeyword` - `OptionalDoubleKeyword` - which serves the purpose of storing an optional `double` value in cases where a default value is available. The `ExponentialSpin` widget is modified to display "special value text" when the minimum value is reached, which is taken to indicate the "default" relevant value should be used.

This PR also removes the monolithic keyword includes, favouring inclusion of specific individual files where necessary.

Closes #901.


rprospero:
I've told the unit test to do all of the rendering offscreen, so that fixes the xcb issue.  I've also snuck in my update to that new Nixos release.

rprospero (approved):

PR #907, opened 2021-12-10
Keyword Saving

### LINKED ISSUE #897: Test input file re-write
We don't currently have any tests that would catch errors arising from writing input files, in particular writing files badly so that they cann't be read back in.

There are a few ways in which this could be achieved. One would be to sprinkle specialised test cases amongst the current batch of system tests, which load and rewrite the file before running the related test again.

trisyoungs:
This PR addresses issues caused by recent keyword upgrade work, and hardens our testing. Specifically, keyword data was not written consistently to the input file, with data either in the wrong format or completely missing (due to it not being recognised as "set" and containing valid data).

Summary or major changes:
- An `OptionalInteger` keyword and widget have been added, and uses of this and it's double equivalent are now more prevalent in the code.
- The `KeywordBase` member `set_` and the virtual `isDataEmpty()` have been removed in favour of local members on PODs, and simply checking of the data everywhere else.
- A new CLI option to reload a just-written input file has been added, and which is used throughout the system tests in order to check our consistency when writing data.

Closes #897.


rprospero (approved):
This is clearly a much cleaner way of handling the keywords.

Issue #871, opened 2021-11-16
Procedure editing in GUI


trisyoungs:
**Describe the feature / issue**
We *really* need the ability to edit `Procedure`s (e.g. configuration generators) in the GUI.  This would be a good opportunity to move the display off of the custom "coloured boxes" and on to a `QListWidget` base (a model would be best, and complicated!)

rprospero:
I've put together a very rough proof of concept.  I've put in a terrible delegate here for the various data types, but it will require a much more complicated delegate to get full functionality.

![20211209_17h43m22s_grim](https://user-images.githubusercontent.com/1728853/145448262-3df4900d-48a9-424a-aef5-afc6f9a160f3.png)

rprospero:
![20220110_11h53m16s_grim](https://user-images.githubusercontent.com/1728853/148762273-45129642-4089-449e-85ba-8b1f9bce74f9.png)
@trisyoungs  I've attached a screenshot of my most recent prototype.  Selecting a different keyword off the list will update the Keyword Widget beneath the list.

Some changes that I am considering:

* I've moved everything into a list for simplicity, but I could move it back into a tree.  This would allow the user to collapse the nodes they aren't interested in and only see the keywords for the relevant node.

* I might extend the widget to exist not just below the list but below the entire panel.  This would provide more space for the larger widgets

trisyoungs:
@rprospero That's a neat solution, however I have a couple of reservations about it. Displaying only a single keyword and its data at a time doesn't help get an overview of the setup of an individual node.  We could consider leveraging the keyword's `serialise()` and `deserialise()` functions to give richer info in the list, but I am inclined to think that using `KeywordsWidget` (`src/gui/keywordwidgets/widget_funcs.cpp`) is the way forward, and creates a familiar UX since this is how modules display their options.  We would be wise to move to a side-by-side layout, perhaps on a separate tab from the viewer, or at the very least in a collapsible widget so that we can go "full window" for the viewer.

Secondly, the use of a tree model is absolutely essential - while generator procedures for configurations are (at present) linear, the same classes are used to provide analysis routines which have nested loop-like structures.  So, a `QTreeView` is the natural solution here.

rprospero:
![Screenshot from 2022-01-19 12-10-23](https://user-images.githubusercontent.com/1728853/150127283-012fd738-c222-413a-b42e-849a0f1b1277.png)

[...]

PR #915, opened 2021-12-15
Keyword Signals

### LINKED ISSUE #913: Update reference data when changing FT limits
**Describe the bug**
Modules accepting reference data and which Fourier transform that data (principally `NeutronSQ` and `XRaySQ`) need to run the `setUp()` stage when the FT limits and delta are changed, as well as when the filename is edited.

trisyoungs:
This PR attempts to introduce a suitable mechanism for "reacting" to modifications to keywords when those modifications affect underlying data. In particular, data that may have already been created and stored in the main processing list.

A simple and extensible "signal" enum and class is added, and whose enumerations can be set on individual keywords in order to promote the desired effect when caught by the parent `Module` or `Procedure` (although full implementation of the latter will not be performed until editing is moved off of the custom chart classes - #871).

Rather than implement all probable / possible uses of signals in this PR, exemplar signalling is made on the `NeutronSQ` and `XRaySQ` modules (closes #913). Future required uses will be added as and when required locally in other work.

### DIFF
--- a/src/keywords/signals.h
+++ b/src/keywords/signals.h
@@ -3,33 +3,36 @@
 
 #pragma once
 
+#include <bitset>
+
 // Keyword Signals
 class KeywordSignals
 {
     public:
     KeywordSignals();
+    ~KeywordSignals() = default;
     explicit KeywordSignals(int signalMask);
     KeywordSignals(const KeywordSignals &source);
     KeywordSignals &operator=(int signalMask);
     void operator+=(int signalMask);
     operator int() const;
-    ~KeywordSignals() = default;
 
     public:
     enum KeywordSignal
     {
-        ClearData = 1,
-        RecreateRenderables = 2,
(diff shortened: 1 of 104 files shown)


src/keywords/signals.h
```diff
@@ -0,0 +1,35 @@
...
+    enum KeywordSignal
+    {
+        ClearData = 1,
+        RecreateRenderables = 2,
+        ReloadExternalData = 4
+    };
+
+    private:
+    // Signals to be emitted (via Qt) when editing this keyword in the GUI
+    int signalMask_{0};
```
DanNixon, line 28: I tend to prefer `std::bitset` for things like this.

DanNixon (approved):

src/keywords/signals.h
trisyoungs: Old habit! Gotta love a lazy `int`...

PR #918, opened 2021-12-15
Filter Partials

### LINKED ISSUE #604: Add search / pagination for PartialSet viewers
For small systems it is easy enough to display all the possible partials on `RDFModuleWidget`, `SQModuleWidget` etc., but in cases where the number of atom types approaches and exceeds 5/6 plotting all at once on the display is unfeasible.

Add pagination of partials data, as well as a typename-based search, to all relevant widgets.

trisyoungs:
Short and sweet PR to add basic (but effective) filtering of partial structure factors and radial distribution functions, greatly improving UX when there are lots of partials in the simulation.

Closes #604.


src/modules/rdf/gui/rdfwidget_funcs.cpp
```diff
@@ -59,13 +59,22 @@ void RDFModuleWidget::createPartialSetRenderables(std::string_view targetPrefix)
...
     for (auto [first, second] : pairs)
     {
         auto &at1 = ps.atomTypeMix()[first];
         auto &at2 = ps.atomTypeMix()[second];
         const std::string id = fmt::format("{}-{}", at1.atomTypeName(), at2.atomTypeName());
 
+        // Filtering - does this 'id' match our filter?
+        if (filterText && id.find(filterText.value()) == std::string::npos)
+            continue;
+
```
rprospero, line 77: These lines aren't present in the other modules.  Without them, is any filtering actually performed?

rprospero (approved):
There's one question where I didn't understand the logic, but it otherwise looks good.

src/modules/rdf/gui/rdfwidget_funcs.cpp
trisyoungs: Well, this neatly answers the question "Why code review?". An error made in copy/paste haste. Good spot.

PR #924, opened 2021-12-21
Parameter Naming

### LINKED ISSUE #734: Accept / Display Parameter Names in Tables
Following the use of a comma-separated list for specifying numerical parameters for forcefield functions, we now require the parameter names to be displayed alongside the numbers themselves in order to transparently inform the user which parameter is which.

- [x] Store parameters names on functions (how?)
- [x] Change get / set form of parameters to `name1=value1, name2=value2, ...`
- [x] Allow any ordering of parameters to be supplied when editing, but always display parameters in the defined order.
- [ ] When adding an atom type, for example, and setting the functional form the parameters cell is initially empty, but should be populated with at least `name1=0, name2=0, ...`

trisyoungs:
This PR implements naming / editing of parameters within the `SpeciesIntra` context.

The thin functional classes added in #922 are extended to provide access to parameter names and indices per functional form, and parsing routines have been rewritten (specifically, the templated `splitStrings` is fixed-up and moved to `DissolveSys`).

Unit tests covering the new parameter parsing as well as existing string functions in `DissolveSys` are added, and some general tidying is performed (namely adding specific input files for the tests and asserting that files are read correctly).

Extension of parameter passing to the `AtomType` class is made in a subsequent PR (#925).

Closes #734.


rprospero (approved):
I really like this new architecture.  It's much cleaner than before.  It was also the right move to put `splitString` into `DissolveSys`, instead of having it in the `algorithms` file.  I'd briefly thought of fighting for the inclusion of the more generic version of `splitString` that could handle any container, but it's obvious that we haven't needed it and we could always bring it back if we ever do.

PR #929, opened 2022-01-04
Force more free space

rprospero:
We freed up enough space before to get half the packages to build.  Just need to find enough space for the other half.


rprospero:
It looks like free space is no longer the problem (I've freed over 20 Gb).  Instead, I've allowed more time for building the singularity images to see if it is a timeout image.

PR #939, opened 2022-01-07
Keyword Signals Update

trisyoungs:
Some bugs were introduced into #915 after converting to use `std::bitset`. This PR resolves those issues, and belatedly adds a unit test.


DanNixon (commented):
:+1:

DanNixon (approved):

Issue #944, opened 2022-01-12
Update fmt library


rprospero:
We're currently pulling fmt version 7 from conan that they're up to version 8 now.  We should ensure that there's no issues, then upgrade to the latest version.

rprospero:
This was fixed in pr #945

PR #945, opened 2022-01-12
Upgrade to fmt8

rprospero:
The library upgrade also removed some implicit imports that we had not meant to rely upon.  Those have been corrected.

This closes issue #944.

### DIFF
--- a/src/base/enumoptions.h
+++ b/src/base/enumoptions.h
@@ -140,7 +140,7 @@ template <class E> class EnumOptions : public EnumOptionsBase
                                [enumeration](auto &option) { return enumeration == option.enumeration(); });
         if (it == options_.cend())
             throw(std::runtime_error(fmt::format(
-                "Enumerated options '{}' missing enumeration {}, so can't return its keyword.\n", name_, enumeration)));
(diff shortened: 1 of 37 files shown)


src/base/enumoptions.h
```diff
@@ -162,7 +162,7 @@ template <class E> class EnumOptions : public EnumOptionsBase
         auto it = std::find_if(options_.cbegin(), options_.cend(),
                                [enumeration](auto &option) { return enumeration == option.enumeration(); });
         if (it == options_.cend())
-            throw(std::runtime_error(fmt::format("No option set for enumeration '{}'.\n", enumeration)));
+            throw(std::runtime_error(fmt::format("No option set for enumeration '{}'.\n", (int)enumeration)));
```
trisyoungs, line 165: Gyah. I expect there will be a few of these not caught by the test suite, and which will throw themselves at us in practical testing.

trisyoungs (approved):
Looks good.   I assume the nix upgrade is to a version above that which caused us the hangs in the CI (and which hopefully doesn't have the same issue!).

trisyoungs (approved):
I know that I must be responsible for the majority of the uses of `#include <fmt/core.h>` - I fully blame the manual in this instance!

PR #948, opened 2022-01-19
Remove RefDataList

rprospero:
This removes another part of #245


trisyoungs (approved):
Looks good. Also highlights some of the other custom class helpers in the GUI that are probably now (nearly) ready for removal following the move to models.

PR #949, opened 2022-01-19
Switch to completion in Data Manager Dialog

### LINKED ISSUE #723: Replace Regular Expression searches with proper completion
Currently, we are using QRegExp in several dialogs, which has been deprecated.  The upcoming PR to move to Qt 6 will remove QRegExp support, forcing us to switch to QRegularExpression, which does not have wildcard search support.  All of these cases would be better served with a QCompleter on the input prompt, which can preform a wildcard search and also allows the user a more intutive way to select the option that they were searching for.

rprospero:
This removes the regex from the Data Manager Dialog and, instead, causes the dialog to pop up suggestions as you type.  Selecting one of the suggestions selects the item in the table.

If we don't like this setup, we may just stick with the regex and close #723.

### DIFF
--- a/src/gui/datamanagerdialog.h
+++ b/src/gui/datamanagerdialog.h
@@ -44,8 +44,6 @@ class DataManagerDialog : public QDialog
     Ui::DataManagerDialog ui_;
 
     private:
-    // Update the specified table of GenericItems, optionally filtering them by name and description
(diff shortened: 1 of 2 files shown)


rprospero:
After group discussion, we decided that this was not a useful path forward from a UX experience.

PR #950, opened 2022-01-20
Remove RefList

### LINKED ISSUE #245: Move to standard STL Containers

rprospero:
This is the last remaining custom template class to close #245


src/module/groups.h
```diff
@@ -40,7 +41,7 @@ class ModuleGroups
     private:
     // Current list of groups
     std::vector<std::unique_ptr<ModuleGroup>> groups_;
-    // RefList of all Modules present in all groups
+    // vector of all Modules present in all groups
```
trisyoungs, line 44: Picky!
```suggestion
    // Vector of all Modules present in all groups
```

src/modules/rdf/functions.cpp
```diff
@@ -505,7 +505,7 @@ bool RDFModule::sumUnweightedGR(GenericList &processingData, ProcessPool &procPo
     summedUnweightedGR.setUpPartials(combinedAtomTypes);
 
     // Determine total weighting factors and combined density over all Configurations, and set up a Configuration/weight
-    // RefList for simplicity
+    // vector for simplicity
```
trisyoungs, line 508: Awakening my OCD:
```suggestion
    // Vector for simplicity
```

src/procedure/nodes/select.cpp
```diff
@@ -237,7 +237,7 @@ bool SelectProcedureNode::execute(ProcessPool &procPool, Configuration *cfg, std
                 continue;
 
             // Check Site exclusions
-            if (excludedSites_.contains(&site))
+            if (excludedSites_.end() != std::find(excludedSites_.begin(), excludedSites_.end(), &site))
```
trisyoungs, line 240: Again, reversing the compared quantities makes more sense to me:
```suggestion
            if (std::find(excludedSites_.begin(), excludedSites_.end(), &site) != excludedSites_.end())
```

trisyoungs (approved):
Marvellous. Some small comments / suggested changes (all of which I admit are very picky!).

PR #955, opened 2022-01-22
Automatic targets

### LINKED ISSUE #507: Automatically set relevant local module targets in added modules
Adding a new module to a layer in the GUI does not currently set any relevant target keywords (e.g. `SourceSQs` for the `NeutronSQ` and `XRaySQ` modules, for example.

Possible the cleanest way to set this kind of data automatically (i.e. to the relevant module appearing in the same layer) is to have a "post-wizard" function in the GUI that performs necessary setup after the module has been added.

trisyoungs:
Here we make the GUI a little more user-friendly and automatically set relevant targets (e.g. `Configuration`s and specific module types) as soon as a new module is created in a layer.

This is achieved with a simple virtual function in the `Module` class which handles setting of `Configuration` targets by searching the `Keyword::targetsGroup_` vector for relevant keyword types (`ConfigurationKeyword` and `ConfigurationVectorKeyword`) and which is overridden in modules when specific module targets (e.g. `RDF`, `SQ`) are required. In those instances, target `Configuration`s are typically not required.

Some other changes are made to harmonise other aspects of target keywords, specifically forcing all modules taking `Configuration` targets to accept only one, the exception being `RDF` which is able to combine data from multiple `Configuration`s.

Closes #507.

### DIFF
--- a/src/module/layer.cpp
+++ b/src/module/layer.cpp
@@ -88,6 +88,21 @@ bool ModuleLayer::contains(Module *searchModule) const
 // Return vector of Modules
 std::vector<std::unique_ptr<Module>> &ModuleLayer::modules() { return modules_; }
 
+// Return map of modules in the layer, optionally preceding the specified module
+std::map<std::string, std::vector<const Module *>> ModuleLayer::modulesAsMap(const Module *beforeThis) const
+{
+    std::map<std::string, std::vector<const Module *>> moduleMap;
+
+    for (auto &module : modules_)
+    {
+        if (module.get() == beforeThis)
+            break;
+        moduleMap[std::string(module->type())].emplace_back(module.get());
+    }
+
+    return moduleMap;
+}
+
 /*
  * General Actions
  */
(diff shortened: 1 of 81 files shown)


src/module/layer.cpp
```diff
@@ -88,6 +88,21 @@ bool ModuleLayer::contains(Module *searchModule) const
...
+std::map<std::string, std::vector<const Module *>> ModuleLayer::modulesAsMap(const Module *beforeThis) const
+{
+    std::map<std::string, std::vector<const Module *>> moduleMap;
+
+    for (auto &module : modules_)
+    {
+        if (module.get() == beforeThis)
+            break;
+        moduleMap[std::string(module->type())].emplace_back(module.get());
+    }
```
rprospero, line 101: There's a way of doing this loop with a `std::find_if` and a `std::transform`, but it's less clear and doesn't buy us anything.

rprospero (approved):
Moving to a single Configuration does simplify things.

src/module/layer.cpp
trisyoungs: Happy to stick with clarity over cleverness!

PR #960, opened 2022-01-26
Patch 01 2022 3

### LINKED ISSUE #942: Windows build breaks because of Freetype
**Describe the bug**
Windows builds fail because the current (recently-updated) version of MSVC breaks the Freetype build - the issue is documented by the Freetype project here: https://gitlab.freedesktop.org/freetype/freetype/-/issues/1124

PR #941 hacks a solution to this issue, but once the issue is resolved by the Freetype maintainers it should be reverted.

trisyoungs:
Final patch PR for January (but submitted a bit late).

Closes #942.

### DIFF
--- a/src/gui/models/sitesFilterProxy.h
+++ b/src/gui/models/sitesFilterProxy.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <QSortFilterProxyModel>
+#include <bitset>
 
 // Forward Declarations
 class QModelIndex;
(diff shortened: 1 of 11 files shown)


src/gui/models/sitesFilterProxy.h
```diff
@@ -13,19 +14,21 @@ class SitesFilterProxy : public QSortFilterProxyModel
...
-
-    public:
+    SitesFilterProxy(int flags = 0);
     // Filter flags
     enum FilterFlags
     {
-        None = 0,
-        IsOriented = 1
+        HasAxes,
+        nFilterFlags
```
IVlaD17, line 22: Could we, for the sake of consistency call `nFilterFlags` `NFilterFlags` instead or find a better name for it?

src/gui/models/sitesFilterProxy.h
```diff
@@ -13,19 +14,21 @@ class SitesFilterProxy : public QSortFilterProxyModel
...
+    private:
     // Current filter flags
-    int filterFlags_;
-    // Set filter flags
-    void setFlags(int flags);
+    std::bitset<nFilterFlags> filterFlags_;
+
+    public:
+    // Add filter flag
+    void addFlag(FilterFlags flag);
```
IVlaD17, line 31: Is this meant to be a `set` method? If yes, could we perhaps rename it to `setFlag` instead? If not, all good.

src/gui/models/sitesFilterProxy.h
trisyoungs: I see what you're saying, but in some respects having the "size" member start with a lowercase `n` nicely differentiates it from the (usable) other enum values.

IVlaD17: Up to you, pick whichever option you like the most, I personally usually go for consistency (aesthetics lies in the GUI) but whatever you prefer is fine by me. Feel free to resolve this whenever.

PR #963, opened 2022-01-31
Add a proper model for the procedure nodes

### LINKED ISSUE #871: Procedure editing in GUI

rprospero:
This adds a proper, recursive tree model for the procedure nodes.  It's currently only used for the generators, but it can also later be used for analysis nodes.  This includes adding a `parent` and `child` function to the procedure nodes to more easily enable navigating the structure.  This should close #871.


rprospero:
This PR also includes a small change to the build system of the unit test setup.  Tests are now only linked against the gui parts of dissolve if they are, in fact, testing gui components.  This prevents unit tests from being repeatedly recompiled and linked when nothing has changed within them or their dependencies.

src/procedure/nodes/process1d.cpp
```diff
@@ -94,6 +96,13 @@ bool Process1DProcedureNode::hasBranch() const { return (normalisationBranch_ !=
 // Return SequenceNode for the branch (if it exists)
 std::shared_ptr<SequenceProcedureNode> Process1DProcedureNode::branch() { return normalisationBranch_; }
 
+std::vector<ConstNodeRef> Process1DProcedureNode::children() const
+{
+    if (normalisationBranch_)
+        return {std::static_pointer_cast<const ProcedureNode>(normalisationBranch_)};
+    return {};
```
trisyoungs, line 103: Perhaps I'm missing something here, but why does this particular branch need an explicit cast where in other nodes (e.g. `Collect1D`) it doesn't?

src/procedure/nodes/process2d.cpp
```diff
@@ -90,6 +92,13 @@ bool Process2DProcedureNode::hasBranch() const { return (normalisationBranch_ !=
 // Return SequenceNode for the branch (if it exists)
 std::shared_ptr<SequenceProcedureNode> Process2DProcedureNode::branch() { return normalisationBranch_; }
 
+std::vector<ConstNodeRef> Process2DProcedureNode::children() const
+{
+    if (normalisationBranch_)
+        return {std::static_pointer_cast<const ProcedureNode>(normalisationBranch_)};
+    return {};
```
trisyoungs, line 99: And here.

src/procedure/nodes/select.cpp
```diff
@@ -78,6 +79,9 @@ std::vector<const SpeciesSite *> &SelectProcedureNode::speciesSites() { return s
 // Return whether sites must have a defined orientation
 bool SelectProcedureNode::axesRequired() { return axesRequired_; }
 
+// Return nodes owned by this node
+std::vector<ConstNodeRef> SelectProcedureNode::children() const { return {forEachBranch_}; }
```
trisyoungs, line 83: No cast here, but in other places there is an `if`/`else` clause.  Can all of the `children()` functions be similar to this one?

trisyoungs (changes requested):
An excellent solution to the problem.

I have one comment to make on the variance of the `children()` functions, and a small change to request. The `ProcedureGroup` `QGroupBox` containing the keywords widget always has the title `ProcedureNode` - this should really be the node type followed by the explicit name of the node in parentheses if it has one.

Issue #964, opened 2022-02-01
Review Remaining Templates


rprospero:
We have cleaned out most of the custom containers.  However, we need to make a review of our templates to see if there are any extraneous templates.  These are either other classes that could be replaced with something from the standard library or classes that are only use in one location and could be folded into their usage site.

IVlaD17:
Temporary Comment

PR #968, opened 2022-02-04
Cleaned out some of the Remaining Templates

IVlaD17:
Tackling #964, this PR cleans out some of the templates present there.

The following templates have been removed:
- `EnumContainer` was not used at all and just got deleted
- `Venum` was not used at all and just got deleted
- `DynamicArray` was not used at all and just got deleted
- `DynamicArrayObjectBase` was used as the base of `DynamicArrayObject` therefore its functionality was replicated as part of replacing `DynamicArrayObject`
- `DynamicArrayObject` was used within `Atom` and `Molecule` and its only useful functionality was that it kept an `int arrayIndex` and provided a getter and a setter for it. All of this was replicated in `Atom` and `Molecule`
- `EnumCast` which was being used in `LineParser` was replaced by two methods: `getIntFromParseReturnValue(LineParser::ParseReturnValue value)` and `getParseReturnValueFromInt(int value)`. Additionally, as part of this change, the logic where `EnumCast` was used was changed to stop using `if` statements. It now relies on the two methods and it ultimately returns a `LineParser::ParseReturnValue`

Additional changes that were done as part of this PR:
- Removed all template includes where the template was not actually used.
- Changed variable names in a couple of places from `Atom i` to `Atom atom` for the sake of improving code readability.
- Changed the name of the `arrayIndex` getter from `arrayIndex()` to `getArrayIndex()` to better reflect its functionality and to keep in line with standards.


rprospero:
I'm not entirely sold on the change from `arrayIndex` to `getArrayIndex`.  You'll notice that the `Atom` class has member functions `x`, `y`, and `z`, instead of `getX`, `getY`, and `getZ`.  We've largely reserved the `get` prefix for members that perform some calculation, instead of just returning a value.

Granted, I'm a bit biased here, as I've never been a fan of getters and setters in general.  If you'd like to make a case for the `get` prefix, I'd be happy to hear it.

IVlaD17:
> I'm not entirely sold on the change from `arrayIndex` to `getArrayIndex`. You'll notice that the `Atom` class has member functions `x`, `y`, and `z`, instead of `getX`, `getY`, and `getZ`. We've largely reserved the `get` prefix for members that perform some calculation, instead of just returning a value.
> 
> Granted, I'm a bit biased here, as I've never been a fan of getters and setters in general. If you'd like to make a case for the `get` prefix, I'd be happy to hear it.

It's not a major thing so I can easily revert it to `arrayIndex` instead. I've noticed your approach where you have member functions called the same name as the variable itself which is why I wasn't too sure about this change. Here's a couple of things to think about though:

Even though you say you aren't a fan of getters and setters (which I understand), the Dissolve code still uses both of them, with setters having the `set` prefix but getters not having the `get` prefix (which is why I thought it might be a good idea to add it). 

In my experience so far, I have noticed two main approaches, depending on the language the project was written in.

With Java, you'd have both getters and setters using `type getVariableName()` and `void setVariableName(type varName)`. This seems to be the closest we'd get to an implementation in C/C++.

With C#, you've got properties: 
`private int x`
`public int X          ` 
`{                           ` 
`   get { return x; } `
`   set { x = value; }` 
`}                           ` 

I think we could achieve properties-like declarations within C/C++ as well. We could have two methods, `int x()` and `void x(int value)` although, to me, it seems like we'd have to go out of our way to implement this whereas the `get` prefix is a bit more natural and probably easier to transition to.

rprospero:
As my own personal preference, I like the properties style arrangement that you mentioned.  Honestly, I really prefer genuine profunctor optics, but we're not getting those until c++65.

However, changing our naming convention is pretty far outside the scope of this PR.  It might be worth discussing at the TSC, but it should still be handled in a separate issue.  In the mean time, I feel that we should keep the old `arrayIndex`, since it does match the naming convention of the rest of the code.

src/base/lineparser.cpp
```diff
@@ -585,8 +584,9 @@ LineParser::ParseReturnValue LineParser::readNextLine(int optionMask)
     // Broadcast result of file check
     if (processPool_)
     {
-        if (!processPool_->broadcast(EnumCast<LineParser::ParseReturnValue>(result)))
-            return LineParser::Fail;
+        int enumValue = getIntFromParseReturnValue(result);
+        processPool_->broadcast(enumValue);
```
trisyoungs, line 588: Should check the broadcast result as well:
```suggestion
        if (!processPool_->broadcast(enumValue))
            return LineParser::Fail;
```

src/base/lineparser.cpp
```diff
@@ -668,8 +668,9 @@ LineParser::ParseReturnValue LineParser::readNextLine(int optionMask)
     // Broadcast result
     if (processPool_)
     {
-        if (!processPool_->broadcast(EnumCast<LineParser::ParseReturnValue>(result)))
-            return LineParser::Fail;
+        int enumValue = getIntFromParseReturnValue(result);
+        processPool_->broadcast(enumValue);
```
trisyoungs, line 672: And here.

src/base/lineparser.cpp
```diff
@@ -948,3 +949,24 @@ bool LineParser::hasArg(int i) const
...
+    LineParser::ParseReturnValue valueEnum;
+
+    if (value == -1)
+        valueEnum = LineParser::ParseReturnValue::EndOfFile;
+    else if (value == 0)
+        valueEnum = LineParser::ParseReturnValue::Success;
+    else
+        valueEnum = LineParser::ParseReturnValue::Fail;
+
+    return valueEnum;
```
trisyoungs, line 971: Just condensing things a bit and removing the temporary:
```suggestion
    if (value == -1)
        return LineParser::ParseReturnValue::EndOfFile;
    else if (value == 0)
        return LineParser::ParseReturnValue::Success;

    return LineParser::ParseReturnValue::Fail;
```

IVlaD17:
@trisyoungs All your comments should be addressed with my last commit.

trisyoungs:
Looking all OK now, but the lack of CI/CD means we are blind to any issues that might arise on merge...

Issue #884, opened 2021-11-25
Add forcefield wizard needs updating


trisyoungs:
**Describe the bug**
The Add Forcefield wizard relies on the messages output to show errors for missing parameters etc. - since the messages are now in their own separate tab, there is no visible feedback to the user when missing parameters are encountered.

Errors / missing parameters lists should be redirected to the wizard itself.

rprospero:
![20220210_15h10m11s_grim](https://user-images.githubusercontent.com/1728853/153436577-5ab3de57-338e-4c4a-a5c7-93148d7677db.png)

@trisyoungs  Would this error message be sufficient?

trisyoungs:
@rprospero Change "forcefields" for "atom types" and I think that will do it!

rprospero:
Fixed in PR #971

PR #971, opened 2022-02-10
Display error message in Add Forcefield Dialog

rprospero:
This closes issue #884.  There's a screenshot on the issue that shows the current status of the messages in the GUI.

### DIFF
--- a/src/data/ff/ff.cpp
+++ b/src/data/ff/ff.cpp
@@ -289,7 +289,7 @@ std::vector<Elements::Element> Forcefield::assignAtomTypes(Species *sp, CoreData
     Messenger::print("Assigning atomtypes to species '{}' from forcefield '{}'...\n", sp->name(), name());
 
     // Loop over Species atoms
-    std::vector<Elements::Element> failed;
+    std::vector<Elements::Element> failedElements;
     for (auto &i : sp->atoms())
     {
         // Obey the supplied strategy:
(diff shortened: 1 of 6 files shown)


src/data/ff/ff.cpp
```diff
@@ -304,15 +304,15 @@ int Forcefield::assignAtomTypes(Species *sp, CoreData &coreData, AtomTypeAssignm
...
+            failed.push_back(i.Z());
         }
     }
 
-    if (nFailed)
-        Messenger::error("Failed to assign atom {} to {} {}.\n", (nFailed == 1 ? "type" : "types"), nFailed,
-                         (nFailed == 1 ? "atom" : "atoms"));
+    if (!failed.empty())
+        Messenger::error("Failed to assign atom {} to {} {}.\n", (failed.size() == 1 ? "type" : "types"), failed.size(),
+                         (failed.size() == 1 ? "atom" : "atoms"));
```
trisyoungs, line 313: Gets a bit awkward here if one thinks about all of the possibilities.  For instance, some carbon atoms could be successfully assigned types, while others might not. We could hedge our bets and change the message to `Failed to assign atom types to (some)...`?

src/data/ff/ff.h
```diff
@@ -185,7 +185,9 @@ class Forcefield
         SelectionOnlyFlag = 4      /* Only assign terms where all atoms are in the current selection */
     };
     // Assign suitable AtomTypes to the supplied Species, returning the number of failures
-    int assignAtomTypes(Species *sp, CoreData &coreData, AtomTypeAssignmentStrategy strategy, bool setSpeciesAtomCharges) const;
+    // Returns any elements that were unassigned
+    std::vector<Elements::Element> assignAtomTypes(Species *sp, CoreData &coreData, AtomTypeAssignmentStrategy strategy,
+                                                   bool setSpeciesAtomCharges) const;
```
trisyoungs, line 190: A possible future extension here could be to return the list of atom indices that failed to type correctly, and adding a new page into the wizard to display a viewer with those atoms highlighted in the model.  But definitely not for this PR!

src/gui/addforcefieldtermsdialog.h
```diff
@@ -103,7 +103,8 @@ class AddForcefieldTermsDialog : public WizardDialog
     bool intramolecularTermsAssigned_;
     // Assign intramolecular terms to species
     bool assignIntramolecularTerms(const Forcefield *ff);
-
+    // Helper function for error messages
+    bool alertAboutAtomTypeErrors(std::vector<Elements::Element> errs);
```
trisyoungs, line 107:
```suggestion
    bool alertAboutAtomTypeErrors(std::vector<Elements::Element> errs);

```

trisyoungs (approved):
:+1: Just some small suggestions.

src/data/ff/ff.h
trisyoungs: I guess this is sensible - sorry for dragging this PR out!  Don't forget to return the "user" indices (1:N) for the atoms so as not to confuse us chemists who aren't zero-indexed...

rprospero: I've moved over to the index based reporting

src/data/ff/ff.cpp
rprospero: We now return the specific atomic indices, so this shouldn't be an issue.

PR #972, opened 2022-02-13
Enum Flags

trisyoungs:
This PR refactors the `KeywordSignals` class introduced in #915 and #939 and generalises it into a simple new template class `Flags`, the purpose of which is to nicely manage an or'd set of enumerations.

This essentially reproduces some functionality of the `QFlags` class, but of course that could not be used with CLI version of the code which has no dependence on Qt. Unlike Qt, however, the enumerated items don't have to be assigned unique int values to represent bits. Also, the use of sizing enumerations such as `nEnumValues` is not necessary since the `Flags` class assumes a default maximum size which should be suitable for most purposes.  While this isn't necessarily ideal, it does make for clean enum definitions with no extraneous values.

Community alternatives exist - e.g. the header-only http://aantron.github.io/better-enums/, but this add 1k lines of code compared to the 51 lines here.


DanNixon (approved):
:+1: for the unit test.

IVlaD17 (approved):

Issue #954, opened 2022-01-20
Don't run RDF module is SizeFactor > 1


trisyoungs:
**Describe the bug**
There is no purpose or utility in calculating RDFs for a configuration if the size factor is anything other than 1.0.

The RDF module should be modified to check for this and refuse to calculate RDFs if that is the case (returning `true` from its processing).

rprospero:
Closed in #975

PR #975, opened 2022-02-24
Check at start of RDF run for size factor issues

rprospero:
This should close issue #954


src/modules/rdf/process.cpp
```diff
@@ -48,6 +48,12 @@ bool RDFModule::process(Dissolve &dissolve, ProcessPool &procPool)
      * multiple independent Configurations, we must loop over the specified targetConfigurations_ and calculate the partials
      * for each.
      */
+
+    // Check that Configurations have unmodified size factor
+    if (std::all_of(targetConfigurations_.begin(), targetConfigurations_.end(),
+                    [](const auto *cfg) { return cfg->appliedSizeFactor() != 1.0; }))
```
trisyoungs, line 54: We're comparing two `double`s here, which could be seen as risky - should we do a check on the difference between the applied factor and `1.0` instead?

src/modules/rdf/process.cpp
```diff
@@ -48,6 +48,12 @@ bool RDFModule::process(Dissolve &dissolve, ProcessPool &procPool)
...
+        return true;
```
trisyoungs, line 55: Printing a message here would be good and informative, I think:
```suggestion
    if (std::all_of(targetConfigurations_.begin(), targetConfigurations_.end(),
                    [](const auto *cfg) { return cfg->appliedSizeFactor() != 1.0; }))
    {
        Messenger::print("One or more configurations have an applied size factor, so RDF calculation will be skipped.\n");
        return true;
    }
```

trisyoungs (approved):
Two small comments (on three lines of code!).  This will achieve what we want in the immediate sense.  We should consider a more general implementation of this kind of functionality across all modules, for which I'll submit a new issue.

src/modules/rdf/process.cpp
rprospero: I've updated this to give us two free bits of slack.  That should handle most issues, but there's a 52 bit mantissa, so we might be able to afford a third bit.

Issue #947, opened 2022-01-18
Remove MinimiserBase<T>


trisyoungs:
**Describe the feature / issue**
Minimisation algorithms (Monte Carlo, praxis etc.) are all built on a template class (`MinimiserBase`) which has a historically bad and unnecessary design.

Proposal is to remove the use of `MinimiserBase` completely and move to independent classes specific to each minimisation algorithm, and replace the use of cost function callbacks with lambdas.

rprospero:
Closed in #976

PR #976, opened 2022-02-24
Remove MinimiserBase<T>

rprospero:
The Pr eliminates the MinimiserBase class and just leaves in the two subclasses as independent classes.

Closes issue #947


trisyoungs (commented):
All of what's been done here looks good.  Is there room in this PR to remove the need for the `T& object` and function pointer and move to lambdas instead?

rprospero:
@trisyoungs  I will put something together that will move us to lambdas.  I didn't do it originally because C++ gives each lambda its own specific type (instead of something sane based on parameters and return values), so we wind up needing a little bit of template magic to do this.  That's not that large of an issue on its own, but someone making the call incorrectly will get a screen-full of template giberation, while the current implementation gives a more comprehensible error message.

An alternative is to use `std::function`, which avoids the template and still gives good error messages.  Unfortunately, there's a slight overhead on `std::function` calls which I initially didn't want to put into a hot loop.  However, looking over the code, I'm thinking that this overhead might be insignificant compared to the body of the function being called.

Any preferences between the two options?

trisyoungs:
@rprospero I think the `std::function` approach fits better, given what you've mentioned above. The overhead shouldn't be an issue since it's only a lukewarm loop, and the meat of the processing is within the function itself.

PR #986, opened 2022-03-08
Recalculate Bragg reflections after changes to Bragg module

### LINKED ISSUE #914: Changing Bragg limits must reinitialise data
**Describe the bug**
Changing the Q limits / multiplicity of the Bragg calculation needs to re-init the reflections vector, otherwise the same vector of reflections is used for the new Q limit and/or multiplicity.

rprospero:
This PR will trigger a recalculation of all of the Bragg reflection vectors after ANY changes occurs to the Bragg module's keywords.  The original issue only requested the recalculation after certain terms.  Checking only the desired terms would be possible, but will require a significantly larger PR.

Closes #914


rprospero:
I believe that I have a better solution here.  The `Q` terms now all throw a `RecalculateBraggReflections` flag when they change.  The setup for the Bragg module will detect this and clear the KVectors, triggering a the calculation of the new, correct values.

PR #987, opened 2022-03-10
MC Tidy

trisyoungs:
This PR started off as a bug hunt attempting to fix the `MonteCarloMinimiser` class which no longer appeared to modify any values, and ended up as a complete simplification of said class, removing redundant functionality and improving the basic method.

### DIFF
--- a/src/math/poissonfit.cpp
+++ b/src/math/poissonfit.cpp
@@ -299,11 +299,8 @@ double PoissonFit::sweepFitC(FunctionSpace::SpaceType space, double xMin, int sa
             // Generate the approximate data - we will subtract the functions that we are fitting in the next loop
             generateApproximation(space);
 
-            // Clear the reference array of function indices
-            alphaIndex_.clear();
-
             // Set up minimiser for the next batch
-            MonteCarloMinimiser poissonMinimiser([this](const std::vector<double> &alpha) {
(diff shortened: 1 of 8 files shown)


IVlaD17 (approved):
I can't comment too much on the scientific side of this, but, in terms of the code itself, it looks clean enough for me :)

src/math/poissonfit.cpp
```diff
@@ -436,28 +445,44 @@ double PoissonFit::constructReciprocal(double rMin, double rMax, const std::vect
...
+
+            // Add in contributions from our Gaussians
+            for (auto n = 0; n < C_.size(); ++n)
+                y += functions_[{n, i}] * C_[n];
+
+            dy = referenceData_.value(i) - y;
+            sose += dy * dy;
+        }
+
+        return sose;
```
rprospero, line 472: This lambda appears twice.  Maybe we should make it a member function and just pass is into the minimiser?

rprospero (approved):
I'd tried to work on doing some of this clean-up, but this is much nicer than what I had tried.  Besides one place where I think we can factor a repeated lambda out into a single function, I think that this is good to go in.

src/math/poissonfit.cpp
trisyoungs: :+1:

PR #992, opened 2022-03-18
Internal MD

trisyoungs:
This PR adds an option to calculate only intramolecular (internal) forces occurring between atoms in the same molecule, giving another option for efficient structure evolution as well as optimising the necessary code for application to isolated `Species` (useful in future features).

Along the way the force routines have been adjusted to provide better clarity on what they do, and documentation for the module has been added.

### DIFF
--- a/src/modules/forces/functions.cpp
+++ b/src/modules/forces/functions.cpp
@@ -19,9 +19,68 @@ dissolve::CombinableContainer<std::vector<Vec3<double>>> createCombinableForces(
 }
 } // namespace
 
-// Calculate interatomic forces within the supplied Configuration
-void ForcesModule::interAtomicForces(ProcessPool &procPool, Configuration *cfg, const PotentialMap &potentialMap,
-                                     std::vector<Vec3<double>> &f)
+// Calculate internal molecule forces, optionally only over the supplied molecules
+void ForcesModule::internalMoleculeForces(ProcessPool &procPool, Configuration *cfg, const PotentialMap &potentialMap,
+                                          bool includePairPotentialTerms, std::vector<Vec3<double>> &f,
+                                          OptionalReferenceWrapper<std::vector<const Molecule *>> targetMolecules)
+{
+    // Create a ForceKernel
+    ForceKernel kernel(procPool, cfg->box(), cfg->cells(), potentialMap);
+
+    // Molecule Force Operator
+    auto combinableForces = createCombinableForces(f);
+    auto moleculeForceOperator = [&combinableForces, &kernel, includePairPotentialTerms](const auto &mol) {
+        auto &fLocal = combinableForces.local();
+        // Loop over bonds
+        for (const auto &bond : mol->species()->bonds())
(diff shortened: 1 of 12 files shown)


src/modules/forces/functions.cpp
```diff
@@ -234,9 +182,57 @@ void ForcesModule::totalForces(ProcessPool &procPool, Configuration *cfg, const
...
+            if (magjisq > cutoffSq)
+                continue;
+            r = sqrt(magjisq);
+            vecij /= r;
+
+            vecij *= potentialMap.force(&i, &j, r) * scale;
+            f[indexI] += vecij;
+            f[indexJ] -= vecij;
+        }
+    }
```
rprospero, line 217:
```suggestion
    dissolve::for_each_pair(ParallelPolicies::seq, sp->atoms.begin(), sp->atoms.end(), [&](int indexI, auto &i, int indexJ, auto&j) {
        if (i == j) return;
        
        // Get intramolecular scaling of atom pair
        scale = i.scaling(&j);
        if (scale < 1.0e-3) return;

        // Determine final forces
        vecij = j.r() - i.r();
        magjisq = vecij.magnitudeSq();
        if (magjisq > cutoffSq) return;
        r = sqrt(magjisq);
        vecij /= r;

        vecij *= potentialMap.force(&i, &j, r) * scale;
        f[indexI] += vecij;
        f[indexJ] -= vecij;
    });
```

I don't think that this is safe to run in parallel, since we might get a race condition on different threads updating the forces.

src/modules/md/process.cpp
```diff
@@ -141,25 +148,26 @@ bool MDModule::process(Dissolve &dissolve, ProcessPool &procPool)
...
-    // If ke is in units of [g mol-1 Angstroms2 ps-2] then must use kb in units of 10 J mol-1 K-1 (= 0.8314462)
-    const auto kb = 0.8314462;
-    ke = 0.0;
-    for (auto &&[m, v] : zip(mass, velocities))
-        ke += 0.5 * m * v.dp(v);
-    tInstant = ke * 2.0 / (3.0 * atoms.size() * kb);
+        // Calculate instantaneous temperature
+        ke = 0.0;
+        for (auto &&[m, v] : zip(mass, velocities))
+            ke += 0.5 * m * v.dp(v);
```
rprospero, line 164:
```suggestion
        std::transform_reduce(ParallelPolicies::par_unseq, mass.begin(), mass.end(), velocities.begin(), 0.0, std::plus<double>(), [](double m, double v){return 0.5 * m * v.dp(v);});
```

Even if the arrays are small, this still might manage to benefit from an `unseq` policy.
Depending on the

rprospero (approved):
A couple of algorithm suggestions, but that's about all.  This code does feel cleaner and easier to follow than what we had before.

src/modules/forces/functions.cpp
trisyoungs: You're probably right - I'll implement a combinable here.

PR #995, opened 2022-03-22
Constify ProcessPool

trisyoungs:
This *incredibly tedious* PR `const`-ifies usage of the `ProcessPool` class, making it more easily applicable throughout the code. To achieve this the following transformations were made:
- Shared random number buffers have been removed from `ProcessPool` and placed in a separate `RandomBuffer` class, which leverages a `const ProcessPool` for communication.
- Internal `Timer` objects within `ProcessPool` have been removed in favour of passing in an (optional) `Timer&` as arguments to the various communication functions.
- One of the first actions of `Module`s was to reorganise the `ProcessPool` in accordance with the local `ProcessPool` specified on the target `Configuration`.  This is a legacy behaviour from back when it was intended to have process groups operating independently on different `Configuration` targets at the same time, but this is no longer in scope so the behaviour has been removed.

It was a lot of fun.


rprospero (approved):
Just a couple of algorithm suggestions.  I do like the new const system.  I cannot imagine how tedious this was to implement.

trisyoungs:
Good catch on the algorithms - had intended to go back and do that but totally forgot.

Overall tedium rating: 9.5.

PR #997, opened 2022-03-23
Push to Harbor on release

### LINKED ISSUE #996: Push new builds to IDaaS
Pushing the SIF files to the IDaaS harbor will allow users on IDaaS to access the latest versions of Dissolve.

rprospero:
Obviously, I can't test this, so a look through would be appreciated to see if I missed anything.  The general idea is that both the continuous and release pipelines will update container tagged `latest`.  However, the release pipeline will also add a container tagged with the version number.

This will close #996.

### DIFF
--- a/.azure-pipelines/templates/create-release.yml
+++ b/.azure-pipelines/templates/create-release.yml
@@ -102,3 +102,20 @@ jobs:
         env:
           GITHUB_TOKEN: $(REPO_SECRET)
         displayName: 'Update Web Release Info'
+      - bash: |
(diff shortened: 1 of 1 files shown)


trisyoungs (changes requested):
Looks fine to me, but I'd suggest to have release builds tagged with `latest` and continuous builds tagged with `continuous`, and forget the version numbers.  This seems to be the best compromise, and means that we can have effectively static scripts on the IDAaaS side that don't need to know which version numbers we require to be installed.  When we're frequently pushing patch releases (this happened every 2-3 weeks for a good period of 0.8's existence) removing the need to raise a support ticket to IDAaaS every time feels like a sensible idea.

trisyoungs (approved):

PR #1000, opened 2022-03-25
Cleaned up the input files

IVlaD17:
Input files are no longer creating `Natural` isotopologues. Hopefully.


trisyoungs (approved):
Looks good.  And congratulations on being the `1000`th customer!

PR #1002, opened 2022-03-30
Remove random seed

trisyoungs:
This PR focuses on the random seed used in Dissolve, and makes the following changes:

- Corrects the order of initialisation of the random seed relative to other functionality. Previously, generators for `Configuration`s were run before the random seed was set (if specified in the input file), meaning that MPI processes were working on different sets of coordinates.
- The `Dissolve::seed_` member has been removed in favour of a CLI option.
- The input file `Simulation` block has been removed. It was largely redundant, with the final use for it being to set `Dissolve::seed_`.

This PR fixes a failing test (`molshake`) cropping up from previous PRs (#995).


trisyoungs:
This is a good point - in fact it made me realise that I'd forgotten to explicitly set the seed for the MolShake test, which I hope is why it failed again in the follow-on PR.  We shouldn't have to worry about setting the seed for every test, since most rely on reading in static reference points (i.e. coordinates).

PR #1004, opened 2022-04-01
Coordinate Sets for Species

trisyoungs:
This PR implements the new `CoordinateSetsProcedureNode` which, when used in a generator context, performs MD on a given species in order to generate a set of "different" geometries.  When an `AddProcedureNode` is given this node instead of a target species, copies of the different geometries are used sequentially instead, resulting in a more diverse geometry in the starting configuration (which is a good thing).

### DIFF
--- a/src/modules/md/functions.cpp
+++ b/src/modules/md/functions.cpp
@@ -58,26 +58,19 @@ std::vector<Vec3<double>> MDModule::evolve(const ProcessPool &procPool, const Po
     const auto kb = 0.8314462;
 
     // Store atomic masses for future use
-    for (auto &&[i, m] : zip(atoms, mass))
-        m = AtomicMass::mass(i.Z());
+    std::transform(atoms.begin(), atoms.end(), mass.begin(), [](const auto &atom) { return AtomicMass::mass(atom.Z()); });
 
     // Calculate total velocity and mass over all atoms
-    Vec3<double> vCom;
-    auto massSum = 0.0;
-    for (auto &&[v, m] : zip(velocities, mass))
-    {
-        vCom += v * m;
-        massSum += m;
-    }
+    auto massSum = std::accumulate(mass.begin(), mass.end(), 0.0);
+    auto vCom = std::transform_reduce(velocities.begin(), velocities.end(), mass.begin(), Vec3<double>());
 
     // Remove any velocity shift
     vCom /= massSum;
     std::transform(velocities.begin(), velocities.end(), velocities.begin(), [vCom](auto vel) { return vel - vCom; });
 
     // Calculate instantaneous temperature
-    ke = 0.0;
-    for (auto &&[m, v] : zip(mass, velocities))
-        ke += 0.5 * m * v.dp(v);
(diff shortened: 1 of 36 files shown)


trisyoungs:
At this stage probably not at all. It's the addition of another class derived from `ProcedureNode`, so any (de)serialisation would be handled by the `KeywordStore` and its contents in a generic manner.

IVlaD17:
> At this stage probably not at all. It's the addition of another class derived from `ProcedureNode`, so any (de)serialisation would be handled by the `KeywordStore` and its contents in a generic manner.

Alrighty then.

src/modules/md/functions.cpp
```diff
@@ -32,3 +35,107 @@ double MDModule::determineTimeStep(const std::vector<Vec3<double>> &f)
...
+    double tInstant, ke, tScale;
+
+    // Units
+    // J = kg m2 s-2  -->   10 J = g Ang2 ps-2
+    // If ke is in units of [g mol-1 Angstroms2 ps-2] then must use kb in units of 10 J mol-1 K-1 (= 0.8314462)
+    const auto kb = 0.8314462;
+
+    // Store atomic masses for future use
+    for (auto &&[i, m] : zip(atoms, mass))
+        m = AtomicMass::mass(i.Z());
```
rprospero, line 62:
```suggestion
    std::transform(atoms.begin(), atoms.end(), mass.begin(), 
      [](const auto atom){return AtomicMass::mass(atom.Z());});
```

src/modules/md/functions.cpp
```diff
@@ -32,3 +35,107 @@ double MDModule::determineTimeStep(const std::vector<Vec3<double>> &f)
...
+
+    // Calculate total velocity and mass over all atoms
+    Vec3<double> vCom;
+    auto massSum = 0.0;
+    for (auto &&[v, m] : zip(velocities, mass))
+    {
+        vCom += v * m;
+        massSum += m;
+    }
```
rprospero, line 71:
```suggestion
    auto massSum = std::accumulate(mass.begin(), mass.end());
    auto vCom = std::transform_reduce(mass.begin(), mass.end(), velocities.begin(), 0);
```

This should help the compiler with vectorization.

src/modules/md/functions.cpp
```diff
@@ -32,3 +35,107 @@ double MDModule::determineTimeStep(const std::vector<Vec3<double>> &f)
...
+
+    // Remove any velocity shift
+    vCom /= massSum;
+    std::transform(velocities.begin(), velocities.end(), velocities.begin(), [vCom](auto vel) { return vel - vCom; });
+
+    // Calculate instantaneous temperature
+    ke = 0.0;
+    for (auto &&[m, v] : zip(mass, velocities))
+        ke += 0.5 * m * v.dp(v);
```
rprospero, line 80:
```suggestion
    ke = std::transform_reduce(mass.begin(), mass.end(), velocities.begin(), std::plus<>(),
                               [](const auto m, const auto &v){ return 0.5 * m * v.dp(v); });
```

Again should help with vectorization

PR #1005, opened 2022-04-01
Singularity upload nix

rprospero:
Unfortunately, we need to install nix at multiple places in the pipeline.  While this isn't a particular time sink, it does mean duplicated code.  I've refactored the code a bit to ensure that the logic is held the same everywhere.

This should also help with the pushing of singularity images to the harbor, which I broke again


trisyoungs:
Not that you requested a review, but this looks lots cleaner!

Issue #984, opened 2022-03-08
ProcedureNode edits don't signal input file is modified


trisyoungs:
**Describe the bug**
Editing a ProcedureNode keyword (e.g. a value in a `Parameters` node or length / angle in a `Box` node) does not correctly flag a change to the modification status of the file.

PR #1011, opened 2022-04-04
Connect the keyword widget to the setModified slot

rprospero:
There is no need to connect the actual ProcedureModel, since the model
is read-only.  This will close issue #984.


trisyoungs (approved):
:+1:

PR #1023, opened 2022-04-11
Remove ProcedureEditor

### LINKED ISSUE #989: Use ProcedureNode model in Analyse widget GUI
Following #963 we are now in a position to remove all of the old code from `src/gui/charts`. The only thing stopping us from doing so right now is the GUI widget for the `Analyse` module which needs moving to the new model.

### LINKED ISSUE #1003: Use stack of KeywordWidgets for Procedure node editing
Currently the `ConfigurationTab` displays a `KeywordWidget` for the currently-selected `ProcedureNode` but which is new'd each time.  This would be better implemented as a stack of "viewed" `KeywordWIdget`s as is implemented for modules in the `LayerTab`.

This probably warrants the creation of a compound `ProcedureEditor` widget which encompasses both the node tree and the `KeywordWidget` stack, and which can be reused (e.g. for #989).

rprospero:
This takes removes the old `ProcedureEditor`.  The last place it was called was in the Analyse module GUI.  I've removed it from the module and it seems to run just fine without it.

This closes #989.
This closes #1003.


rprospero:
I've added a cache of widgets that will close #1003.  However, it should get refactored out into its own class.

IVlaD17 (approved):

Issue #654, opened 2021-04-22
Simple option to get / download all required ANTLR4 dependencies


trisyoungs:
Nice feature to have in the build system - an option like `-DANTLR_GET` that will automatically download the ANTLR4 jar / zip from the website without any further options.

trisyoungs:
Could be achieved with something like:
```
if(ANTLR_GET)
  include(FetchContent)
  fetchcontent_populate(
    antlrjar
    URL        https://www.antlr.org/download/antlr-4.8-complete.jar
    URL_HASH   MD5=2016a2a6a10f0dd18eaef8b15c9df779
    DOWNLOAD_NO_UNPACK
    SOURCE_DIR antlrjar
  )
endif(ANTLR_GET)
```

... however the `DOWNLOAD_NO_UNPACK` option was not introduced until CMake 3.18.

trisyoungs:
Closed by #1024.

PR #1024, opened 2022-04-13
Acquire antlr runtime from conan

rprospero:
This simplifies the build system and should make life easier for windows devs.


trisyoungs:
Quick note - this would close #654.

PR #1027, opened 2022-04-14
Add a Pair Potential class wrapper

### LINKED ISSUE #1022: Add a wrapper for PairPotential
As part of #1010 it's become evident that the way `PairPotentials` data is being stored is less than ideal. Adam and I had a talk and we agreed to write in a wrapper that will hold references to the current data and we'll slowly start replacing the current code with that as we go along.

IVlaD17:
Closes #1022 

The `PairPotential` class elements are spread all over the place which makes writing deserializing and serializing code particularly tedious and ugly. This PR adds a class that contains references to all of these elements and this class should set us up to be able to easily transition towards a nicer set-up in the future.


src/main/dissolve.h
```diff
@@ -31,6 +32,7 @@ class Dissolve
     private:
     // Reference to CoreData
     CoreData &coreData_;
+    SerializablePairPotential pairPotential_;
```
trisyoungs, line 35: This is a bit misleading as named.
```suggestion
    SerializablePairPotential serializablePairPotentials_;
```

src/main/dissolve.cpp
```diff
@@ -7,7 +7,9 @@
 #include "classes/neutronweights.h"
 #include "classes/species.h"
 
-Dissolve::Dissolve(CoreData &coreData) : coreData_(coreData)
+Dissolve::Dissolve(CoreData &coreData)
+    : coreData_(coreData),
+      pairPotential_(pairPotentialRange_, pairPotentialDelta_, atomTypeChargeSource_, coreData_.atomTypes())
```
trisyoungs, line 12:
```suggestion
      serializablePairPotentials_(pairPotentialRange_, pairPotentialDelta_, atomTypeChargeSource_, coreData_.atomTypes())
```

src/classes/serializablepairpotential.cpp
```diff
@@ -0,0 +1,34 @@
+#include "serializablepairpotential.h"
```
trisyoungs, line 1: Convention in the code is to precede with the subdir name as well. Also needs the license header.
```suggestion
// SPDX-License-Identifier: GPL-3.0-or-later
// Copyright (c) 2022 Team Dissolve and contributors

#include "classes/serializablepairpotential.h"
```

src/classes/serializablepairpotential.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
```
trisyoungs, line 1:
```suggestion
// SPDX-License-Identifier: GPL-3.0-or-later
// Copyright (c) 2022 Team Dissolve and contributors

#pragma once
```

windbuild.ps1
```diff
@@ -0,0 +1,6 @@
+mkdir build
```
trisyoungs, line 1: Custom build script of limited use - should this file really be in the repo?

trisyoungs (changes requested):
Makes sense as a first step in gathering all the pair potential info together.

windbuild.ps1
IVlaD17: I tried adding it to .gitignore but it seems that that did not remove the file from the PR. I might need some help there.

trisyoungs (approved):

PR #1031, opened 2022-04-26
Patch 04/2022

trisyoungs:
A small collection of fixes and improvements in one PR.


rprospero:
This PR may have a small wait on it.  Homebrew upgraded the ANTLR version on Mac, which seems to be a breaking change, and they have no facility to access old versions.  Until this is fixed, nothing can build on Mac.  I'm currently working on the issue.

src/procedure/nodes/node.cpp
```diff
@@ -300,15 +288,15 @@ bool ProcedureNode::deserialise(LineParser &parser, const CoreData &coreData)
 // Write node data to specified LineParser
 bool ProcedureNode::write(LineParser &parser, std::string_view prefix)
 {
-    // Block Start - does this node have a required name?
-    if (mustBeNamed())
+    // Block Start - node type and name (if specified)
+    if (name_.empty())
```
IVlaD17, line 292: Shouldn't this be `if(!name_.empty())`?

IVlaD17 (commented):
Looks good overall, I can't see anything wrong except for the comment I made.

src/procedure/nodes/node.cpp
trisyoungs: The code in the `if`/`else` is correct based on this initial condition - I guess the argument here would be to consider the "positive" case first, but I kinda prefer the approach of going for the simpler logic test in the initial condition.

IVlaD17: Fair enough.

PR #1033, opened 2022-04-26
Include std::any in CIFImportParser

### LINKED ISSUE #1032: CI is broken by Homebrew upgrade to Antlr 1.40
Homebrew recently upgraded to Antlr 4.10 and there is no mechanism for installing the older version.  This means that we must make our code compatible with 4.10.

rprospero:
Fix the grammars to support ANTLR 4.10.  This will close #1032.

### DIFF
--- a/.azure-pipelines/templates/build-osx.yml
+++ b/.azure-pipelines/templates/build-osx.yml
@@ -22,7 +22,7 @@ steps:
       Qt6_DIR=/tmp/qt/${{ parameters.qtver }}/macos/lib/cmake/Qt6
       QT_BASE_DIR=/tmp/qt/${{ parameters.qtver }}/macos
       QTVER=${{ parameters.qtver }}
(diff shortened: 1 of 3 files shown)


trisyoungs:
Maybe I'm missing something, but what why are we using the Homebrew ANTLR4 as opposed to that from conan? We can download any Java binary version from the ANTLR4 website if that's the issue.

rprospero:
I'm an idiot.  I already moved the antlr runtime into conan as part of cleaning up the Windows development environment.  So downloading the Antlr jar file that matches the runtime version is the only possible solution.

trisyoungs:
Heh - well let's hope this is enough to satisfy OSX. Easily the winner for the most awkward cross-platform target award.

trisyoungs:
Also, don't forget that we already have an `ANTLR_EXECUTABLE` option defined in our CMake.

PR #1034, opened 2022-04-27
Desensitize

### LINKED ISSUE #1030: Don't allow context menu edits in Layer when running
When Dissolve is running the context menu in the module layer model is still enabled and active, which it definitely shouldn't be.

trisyoungs:
This PR renames a bunch of functions throughout the GUI code relating to enabling/disabling widgets that edit data. For some unknown reason dating back to very early days of the project these had been termed "sensitive" controls, which I guess sort of makes sense, but isn't exactly descriptive.

Closes #1030.

### DIFF
--- a/src/gui/charts/chart.hui
+++ b/src/gui/charts/chart.hui
@@ -48,10 +48,10 @@ class ChartBase : public QWidget
     public:
     // Update controls within widget
     void updateControls();
-    // Disable sensitive controls within widget
-    void disableSensitiveControls();
-    // Enable sensitive controls within widget
-    void enableSensitiveControls();
+    // Prevent editing within widget
+    void preventEditing();
+    // Enable editing within widget
+    void enableEditing();
 
     /*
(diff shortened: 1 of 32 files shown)


IVlaD17 (changes requested):
Sorry to be this petty, but if we are calling it `enableEditing`, can we call the other one `disableEditing` instead of `preventEditing`?

trisyoungs:
> Sorry to be this petty, but if we are calling it `enableEditing`, can we call the other one `disableEditing` instead of `preventEditing`?

What about changing `enableEditing` for `allowEditing` instead?

IVlaD17:
> > Sorry to be this petty, but if we are calling it `enableEditing`, can we call the other one `disableEditing` instead of `preventEditing`?
> 
> What about changing `enableEditing` for `allowEditing` instead?

That works too.

IVlaD17 (approved):

PR #1008, opened 2022-04-04
TOML File Format Initial PR

IVlaD17:
Closes #899 

First of all I am making it close #899 because we'll be replacing that issue with others that will better define the overall issue.

Secondly, there's two problems with this PR as I am creating this:
- The linting will be all over the place and I'll have to do this manually after making  the PR (please bear with me on that)
- The code within `io.cpp` code that does the outputting at the moment should be put behind a flag/macro.

PR #1039, opened 2022-05-03
Deserialize Pair Potentials

IVlaD17:
Implemented the `deserialize()` method for the PairPotentials. As part of this PR, I've also added another constructor which I'm fairly sure will change based on other comments. 

Additionally, I changed the node structure to now be `pairPotentials.atomTypes.name` instead of `pairPotentials.name` because TOML bugged out when reading the file without the `atomTypes` node. Doing this required changing the `serialize()` method too.


src/classes/atomtype.cpp
```diff
@@ -114,3 +115,21 @@ toml::basic_value<toml::discard_comments, std::map, std::vector> AtomType::seria
 
     return atomType;
 }
+void AtomType::deserialize(toml::value node)
+{
+    if (!node["z"].is_uninitialized())
```
trisyoungs, line 120: Admittedly I may not have been following the TOML conversation as closely as I would have liked lately, but why can't this be written (more nicely) as:
```
    if (node.contains("z"))
```

src/classes/atomtype.cpp
```diff
@@ -114,3 +115,21 @@ toml::basic_value<toml::discard_comments, std::map, std::vector> AtomType::seria
 
     return atomType;
 }
+void AtomType::deserialize(toml::value node)
```
trisyoungs, line 118: Two comments on style here:
1. For consistency, `serialize` and `deserialize` should be the British spelling.
2. I can live with both method definitions being "back-to-back", but there should at the very least be a comment line before the first function to briefly describe the functions that follow.

src/classes/atomtype.cpp
```diff
@@ -114,3 +115,21 @@ toml::basic_value<toml::discard_comments, std::map, std::vector> AtomType::seria
...
+    if (!node["z"].is_uninitialized())
+        Z_ = Elements::element(std::string(node["z"].as_string()));
+    if (!node["charge"].is_uninitialized())
+        charge_ = node["charge"].as_floating();
+    if (!node["form"].is_uninitialized())
+        interactionPotential_.setForm(ShortRangeFunctions::forms().enumeration(std::string(node["form"].as_string())));
+
+    if (!node["parameters"].is_uninitialized())
+    {
+        std::vector<std::string> parameters = ShortRangeFunctions::parameters(interactionPotential_.form());
```
trisyoungs, line 129:
```suggestion
        auto &parameters = ShortRangeFunctions::parameters(interactionPotential_.form());
```

src/classes/atomtype.h
```diff
@@ -84,4 +85,5 @@ class AtomType
     bool sameParametersAs(const AtomType *other, bool checkCharge = false);
 
     toml::basic_value<toml::discard_comments, std::map, std::vector> serialize();
+    void deserialize(toml::value node);
```
trisyoungs, line 88: As with the definitions, need the same comment line describing the purpose of these two functions.

src/classes/serializablepairpotential.cpp
```diff
@@ -47,6 +47,29 @@ toml::basic_value<toml::discard_comments, std::map, std::vector> SerializablePai
     pairPotentials["shortRangeTruncation"] = PairPotential::shortRangeTruncationSchemes().keyword(shortRangeTruncationScheme_);
     if (!atomTypes_.empty())
         for (auto &atomType : atomTypes_)
-            pairPotentials[atomType->name().data()] = atomType->serialize();
+            pairPotentials["atomTypes"][atomType->name().data()] = atomType->serialize();
```
trisyoungs, line 50: Fairly sure I saw this comment in another PR review somewhere, but line 48 here (checking for the `atomTypes_` vector being `empty()`) is redundant and can be removed.

src/classes/serializablepairpotential.cpp
```diff
@@ -47,6 +47,29 @@ toml::basic_value<toml::discard_comments, std::map, std::vector> SerializablePai
...
     return pairPotentials;
 }
+
+void SerializablePairPotential::deserialize(toml::value node)
```
trisyoungs, line 54: Same comments on style as with the `AtomType` routines.

src/main/io.cpp
```diff
@@ -214,6 +214,26 @@ bool Dissolve::loadInput(std::string_view filename)
...
+            {
+                toml::value pairPotentialsNode = toml::find(file, "pairPotentials");
+                if (!pairPotentialsNode.is_uninitialized())
+                    serializablePairPotential_.deserialize(pairPotentialsNode);
+            }
+
+            toml::basic_value<toml::discard_comments, std::map, std::vector> root;
+            root["pairPotentials"] = serializablePairPotential_.serialize();
+            output << std::setw(40) << root;
+        }
```
trisyoungs, line 236: Feels like chunk of code shouldn't be merged in as it's for local testing and is machine specific.

trisyoungs (changes requested):
Starting to take shape! Some comments on style and a few other bits.

src/main/io.cpp
IVlaD17: I'll fix it so it's no longer machine specific.

src/classes/atomtype.cpp
IVlaD17: 6euyx3

IVlaD17: I'll begrudgingly do the first and happily do the second.

src/main/io.cpp
rprospero: The `if` expression does need to be a `constexpr` if we're going to properly make this a feature flag.  However, if we make the filename cross-platform (i.e. putting the toml file in the same directory as the input file), then I would argue that it is useful to leave this in for now.  It's been quite useful when we collaborate on the TOML code to have this testing code present and identical for both of us.

I'd argue that we should but in an issue for moving this into a unit test, but we also don't want that unit test blocking other valid PRs in the interim.  The feature flag makes it possible, but I feel it's complex enough that it should be its own PR and not added onto this one.

IVlaD17: Created #1045 so I'll mark this as resolved.

src/classes/atomtype.cpp
IVlaD17: I've sorted that out but I also created #1046 to tackle `serialize()` methods added in #1008 because I wanted to keep this PR focused just on the PairPotentials.

src/classes/atomtype.cpp
IVlaD17: Changed it to that.

PR #1040, opened 2022-05-04
Restart File UX

### LINKED ISSUE #828: Better UX when loading simulations in GUI
The GUI (as with the CLI version) automatically loads the restart file named after the input file. However, from the GUI it is not possible to choose a different restart file or choose not to load one at all.  There are CLI switches to achieve this, but would be better implemented as functionality within the GUI.

We should probably do the following: after loading the input file, scan for restart files in the same directory, and present some dialog offering various options (e.g. Load Associated Restart File, Choose Other File,  Don't Load Any Restart File etc.)

Similarly, menu items should be added to offer similar functionality, including the ability to just load the Configuration data and nothing else.

### LINKED ISSUE #845: Change restart file write frequency from GUI
There is no option in the GUI to change the write frequency of the restart file (i.e. a companion to the CLI option `-f`).

trisyoungs:
This PR adds some UX to the GUI in relation to loading restart files.

### Old Behaviour
- When an input file was loaded within the GUI (i.e. via the `Open` or `Open Recent` file menu actions) the associated restart file was always loaded automatically if it was present. No choice was given to the user.
- No choice for an alternative restart file to load was offered.

### New Behaviour
- The GUI now checks which (if any) restart files exist in the input file directory, and a choice is offered to the user.
- If multiple restart files exist, a choice is offered over which to load.
- It is now possible to choose to load no restart data when loading an input file.

Some other tidying and bugfixing is performed along the way.

Closes #828.
Closes #845.


src/dissolve-gui.cpp
```diff
@@ -52,28 +49,40 @@ int main(int args, char **argv)
...
+    if (options.inputFile() && loadSuccessful && options.nIterations() > 0)
+    {
+        // Prepare for run
+        if (!dissolve.prepare())
+            return 1;
+
+        // Run main simulation
+        dissolve.iterate(options.nIterations());
+    }
+
```
rprospero, line 85: Could this go inside the previous conditional?  They share the first two conditions, so you'd only need to check the `nInterations`.  It also protects us if we need to perform another check for a successful load.

trisyoungs: :+1:

PR #1042, opened 2022-05-04
Remove Workspaces

### LINKED ISSUE #147: Make renaming gizmos easier
The current (focussed) gizmo in a workspace tends to follow the mouse, meaning that the present "Rename X" option in the Workspace menu is not ideal.

Could instead have the Rename action as a menu of the current gizmos, with any selected item triggering the rename dialog.

### LINKED ISSUE #177: Fix linked view

trisyoungs:
A PR to remove entirely the existing "workspace" functionality of Dissolve, which was: 1) incomplete; 2) sparingly used, and; 3) way out of scope for the project. Intended on being a sort of "better Excel" for comparing and analysing calculation results, it offers little that can't be better achieved with specific `Module`s should it be decided that such functionality is required in future.

The benzene example is the only real place where workspaces are used, and will be updated in #1038 to use loading of specific restart files instead (following merge of #1040).

Closes #147 by redundancy.
Closes #177 by redundancy.


rprospero (approved):
PRs that delete code are always my favourites.

This is the point where I will also confess that both the Workspaces and Gizmo code were bits that I didn't understand and was a little afraid to touch, so I'm happier than I should be to see them gone.

trisyoungs:
Very happy to oblige in this instance!  It definitely needed pruning...

PR #1047, opened 2022-05-05
Module icons

### LINKED ISSUE #130: Regroup calculation modules and add icons
Many Calculate* modules are missing icons, and are in the same group in the module palette, but would be better organised to reflect general type of calculation performed.

trisyoungs:
PR to add icons for all modules that still required one, and updated some others.  Also moves the module category map from `std::vector<std::pair>>` to an actual `std::map`, albeit at the expense of having to use `std::next` in order to translate model row indices into positions within the map. Still, this is not a time critical or heavily used piece of code, so should be acceptable.

Closes #130.


IVlaD17 (approved):
They look much better than I expected them to!

trisyoungs:
> They look much better than I expected them to!

Thanks! I think... :)

PR #1048, opened 2022-05-05
Keep CellArrays consistent with pair potential range

### LINKED ISSUE #957: Changing cutoff must regenerate Configuration cells
**Describe the bug**
Changing the cutoff mid-simulation uses the current `Configuration` contents along with the `CellArray` generated at the previous cutoff distance, causing artifacts in the simulation box (tendency for molecules to align at cell edges).

The current cutoff distance value at the `prepare()` stage could be stored when the cell array(s) are generated, and checked for consistency at the next call to `prepare()`.

trisyoungs:
A straightforward PR to ensure that the `Configuration` `CellArray`s are updated if the user changes the pair potential range.

Closes #957.

### DIFF
--- a/src/main/simulation.cpp
+++ b/src/main/simulation.cpp
@@ -57,7 +57,7 @@ bool Dissolve::prepare()
             return Messenger::error("PairPotential range ({}) is longer than the shortest non-minimum image distance ({}).\n",
                                     pairPotentialRange_, maxPPRange);
 
-        // Accumulate the total species usage counts for the next check
+        // Update species usage for the next check
         for (auto &[sp, pop] : cfg->speciesPopulations())
             globalUsedSpecies.emplace(sp);
     }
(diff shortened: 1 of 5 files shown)


src/main/simulation.cpp
```diff
@@ -47,17 +47,33 @@ bool Dissolve::prepare()
...
+        newPairPotentialRange = pairPotentialRange_;
+    }
+
     // Check Configurations
     std::set<const Species *> globalUsedSpecies;
     for (auto &cfg : configurations())
     {
+        // Regenerate cell array if the pair potential range has changed
+        if (newPairPotentialRange)
+            cfg->updateCells(7.0, newPairPotentialRange.value());
```
rprospero, line 68: There's a nice way to do this with the new monadic operations on `optional` in C++23, but that will have to wait.  Meanwhile, I thought that the above with clearer, but I'm open to disagreement.

```suggestion
        if (newPairPotentialRange)
            cfg->updateCells(7.0, *newPairPotentialRange.);
```

rprospero (approved):
One bit of opinionated rambling on my part, but this is very clear code.

src/main/simulation.cpp
trisyoungs: Happy to dereference rather than `value()` - don't mind either approach.

PR #1049, opened 2022-05-05
Clear additional potentials

trisyoungs:
Quick PR to add a menu item to clear all additional potentials from the simulation, reverting to the reference (input) forcefield.


rprospero (approved):
Clean and clear.

PR #1050, opened 2022-05-09
1045 toml unit test

### LINKED ISSUE #1045: Implement proper unit tests for TOML
Based on @rprospero's comments in https://github.com/disorderedmaterials/dissolve/pull/1039#discussion_r864602568, this issue should be keeping track of the process of implement proper unit tests for serialising and deserialising objects to and from TOML.

rprospero:
I've added a proper unit test for the TOML code.  As part of this process, I've also done the following:

- I've factored out the primary Dissolve serialisation and deserialisation into their own functions.  This makes it easier to separate the testing of the TOML code from the rest of Dissolve
- I've made the toml_testing_flag public.  This allows us to access if from the unit test.

Currently, it only performs some *very* basic tests on a couple of files because the current development branch doesn't particularly support deserialisation at the moment.  The test can be expanded the serialisation code is merged in.

This closes #1045.

### DIFF
--- a/src/main/io.cpp
+++ b/src/main/io.cpp
@@ -142,6 +142,66 @@ bool Dissolve::loadInputFromString(std::string_view inputString)
     return result;
 }
 
+toml::basic_value<toml::discard_comments> Dissolve::serialise()
+{
+    toml::basic_value<toml::discard_comments, std::map, std::vector> root;
+    if (!coreData_.masterBonds().empty() || !coreData_.masterAngles().empty() || !coreData_.masterTorsions().empty() ||
+        !coreData_.masterImpropers().empty())
+    {
+        toml::basic_value<toml::discard_comments, std::map, std::vector> masterNode;
+        if (!coreData_.masterBonds().empty())
+        {
+            toml::basic_value<toml::discard_comments, std::map, std::vector> bonds;
+            for (auto &bond : coreData_.masterBonds())
+                bonds[bond->name().data()] = bond->serialize();
+            masterNode["bonds"] = bonds;
+        }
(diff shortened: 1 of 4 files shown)


src/main/io.cpp
```diff
@@ -142,6 +142,43 @@ bool Dissolve::loadInputFromString(std::string_view inputString)
...
+template <typename T>
+void addVectorToNode(std::vector<T> &vector, std::string name, toml::basic_value<toml::discard_comments, std::map> &node)
+{
+    if (vector.empty())
+        return;
+    toml::basic_value<toml::discard_comments, std::map, std::vector> group;
+    for (auto &item : vector)
+        group[item->name().data()] = item->serialize();
+    node[name] = group;
+};
```
trisyoungs, line 155: Perhaps a candidate for being moved to the interface class once it exists?

src/main/io.cpp
```diff
@@ -142,6 +142,43 @@ bool Dissolve::loadInputFromString(std::string_view inputString)
...
+
+toml::basic_value<toml::discard_comments> Dissolve::serialise()
+{
+    toml::basic_value<toml::discard_comments, std::map, std::vector> root;
```
trisyoungs, line 159: Typedef?

trisyoungs (approved):

src/main/io.cpp
rprospero: We both had the same idea!

src/main/io.cpp
rprospero: I moved that typedef into the file with the interface class.

PR #1052, opened 2022-05-10
1046 toml cleanup

rprospero:
This is a clean-up of some stylistic issues on the TOML code.

1) Following the suggestion from @IVlaD17, the `serialise` and `deserialise` routines have been moved into an interface class.  This class also provides a convenient namespace for utility functions

2) Move `serialise` and `deserialise` onto the British spelling

3) Ensure that `serialise` is always const

4) Use a type alias for the fairly complicated TOML type.  As an added bonus, I've tried to structure this so that we could swap out TOML for YAML and only need to edit the `base/toml.h` file.  Not that we ever want to do that.

5) Move more code into helper functions.  I think that there's a way that we can use regular overloading to escape needing so many templates, but that is still a work in progress.

Questions to think about during your review:

1) Is there a better name than TomlTable?  I assume that there has to be.
2) Is there a better place for the `Serialisable` class than `base/toml.h`.  I don't like that the class name and the file name are unrelated.

### DIFF
--- a/src/base/toml.h
+++ b/src/base/toml.h
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2022 Team Dissolve and contributors
+
+#pragma once
+
+#include <toml11/toml.hpp>
+
+#include <map>
+#include <vector>
+
+using TomlTable = toml::basic_value<toml::discard_comments, std::map, std::vector>;
+
+class Serialisable
+{
+    public:
+    virtual TomlTable serialise() const = 0;
+    void deserialise(TomlTable node) { return; }
+};
--- a/src/base/toml.h
+++ b/src/base/toml.h
@@ -8,11 +8,60 @@
 #include <map>
 #include <vector>
 
+// The type we use for the nodes of our serialisation tree
 using TomlTable = toml::basic_value<toml::discard_comments, std::map, std::vector>;
 
+// An interface for classes that can be serialised into an input file
 class Serialisable
 {
     public:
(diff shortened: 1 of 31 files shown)


src/base/toml.h
```diff
@@ -0,0 +1,67 @@
...
+
+#pragma once
+
+#include <toml11/toml.hpp>
+
+#include <map>
+#include <vector>
+
+// The type we use for the nodes of our serialisation tree
+using TomlTable = toml::basic_value<toml::discard_comments, std::map, std::vector>;
```
trisyoungs, line 12: Perhaps `SerialisedData`? Slightly clunky, I admit, but it does abstract things even further away from Toml should we ever decide to change.

src/classes/atomtype.h
```diff
@@ -4,11 +4,11 @@
 #pragma once
 
 #include "base/enumoptions.h"
+#include "base/toml.h"
```
trisyoungs, line 7: If we move to `SerialisedData`, the include could be `base/serialiser.h` or similar.

trisyoungs (commented):
Lots of nice cleaning up here. Some suggestions for renaming made in the comments - feel free to reject!

rprospero:
I settled on `SerialisedValue` as part of a move to using `toml::value`.  This covers the fact that it might contain different types of values (e.g. table or array).

IVlaD17:
> I settled on `SerialisedValue` as part of a move to using `toml::value`. This covers the fact that it might contain different types of values (e.g. table or array).

That sounds appropriate.

PR #1055, opened 2022-05-18
Remove renameat2 call from Qt6

### LINKED ISSUE #1054: Dissolve cannot run on old Kernels
**Describe the bug**
On linux, Qt 6 (and later editions of Qt 5) use the [renameat2](https://manpages.debian.org/testing/manpages-dev/renameat2.2.en.html) kernel call, which was added in Kernel 3.15.  Unfortunately, CentOS 7 is on kernel 3.10.

**To Reproduce**
Steps to reproduce the behaviour (if necessary):
1. Find a system with CentOS 7
2. Download the Dissolve singularity image
3. Run the singularity image
4. Get an error message

**Expected behaviour**
Dissolve loads instead of an error message

**Operating System:**
 - OS:   Linux
 - OS version / flavour: CentOS 7

rprospero:
This strips the offending kernel call out of Qt6 for our binary and singularity builds.  I have run the new image on the testing server and confirmed that Dissolve loads again.  This closes #1054

### DIFF
--- a/nix/qt6.nix
+++ b/nix/qt6.nix
@@ -40,6 +40,9 @@ in pkgs.stdenv.mkDerivation {
     ${pkgs.p7zip}/bin/7z x ${qttools} -o$out
     ${pkgs.p7zip}/bin/7z x ${qtwayland} -o$out
   '';
+  buildPhase = ''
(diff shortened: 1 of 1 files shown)


trisyoungs (approved):
Simple enough, but ultimately not what we should have to be doing!

PR #1058, opened 2022-05-19
Overcome Flat Geometries in Minimiser

### LINKED ISSUE #1056: Minimizer creating flat molecules
If you draw the molecule "flat" (i.e. with all atom coordinates at z = 0.0 which is easy to do on a flat screen!) the minimiser can't move atoms out of that plane.

trisyoungs:
Quick PR to allow the geometry optimise species to overcome the edge case where all atoms are in a single plane.  This is actually a relatively common case since most users draw molecules in the plane of the screen.

Closes #1056.


rprospero (approved):
Ingenious

PR #1060, opened 2022-05-20
Manual Procedure Creation

trisyoungs:
This PR implements templated node creation functions to allow more robust creation of manual procedures.

Major changes:
- Template `Procedure::createRootNode()` and `SequenceProcedureNode::create()`  replace `Procedure::addRootNode()` and SequenceProcedureNode::addNode()` respectively, with the former being removed completely and the latter being made `private` and called from the new templated function.
- `SequenceProcedureNode::create()` is now solely responsible for setting of the `scope_` of nodes and performing additional checks on validity.
- Long comments reflecting the structure of procedures in the `Calculate*` nodes have been removed, since the reduced verbosity of the procedure creation essentially reflects the same information.
- A `ProcedureNodeRegistry`, analogous to `ModuleRegistry`, has been added and handles creation of nodes from their enumerated type, as well as allowing nicer display of node types within models in the GUI in future.

Outcomes:
- Manual procedure creation is significantly cleaner.
- Procedure creation is more robust - duplicate node names are now not allowed within `Procedure`-scope since some nodes create processing data named after themselves.
- A bug introduced in #1031 meant that unique node names were only generated within local scope, rather than the required `Procedure`-scope. This was not possible to fix until proper scope-setting was remedied in the present PR.

### DIFF
--- a/src/gui/menu_configuration.cpp
+++ b/src/gui/menu_configuration.cpp
@@ -27,8 +27,6 @@ std::vector<std::shared_ptr<AddProcedureNode>> createRelativeMix(const std::vect
     auto count = 0;
     for (auto *sp : mixSpecies)
     {
-        std::shared_ptr<AddProcedureNode> add;
-
         // Set the population equation, creating a new ratio parameter if we need one
         std::string popString;
         if (count == 0)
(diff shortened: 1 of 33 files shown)


rprospero (approved):
This is much cleaner and more maintainable.

PR #1074, opened 2022-05-24
Nice, unique names

### LINKED ISSUE #1073: Disallow Spaces in Module Names
**Describe the bug**
Spaces in module names creates havoc in the restart file since this disrupts the (assumed) continuous name for the data item.

### LINKED ISSUE #1075: Enforce unique, nice names for Isotopologues
**Describe the bug**
Currently isotopologues within the same species can have the same name. Furthermore, nice names (i.e. no spaces or certain special characters) should be enforced.

trisyoungs:
This PR is a simple fix to enforce "nice names" (i.e. no spaces or specific special characters) for modules and isotopologues, and to also enforce unique names for the latter, when edited through the GUI and when loaded in initially.

At the same time, the `Module` variable `uniqueName_` is renamed to just `name_`, since the uniqueness is implicit (and we don't call any other object names by the same convention).

Closes #1073.
Closes #1075.

### DIFF
--- a/src/classes/species_io.cpp
+++ b/src/classes/species_io.cpp
@@ -428,8 +428,8 @@ bool Species::read(LineParser &parser, CoreData &coreData)
                 }
                 break;
             case (Species::SpeciesKeyword::Isotopologue):
-                iso = addIsotopologue(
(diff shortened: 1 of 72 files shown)


rprospero (approved):
I really hope that you didn't have to do this all by hand.

Issue #1072, opened 2022-05-24
OSX DMG Misses ANTLR4


trisyoungs:
**Describe the bug**
Packaging for OSX is broken, since the ANTLR4 runtime library is not included in the DMG file.

The CI logs show that the search path for the runtime is not the (correct) conan location:
```
ERROR: no file at "/usr/lib/libantlr4-runtime.4.9.3.dylib"
```

trisyoungs:
Snippet from the error log dumped on OSX:
```
Dyld Error Message:
dyld: Using shared cache: DC9219E8-D54C-369D-81BB-CF54BFB2F6CD
Library not loaded: libantlr4-runtime.4.9.3.dylib
Referenced from: /Applications/Dissolve-GUI 3.app/Contents/MacOS/dissolve-gui
Reason: image not found
```

trisyoungs:
Closed by #1079.

PR #1079, opened 2022-05-26
Move ANTLR to where macqtdeploy can find it

rprospero:
`macqtdeploy` is very dumb about finding dependencies.  Since we're now installing the antlr runtime library through conan instead of brew, it gets lost.
The general consensus online is that it's best to just copy the dependencies to where `macqtdeploy` will look for them.
This is a little ugly, since conan puts the dependencies in a hashed path, but a little `jq` magic got that for us.

This will come out of draft when I'm sure that it works.

### DIFF
--- a/.azure-pipelines/pr.yml
+++ b/.azure-pipelines/pr.yml
@@ -21,8 +21,7 @@ stages:
   - stage: "build"
     displayName: "Build"
     jobs:
-      - template: templates/build-linux-releases.yml
-        condition: eq( 1, 0)
+      # - template: templates/build-linux-releases.yml
       - job: "test_linux_mpi"
         displayName: "Testing MPI, CLI"
         condition: eq( 1, 0)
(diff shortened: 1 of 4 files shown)


rprospero:
So, the good news is that the current commit fixes the ANTLR issue.  Unfortunately, TBB has exactly the same issue and will also need special handling.

rprospero:
This will close #1072

trisyoungs (approved):
Not so ugly!  I was going to suggest copying the libs to find to a build-local directory rather than "polluting" a system directory, but it makes no real odds. Saves a `mkdir` anyway!

rprospero:
@trisyoungs Before I merge this, is `develop` the right place for this, or should it go onto the 0.9.1-pre branch?

trisyoungs:
Go straight into develop. I need to cherry pick the last two small prs across to 0.9.1-pre so I'll do this one at the same time.

PR #1080, opened 2022-05-26
Const correctness of deserialiser

rprospero:
This started as just a small PR to get the Vec3 class to support TOML serialisation.  However, it kind of spiraled out a bit from there due to some convenience issues and `const` correctness.

The major points are as follows

1. the `deserialise` function of the `Serialisable` class now takes a `const Serialised&` instead of just a `Serialisaed&`.  This was the right choice in general, so it putting this off was only going to make things more painful.
2. Remove calls to `operator[]` on `SerialisedValue`.  Unfortunately, for reasons I cannot quite comprehend, the oeprator overload isn't supported on const objects by the library.  I'm honestly looking at submitting a PR upstream to fix the issue, because it seems to just be an oversight.
3. Add the `from_toml` and `into_toml` members to the `Serialisable` class.  The library has some templates that hook into these members and will save us quite a bit of boilerplate, as well as making the code more readable.
4. Make Vec3 serialisable
5. Use clean up a couple of deserialiser using the new possibilities from `from_toml` and `into_toml`.

### DIFF
--- a/src/classes/serializablepairpotential.cpp
+++ b/src/classes/serializablepairpotential.cpp
@@ -52,25 +52,22 @@ SerialisedValue SerializablePairPotential::serialise() const
 }
 
 // This method populates the object's members with values read from a 'pairPotentials' TOML node
-void SerializablePairPotential::deserialise(SerialisedValue &node)
+void SerializablePairPotential::deserialise(const SerialisedValue &node)
 {
     if (node.contains("range"))
-        range_ = node["range"].as_floating();
+        range_ = toml::find<double>(node, "range");
     if (node.contains("delta"))
-        delta_ = node["delta"].as_floating();
+        delta_ = toml::find<double>(node, "delta");
     if (node.contains("includeCoulomb"))
-        atomTypeChargeSource_ = node["includeCoulomb"].as_boolean();
+        atomTypeChargeSource_ = toml::find<bool>(node, "includeCoulomb");
     if (node.contains("coulombTruncation"))
-        coulombTruncationScheme_ =
-            PairPotential::coulombTruncationSchemes().enumeration(std::string(node["coulombTruncation"].as_string()));
+        coulombTruncationScheme_ = PairPotential::coulombTruncationSchemes().enumeration(
+            std::string(toml::find<std::string>(node, "coulombTruncation")));
     if (node.contains("shortRangeTruncation"))
-        shortRangeTruncationScheme_ =
-            PairPotential::shortRangeTruncationSchemes().enumeration(std::string(node["shortRangeTruncation"].as_string()));
+        shortRangeTruncationScheme_ = PairPotential::shortRangeTruncationSchemes().enumeration(
+            std::string(toml::find<std::string>(node, "shortRangeTruncation")));
 
     if (node.contains("atomTypes"))
-    {
-        toml::value atomTypesNode = node["atomTypes"];
-        for (auto &[name, data] : atomTypesNode.as_table())
+        for (auto &[name, data] : toml::find(node, "atomTypes").as_table())
             atomTypes_.emplace_back(std::make_unique<AtomType>(name))->deserialise(data);
-    }
 }
--- a/src/classes/species.cpp
+++ b/src/classes/species.cpp
@@ -223,39 +223,51 @@ SerialisedValue Species::serialise() const
 }
 
 // Read values from a tree node
-void Species::deserialise(SerialisedValue &node, CoreData &coreData)
+void Species::deserialise(const SerialisedValue &node, CoreData &coreData)
 {
     auto tomlAtoms = toml::find(node, "atoms").as_array();
     for (auto &tomlAtom : tomlAtoms)
         atoms_.emplace_back().deserialise(tomlAtom);
 
-    Serialisable::toVector(node, "bonds", [this, &coreData](SerialisedValue &bond) {
-        bonds_.emplace_back(&atoms_[bond["i"].as_integer() - 1], &atoms_[bond["j"].as_integer() - 1])
-            .deserialise(bond, coreData);
-    });
-    Serialisable::toVector(node, "angles", [this, &coreData](SerialisedValue &angle) {
-        angles_
-            .emplace_back(&atoms_[angle["i"].as_integer() - 1], &atoms_[angle["j"].as_integer() - 1],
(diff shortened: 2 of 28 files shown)


rprospero:
I've added an initial serialiser for NodeValue, since it's now required to create a Vec3<NodeValue>.  However, that serialiser is not called at the moment, so the code isn't properly tested and I've added a FXIME comment on it.  Once we start serialising expressions,  which should come soon, then that comment can be removed and we NodeValue can have a proper deserialiser function.

src/classes/serializablepairpotential.cpp
```diff
@@ -52,25 +52,22 @@ SerialisedValue SerializablePairPotential::serialise() const
 }
 
 // This method populates the object's members with values read from a 'pairPotentials' TOML node
-void SerializablePairPotential::deserialise(SerialisedValue &node)
+void SerializablePairPotential::deserialise(const SerialisedValue &node)
 {
     if (node.contains("range"))
-        range_ = node["range"].as_floating();
+        range_ = toml::find<double>(node, "range");
```
trisyoungs, line 58: :+1:
I much prefer this syntax.

trisyoungs: Thinking more about it I wonder if there's a case for allowing a one-line form for these POD-focussed calls by wrapping the `contains()` and `find<T>()` calls in a trivial template function in the `Serialisable` class which takes a reference to the member variable we _might_ set, e.g.:
```
template <class T> bool Serialisable::getIfPresent(const Serialisable &node, std::string_view key, T &target)
{
    if (!node.contains(key))
        return false;
    target = toml::find<T>(node, key);
    return true;
}
```

src/classes/species.cpp
```diff
@@ -223,39 +223,37 @@ SerialisedValue Species::serialise() const
 }
 
 // Read values from a tree node
-void Species::deserialise(SerialisedValue &node, CoreData &coreData)
+void Species::deserialise(const SerialisedValue &node, CoreData &coreData)
 {
-    auto tomlAtoms = toml::find(node, "atoms").as_array();
-    for (auto &tomlAtom : tomlAtoms)
-        atoms_.emplace_back().deserialise(tomlAtom);
+    atoms_ = toml::find<std::vector<SpeciesAtom>>(node, "atoms");
```
trisyoungs, line 228: No `contains()` check here, so does this `throw` if no "atoms" key exists? No atoms in a Species might be invalid from the perspective of the simulation, but I suppose it *is* valid TOML.

src/classes/species.cpp
```diff
@@ -223,39 +223,37 @@ SerialisedValue Species::serialise() const
...
 
-    Serialisable::toVector(node, "bonds", [this, &coreData](SerialisedValue &bond) {
-        bonds_.emplace_back(&atoms_[bond["i"].as_integer() - 1], &atoms_[bond["j"].as_integer() - 1])
+    Serialisable::toVector(node, "bonds", [this, &coreData](const SerialisedValue &bond) {
+        bonds_.emplace_back(&atoms_[toml::find<int>(bond, "i") - 1], &atoms_[toml::find<int>(bond, "j") - 1])
```
trisyoungs, line 231: Same comment as for the "atoms" retrieval above. Also we're not doing any range checking on the supplied atom indices, so we risk crashing hard here. We should probably capture the atom indices first and check them before emplacing the intramolecular term.

src/classes/speciesintra.h
```diff
@@ -156,23 +156,24 @@ template <class Intra, class Functions> class SpeciesIntra : public Serialisable
...
     }
     // Load form from tree node
-    template <typename Lambda> void deserialiseForm(SerialisedValue &node, Lambda lambda)
+    template <typename Lambda> void deserialiseForm(const SerialisedValue &node, Lambda lambda)
     {
         if (node.contains("form"))
         {
-            std::string form = node["form"].as_string();
+            auto form = toml::find<std::string>(node, "form");
             if (form.find("@") != std::string::npos)
```
trisyoungs, line 177: The '@' only indicates a master term if it is the very first character in the name, so this check can be more strict.

src/procedure/nodevalue.h
```diff
@@ -61,4 +62,9 @@ class NodeValue
     double asDouble();
     // Return value represented as a string
     std::string asString(bool addQuotesIfRequired = false) const;
+
+    // Express as a tree node
```
trisyoungs, line 66: I'm a sucker for grouping functions together in headers under big banner comments to aid readability and casual location of functions, as I'm sure you've noticed! So...
```suggestion

    /*
     * Serialisable
     */
    public:
    // Express as a tree node
```

trisyoungs (approved):
Looks good. I don't think I made any comments that stop this from being approved, and there are one or two that, if you think they're worthwhile, can go as future issues.

src/classes/species.cpp
rprospero: You're right that it is valid TOML, but I thought that it would be an invalid input file, so the best solution would be to give an error to the user at that point instead of continuing to try and load a file that we couldn't make sense of.

src/classes/species.cpp
rprospero: You're absolutely right about the atom indices and I intend to fix this before the merge.

src/classes/speciesintra.h
rprospero: Agreed.  I will change this to enforce the first character.

src/classes/species.cpp
rprospero: I've switched from using the `[]` operator to using the `at` member function, which performs bounds checking.

src/classes/serializablepairpotential.cpp
rprospero: I've marked this in issue #1087.

rprospero: However, looking over the code, many of these instances would be better served with `toml::find_or`, which provides a default value when the key isn't present.

trisyoungs: I thought that too, but I think there's an argument for only setting values if they're present, and not enforcing defaults if they aren't. If we do enforce defaults then we have to set (or have access to) the default values in the `deserialise` function which may not always be convenient. There probably aren't many cases where a default value is required, but I think the intent is clearer if we only set provided values.

PR #1082, opened 2022-05-27
Improved CIF Import

trisyoungs:
This PR improves the overall ability of the CIF Import Wizard, and fixes some bugs along the way.

### DIFF
--- a/src/classes/species.cpp
+++ b/src/classes/species.cpp
@@ -29,14 +29,14 @@ void Species::clear()
 }
 
 // Copy basic information (atoms and intramolecular terms)
-void Species::copyBasic(const Species *source)
+void Species::copyBasic(const Species *source, bool copyAtomTypes)
 {
     clear();
 
     name_ = source->name_;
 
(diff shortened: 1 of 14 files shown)


src/classes/species.cpp
```diff
@@ -6,7 +6,7 @@
 #include "data/ff/ff.h"
 #include "data/isotopes.h"
 
-Species::Species(std::string name) : attachedAtomListsGenerated_(false), forcefield_(nullptr), name_(name)
+Species::Species(std::string name) : name_(name), attachedAtomListsGenerated_(false), forcefield_(nullptr)
```
rprospero, line 9: Just for my own education, did the order matter?

rprospero (approved):

src/classes/species.cpp
trisyoungs: Not here. But the order of initialisation here was different from the definition order in the header, so `gcc` generates a warning.  I like to strive for the ideal of zero warnings during compilation!

PR #1088, opened 2022-05-30
Cu-BTC Forcefield (Zhao 2010)

trisyoungs:
This PR adds the forcefield of Zhao et al. and which handles the Cu-BTC metal organic framework.

Some modifications to underlying routines have been made, so a cursory review is required.


rprospero (approved):
Went over the relationship between force and energy three times before I figured out that I was doing it wrong.

PR #1090, opened 2022-05-30
Toml contains

rprospero:
This PR remove a large amount of boilerplate around checking whether nodes exist while deserialising.  Nodes that have a default value now use `toml::find_or`.  Nodes that are mandatory will just throw an exception.

Originally, I had added a `Serialisable::getIfPresent` function for nodes which were not mandatory but did not have a default value.  In practice, this hasn't occurred yet.


trisyoungs (approved):
:+1:

PR #1101, opened 2022-06-07
Harbor Release fix

rprospero:
This should fix the issue with releases not getting pushed to harbor.  It's still a mystery why certain parts of the build process default to version 1.0.0, but this patch will indiscriminately push whatever version has been built.


trisyoungs (approved):
:+1:

PR #1114, opened 2022-06-10
Nix 22.05 upgrade

rprospero:
This move pulls the nix flake to the latest version.  Hopefully, this new version should include several bug fixes that will simplify our build


rprospero:
The biggest upgrade in this PR is that we can now pull Qt 6 from the package manager, instead of slurping the binaries off the Qt FTP server.  This means that adding more Qt packages, upgrading Qt 6 minor versions, and accepting bug fixes will happen essentially for free.

PR #1146, opened 2022-06-28
Fix latest Mac Antlr build issue

rprospero:
Use the same conan deploy framework on Mac that we use on Windows.  This should fix the build problems and leave us with a more stable mac build.


trisyoungs (approved):
Lovely! Although I'll miss the `jq`-bashing...

PR #1159, opened 2022-07-04
979 Layer Awareness

### LINKED ISSUE #979: Layer Awareness
PR #975 prevents the `RDF` module from running if a size factor for any targeted configuration is greater than 1.0. This is a more general issue, and applies to many more (but not all) modules (e.g. calculation modules in particular). Rather than adding these checks to all relevant modules, we could do this at the `Layer` level instead, preventing entire sets of modules from running depending on various criteria. The size factor is one, energy stability is another (as used by the `EPSR` module for instance).

trisyoungs:
This PR introduces additional run control flags at the layer level, allowing them to be skipped if the energies of all module-targeted configurations are not yet stable, or their size factors are anything other than 1.0.

Existing code in the `RDFModule` and `EPSRModule` has been removed since it is now redundant.

Closes #979.


DanNixon (approved):
:+1:

PR #1163, opened 2022-07-06
Keyword deprecation

trisyoungs:
This PR provides a simple facility for deprecating keywords in modules and nodes, allowing them to be recognised and deserialised, but which are not serialised when saving a new file.  This allows compatibility with old input files to be temporarily retained until the next significant version comes around.


DanNixon (approved):
:+1:

Issue #1164, opened 2022-07-06
Deprecated Keywords (v1.0)


trisyoungs:
Stub issue for tracking variables deprecated by PRs leading up to release.

PR #1165, opened 2022-07-06
1112 MD Starts Early

### LINKED ISSUE #1112: MD starts too early
**Describe the bug**
One can observe that the `MD` module can start too early - when the energy is "stable" but the configuration is still in a relatively bad state. Since we default to using a fixed timestep, this leads to molecule explosion from which there is no return.

We could consider implementing an additional check on the maximum force component at the beginning of the `MD` module if this exceeds our preset force-capping threshold then we can override the fixed timestep and use a variable one instead.

trisyoungs:
This PR refactors the timestep handling in the molecular dynamics module, aiming to solve #1112. A fixed timestep is not suitable for "bad" configurations where atoms overlap (e.g. as is the case after randomly adding molecules in to a box) hence a variable timestep was introduced some time ago.  However, the algorithm implemented at the time was derived from atomic collision simulations, whereas we have bound molecules, and so it tends to generate timesteps roughly 1.5 orders of magnitude smaller than accepted standard values. Here, a new algorithm is implemented which does not suffer from the same issues and appears to be robust in testing on a variety of small and large systems.

Deprecates the `VariableTimestep` keyword of `MDModule` (#1164).
Closes #1112.


src/modules/md/md.h
```diff
@@ -46,17 +60,16 @@ class MDModule : public Module
...
      */
     private:
     // Cap forces in Configuration
-    int capForces(double maxForceSq, std::vector<Vec3<double>> &f);
-    // Determine timestep based on maximal force component
-    double determineTimeStep(const std::vector<Vec3<double>> &f);
+    int capForces(double maxForceSq, std::vector<Vec3<double>> &fInter, std::vector<Vec3<double>> &fIntra);
+    // Determine timestep to use
+    std::optional<double> determineTimeStep(const std::vector<Vec3<double>> &fInter,
+                                            const std::vector<Vec3<double>> &fIntra) const;
```
rprospero, line 72: Nice use of optional

rprospero (approved):

PR #1175, opened 2022-07-13
Add QtSvg to Singularity

### LINKED ISSUE #1166: Singularity doesn't display svg icons
**Describe the bug**
The current continuous build Singularity artifacts do not display SVG icons in the GUI - most likely cause is that the QtSvg library is not being correctly packaged.

Seen on IDAaaS.

trisyoungs:
One-line PR which fixes display of svg icons in the Singularity image via addind QtSvg as a nix build dependency.

Closes #1166.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -90,6 +90,7 @@
             nativeBuildInputs = pkgs.lib.optionals gui [
               pkgs.wrapGAppsHook
               pkgs.qt6Packages.wrapQtAppsHook
+              pkgs.qt6Packages.qtsvg
             ];
 
             TBB_DIR = "${pkgs.tbb}";


flake.nix
```diff
@@ -90,6 +90,7 @@
             nativeBuildInputs = pkgs.lib.optionals gui [
               pkgs.wrapGAppsHook
               pkgs.qt6Packages.wrapQtAppsHook
+              pkgs.qt6Packages.qtsvg
```
rprospero, line 93: This absolutely will work.  However, it might be better to put it in the gui libs list that starts on line 54.  The `nativeBuildInput` are run time dependencies while `buildInputs` are  compile time dependencies.  This *should* only be a compile time dependency, so it probably belongs in the `buildInputs`, but Qt loves blurring those lines, so I'm already prepared for you to say that you tried that and it didn't work.

rprospero (approved):

flake.nix
trisyoungs: Nope, hadn't tried that.  Still getting my eyes used to the nix stuff.  I'll add the package to `buildInputs` and test to make sure.

trisyoungs: Ah, hang on - just reading your comment again. The `QtSvg` dependency is a funny one, and it's the runtime dependency that is causing the issue here, so I think it belongs in the `nativeBuildInputs` stage after all.

Issue #1176, opened 2022-07-13
Epic / Procedure Editing


trisyoungs:
### Focus
Work towards full editability of `Procedure`s within the GUI.

#### Part 1
Implemented in #1177.
- [x] Refactor `ProcedureNodeStack` to bring in line with `ModuleControlWidget`-style operation and close #1171
#### Part 2
Implemented in #1180.
- [x] Retrofit `AnalyseModule` GUI to use a `ProcedureWidget`.
#### Part 3
Implemented in #1182.
- [x] Move model to produce a tree rather than a list.
#### Part 4
Implemented in #1224.
- [x] Don't require sequence nodes to store pointer to parent `Procedure`
- [x] Make `SequenceProcedureNode` a non-node class for use as a member variable.
#### Part 5
Implemented in #1225.
- [x] Allow drag/drop of nodes from a palette, similar to module layer editing.
- [x] Allow drag/drop reordering of nodes within a procedure
- [x] Allow renaming of nodes through model
- [x] Update `Procedure` after internal move to ensure validity of node keyword data
- [x] Fix internal move of root sequence nodes
#### Part 6
Implemented in #1238.
- [x] #923
- [x] Add controls to collapse / expand the tree
- [x] Add control to show/hide context column (hide by default too) to simplify view
- [x] Bug: OperateSitePopulationNormalise node appears to have no keywords
#### Part 7
- [x] Remove `ProcedureChart`, `MimeTreeWidget`, `MimeTreeWIdgetItem`, and `MimeStrings` classes (all defunct)

PR #1180, opened 2022-07-14
Procedure Editing - Part 2

trisyoungs:
This PR follows on from #1177 and implements the display of a `ProcedureWidget` for a relevant keyword in a module.

Previously, `ProcedureKeyword`s had no associated control widget in the GUI, meaning that no editability of the procedure was possible. It is not possible to simply add a `ProcedureKeywordWidget` which displays a `ProcedureWidget` editor since this creates a circular dependency between the keyword widgets and the main GUI lib.  Therefore, an alternative solution is implemented here, detailed below. Suggestions for other, neater approaches are welcome!

A `ModuleControlWidget` has two buttons to access "Options" (i.e. target and control keywords, the latter via a `KeywordsWidget`) and "Output" (which constituted the `ModuleWidget` associated with the module, if one existed).  A third button, "Procedure", has been added, offering access to the procedure targeted by (currently) the first `ProcedureKeyword` defined in the keywords for the module.

If the module does not have a module control widget and/or a procedure keyword, the relevant button(s) is/are hidden.

Works towards #1176.

### DIFF
--- a/src/gui/modulecontrolwidget.h
+++ b/src/gui/modulecontrolwidget.h
@@ -12,6 +12,7 @@ class Dissolve;
 class DissolveWindow;
 class Module;
 class ModuleWidget;
+class ProcedureWidget;
 
 // Module Control Widget
 class ModuleControlWidget : public QWidget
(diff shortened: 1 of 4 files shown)


rprospero (approved):
Beautifully written

PR #1182, opened 2022-07-15
Procedure Editing - Part 3

trisyoungs:
This PR follows #1180 and addresses the most tedious part of the procedure editing work - transitioning `ProcedureModel` from a simple list to a tree. As you can readily imagine this was generally a pain, but after some light refactoring the results are rather nice:

image

Some modest simplification work was undertaken on `ProcedureNode`s in order to clarify the clashing roles of the `ProcedureNode::parent_` and `SequenceProcedureNode::parentNode_`. All nodes exist in a `SequenceProcedureNode` (either the `Procedure::rootSequenceNode_` or a branch local to another node, but the parent/child relationship that we want in order to aid representation in any model is the one between a given node and the parent node which owns the `SequenceProcedureNode` it exists in. Thus, `SequenceProcedureNode::parentNode_` is renamed to `owner_`, and `ProcedureNode::parent_` is removed completely. The `ProcedureNode::parent()` function now uses the stored sequence `scope_` variable to determine ownership at the higher level.

Works towards #1176.

### DIFF
--- a/src/gui/models/isotopologueSetModel.cpp
+++ b/src/gui/models/isotopologueSetModel.cpp
@@ -13,6 +13,104 @@ void IsotopologueSetModel::setSourceData(IsotopologueSet &set)
     endResetModel();
 }
 
+std::vector<QModelIndex> IsotopologueSetModel::addMissingSpecies(const std::vector<std::unique_ptr<Species>> &species)
+{
+    if (!set_)
+        return {};
+
+    auto &set = set_->get();
+
+    auto nAdded = 0;
+    for (const auto &sp : species)
+    {
+        if (!set.contains(sp.get()))
+        {
+            beginInsertRows({}, set.nSpecies(), set.nSpecies());
+            set.add(sp->naturalIsotopologue(), 1.0);
+            endInsertRows();
+            ++nAdded;
+        }
+    }
+
+    if (nAdded > 0)
+    {
+        emit(dataChanged(createIndex(set.nSpecies() - nAdded, 0), createIndex(set.nSpecies() - 1, 0)));
+        std::vector<QModelIndex> newIndices;
+        for (auto n = set.nSpecies() - nAdded; n < set.nSpecies(); ++n)
+            newIndices.push_back(createIndex(n, 0));
+        return newIndices;
+    }
+
+    return {};
+}
+
+void IsotopologueSetModel::addIsotopologueWeight(const QModelIndex index)
+{
+    // Check index validity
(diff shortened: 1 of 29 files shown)


src/gui/models/isotopologueSetModel.cpp
```diff
@@ -13,6 +13,104 @@ void IsotopologueSetModel::setSourceData(IsotopologueSet &set)
...
+    if (nAdded > 0)
+    {
+        emit(dataChanged(createIndex(set.nSpecies() - nAdded, 0), createIndex(set.nSpecies() - 1, 0)));
+        std::vector<QModelIndex> newIndices;
+        for (auto n = set.nSpecies() - nAdded; n < set.nSpecies(); ++n)
+            newIndices.push_back(createIndex(n, 0));
+        return newIndices;
+    }
+
+    return {};
```
rprospero, line 44:
```suggestion
    std::vector<QModelIndex> newIndices;
    for (auto n = set.nSpecies() - nAdded; n < set.nSpecies(); ++n)
        newIndices.push_back(createIndex(n, 0));
    if (!newIndices.empty())
        emit(dataChanged(createIndex(set.nSpecies() - nAdded, 0), createIndex(set.nSpecies() - 1, 0)));
    return newIndices;
```

This arrangement does reduce the complexity slightly.

src/gui/models/procedureModel.cpp
```diff
@@ -31,57 +31,86 @@ int ProcedureModel::rowCount(const QModelIndex &parent) const
...
+    return 0;
 }
 
 int ProcedureModel::columnCount(const QModelIndex &parent) const
 {
     if (!procedure_)
         return 0;
 
-    Q_UNUSED(parent);
-
```
rprospero, line 48: We still don't use the parent value.  Do we still need the `Q_UNUSED`?  I'll admit that I've always been a little vague on when exactly it is required.

rprospero (approved):
I'm quite excited by how this cleanup clarifies the structure of the nodes.  I've included a couple of optional comments, but I believe that everything else is good.

src/gui/models/procedureModel.cpp
trisyoungs: AFAIK it's just to suppress compiler warnings for unused vars, but for sure we don't use it everywhere where that's the case.  I suppose the question is whether we remove all such uses and suffer the indignity of compiler warnings, or go through the code and add (perhaps thousands) of them to suppress them consistently?

PR #1202, opened 2022-07-29
CIF Upgrades

### LINKED ISSUE #1198: CIF import should handle missing _atom_site_type_symbol
When importing a CIF, if only the _atom_site_label key is present and the _atom_site_type_symbol key is missing, the elements of the atoms are not detected.  Typically, if the latter key is missing the _atom_site_label contains a (potentially) pure element symbol, so we should use that instead.

### LINKED ISSUE #1200: CIF imported needs better unit tests
Need better coverage of CIF import / parsing via unit tests.

### LINKED ISSUE #1201: Fail to pack Faujasite crystal structure correctly
**Describe the bug**
User reports that faujasite CIFs (spacegroup Fm-3d) do not pack correctly and look odd compared to, e.g. , Mercury.
image
image

trisyoungs:
This PR makes some significant changes to the way space group information is stored and handled, following on from the discovery that Dissolve fails to detect (many) crystal space groups correctly. The reason behind the failures was the primitive way in which the CIF data was used - only the Hermann-Mauginn symbols were ever really probed, and the Hall symbols were ignored. In the present version, a full list of Hall symbols is now tabulated and used as a reference, with the associated Hermann-Mauginn symbols, international table number and space group setting code associated to them.  This also prompted a refactoring of the way `SGinfo` was utilised in the code, resulting in the `namespace`-ing of space group-related data, and the removal of an extraneous class in favour of namespaced functions.

Closes #1201.
Closes #1200 via the refactoring of space group detection and the introduction of a new unit test for space group determination.
Closes #1198.


DanNixon (approved):
:+1:

PR #1208, opened 2022-08-01
Fix intrashake energy

trisyoungs:
This PR fixes an issue with the energies calculated in the `IntraShake` processing, which failed to include the internal molecule pair potential contributions. This resulted in bad overlaps between atoms within molecules, which obviously ain't a good thing.


DanNixon (approved):
:+1:

PR #1212, opened 2022-08-08
Include antlr4 in Windows installer

rprospero:
So we enabled the installer to pull in the dependencies from conan, but the Windows installer didn't realise that it needed antlr.  Now that we've lead the horse to water, this PR should make it drink.


trisyoungs (approved):
:+1:

PR #1218, opened 2022-08-15
1215 Instantaneous CN

### LINKED ISSUE #1214: Export CN data from module
Add option to export CN data to disk in `CNModule`.

### LINKED ISSUE #1215: Allow instantaneous CN calculation
In most cases the averaged CN over an equilibrated simulation is the desired effect, however, the instantaneous data (i.e. for the current frame alone) is also useful in some cases.

Add an option / mechanism to store / export this data.

trisyoungs:
This PR adds the facility to calculate the instantaneous coordination numbers in the `CalculateCN` module, as opposed to the "time-averaged" coordination number.  A few places in the code are touched, namely the `Sum1DProcedureNode` and `CalculateCNModule` where relevant options / checks are made. An option for saving the instantaneous data is also added to the module.

Closes #1215.
Closes #1214.

### DIFF
--- a/src/procedure/nodes/sum1d.cpp
+++ b/src/procedure/nodes/sum1d.cpp
@@ -24,6 +24,7 @@ Sum1DProcedureNode::Sum1DProcedureNode(std::shared_ptr<Process1DProcedureNode> t
     keywords_.add<BoolKeyword>("Control", "RangeCEnabled", "Whether the second summation region is enabled", rangeEnabled_[2]);
     keywords_.add<RangeKeyword>("Control", "RangeC", "X range for third summation region", range_[2],
                                 Vec3Labels::MinMaxDeltaLabels);
+    keywords_.add<BoolKeyword>("Control", "Instantaneous", "Calculate instantaneous sums rather than averages", instantaneous_);
 }
 
 /*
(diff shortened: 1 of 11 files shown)


src/procedure/nodes/sum1d.cpp
```diff
@@ -93,7 +94,10 @@ bool Sum1DProcedureNode::finalise(const ProcedureContext &procedureContext)
                 sum_[i] = procedureContext.dataList().realise<SampledDouble>(
                     fmt::format("Sum1D//{}//{}", name(), rangeNames[i]), procedureContext.dataPrefix(),
                     GenericItem::InRestartFileFlag);
-            sum_[i]->get() += Integrator::sum(sourceData_->processedData(), range_[i]);
+            if (instantaneous_)
+                sum_[i]->get() = Integrator::sum(sourceData_->processedData(), range_[i]);
+            else
+                sum_[i]->get() += Integrator::sum(sourceData_->processedData(), range_[i]);
```
rprospero, line 100:
```suggestion
            sum_[i]->get() = Integrator::sum(sourceData_->processedData(), range_[i]) + (instantaneous_ ? 0 : sum_[i]->get());
```

A possible suggestion for reducing some duplication.

rprospero (approved):

src/procedure/nodes/sum1d.cpp
trisyoungs: This unfortunately loses information as the `=` operator resets the `SampledDouble` to a single data point (i.e. count of 1) while the `+=` accumulates the average and standard deviation while absorbing the value.

PR #1223, opened 2022-08-19
Toml configuration

rprospero:
This is the current work in progress on adding Toml support for the configuration part of the input file.  Currently, only Add and Box parts are supported.  The Add node will currently fail if a CoordinateSet or region is used.  However, I wanted to make this available so that the design decisions could be checked before we get too deep into this.


rprospero:
A couple of deisgn issues worth mentioning:

1) I have a factory function in `generator.h` that creates ProcedureNodes of the correct type out of a Toml node.  It handles this by parsing the type and then using a giant switch statement that creates the correct type of object then calls deserialise on that object.  However, this function should probably be namespaced somewhere.

2) Deserialising Add nodes requires access to the CoreData in order to pull in the species pointer.  However, the Serialisable class doesn't have this extra parameter, so the ProcedureNodes can't be a member of Serialisable.  It largely doesn't matter, since the templates handle the rest.  That said, this isn't the first time that this has come up (e.g. SpeciesAngle).  It would be possible to make the Serialisable class templated on the type of the extra context.

rprospero:
Superceded by #1256

PR #1224, opened 2022-08-22
Procedure Editing - Part 4

trisyoungs:
Here we work on one of the core structural pieces of `Procedure` assembly - the `SequenceProcedureNode` - in order to promote tighter ownership and more facile usage going forward.

Summary of changes:
- The sequence nodes previously stored a pointer to the parent `Procedure`, meaning that constructing sequence nodes (branches) required this knowledge. In practice, this information was used only sparingly, and was trivially replaced by using scope instead.
- `SequenceProcedureNode` derived from `ProcedureNode`, requiring explicit instantiation via `std::make_shared`. For branches, this was only performed when they were required (e.g. on deserialise) making it difficult to manage this information within a MVC setting. This PR moves the sequence class to a non-`ProcedureNode` class to be used as a member variable.
- The new `ProcedureNodeSequence` still requires knowledge of the owner node which was previously was stored as a `shared_ptr`. However, since `PrcedureNodeSequence` is now used as a member variable this requires the use of `shared_from_this()` in the parent node's constructor, which is invalid. Hence, here a reference to the parent node is stored instead.
- Since the `owner_` of a `ProcedureNodeSequence` is now stored as a reference, the `ProcedureNode::parent()` function now returns a raw `ProcedureNode` pointer rather than a `shared_ptr`. This is not ideal behaviour, but is the easiest solution to the problem cited above.

Works towards #1176.


rprospero (approved):
This is clearer to follow than what we had before.  I'll probably have to cherrypick this into the TOML branch.

trisyoungs:
> I'll probably have to cherrypick this into the TOML branch.

Yeah, sorry about that!

PR #1225, opened 2022-08-25
Procedure Editing - Part 5

trisyoungs:
This PR adds in full editability (addition of new nodes, rearrangement of existing nodes) within a `Procedure` via the `ProcedureModel`.

Additional testing / checking is also added.

Works towards #1176.

### DIFF
--- a/src/gui/models/nodePaletteModel.cpp
+++ b/src/gui/models/nodePaletteModel.cpp
@@ -2,6 +2,7 @@
 // Copyright (c) 2022 Team Dissolve and contributors
 
 #include "gui/models/nodePaletteModel.h"
+#include "gui/models/procedureModelMimeData.h"
 #include "procedure/nodes/registry.h"
 #include <QIODevice>
 #include <QIcon>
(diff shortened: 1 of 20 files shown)


src/gui/models/nodePaletteModel.cpp
```diff
@@ -0,0 +1,117 @@
...
+        auto [nodeType, brief] =
+            std::next(ProcedureNodeRegistry::categoryMap().begin(), index.parent().row())->second[index.row()];
+        if (role == Qt::DisplayRole)
+            return QString::fromStdString(std::string(ProcedureNode::nodeTypes().keyword(nodeType)));
+        else if (role == Qt::ToolTipRole)
+            return QString::fromStdString(brief);
+        else if (role == Qt::DecorationRole)
+            return QIcon((QPixmap(
+                QString(":/nodes/icons/nodes_%1.svg")
+                    .arg(QString::fromStdString(std::string(ProcedureNode::nodeTypes().keyword(nodeType))).toLower()))));
```
rprospero, line 67: Might this be cleaner as a `switch` statement?

src/gui/models/procedureModel.cpp
```diff
@@ -102,6 +126,63 @@ QVariant ProcedureModel::data(const QModelIndex &index, int role) const
...
 }
 
+bool ProcedureModel::setData(const QModelIndex &index, const QVariant &value, int role)
+{
+    if (!procedure_)
+        return false;
+
+    if (role == Qt::EditRole)
+    {
+        auto *node = static_cast<ProcedureNode *>(index.internalPointer());
```
rprospero, line 136:
```suggestion
        auto *node = rawData(index);
```
I like your `rawData` function and it seems a shame not to use it.

rprospero (approved):
I can't imagine what a pain the drag and drop bits where.  I've got a couple of minor tweaks, but everything looks good.

src/gui/models/nodePaletteModel.cpp
trisyoungs: Yes, I believe it would.  We use `if`/`switch` quite interchangeably throughout all the GUI models, and of course we should be consistent.  I'll add an issue to go back and change them all.

trisyoungs: #1241

PR #1230, opened 2022-08-26
1219 IntraShake Restrict to Species

### LINKED ISSUE #1219: Restrict IntraShake to specific species types
As with the `MD` module, the ability to restrict the `IntraShake` module to a subset of molecules would be useful.

### LINKED ISSUE #1229: IntraShake destroys cyclic molecules
**Describe the bug**
Adjusting the torsions with `IntraShake` causes terms within cycles to be adjusted, destroying the geometry of the molecule.

trisyoungs:
This PR adds the option to restrict the `IntraShake` module to a subset of species, and also fixes an issue with the adjustment of torsions within cycles.

Closes #1219.
Closes #1229.

### DIFF
--- a/src/modules/intrashake/process.cpp
+++ b/src/modules/intrashake/process.cpp
@@ -281,7 +281,7 @@ bool IntraShakeModule::process(Dissolve &dissolve, const ProcessPool &procPool)
 
                         // Trial the transformed Molecule
                         delta = (newPPEnergy + newIntraEnergy) - (ppEnergy + intraEnergy);
-                        accept = delta < 0 ? true : (randomBuffer.random() < exp(-delta * rRT));
+                        accept = delta < 0 || (randomBuffer.random() < exp(-delta * rRT));
 
                         // Accept new (current) positions of the Molecule's Atoms?
                         if (accept)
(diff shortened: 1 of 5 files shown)


src/modules/intrashake/process.cpp
```diff
@@ -59,6 +63,20 @@ bool IntraShakeModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
+        std::vector<int> targetIndices;
+        auto id = 0;
+        for (const auto &mol : targetConfiguration_->molecules())
+        {
+            if (std::find(restrictToSpecies_.begin(), restrictToSpecies_.end(), mol->species()) != restrictToSpecies_.end())
+                targetIndices.push_back(id);
+            ++id;
+        }
+        distributor.setTargetMolecules(targetIndices);
+    }
```
rprospero, line 78: There's a way of doing this with algorithms, but it's very ugly and confusing.  It gets clearer with C++20 ranges, so we'll just put that off until some future date when we upgrade.

rprospero (approved):

src/modules/intrashake/process.cpp
trisyoungs: I felt that there should be a better way when I copy/pasted this bit in.  Roll on C++20.

PR #1232, opened 2022-08-26
1228 Tab Signalling

### LINKED ISSUE #1228: Tabs don't change while Dissolve is running
**Describe the bug**
Fixing #985 via #1220 has introduced an issue where the tab contents do not change when Dissolve is running. Obviously this is due to all signals being blocked, so we need a more subtle mechanism to achieve data safety.

trisyoungs:
Quick PR to fix an issue introduced in #1220 whereby signals were blocked in the main tabs widget when a simulation was running. However, this prevented switching between tabs.

Closes #1228.

### DIFF
--- a/src/gui/maintabswidget.hui
+++ b/src/gui/maintabswidget.hui
@@ -104,6 +104,10 @@ class MainTabsWidget : public QTabWidget
     /*
      * Update
      */
+    private:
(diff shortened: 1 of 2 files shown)


DanNixon (approved):
:+1:

PR #1233, opened 2022-08-26
896 OSX dmgbuild

### LINKED ISSUE #896: Building dmg files broken in Big Sur
Can't move to Big Sur until fix or workaround for `dmgbuild` / `hdiutil` bug is resolved (#895 fixes pipelines to Catalina in the interim).

### LINKED ISSUE #1226: OSX 10.15 Deprecation
Mac OS 10.15 will be fully deprecated on 30th September 2022 - see https://github.com/actions/runner-images/issues/5583.

Unfortunately we rely on this version for packaging - see #896.

trisyoungs:
Here we address errors in dmg creation on the latest OSX versions. The root of the problem seems to be the addition of a license to the generated dmg, which requires the native OSX utility `hdiutil` to use a now-deprecated keyword.

Simply not adding a license to the dmg solves the issue, but how do we feel about this?

Closes #1226.
Closes #896.

### DIFF
--- a/.azure-pipelines/pr.yml
+++ b/.azure-pipelines/pr.yml
@@ -60,7 +60,7 @@ stages:
         displayName: "OSX, Threads, CLI/GUI"
         timeoutInMinutes: 120
         pool:
-          vmImage: "macOS-10.15"
+          vmImage: "macos-latest"
         steps:
           - checkout: self
             fetchDepth: 1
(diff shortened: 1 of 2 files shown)


rprospero (approved):
Since we're GPL, I'm almost happier NOT having the licence click through.

PR #1234, opened 2022-08-27
Unknown Module Types

trisyoungs:
Here's a tiddly tiny PR to address a UX issue and update the example files. If an input file was loaded which referenced an unknown module type, the GUI would complain that the file failed to load but didn't say why, and the CLI version would throw an exception. Now we report the error and exit nicely.

Also fixes two example input files which still referenced the recently-removed `CalculateCN` module (#1222).


DanNixon (approved):
:+1:

PR #1253, opened 2022-10-10
1227 Reference data G(r)

### LINKED ISSUE #1227: Fourier transform of reference data not visible
**Describe the bug**

In the `NeutronSQ` module (and presumably `XRaySQ` as well) occasionally users notice that the graph displaying the FT of the supplied reference data (i.e. the F(r)) is "broken" in the sense that nothing appears at all.  Internally, this can be causes by NaNs in the data.

I suspect that the issue is related to how we now handle `Configuration` generation - previously this was always performed, but an empty configuration will return a zero atomic number density.  If this is used in the FT then NaNs wlll occur and may be the cause of the issue.

trisyoungs:
This PR addresses #1227, which is a bug caused by a previous PR removing the automatic generation of configuration contents on file load.  Thus, when any module read in from the file tries to Fourier transform associated reference data, the density value retrieved from the associated `Configuration` is zero, leading to divide-by-zero errors in the Fourier transform and subsequent death of any graphed values.

We solve this issue by returning `std::optional<double>` from the relevant density routines, allowing checks against the validity of the value to be made.

Closes #1227.

### DIFF
--- a/src/gui/configurationtab_funcs.cpp
+++ b/src/gui/configurationtab_funcs.cpp
@@ -100,8 +100,11 @@ void ConfigurationTab::updateDensityLabel()
     if (!configuration_)
         ui_.DensityUnitsLabel->setText("N/A");
     else
-        ui_.DensityUnitsLabel->setText(QString::number(
-            ui_.DensityUnitsCombo->currentIndex() == 0 ? configuration_->atomicDensity() : configuration_->chemicalDensity()));
+    {
+        auto rho =
+            ui_.DensityUnitsCombo->currentIndex() == 0 ? configuration_->atomicDensity() : configuration_->chemicalDensity();
+        ui_.DensityUnitsLabel->setText(rho ? QString::number(*rho) : "--");
+    }
 }
 
 // Update controls in tab
--- a/src/modules/epsr/process.cpp
+++ b/src/modules/epsr/process.cpp
@@ -164,6 +164,8 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
     if (!targetConfiguration_)
         return Messenger::error("No target configuration is set.\n");
 
+    if (!targetConfiguration_->atomicDensity())
+        return Messenger::error("No density available for target configuration '{}'\n", targetConfiguration_->name());
     auto rho = *targetConfiguration_->atomicDensity();
 
     /*
--- a/src/modules/epsr/process.cpp
+++ b/src/modules/epsr/process.cpp
@@ -32,7 +32,7 @@ bool EPSRModule::setUp(Dissolve &dissolve, const ProcessPool &procPool, Flags<Ke
 {
     // Check for exactly one Configuration referenced through target modules
     targetConfiguration_ = nullptr;
-    auto rho = 0.0;
+    std::optional<double> rho;
     for (auto *module : targets_)
     {
         // Retrieve source SQ module, and then the related RDF module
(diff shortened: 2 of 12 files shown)


src/modules/epsr/process.cpp
```diff
@@ -162,6 +164,8 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
     if (!targetConfiguration_)
         return Messenger::error("No target configuration is set.\n");
 
+    auto rho = *targetConfiguration_->atomicDensity();
```
rprospero, line 167: I feel we should have a check here to return an error message if `rho` does not have a value.

src/modules/neutronsq/process.cpp
```diff
@@ -220,7 +225,7 @@ bool NeutronSQModule::process(Dissolve &dissolve, const ProcessPool &procPool)
     Filters::trim(repGR, referenceFTQMin_.value_or(0.0), ftQMax);
     auto rMin = weightedGR.total().xAxis().front();
     auto rMax = weightedGR.total().xAxis().back();
-    auto rho = rdfModule->effectiveDensity();
+    auto rho = *rdfModule->effectiveDensity();
```
rprospero, line 228: Again, we should return an error if rho is unset.

src/modules/rdf/functions.cpp
```diff
@@ -530,7 +540,7 @@ bool RDFModule::sumUnweightedGR(GenericList &processingData, const ProcessPool &
             fingerprint.empty() ? fmt::format("{}", cfg->contentsVersion()) : fmt::format("_{}", cfg->contentsVersion());
 
         // Calculate weighting factor
-        double weight = ((cfgWeight / totalWeight) * cfg->atomicDensity()) / rho0;
+        double weight = ((cfgWeight / totalWeight) * *cfg->atomicDensity()) / rho0;
```
rprospero, line 543: Error check

src/modules/rdf/functions.cpp
```diff
@@ -517,7 +527,7 @@ bool RDFModule::sumUnweightedGR(GenericList &processingData, const ProcessPool &
 
     // Calculate overall density of combined system
     double rho0 = std::accumulate(configWeights.begin(), configWeights.end(), 0.0, [totalWeight](double acc, auto pair) {
-        return acc + pair.second / totalWeight / pair.first->atomicDensity();
+        return acc + pair.second / totalWeight / pair.first->atomicDensity().value();
```
rprospero, line 530: Error check, but this is going to be a bit trickier, since it's inside the lambda.  My initial suggestion would be
```suggestion
        if (!acc) return {};
        auto density = pair.first->atomicDensity{};
        if (!density) return {};
        return *acc + pair.second / totalWeight / *density;
```

And then return an error message if the overall accumulate returns null.

src/procedure/nodes/add.cpp
```diff
@@ -329,8 +329,8 @@ bool AddProcedureNode::execute(const ProcedureContext &procedureContext)
         }
     }
 
-    Messenger::print("[Add] New box density is {:e} atoms/Angstrom**3 ({} g/cm3).\n", cfg->atomicDensity(),
-                     cfg->chemicalDensity());
+    Messenger::print("[Add] New box density is {:e} atoms/Angstrom**3 ({} g/cm3).\n", *cfg->atomicDensity(),
+                     *cfg->chemicalDensity());
```
rprospero, line 333: Switch these to `value_or` so that we can handle a missing rho

rprospero (changes requested):
Just some error handling bits with the optional.  I like the setup, but we might as well take advantage of the information that the optional is giving us.

src/gui/configurationtab_funcs.cpp
```diff
@@ -100,8 +100,11 @@ void ConfigurationTab::updateDensityLabel()
     if (!configuration_)
         ui_.DensityUnitsLabel->setText("N/A");
     else
-        ui_.DensityUnitsLabel->setText(QString::number(
-            ui_.DensityUnitsCombo->currentIndex() == 0 ? configuration_->atomicDensity() : configuration_->chemicalDensity()));
+    {
+        auto rho =
+            ui_.DensityUnitsCombo->currentIndex() == 0 ? configuration_->atomicDensity() : configuration_->chemicalDensity();
+        ui_.DensityUnitsLabel->setText(rho ? QString::number(*rho) : "--");
```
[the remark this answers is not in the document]
trisyoungs: Cool feature tho.

src/modules/rdf/functions.cpp
trisyoungs: Rather than "complicate" the `accumulate` call here, I've elected to perform a direct check in the preceding loop so that no lambda-based check is necessary.

PR #1256, opened 2022-11-02
Toml keywords

rprospero:
This PR adds support for Procedures and Modules to TOML.  This completes the last of the updates needed for round trip support.  Importantly, it does *NOT* fully implement TOML, as the TOML loaded version does not return the same results as the input file version on all the test files.


src/keywords/enumoptions.h
```diff
@@ -53,6 +53,7 @@ template <class E> class EnumOptionsKeyword : public EnumOptionsBaseKeyword
     private:
     // Reference to data
     E &data_;
+    const E default_;
```
trisyoungs, line 56:
```suggestion
    // Initial value
    const E default_;
```

src/keywords/geometrylist.cpp
```diff
@@ -77,3 +78,71 @@ bool GeometryListKeyword::serialise(LineParser &parser, std::string_view keyword
...
+            case Geometry::GeometryType::AngleType:
+                geo.set(value, indices[0], indices[1], indices[2]);
+                break;
+            case Geometry::GeometryType::DistanceType:
+                geo.set(value, indices[0], indices[1]);
+                break;
+            case Geometry::GeometryType::TorsionType:
+            default:
+                geo.set(value, indices[0], indices[1], indices[2], indices[3]);
+                break;
```
trisyoungs, line 141: This should really be an explicit `set()` for `Geometry::GeometryType::TorsionType` and an explicit error in the `default` clause.

src/keywords/geometrylist.h
```diff
@@ -19,6 +19,8 @@ class GeometryListKeyword : public KeywordBase
     private:
     // Reference to vector of data
     std::vector<Geometry> &data_;
+    // Initial Value
+    const std::vector<Geometry> default_;
```
trisyoungs, line 23: Since this keyword is a user-controlled one which always defaults to being empty, you don't really need to store the `default_` state and check against it.

src/keywords/integer.h
```diff
@@ -19,6 +19,8 @@ class IntegerKeyword : public KeywordBase
     private:
     // Reference to target data
     int &data_;
+    // Initial value
+    int default_;
```
trisyoungs, line 23:
```suggestion
    const int default_;
```

src/keywords/store.cpp
```diff
@@ -188,3 +188,34 @@ std::vector<Module *> KeywordStore::getVectorModule(std::string_view name) const
...
+{
+    auto result = std::string(original);
+    result[0] = tolower(result[0]);
+    return result;
+}
+
+// Apply the terms in the keyword store to a node
+SerialisedValue KeywordStore::serialiseOnto(SerialisedValue node) const
+{
+
```
trisyoungs, line 204:
```suggestion
```

src/keywords/vec3integer.h
```diff
@@ -23,10 +23,14 @@ class Vec3IntegerKeyword : public KeywordBase
     private:
     // Reference to target data
     Vec3<int> &data_;
+    // Initial Value
+    Vec3<int> default_;
```
trisyoungs, line 27:
```suggestion
    const Vec3<int> default_;
```

src/keywords/vec3nodevalue.h
```diff
@@ -23,6 +23,8 @@ class Vec3NodeValueKeyword : public KeywordBase
      * Data
      */
     private:
+    // Initial value
+    Vec3<NodeValue> default_;
```
trisyoungs, line 27:
```suggestion
    const Vec3<NodeValue> default_;
```

src/keywords/geometrylist.cpp
```diff
@@ -77,3 +78,71 @@ bool GeometryListKeyword::serialise(LineParser &parser, std::string_view keyword
...
+        }
+        data_.push_back(geo);
+    }
+}
+
+// Has not changed from initial value
+bool GeometryListKeyword::isDefault() const { return data_ == default_; }
```
trisyoungs, line 148:
```suggestion
bool GeometryListKeyword::isDefault() const { return data_ .empty(); }
```

src/keywords/vector_intdouble.cpp
```diff
@@ -84,3 +85,21 @@ bool IntegerDoubleVectorKeyword::serialise(LineParser &parser, std::string_view
...
+
+// Read values from a tree node
+void IntegerDoubleVectorKeyword::deserialise(const SerialisedValue &node, const CoreData &coreData)
+{
+    for (auto item : node.as_array())
+        data_.emplace_back(toml::find<std::vector<int>>(item, "indices"), toml::find<std::vector<double>>(item, "values"));
+}
+
+// Has not changed from initial value
+bool IntegerDoubleVectorKeyword::isDefault() const { return KeywordBase::isDefault(); }
```
trisyoungs, line 105:
```suggestion
bool IntegerDoubleVectorKeyword::isDefault() const { return data_.empty(); }
```

src/math/data2d.cpp
```diff
@@ -419,3 +419,23 @@ bool Data2D::serialise(LineParser &parser) const
...
+
+// Read values from a tree node
+void Data2D::deserialise(const SerialisedValue &node)
+{
+    tag_ = toml::find<std::string>(node, "tag");
+    x_ = toml::find<std::vector<double>>(node, "x");
+    y_ = toml::find<std::vector<double>>(node, "y");
+    values_.linearArray() = toml::find<std::vector<double>>(node, "values");
+    if (node.contains("errors"))
+        errors_.linearArray() = toml::find<std::vector<double>>(node, "errors");
```
trisyoungs, line 440: Need to be careful here - the `values_` and `errors_` are of type `Array2D<double>`, and which needs to be initialised to the correct dimensions first otherwise element lookup will not work correctly.

src/module/layer.cpp
```diff
@@ -185,4 +185,36 @@ std::vector<Configuration *> ModuleLayer::allTargetedConfigurations() const
...
+
+// Read values from a tree node
+void ModuleLayer::deserialise(const SerialisedValue &node, const CoreData &coreData)
+{
+    frequency_ = toml::find_or<int>(node, "frequency", 1);
+    if (toml::find_or<bool>(node, "disabled", false))
+        runControlFlags_.setFlag(ModuleLayer::RunControlFlag::Disabled);
+    if (toml::find_or<bool>(node, "requireEnergyStability", false))
+        runControlFlags_.setFlag(ModuleLayer::RunControlFlag::EnergyStability);
+    if (toml::find_or<bool>(node, "requireSieFactors", false))
```
trisyoungs, line 213:
```suggestion
    if (toml::find_or<bool>(node, "requireSizeFactors", false))
```

src/procedure/nodes/generator.cpp
```diff
@@ -0,0 +1,158 @@
...
+NodeRef nodeGenerator(const SerialisedValue &node, const std::string &name, const CoreData &data,
+                      const std::shared_ptr<ProcedureNode> parent)
+{
+    NodeRef result;
+    ProcedureNode::NodeType type = ProcedureNode::nodeTypes().deserialise(node.at("type"));
+    switch (type)
+    {
+        case ProcedureNode::NodeType::Add:
+            result = std::make_shared<AddProcedureNode>();
+            break;
```
trisyoungs, line 48: Can we leverage the `ProcedureNodeRegistry` class to do this?

src/classes/configuration.cpp
```diff
@@ -155,8 +155,16 @@ SerialisedValue Configuration::serialise() const
...
     return configuration;
 }
+
+// Read values from a tree node
+void Configuration::deserialise(const SerialisedValue &node, const CoreData &data)
+{
+    setTemperature(toml::find_or<double>(node, "temperature", defaultTemperature_));
+    requestedSizeFactor_ = toml::find_or<double>(node, "sizeFactor", defaultSizeFactor_);
+    requestedCellDivisionLength_ = toml::find_or<double>(node, "cellDivisionLength", defaultCellDivisionLength_);
+    generator_.deserialise(node.at("generator"), data);
```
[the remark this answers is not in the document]
rprospero: If there is no "generator", then the toml library will throw an exception complaining about the missing value.  This exception provides some extra context, such as the location within the file where the missing key was expected.

I know that we generally avoid exceptions in Dissolve, but the deserialisation routine was found to be an excellent case for them.  A single `try` block can wrap the reading of the input file and perform the correct clean-up on an invalid file (i.e. exiting with an error on the command line and displaying a message to the user in the GUI).  This quarantines the exceptions from appearing elsewhere in the program.

src/classes/data1dstore.cpp
```diff
@@ -43,3 +44,27 @@ OptionalReferenceWrapper<const Data1D> Data1DStore::data(std::string_view name)
 
 // Return vector of all data
 const std::vector<std::shared_ptr<std::pair<Data1D, Data1DImportFileFormat>>> &Data1DStore::data() const { return data_; }
+
+// Express as a tree node
```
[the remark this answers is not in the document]
rprospero: That does feel easier to read.  I've pushed a change to enact this.

src/classes/isotopologue.cpp
```diff
@@ -88,15 +88,19 @@ const std::vector<std::tuple<std::shared_ptr<AtomType>, Sears91::Isotope>> &Isot
 SerialisedValue Isotopologue::serialise() const
 {
     SerialisedValue isotopologue;
```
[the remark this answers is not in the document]
rprospero: Agreed.

src/keywords/configuration.cpp
```diff
@@ -55,3 +55,16 @@ void ConfigurationKeyword::removeReferencesTo(Configuration *cfg)
     if (data_ == cfg)
         data_ = nullptr;
 }
+
+SerialisedValue ConfigurationKeyword::serialise() const
+{
+    // isDefault is checked before serialisation of keywords
+    // so we have checked for the null pointer
```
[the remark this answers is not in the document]
rprospero: I've left in the comment, but added the assertion.

src/math/data2d.cpp
rprospero: Good catch. We now initialise the arrays before setting the values.

PR #1263, opened 2022-11-14
EPSR Coefficient Smoothing

trisyoungs:
A small PR to change the way EPSR smooths coefficients used in the potential generation. Previously this was performed just prior to potential generation, but this was not terribly good at eliminating noise from being carried over into the potentials. Now the smoothing is performed after the delta F(Q) fitting, and the smoothed parameters are stored for posterity.

### DIFF
--- a/src/modules/epsr/process.cpp
+++ b/src/modules/epsr/process.cpp
@@ -139,8 +139,7 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
     Messenger::print("EPSR: Range for potential generation is {} < Q < {} Angstroms**-1.\n", qMin_, qMax_);
     Messenger::print("EPSR: Weighting factor used when applying fluctuation coefficients is {}\n", weighting_);
     if (fluctuationSmoothing_)
-        Messenger::print("EPSR: Fluctuation coefficients will be smoothed (average length = 2N+1, N = {})",
-                         *fluctuationSmoothing_);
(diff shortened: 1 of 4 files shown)


src/modules/epsr/process.cpp
```diff
@@ -636,23 +644,9 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
-                                        // Add in fluctuation coefficients
-                                        for (auto n = 0; n < nCoeffP_; ++n)
-                                            potCoeff[n] += weighting_ * smoothedCoefficients.value(n);
-                                    }
-                                    else
-                                        for (auto n = 0; n < nCoeffP_; ++n)
-                                            potCoeff[n] += weighting_ * fluctuationCoefficients[{i, j, n}];
+                                    // Add in fluctuation coefficients
+                                    for (auto n = 0; n < nCoeffP_; ++n)
+                                        potCoeff[n] += weighting_ * fluctuationCoefficients[{i, j, n}];
```
rprospero, line 649: This sent me down a bit of a rabbit hole.  We group our Array3D by the first coordinate, then the second, then the third.  If we grouped by the third coordinate, then it would be possible to make an overload `fluctuationCoefficients[{i, j}]` that returned the vector at that coordinate.  This could enable some SIMD optimizations that aren't currently possible and also lead to some cleaner code.  It should also have much better cache locality.

I'm going to mark a separate issue for trying this out and checking the benchmarks to see if this does produces a performance benefit.

rprospero (approved):

PR #1273, opened 2022-11-23
Interpolator Speed

trisyoungs:
This PR improves the speed of the `Interpolator` class when calculating and returning interpolated values at a given x. Previously, the interval in which the target x value was determined with a binary chop - this is fine, but adds significant overhead to the process.  In common use cases the supplied x values are reliably increasing (i.e. it is not random access).

Here we add a quick check on the last used interval in the interpolator, and try to trivially determine the required valid interval and avoid the binary chop wherever possible.  A unit test is added.  Anecdotally, this provides at least a 50% speed up when providing x-ordered values.


rprospero (approved):
Everything looks good.  However, I'm thinking that we need to add another overloay for the `y` function that takes a vector of doubles and returns a vector of interpolated doubles.  There's multiple places where this comes up within the class itself.  It might also allow us to simplify some of the logic since we can analyse the full set of input coordinates before doing the calculations (e.g. check if the array is monotonically increasing)

Issue #1268, opened 2022-11-16
Reverse index ordering on Array3D


rprospero:
Currently, the Array3D gives the tightest grouping to the first coordinate.  However, most of our inner loops are over the final coordinate.  This leads to two problems

- [X] We're pulling out disparate array elements, thrashing the cache.  Having the tightest grouping on the final coordinate would keep the memory together.  It might also enable some SIMD operations.

- [x] It would enable us to implement an operator `A[{i, j}]` that returned an iterator that would walk through the elements of the final coordinate.  This would enable using standard algorithms on the terms inside the array.  These would help enabling the SIMD options previously discussed.

PR #1263 provides a good example of where this could be used, but there are likely to be many more locations through the code.  Still, this should only be performed alongside benchmarks that show that there is no loss of performance (and preferably a gain).

PR #1275, opened 2022-11-28
Scattering matrix inversion

### LINKED ISSUE #1270: Potential generation is unusably slow for modest system complexities
**Describe the bug**
For a fairly modest system complexities of around 16 atom types it was noticed that the EPSR module takes perhaps a day to generate the potentials, appearing to hang, but it's just painfully slow.

The reason boils down to the continual, repeated inversion of the 16x16 scattering matrix, which is performed for each atom type pair on each pair potential. For smaller system sizes this is negligible compute, but there is clearly a limit where it becomes a severe bottleneck.

The solution has to be to precalculate the necessary inversion matrices and store them somehwere - simply optimising out the inversion is not viable since the problem is compounded even further when considering Q-dependent scattering matrices (i.e. for X-ray) although the potentials are generated with the Q = 0 weightings. Nevertheless, this will still slow up earlier parts of the EPSR process.

trisyoungs:
This PR addresses #1270 which stems from the fact that, at present, scattering matrices from the functions `ScatteringMatrix::matrix(double)` and `ScatteringMatrix::inverse(double)` are generated on the fly each call, which quickly becomes a significant performance overhead for modestly-sized systems (in terms of atom types).

Here we change the behaviour of the `ScatteringMatrix` class to remove that bottleneck entirely, simply by precalculating and storing the matrices in the `ScatteringMatrix` object.  This information is generated on first use and then remains static.

Closes #1270.

### DIFF
--- a/src/modules/epsr/process.cpp
+++ b/src/modules/epsr/process.cpp
@@ -529,6 +529,10 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
             return EarlyReturn<bool>::Continue;
         });
 
+    // If the scattering matrix was not set-up, need to generate the necessary inverse matrix or matrices here
+    if (!scatteringMatrixSetUp)
+        scatteringMatrix_.generateMatrices();
+
     scatteringMatrix_.print();
 
     if (Messenger::isVerbose())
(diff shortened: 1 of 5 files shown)


src/modules/epsr/process.cpp
```diff
@@ -611,31 +622,32 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
+                 */
+
+                // In the original EPSR the off-diagonal elements in the inverse matrix have also been
+                // halved so as not to double-count the i != j terms
+                if (i != j)
+                    weight *= 0.5;
+
+                // Store fluctuation coefficients ready for addition to potential coefficients later on.
+                for (auto n = 0; n < nCoeffP_; ++n)
+                    fluctuationCoefficients[{i, j, n}] += weight * feedback_ * fitCoefficients[n];
```
rprospero, line 649: This looks to be a good target for some of the Array3D work that we're planning in #1268.

PR #1279, opened 2022-12-09
Update LLVM Repo

trisyoungs:
This PR bumps the version of `clang-format` we use from `7` to `13`. `13` is not the most recent release, but it seems to be one which is consistently available across recent Linux distributions. In addition, version `7` is not available for Ubuntu 22.04 which will become the default version in pipelines over Azure and GitHub Actions soon.

There are, of course, some subtle formatting changes that this enforces within the code base, but at around ~1k lines it isn't _too_ bad.


DanNixon (approved):
:+1: 

Just curious, what was the reason for going with the Ubuntu repos rather than the LLVM repo?

trisyoungs:
@DanNixon - Just seemed sensible to use the distro-packaged version at the time.  Admittedly this version-locks us to whatever Ubuntu decide to distribute.  You think it's wiser to explicitly select the version from the LLVM repos instead?

DanNixon:
Only if you cared about newer versions, if the version in the Ubuntu repo is new enough then that's fine.

PR #1286, opened 2023-01-09
NETA Identifiers

trisyoungs:
This PR provides additional functionality to the NETA typing, allowing text identifiers to be assigned to atoms as they are matched. This contextual information can then be used to infer axis systems, for example, in other workflows in the code.

### DIFF
--- a/src/neta/matchedpath.cpp
+++ b/src/neta/matchedpath.cpp
@@ -36,3 +36,6 @@ void NETAMatchedPath::merge(const NETAMatchedPath &other)
         for (auto &i : atoms)
             identifiers_[key].insert(i);
 }
+
+// Add identifier for specified atom
+void NETAMatchedPath::addIdentifier(const SpeciesAtom *i, std::string id) { identifiers_[id].insert(i); }
--- a/src/neta/matchedpath.cpp
+++ b/src/neta/matchedpath.cpp
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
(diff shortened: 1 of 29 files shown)


rprospero (approved):
I've included one small design note, but it looks good otherwise.

src/neta/matchedpath.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+// Return whether the specified atom is the root atom in the path
+bool NETAMatchedPath::isRoot(const SpeciesAtom *i) const { return !path_.empty() && (path_.front() == i); }
+
+// Merge supplied path into this one, appending new atoms and copying any relevant identifiers
+void NETAMatchedPath::merge(const NETAMatchedPath &other)
+{
+    // Copy new, unique atoms to our path
+    std::copy_if(other.path_.begin(), other.path_.end(), std::back_inserter(path_),
+                 [&](const auto *j) { return std::find(path_.begin(), path_.end(), j) == path_.end(); });
```
rprospero, line 32:
```suggestion
    std::copy(other.path_.begin(), other.path_.end(), std::inserter(path_, path_.begin()));
```

Changing `path_` from `std::vector` to `std::set` would clean up the code here and on the `contains` function.  We don't seem to depend on the order of the elements and I can't find any places outside the class implementation where changes would be necessary, but I might be missing something.

trisyoungs: I have now made this change - we do make one "positional" distinction for atoms, and that is the so-called root atom (where the matching process started off from). This is dealt with by a specific member variable, which actually makes more sense anyway.  At the same time I have renamed the class and some of its functions to reflect better what the class contains (i.e. it is no longer any kind of path).

PR #1287, opened 2023-01-09
Nixos 22.11

rprospero:
Update the nix file to build off the latest release.  This release has made changes to the Qt infrastructure, so our old modification of Qt for IDAaaS didn't work.  We're now manually building Qt, but we can drop this entire infrastructure after the next kernel upgrade.


nix/qt/fetch.sh
```diff
@@ -0,0 +1 @@
+WGET_ARGS=( https://download.qt.io/official_releases/qt/6.4/6.4.1/submodules/ -A '*.tar.xz' )
```
trisyoungs, line 1: Just a general comment, but rather than hardcoding a Qt version here (which incidentally is higher than the 6.3.something required by the rest of the CI) we should probably modify it in-place in the GitHub action so we conform to the same version for all builds.

trisyoungs (approved):
I mean, it's hard to say which is larger - my hatred for us having to do this because Qt are a law unto themselves, or my appreciation for you having gone through it all on our behalf!

:+1:

nix/qt/fetch.sh
rprospero: That's a very good point.  I'm going to try and role the version back to 6.3.0 and make sure it doesn't break anything.  I'm putting this back on draft until I've made the update.

PR #1289, opened 2023-01-12
GitHub actions nix

rprospero:
This PR adds the nix support to the GitHub Actions, enabling the Linux builds.  The following need to occur before we can merge.

- [X] The linux version is built in GitHub Actions
- [x] `qtbase` is not rebuilt for every GUI build
- [x] Nix dependencies are cached between runs


rprospero:
The Mac packaging fixed the network error.  The Windows packaging is still failing with a DLL issue, but that seems to be in the original branch (which doesn't reach the packaging phase because the linux build is broken there).

trisyoungs (approved):
:+1:

PR #1298, opened 2023-01-20
Module Bits

trisyoungs:
Here we fix a an issue with the display of calculated coordination numbers in the GUI, related to the wrong names being used for the target values.

It also removes some additional redundant arguments to various calls to `fmt::format`, left over from when analysis modules could target multiple configurations (which was a looooong time ago....)

### DIFF
--- a/src/modules/calculate_angle/gui/calculateanglewidget_funcs.cpp
+++ b/src/modules/calculate_angle/gui/calculateanglewidget_funcs.cpp
@@ -121,25 +121,25 @@ void CalculateAngleModuleWidget::setGraphDataTargets(CalculateAngleModule *modul
         return;
 
     // Calculated A...B RDF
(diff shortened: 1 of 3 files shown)


rprospero (approved):
It's interesting that there's not safety check on passing excessive arguments to fmt.

trisyoungs:
Indeed, my initial thought as well. But there are long discussions on the interweb which detail at length why we are wrong.

This email and any attachments are intended solely for the use of the named recipients. If you are not the intended recipient you must not use, disclose, copy or distribute this email or any of its attachments and should notify the sender immediately and delete this email from your system. UK Research and Innovation (UKRI) has taken every reasonable precaution to minimise risk of this email or any attachments containing viruses or malware but the recipient should carry out its own virus and malware checks before opening the attachments. UKRI does not accept any liability for any losses or damages which the recipient may sustain due to presence of any viruses.

PR #1302, opened 2023-01-23
1274 Interpolation over full array speed increase

Danbr4d:
Added a test function loop to the interpolation function


rprospero:
[original.csv](https://github.com/disorderedmaterials/dissolve/files/10487944/original.csv)

I've attached an initial benchmark run that we can compare against as we iterate on the `y` function

trisyoungs:
Nice. Every 1% counts! :smile:

scripts/linux-benchmark.sh
```diff
@@ -0,0 +1,17 @@
...
+  exit 1
+fi
+
+echo performance | sudo tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor > /dev/null
+echo 1 | sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo > /dev/null
+
+$1 --benchmark_filter=Interpolator --benchmark_out_format=csv --benchmark_out=$2
+
+echo powersave | sudo tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor > /dev/null
+echo 0 | sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo > /dev/null
```
trisyoungs, line 17: If I was in a particularly non-destructive and polite scripting mood I would store the current contents of the two system values rather than assuming these defaults.

src/math/interpolator.cpp
```diff
@@ -474,3 +474,105 @@ void Interpolator::addInterpolated(Interpolator &source, Data1D &dest, double fa
     for (auto &&[x, y] : zip(dest.xAxis(), dest.values()))
         y += source.y(x) * factor;
 }
+
+// Add test function for benchmarks
+std::vector<double> Interpolator::y(std::vector<double> xs)
```
trisyoungs, line 479: We can pass in `xs` as a nice `const` reference to avoid copying it unnecessarily.
```suggestion
std::vector<double> Interpolator::y(const std::vector<double> &xs)
```

src/math/interpolator.h
```diff
@@ -56,6 +56,8 @@ class Interpolator
     double y(double x);
     // Return interpolated y value for supplied x, specifying containing interval
     double y(double x, int interval);
+    // Return interpolated y values for all supplied x values
+    std::vector<double> y(std::vector<double> xs);
```
trisyoungs, line 60:
```suggestion
    std::vector<double> y(const std::vector<double> &xs);
```

trisyoungs (approved):
A comment and a small suggestion, but otherwise fine. :)

src/math/interpolator.cpp
```diff
@@ -474,3 +474,105 @@ void Interpolator::addInterpolated(Interpolator &source, Data1D &dest, double fa
...
+
+            double vk0 = y_[lastInterval_];
+            double vk1 = y_[lastInterval_ + 1];
+            double vk2 = y_[lastInterval_ + 2];
+            double t1 = vk0 + (vk1 - vk0) * ppp;
+            double t2 = vk1 + (vk2 - vk1) * (ppp - 1.0);
+            result.push_back(t1 + (t2 - t1) * ppp * 0.5);
+        }
+
+        result.push_back(0.0);
```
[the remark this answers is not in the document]
Danbr4d: So on the original function that I copied over after the if statements there is a return 0.0. I assumed that then if that is the default if there is no interpolation scheme

src/math/interpolator.cpp
```diff
@@ -474,3 +474,105 @@ void Interpolator::addInterpolated(Interpolator &source, Data1D &dest, double fa
...
+                    right = i;
+                else
+                    lastInterval_ = i;
+            }
+        }
+
+        if (lastInterval_ < 0)
+            result.push_back(y_.front());
+
+        if (scheme_ == Interpolator::SplineInterpolation)
```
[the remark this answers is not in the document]
Danbr4d: yeah I thought this too but wasn't sure whether or not to make the change. Have done now

PR #1304, opened 2023-01-24
Cache Workflow

trisyoungs:
This PR adds a "Generate Cache" workflow which can be manually run on the `develop` branch in order to create caches that can be accessed by any future workflow, from any branch. Of course, this is one of those PRs that can't be realistically tested until it gets merged in to `develop`!

@rprospero - A question on the GitHub Action caches generated for nix: The `get-nix` action is called for each distinct `$target`, but the actual value of that variable doesn't seem to influence what happens in the action, just the final name of the resulting cache.  Is the cache the same regardless of CLI/GUI/threading/MPI etc. build, and can we reduce the number of caches we need to just one for all nix builds?

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -1,0 +1,31 @@
+name: Generate Caches
+
+on:
+  workflow_dispatch:
+
+env:
+  pythonVersion: 3.9
+  qtVersion: 6.2.2
+  nixVersion: 2.4
+  nixOSVersion: 22.11
+
+jobs:
+
+  Cache:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        target: [dissolve-gui]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: "Build (Cache Only) (${{ matrix.os }})"
+        uses: "./.github/workflows/build"
+        env:
+          CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_TOKEN }}
+        with:
+          target: ${{ matrix.target }}
+          cacheOnly: true
+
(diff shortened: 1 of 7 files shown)


rprospero:
@trisyoungs Each cache covers the same files, but they'll have different contents based on what dependencies get pulled during compilation.  More specifically, the `Linux-dissolve-gui-cache` will contain the rebuilt Qt, but the `Linux-dissolve-cache` will not.  In theory, the CLI version should be able to be built with the GUI cache, but I was concerned about the CLI run over-writing the GUI cache, eliminating the custom Qt, and requiring Qt to be rebuilt again the next time the GUI is called.

We can certainly try switching to the GUI cache for all the builds.  I'm about 80% sure that it would be fine, but I left it out of the original PR so we didn't have to worry about the other 20%.

.github/workflows/cache.yml
```diff
@@ -0,0 +1,31 @@
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        target: [dissolve-gui]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: "Build (Cache Only) (${{ matrix.os }})"
+        uses: "./.github/workflows/build"
+        env:
+          CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_TOKEN }}
```
rprospero, line 27: We should no longer need the CACHIX_AUTH_TOKEN.

rprospero (approved):
Just one small removal

trisyoungs:
> 



> @trisyoungs Each cache covers the same files, but they'll have different contents based on what dependencies get pulled during compilation. More specifically, the `Linux-dissolve-gui-cache` will contain the rebuilt Qt, but the `Linux-dissolve-cache` will not. In theory, the CLI version should be able to be built with the GUI cache, but I was concerned about the CLI run over-writing the GUI cache, eliminating the custom Qt, and requiring Qt to be rebuilt again the next time the GUI is called.
> 
> We can certainly try switching to the GUI cache for all the builds. I'm about 80% sure that it would be fine, but I left it out of the original PR so we didn't have to worry about the other 20%.

80% is high enough to risk it for me!  I've specified a single cache for nix builds (i.e. not named after the target) and I'll see what happens.

Issue #1161, opened 2022-07-06
Separation / sub-grouping of control keywords


trisyoungs:
**Describe the feature / issue**
All modules have a "Control" section which contains the bulk of important options in a single, unseparated list.

Implementing a read-only dummy keyword type that displays, for instance, a subheading and horizontal line in order to group keywords would make the UX significantly better.

PR #1305, opened 2023-01-24
1161 Keyword Subgroups

trisyoungs:
This PR focuses on #1161, allowing subgrouping of keywords for a given module into "sections" rather than in separate pages of a `QToolBox`.  This is attempt four at designing a suitable mechanism for doing so - briefly, keywords are stored in a `KeywordStore` in a straight vector with group/section information attached, rather than implementing a hierarchical tree network of keyword groups / sections (which formed the basis of attempts 1 and 2!). The disadvantage of this flat storage is that the actual organisation and group of keywords in the GUI must be determined from the keyword vector at runtime, but since this is only performed once when a given module / node widget is created, it feels like an acceptable compromise.

This does not close the related issue - in fact, this PR will fail all of the CI! This is because the changes herein enforce fairly broad modification to each and every module, and this is left for a separate PR based off this one, and whose CI results should be used to validate this PR.

Various aesthetic improvements have also been made along the way - previous and current look-and-feel are shown below. Note additional buttons at the top to reflect different groups of keywords, with each group having its own sections, and an "Advanced" group stashed away to the far right.

image


rprospero:
I do like moving the `Export` section to a tab at the top.  I felt it tended to get lost at the bottom of the window.

rprospero (approved):
Approved with #1308

PR #1307, opened 2023-01-25
Nix Import / Export

trisyoungs:
Retry nix caching using `nix-store` import/export functions, since the previous approach (permission changing) seemed not to work. Again this is one that needs merging in before proper testing can be performed.

@rprospero I think I ended up on the same webpage as you did on this one!  This solution was suggested way further down the thread.  Hopefully you hadn't tried / rejected it already.

### DIFF
--- a/.github/workflows/get-nix/action.yml
+++ b/.github/workflows/get-nix/action.yml
@@ -8,29 +8,22 @@ runs:
   using: "composite"
   steps:
 
-  - name: Create nix store
-    shell: bash
-    run: |
-      # Create with liberal rights, otherwise cache action will complain
-      # about permission errors.
-      sudo mkdir -p /nix/store
-      sudo chmod -R 777 /nix
-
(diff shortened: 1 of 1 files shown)


rprospero (approved):
I remember seeing a similar version of this, but it didn't correctly export the entire nix-store.  However, this seems to not suffer the same issues as that setup.

The one concern that I have is checking on the `cache-hit` variable.  Specifically, the command line version will produce a cache that doesn't contain Qt.  This will count as a cache hit, so the GUI build won't export the new cache that does have Qt.  What we may need to do (just once) is delete the caches, manually run the GUI build to ensure that we have the cache with all the dependencies that we need.

trisyoungs:
@rprospero  - Indeed, that's exactly the plan - run the `cache.yml` manually once to generate all the required caches, and which should then never be overwritten. We'll see if it works!

PR #1310, opened 2023-01-27
1092 - set parameters to correct size when setForm

### LINKED ISSUE #1092: Extraneous parameters left in functional form text
**Describe the bug**
When editing the functional form of an interaction in the GUI, changing the form to one with fewer parameters than the current one (e.g. going from `Cos2` to `Harmonic` angles) leaves the additional parameters unnamed and still in the generated string (i.e. `A=1.0 B=2.0 =3.0`).

Danbr4d:
Closes #1092 . When editing the functional form of an interaction in the GUI, changing the form to one with fewer parameters than the current one (e.g. going from Cos2 to Harmonic angles) leaves the additional parameters unnamed and still in the generated string (i.e. A=1.0 B=2.0 =3.0).

### DIFF
--- a/src/classes/interactionpotential.h
+++ b/src/classes/interactionpotential.h
@@ -37,11 +37,11 @@ template <class Functions> class InteractionPotential
         form_ = form;
         // Set parameters vector to correct size
         while (parameters_.size() != Functions::forms().minArgs(form_))
-        if (parameters_.size() < Functions::forms().minArgs(form_))
-            parameters_.push_back(0.0);
+            if (parameters_.size() < Functions::forms().minArgs(form_))
+                parameters_.push_back(0.0);
 
-        else
-            parameters_.pop_back();
+            else
+                parameters_.pop_back();
     }
     // Return functional form of interaction
     typename Functions::Form form() const { return form_; }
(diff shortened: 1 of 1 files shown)


src/classes/interactionpotential.h
```diff
@@ -32,7 +32,17 @@ template <class Functions> class InteractionPotential
...
     // Set functional form of interaction
-    void setForm(typename Functions::Form form) { form_ = form; }
+    void setForm(typename Functions::Form form)
+    {
+        form_ = form;
+        // Set parameters vector to correct size
+        while (parameters_.size() != Functions::forms().minArgs(form_))
+            if (parameters_.size() < Functions::forms().minArgs(form_))
+                parameters_.push_back(0.0);
+
```
trisyoungs, line 42:
```suggestion
```

src/classes/interactionpotential.h
```diff
@@ -32,7 +32,17 @@ template <class Functions> class InteractionPotential
 
     public:
     // Set functional form of interaction
-    void setForm(typename Functions::Form form) { form_ = form; }
+    void setForm(typename Functions::Form form)
+    {
+        form_ = form;
+        // Set parameters vector to correct size
+        while (parameters_.size() != Functions::forms().minArgs(form_))
+            if (parameters_.size() < Functions::forms().minArgs(form_))
```
trisyoungs, line 40: Ah! I think you can blame me for this one. The logic here is sound, but the use of the `minArgs()` function is the problem.  If you look in `src/base/enumoptions.h` on line 169 where this function is defined you'll see that it actually returns `std::optional<int>` rather than just an `int`.  So, if no minimum number of arguments is defined in the option, it returns `std::nullopt` (i.e. no `int` value at all) rather than zero (which is a specific minimum value).

So, our check on the number of parameters can actually end up comparing `parameters_size()` with `std::nullopt`, which doesn't end well and causes either an infinite `pop_back()` or an infinite `push_back()` loop.  Neither are good!

The solution is to use a default value (zero) when no minimum value is defined:
```suggestion
        while (parameters_.size() != Functions::forms().minArgs(form_).value_or(0))
            if (parameters_.size() < Functions::forms().minArgs(form_).value_or(0))
```

trisyoungs (approved):
:+1:

This would be a good opportunity to add the "Closes #XXX." line into the PR description text.

PR #1303, opened 2023-01-24
1295 Finalise GitHub Actions

trisyoungs:
This PR corrects the build, test, and package stages of the new GitHub Actions CI.

Publishing and web deployment to follow in separate PRs.

Works towards #1295.

Issue #1295, opened 2023-01-19
Complete GitHub Actions Integration


trisyoungs:
GitHub Actions integration requires the following in order to be considered complete:

- [x] Release automation action
- [x] Introduction of a Release pipeline
- [x] Introduction of a Continuous pipeline
- [x] Web publishing
- [x] Nix store caching
- [x] Automatic opening of PR following push to version update branch (`automation/release-version-X.X.X`)
- [x] Reimplementation of `ci/scripts/list-releases` as an action to be run as part of `release` CI? Note incorrect use of `${CURRENT_VERSION}` in that script in `release/0.9.12` branch - should be `${CURRENT_RELEASE}`.
- [x] Remove old Azure stuff
- [x] Fix paths for `osc-conan` cache (wrong slash)
- [x] Centralise core env variables into single file / action

PR #1312, opened 2023-01-31
1295 Nix Cache Tidying

trisyoungs:
This PR addresses an issue introduced in #1303 which used some "clever" commands to remove the build dissolve derivation and image from the nix store before storing the cache.  However, while this worked on a local machine it did not in the CI since local tests were made using a single-user install of nix (CI was multi-user).

Works towards #1295.

### DIFF
--- a/.github/workflows/build/linux/action.yml
+++ b/.github/workflows/build/linux/action.yml
@@ -19,6 +19,18 @@ runs:
     with:
       target: ${{ inputs.target }}
 
+  - name: Cache Nix Store
+    uses: actions/cache@v3
+    id: nix-cache
+    with:
+      path: /tmp/nixcache
+      key: ${{ runner.os }}-nix-cache
+
+  - name: Import Nix Store Cache
+    if: "steps.nix-cache.outputs.cache-hit == 'true'"
+    shell: bash
+    run: nix-store --import < /tmp/nixcache
+
   - name: Build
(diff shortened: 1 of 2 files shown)


.github/workflows/build/linux/action.yml
```diff
@@ -19,6 +19,18 @@ runs:
...
+    id: nix-cache
+    with:
+      path: /tmp/nixcache
+      key: ${{ runner.os }}-nix-cache
+
+  - name: Import Nix Store Cache
+    if: "steps.nix-cache.outputs.cache-hit == 'true'"
+    shell: bash
+    run: nix-store --import < /tmp/nixcache
+
```
rprospero, line 33: I'd be tempted to put these lines at the end of the `git-nix.yml` file, in case we need to use nix again in a different action.  On the other hand, since the clean-up isn't automatic, it's not nearly as critical

rprospero (approved):
One small suggestion, but otherwise looks good.

.github/workflows/build/linux/action.yml
trisyoungs: These commands were actually originally there, but I came to the realisation that this would only cache the basic nix store contents, and would contain nothing related to the actual build dependencies of dissolve(-gui).  The build stage thus seemed like the logical home for it.

PR #1313, opened 2023-02-01
1268-reverse-index-ordering-Array3D 

Danbr4d:
Currently, the Array3D gives the tightest grouping to the first coordinate. However, most of our inner loops are over the final coordinate.

### DIFF
--- a/src/templates/array3d.h
+++ b/src/templates/array3d.h
@@ -68,11 +68,11 @@ template <class A> class Array3D
         array_.resize(nX_ * nY_ * nZ_);
 
         // Create slice offsets array
-        sliceOffsets_.resize(nZ_);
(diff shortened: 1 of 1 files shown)


rprospero:
array3d

I've done some benchmarks on the changes from the indexing on Array 3d.  For array sizes up to 16 Mb, the difference is negligible.  This isn't that surprising, as the machine I'm performing the tests on has 12 Mb of processor cache, so memory fetches aren't a factor.  For the larger arrays, we consistently see a few percentages of speed-up.  Not bad for twelve lines of code.

I think that's it's probably best to merge this in now and put the second part of #1268 into its own PR.

rprospero (approved):

trisyoungs (approved):
:+1:

PR #1319, opened 2023-02-03
1246 Use optional numeric controls where appropriate

### LINKED ISSUE #1246: Use optional numeric controls where appropriate
The `OptionalIntegerKeyword` and `OptionalDoubleKeyword`s can be used in more places:

- [x] `RDFModule`, "Cutoff" keyword (eliminates "UseHalfCellWidth" bool)
- [ ] `EPSRModule`, potential limits
- [x] `MDModule`, energy, output, and trajectory frequencies

Danbr4d:
Closes #1246 
The OptionalIntegerKeyword and OptionalDoubleKeywords can be used in more places:

 RDFModule, "Cutoff" keyword (eliminates "UseHalfCellWidth" bool)
 EPSRModule, potential limits
 MDModule, energy, output, and trajectory frequencies

### DIFF
--- a/src/modules/epsr/process.cpp
+++ b/src/modules/epsr/process.cpp
@@ -286,8 +286,7 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
             GaussFit coeffMinimiser(deltaFQ);
 
             if (status == GenericItem::ItemStatus::Created)
-                fitError =
-                    coeffMinimiser.constructReciprocal(0.0, rmaxpt, ncoeffp, gSigma1_, nIterations, 0.01, false);
+                fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, ncoeffp, gSigma1_, nIterations, 0.01, false);
             else
             {
                 if (fitCoefficients.size() != ncoeffp)
@@ -312,8 +311,8 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
             PoissonFit coeffMinimiser(deltaFQ);
 
             if (status == GenericItem::ItemStatus::Created)
-                fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, ncoeffp, pSigma1_, pSigma2_, nIterations,
-                                                              0.1, false);
+                fitError =
+                    coeffMinimiser.constructReciprocal(0.0, rmaxpt, ncoeffp, pSigma1_, pSigma2_, nIterations, 0.1, false);
             else
             {
                 if (fitCoefficients.size() != ncoeffp)
@@ -321,8 +320,7 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
                     Messenger::warn("Number of terms ({}) in existing FitCoefficients array for target '{}' does "
                                     "not match the current number ({}), so will fit from scratch.\n",
                                     fitCoefficients.size(), module->name(), ncoeffp);
-                    fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, ncoeffp, pSigma1_, pSigma2_,
-                                                                  nIterations, 0.01);
+                    fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, ncoeffp, pSigma1_, pSigma2_, nIterations, 0.01);
                 }
                 else
                     fitError =
--- a/src/modules/epsr/process.cpp
+++ b/src/modules/epsr/process.cpp
@@ -285,8 +285,8 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
             GaussFit coeffMinimiser(deltaFQ);
 
             if (status == GenericItem::ItemStatus::Created)
-                fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, nCoeffP_.value(), gSigma1_, nIterations,
-                                                              0.01, false);
+                fitError =
+                    coeffMinimiser.constructReciprocal(0.0, rmaxpt, nCoeffP_.value(), gSigma1_, nIterations, 0.01, false);
             else
             {
                 if (fitCoefficients.size() != nCoeffP_)
@@ -294,12 +294,10 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
                     Messenger::warn("Number of terms ({}) in existing FitCoefficients array for target '{}' does "
                                     "not match the current number ({}), so will fit from scratch.\n",
                                     fitCoefficients.size(), module->name(), nCoeffP_.value());
-                    fitError =
-                        coeffMinimiser.constructReciprocal(0.0, rmaxpt, nCoeffP_.value(), gSigma1_, nIterations, 0.01);
+                    fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, nCoeffP_.value(), gSigma1_, nIterations, 0.01);
                 }
                 else
-                    fitError =
-                        coeffMinimiser.constructReciprocal(0.0, rmaxpt, fitCoefficients, gSigma1_, nIterations, 0.01);
+                    fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, fitCoefficients, gSigma1_, nIterations, 0.01);
             }
 
             // Store the new fit coefficients
@@ -313,8 +311,8 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
             PoissonFit coeffMinimiser(deltaFQ);
 
             if (status == GenericItem::ItemStatus::Created)
-                fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, nCoeffP_.value(), pSigma1_, pSigma2_,
-                                                              nIterations, 0.1, false);
+                fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, nCoeffP_.value(), pSigma1_, pSigma2_, nIterations,
+                                                              0.1, false);
             else
             {
                 if (fitCoefficients.size() != nCoeffP_)
@@ -326,8 +324,8 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
                                                                   nIterations, 0.01);
                 }
                 else
-                    fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, fitCoefficients, pSigma1_, pSigma2_,
-                                                                  nIterations, 0.01);
+                    fitError =
+                        coeffMinimiser.constructReciprocal(0.0, rmaxpt, fitCoefficients, pSigma1_, pSigma2_, nIterations, 0.01);
             }
 
             // Store the new fit coefficients
@@ -339,8 +337,7 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
                 Filters::movingAverage(fitCoefficients, *fluctuationSmoothing_);
 
                 // Need to pass the smoothed parameters back into the minimiser so we generate the matching approximation
-                fitError =
-                    coeffMinimiser.constructReciprocal(0.0, rmaxpt, fitCoefficients, pSigma1_, pSigma2_, 0, 0.01);
+                fitError = coeffMinimiser.constructReciprocal(0.0, rmaxpt, fitCoefficients, pSigma1_, pSigma2_, 0, 0.01);
             }
 
             deltaFQFit = coeffMinimiser.approximation();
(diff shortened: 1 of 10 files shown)


Danbr4d:
Only thing I am unsure of is what textWhenNull to have and also what values to put in the trajectoryFrequency_ as it seems to be 0 everywhere I can see

src/modules/md/md.cpp
```diff
@@ -48,6 +49,17 @@ MDModule::MDModule() : Module("MD")
         "Only forces arising from intramolecular terms (including pair potential contributions) will be calculated",
         intramolecularForcesOnly_);
 
+    // Output
+    keywords_.add<OptionalIntegerKeyword>("Output", "EnergyFrequency",
```
trisyoungs, line 53: Sorry - my merging in of #1308 means that your definitions here are in the old format! They should now be like:
```suggestion
    keywords_.add<OptionalIntegerKeyword>("EnergyFrequency",
```

src/modules/md/md.cpp
```diff
@@ -48,6 +49,17 @@ MDModule::MDModule() : Module("MD")
...
+                                          "Frequency at which to calculate total system energy (or 0 to inhibit)",
```
trisyoungs, line 54: I think we can now remove the "(or 0 to inhibit)" from the help text and the comments above the variables in the header.
```suggestion
                                          "Frequency at which to calculate total system energy",
```

src/modules/md/md.cpp
```diff
@@ -48,6 +49,17 @@ MDModule::MDModule() : Module("MD")
...
+                                          energyFrequency_, 0, std::nullopt, 10, "Inhibited");
+    keywords_.add<OptionalIntegerKeyword>("Output", "OutputFrequency",
+                                          "Frequency at which to output step information (or 0 to inhibit)", outputFrequency_,
+                                          0, std::nullopt, 5, "Inhibited");
```
trisyoungs, line 58: For simplicity and consistency I would just go with "Off" for the special case here and in the others:
```suggestion
                                          0, std::nullopt, 5, "Off");
```

src/modules/md/md.h
```diff
@@ -45,21 +45,21 @@ class MDModule : public Module
     // Fixed timestep (ps) to use in MD simulation
     double fixedTimestep_{5.0e-4};
     // Frequency at which to calculate total system energy (or 0 to inhibit)
```
trisyoungs, line 47:
```suggestion
    // Frequency at which to calculate total system energy
```
etc.

src/modules/md/process.cpp
```diff
@@ -34,17 +34,17 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
     if (onlyWhenEnergyStable_)
         Messenger::print("MD: Only perform MD if target Configuration energies are stable.\n");
     if (writeTraj)
-        Messenger::print("MD: Trajectory file will be appended every {} step(s).\n", trajectoryFrequency_);
+        Messenger::print("MD: Trajectory file will be appended every {} step(s).\n", trajectoryFrequency_.value());
```
trisyoungs, line 37: Need to be careful here because the frequency can be undefined (i.e. `std::nullopt`) so it might be best to remove the `writeTraj` variable on line 23 and simply do:
```
if (trajectoryFrequency_.value_or(0) > 0)
```

src/modules/md/process.cpp
```diff
@@ -34,17 +34,17 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
     else
         Messenger::print("MD: Trajectory file off.\n");
     if (capForces_)
         Messenger::print("MD: Forces will be capped to {:10.3e} kJ/mol per atom per axis.\n", maxForce / 100.0);
     if (energyFrequency_ > 0)
-        Messenger::print("MD: Energy will be calculated every {} step(s).\n", energyFrequency_);
+        Messenger::print("MD: Energy will be calculated every {} step(s).\n", energyFrequency_.value());
```
trisyoungs, line 43: Same here.

src/modules/md/process.cpp
```diff
@@ -34,17 +34,17 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
     else
         Messenger::print("MD: Energy will be not be calculated.\n");
     if (outputFrequency_ > 0)
-        Messenger::print("MD: Summary will be written every {} step(s).\n", outputFrequency_);
+        Messenger::print("MD: Summary will be written every {} step(s).\n", outputFrequency_.value());
```
trisyoungs, line 47: And here!

src/modules/md/process.cpp
```diff
@@ -308,10 +308,10 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
         ke *= 0.01;
 
         // Write step summary?
-        if ((step == 1) || (step % outputFrequency_ == 0))
+        if ((step == 1) || (step % outputFrequency_.value() == 0))
```
trisyoungs, line 311: And here, but you probably want to move the check for a valid value in `outputFrequency_` out of this `if` statement otherwise it's going to get a little messy.

trisyoungs (changes requested):
Exactly the right idea, but we need to be careful when working with `std::optional`s since we can't assume the presence of a defined value, and need to cater for the `std::nullopt` case explicitly in places.

src/modules/md/process.cpp
Danbr4d: I have these changes locally but I think they didn't get pushed up/got replaced in one of the other commits. Explains why its building locally but failing the tests

Danbr4d:
Made the changes you suggested but getting an error that I don't understand. Thought it was due to the frequency having 8 arguments but I changed that and still get the same error
image

src/modules/md/md.cpp
```diff
@@ -48,6 +49,14 @@ MDModule::MDModule() : Module("MD")
         "Only forces arising from intramolecular terms (including pair potential contributions) will be calculated",
         intramolecularForcesOnly_);
 
+    // Output
+    keywords_.add<OptionalIntegerKeyword>("Output", "EnergyFrequency", "Frequency at which to calculate total system energy",
```
trisyoungs, line 53: Just the initial argument needs removing - this was the old way of specifying the group the keyword would be displayed in in the GUI, but this is now covered by the `setOrganisation()` method.
```suggestion
    keywords_.setOrganisation("Output");
    keywords_.add<OptionalIntegerKeyword>("EnergyFrequency", "Frequency at which to calculate total system energy",
```
etc.

trisyoungs (changes requested):

src/modules/gr/process.cpp
```diff
@@ -25,7 +25,7 @@ bool GRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
     if (useHalfCellRange_)
```
trisyoungs, line 25: As discussed, the `useHalfCellRange_` variable and associated keyword can be removed since using a `std::optional<double>` for the `requestedRange_` variable covers this choice.

src/modules/md/md.cpp
```diff
@@ -30,12 +31,12 @@ MDModule::MDModule() : Module("MD")
...
 
     keywords_.setOrganisation("Options", "Output");
-    keywords_.add<IntegerKeyword>("EnergyFrequency", "Frequency at which to calculate total system energy (or 0 to inhibit)",
-                                  energyFrequency_, 0);
-    keywords_.add<IntegerKeyword>("OutputFrequency", "Frequency at which to output step information (or 0 to inhibit)",
-                                  outputFrequency_, 0);
-    keywords_.add<IntegerKeyword>("TrajectoryFrequency", "Write frequency for trajectory file (or 0 to inhibit)",
-                                  trajectoryFrequency_, 0);
+    keywords_.add<OptionalIntegerKeyword>("EnergyFrequency", "Frequency at which to calculate total system energy",
+                                          energyFrequency_, 0, std::nullopt, 10, "Off");
```
trisyoungs, line 35: For consistency, let's make all the value steps `5`:
```suggestion
                                          energyFrequency_, 0, std::nullopt, 5, "Off");
```

src/modules/md/md.h
```diff
@@ -44,22 +44,22 @@ class MDModule : public Module
...
+    // Frequency at which to output step information
+    std::optional<int> outputFrequency_{5};
     // Whether random velocities should always be assigned before beginning MD simulation
     bool randomVelocities_{false};
     // Species to restrict calculation to
     std::vector<const Species *> restrictToSpecies_;
-    // Write frequency for trajectory file (or 0 to inhibit)
-    int trajectoryFrequency_{0};
+    // Write frequency for trajectory file
+    std::optional<int> trajectoryFrequency_{0};
```
trisyoungs, line 62:
```suggestion
    std::optional<int> trajectoryFrequency_;
```

src/modules/md/process.cpp
```diff
@@ -308,10 +307,10 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
         ke *= 0.01;
 
         // Write step summary?
-        if ((step == 1) || (step % outputFrequency_ == 0))
+        if ((step == 1) || (step % outputFrequency_.value_or(0) == 0))
```
trisyoungs, line 310: I have a suspicion that this will write out every step (or maybe even crash!) if `outputFrequency_` is undefined, since the result of `step % 0` is undefined. It's safer to do:
```suggestion
        if (outputFrequency_ && (step == 1 || (step % outputFrequency_.value() == 0)))
```

src/modules/md/process.cpp
```diff
@@ -308,10 +307,10 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
         {
             // Include total energy term?
-            if ((energyFrequency_ > 0) && (step % energyFrequency_ == 0))
+            if ((energyFrequency_.value_or(0) > 0) && (step % energyFrequency_.value_or(0) == 0))
```
trisyoungs, line 313: And here:
```suggestion
            if (energyFrequency_ && (step % energyFrequency_.value() == 0))
```

src/modules/md/process.cpp
```diff
@@ -324,7 +323,7 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
         }
 
         // Save trajectory frame
-        if (writeTraj && (step % trajectoryFrequency_ == 0))
+        if (trajectoryFrequency_.value_or(0) > 0 && (step % trajectoryFrequency_.value_or(0) == 0))
```
trisyoungs, line 326:
```suggestion
        if (trajectoryFrequency_ && (step % trajectoryFrequency_.value() == 0))
```

src/modules/md/process.cpp
```diff
@@ -333,7 +332,7 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
 
                 // Construct and write header
                 std::string header = fmt::format("Step {} of {}, T = {:10.3e}, ke = {:10.3e}", step, nSteps_, tInstant, ke);
-                if ((energyFrequency_ > 0) && (step % energyFrequency_ == 0))
+                if ((energyFrequency_.value_or(0) > 0) && (step % energyFrequency_.value_or(0) == 0))
```
trisyoungs, line 335:
```suggestion
                if (energyFrequency_ && (step % energyFrequency_.value() == 0))
```

trisyoungs (changes requested):
Looking good.  Suggestions made for the default text values, and some corrections made to variable overwriting / logic in my old code (which your changes highlighted!).

src/modules/epsr/process.cpp
```diff
@@ -124,12 +126,8 @@ bool EPSRModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
 
     // Calculate some values if they were not provided
-    if (rMaxPT_ < 0.0)
-        rMaxPT_ = dissolve.pairPotentialRange();
-    if (rMinPT_ < 0.0)
-        rMinPT_ = rMaxPT_ - 2.0;
-    if (!nCoeffP_)
-        nCoeffP_ = std::min(int(10.0 * rMaxPT_ + 0.0001), mcoeff);
+    auto rmaxpt = rMaxPT_ ? rMaxPT_.value() : dissolve.pairPotentialRange();
+    auto rminpt = rMinPT_ ? rMinPT_.value() : rmaxpt - 2.0;
```
trisyoungs, line 130: Changing this bit also removed the setting of `nCoeffP_` optional which is now causing a crash. We need to do a similar thing as above, and change uses of `nCoeffP_.value()` to use our temporary var:
```suggestion
    auto rminpt = rMinPT_ ? rMinPT_.value() : rmaxpt - 2.0;
    auto ncoeffp = nCoeffP_ ? nCoeffP_.value() : std::min(int(10.0 * rmaxpt + 0.0001), mcoeff);
```

trisyoungs (approved):

Issue #1309, opened 2023-01-26
Calculation Modules assume intermolecular interactions


trisyoungs:
**Describe the feature / issue**
All/most of the current group of calculation modules assume intermolecular interactions are the desired quantity, whereas there are equally good reasons to calculate intramolecular versions of the same thing.  Cleanly separating contributions from inter- and intra-molecular correlations cannot be efficiently done in Dissolve at present.

New calculation modules should be introduced that specifically cater for calculation of properties based on sites within the same molecule. This probably warrants some renaming of the existing modules to prevent names getting to long.

trisyoungs:
Closed by #1314 and #1320.

PR #1320, opened 2023-02-08
1309 Intramolecular RDF

trisyoungs:
This PR adds a new module focussing on calculation of intramolecular RDFs, i.e. between sites located within the same molecule.

Works towards #1309.


trisyoungs:
@Danbr4d This draft PR starts the journey for an intramolecular analogue to the current `SiteRDF` module. All I have done so far is to duplicate that module and rename the classes so we have an equivalent `IntraRDF` module to now work on.

While I would normally combine the three commits into one I have deliberately done it this way so you can see what process I followed. Take a look in the constructor in `intrardf.cpp` file and see if you can get a handle on what's going on in order to create a calculation module like this.

src/modules/intrardf/intrardf.cpp
```diff
@@ -19,7 +19,7 @@
 #include "procedure/nodes/select.h"
 #include "procedure/nodes/sum1d.h"
 
-SiteRDFModule::SiteRDFModule() : Module("SiteRDF"), analyser_(ProcedureNode::AnalysisContext)
+IntraRDFModule::IntraRDFModule() : Module("SiteRDF"), analyser_(ProcedureNode::AnalysisContext)
```
Danbr4d, line 22: Is the Module("SiteRDF") also meant to be IntraRDF now?

trisyoungs: :+1: Yep. Missed that one!

trisyoungs:
This PR is now complete, and includes a system test to verify the calculation.

Danbr4d (approved):

PR #1323, opened 2023-02-17
1097 Parameter Formatting

### LINKED ISSUE #1097: Specifying parameters in built-in forcefields
PR #1096 addresses a situation where the parameters for the cosine torsion potential were given in the wrong order (as a `std::vector<double>`), leading to the forcefield behaving very badly.

Since it is all to easy to make these errors when converting forcefields it seems prudent to make use of the `name=value` scheme we have in place for setting parameters. Alternatively we could accept a `std::map<std::string,double>` of data instead.

trisyoungs:
This PR converts specification of built-in forcefield parameters from `std::vector<double>` to a `std::string` of `name=value` pairs. This greatly aids readability, and eliminates the chances of entering the parameters in the wrong order.

Worth noting that it is still possible to specify parameters as a `std::vector<double>` for the `ForcefieldAtomType` class since otherwise this would break specification of generator-based forcefields such as UFF - these forcefields `ForcefieldAtomType` to store a range of generator parameters, but which do not fit into specification in `ShortRangeFunctions`.

Closes #1097.


rprospero (approved):
It took me a minute to overcome my aversion to stringly-typed code, but you've absolutely made the right call here.  This is a definite improvement and there's no legitimate reason to throw a `std::map` into the middle.

The unit test also sparks a special joy.

PR #1325, opened 2023-02-20
1297 optional keyword exception handling

### LINKED ISSUE #1297: OptionalInt Exception Handling
**Describe the bug**
In 0.9.X a change was made to the `ModifyPotential` keyword of EPSR, moving it from a `bool` to an `OptionalInteger`.  Old input files still containing `False` will silently fail the input file read, annoying the user.

Fix this so that we have considerably more verbosity when a parse error is encountered!

Danbr4d:
Closes #1297. Initial commit for questions

### DIFF
--- a/src/keywords/optionaldouble.cpp
+++ b/src/keywords/optionaldouble.cpp
@@ -3,6 +3,7 @@
 
 #include "keywords/optionaldouble.h"
 #include "base/lineparser.h"
+#include "base/sysfunc.h"
 
 OptionalDoubleKeyword::OptionalDoubleKeyword(std::optional<double> &data, double minValue, std::optional<double> maxValue,
                                              double valueDelta, std::string_view textWhenNull)
@@ -55,21 +56,30 @@ std::string OptionalDoubleKeyword::textWhenNull() const { return textWhenNull_; 
 // Deserialise from supplied LineParser, starting at given argument offset
 bool OptionalDoubleKeyword::deserialise(LineParser &parser, int startArg, const CoreData &coreData)
 {
-    if (parser.hasArg(startArg))
+    try
     {
+        if (!parser.hasArg(startArg))
+            return false;
+
         auto x = parser.argd(startArg);
         if (!setData(x))
         {
             if (maximumLimit_)
-                Messenger::error("Value {} is out of range for keyword. Valid range is {} <= n.\n", x, maximumLimit_.value());
+                Messenger::error("Value {} is out of range for keyword '{}'. Valid range is {} <= n.\n", x, name(),
+                                 maximumLimit_.value());
 
             return false;
         }
-
-        return true;
     }
-
-    return false;
+    catch (...)
+    {
+        // Check explicitly for null text
+        if (DissolveSys::sameString(parser.argsv(startArg), textWhenNull_))
+            setData(std::nullopt);
+        else
+            return Messenger::error("Input {} is invalid for keyword '{}'.\n", parser.argsv(startArg), name());
+    }
+    return true;
 }
 
 // Serialise data to specified LineParser
--- a/src/keywords/optionaldouble.cpp
+++ b/src/keywords/optionaldouble.cpp
@@ -66,10 +66,8 @@ bool OptionalDoubleKeyword::deserialise(LineParser &parser, int startArg, const 
         if (!setData(x))
         {
             if (maximumLimit_)
-                Messenger::error("Value {} is out of range for keyword '{}'. Valid range is {} <= n.\n", x, name(),
-                                 maximumLimit_.value());
-
-            return false;
+                return Messenger::error("Value {} is out of range for keyword '{}'. Valid range is {} <= n.\n", x, name(),
+                                        maximumLimit_.value());
         }
     }
     catch (...)
--- a/src/keywords/optionaldouble.cpp
+++ b/src/keywords/optionaldouble.cpp
@@ -4,6 +4,7 @@
 #include "keywords/optionaldouble.h"
 #include "base/lineparser.h"
 #include "base/sysfunc.h"
+#include "keywords/base.h"
 
 OptionalDoubleKeyword::OptionalDoubleKeyword(std::optional<double> &data, double minValue, std::optional<double> maxValue,
                                              double valueDelta, std::string_view textWhenNull)
--- a/src/keywords/optionalint.cpp
+++ b/src/keywords/optionalint.cpp
@@ -49,9 +49,6 @@ int OptionalIntegerKeyword::valueDelta() const { return valueDelta_; }
 // Return text to display in widget  when value is null
 std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_; }
 
-// Return keyword name
-std::string_view KeywordBase::name() const { return name_; }
-
 /*
  * Arguments
  */
@@ -62,9 +59,8 @@ bool OptionalIntegerKeyword::deserialise(LineParser &parser, int startArg, const
     try
     {
         if (!parser.hasArg(startArg))
-        {
             return false;
-        }
+
         auto x = parser.argi(startArg);
         if (!setData(x))
         {
@@ -83,7 +79,6 @@ bool OptionalIntegerKeyword::deserialise(LineParser &parser, int startArg, const
         else
             return Messenger::error("Input {} is invalid for keyword '{}'.\n", parser.argsv(startArg), name());
     }
-
     return true;
 }
 
--- a/src/keywords/optionalint.cpp
+++ b/src/keywords/optionalint.cpp
@@ -4,6 +4,7 @@
 #include "keywords/optionalint.h"
 #include "base/lineparser.h"
 #include "base/sysfunc.h"
+#include "keywords/base.h"
 
 OptionalIntegerKeyword::OptionalIntegerKeyword(std::optional<int> &data, int minValue, std::optional<int> maxValue,
                                                int valueDelta, std::string_view textWhenNull)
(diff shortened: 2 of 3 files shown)


Danbr4d:
I know this doesn't work but I'm a bit unsure how to access the input that I'm meant to be converting to int/bool. Also, do I try and change it to int before checking if it is the textWhenNull.

Danbr4d:
I've also just realised I didn't rebase this before pushing so will sort that out too

src/keywords/optionalint.cpp
```diff
@@ -55,9 +56,26 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
 // Deserialise from supplied LineParser, starting at given argument offset
 bool OptionalIntegerKeyword::deserialise(LineParser &parser, int startArg, const CoreData &coreData)
 {
+    try
+    {
+        std::stoi(data);
+    }
```
trisyoungs, line 62: So here you can just use the original (actually, fixed!) `parser.argi(startArg)` call to attempt the initial conversion.  This uses `stoi` so will throw an exception if it isn't a valid conversion.

```suggestion
    std::optional<int> newValue;
    try
    {
        newValue = parser.argi(startArg);
    }
```
If this succeeds (i.e. doesn't throw) then your `newValue` optional will contain the value to set.

src/keywords/optionalint.cpp
```diff
@@ -55,9 +56,26 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
+        {
+            if (data == textWhenNull)
+                //here
+            else
+                DissolveSys::stob(data);
+        }
+        catch (...)
+        {
+        }
+    }
```
trisyoungs, line 75: Second part, there's no need to try/catch for the string conversion since we can do that as simple checks or conversions, this time using the string representation of the argument via `parser.argsv(startArg)`. If the string is neither the textWhenNull nor equates to `false` (via `DissolveSys::stob()`) then at this point we must return an error message since we don't know what to do with the input provided by the user.

Danbr4d:
I think the error in the build is coming from DissolveSys::stob() but I'm not sure why

src/keywords/optionalint.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include "keywords/optionalint.h"
 #include "base/lineparser.h"
+#include "base/sysfunc.cpp"
```
trisyoungs, line 6: Eeek! That's not the header...
```suggestion
#include "base/sysfunc.h"
```

src/keywords/optionalint.cpp
```diff
@@ -55,9 +56,28 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
+        newString = parser.argsv(startArg);
+        if (DissolveSys::stob(newString) == false)
+        {
+            if (newString != textWhenNull_)
+                Messenger::error("Unkonwn input, {}, provided\n", newString);
+            else
+                newString = textWhenNull_;
+        }
+        else
+            DissolveSys::stob(newString);
```
trisyoungs, line 76: The logic is a bit convoluted here, and after thinking about it I don't think we need to consider the `stob()` part.  I'd suggest removing the `newString` variable and doing something like:
```suggestion
        // Check explicitly for the null text
        if (DissolveSys::sameString(parser.argv(startArg), textWhenNull_))
            setData(std::nullopt);
        else
                return Messenger::error("Unknown input, {}, provided\n", newString);
```

src/keywords/optionalint.cpp
```diff
@@ -55,9 +56,28 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
+    }
     if (parser.hasArg(startArg))
     {
-        auto x = parser.argd(startArg);
+        auto x = parser.argi(startArg);
```
trisyoungs, line 80: The guts of this bit can be moved into the `try` block, leaving just a `return true;` here (because we've handled all the errors earlier).

src/keywords/optionalint.cpp
```diff
@@ -55,21 +56,32 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
 bool OptionalIntegerKeyword::deserialise(LineParser &parser, int startArg, const CoreData &coreData)
 {
-    if (parser.hasArg(startArg))
+    std::optional<int> newValue;
+    try
     {
-        auto x = parser.argd(startArg);
-        if (!setData(x))
+        newValue = parser.argi(startArg);
+        if (parser.hasArg(startArg))
```
trisyoungs, line 63: This check should actually be made with reverse logic (i.e. `if (!parser.hasArg.....`) as the very first thing we do in the function and immediately returning `false` since, if there's no argument to parse, there's no conversion to be made!

src/keywords/optionalint.cpp
```diff
@@ -55,21 +56,32 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
         {
-            if (maximumLimit_)
-                Messenger::error("Value {} is out of range for keyword. Valid range is {} <= n.\n", x, maximumLimit_.value());
-
-            return false;
+            auto x = parser.argi(startArg);
```
trisyoungs, line 65: With the `if` check moved, you will be left with this bit of the function, and the `newValue` stuff can be removed.

src/keywords/optionalint.cpp
```diff
@@ -55,21 +56,32 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
+            if (!setData(x))
+            {
+                if (maximumLimit_)
+                    Messenger::error("Value {} is out of range for keyword. Valid range is {} <= n.\n", x,
```
trisyoungs, line 69: I should have added the name of the keyword to the error message:
```suggestion
                    Messenger::error("Value {} is out of range for keyword '{}'. Valid range is {} <= n.\n", x, name_,
```

src/keywords/optionalint.cpp
```diff
@@ -55,21 +56,32 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
         }
-
-        return true;
+    }
+    catch (...)
+    {
+        if (DissolveSys::sameString(parser.argsv(startArg), textWhenNull_))
+            setData(std::nullopt);
+        else
+            return Messenger::error("Unknown input, {}, provided\n", parser.argsv(startArg));
```
trisyoungs, line 81: We should make this error message consistent in content and format to the one above.

trisyoungs (commented):
Pretty much there - just tidying.

src/keywords/optionalint.cpp
```diff
@@ -48,28 +49,42 @@ int OptionalIntegerKeyword::valueDelta() const { return valueDelta_; }
 // Return text to display in widget  when value is null
 std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_; }
 
+// Return keyword name
+std::string_view KeywordBase::name() const { return name_; }
+
```
trisyoungs, line 54:
```suggestion
```

src/keywords/optionalint.cpp
```diff
@@ -48,28 +49,42 @@ int OptionalIntegerKeyword::valueDelta() const { return valueDelta_; }
...
 bool OptionalIntegerKeyword::deserialise(LineParser &parser, int startArg, const CoreData &coreData)
 {
-    if (parser.hasArg(startArg))
+    try
     {
-        auto x = parser.argd(startArg);
+        if (!parser.hasArg(startArg))
+        {
+            return false;
+        }
```
trisyoungs, line 67: Stylistically, I prefer to not use curlies around simple statements like this when there's no need to enforce scope for what happens inside the braces.
```suggestion
            return false;
```

trisyoungs (commented):
Small comments, but otherwise :+1: to move on to the `double` version and replicate.

src/keywords/optionaldouble.cpp
```diff
@@ -55,21 +57,30 @@ std::string OptionalDoubleKeyword::textWhenNull() const { return textWhenNull_;
...
+
         auto x = parser.argd(startArg);
         if (!setData(x))
         {
             if (maximumLimit_)
-                Messenger::error("Value {} is out of range for keyword. Valid range is {} <= n.\n", x, maximumLimit_.value());
+                Messenger::error("Value {} is out of range for keyword '{}'. Valid range is {} <= n.\n", x, name(),
+                                 maximumLimit_.value());
 
             return false;
```
trisyoungs, line 72: The `Messenger::error()` function returns `false` so we can do this:
```suggestion
                return Messenger::error("Value {} is out of range for keyword '{}'. Valid range is {} <= n.\n", x, name(),
                                        maximumLimit_.value());
```

src/keywords/optionalint.cpp
```diff
@@ -55,21 +57,30 @@ std::string OptionalIntegerKeyword::textWhenNull() const { return textWhenNull_;
...
+        if (!parser.hasArg(startArg))
+            return false;
+
+        auto x = parser.argi(startArg);
         if (!setData(x))
         {
             if (maximumLimit_)
-                Messenger::error("Value {} is out of range for keyword. Valid range is {} <= n.\n", x, maximumLimit_.value());
+                Messenger::error("Value {} is out of range for keyword '{}'. Valid range is {} <= n.\n", x, name(),
+                                 maximumLimit_.value());
```
trisyoungs, line 70: Use `return` here too.

trisyoungs (approved):
Small suggestion for neatness, otherwise :+1:.

PR #1329, opened 2023-02-22
1309 intra angle

Danbr4d:
Set up the initial intra angle module


Danbr4d:
Not sure how to set up the tests for this I've looked at IntraDistance but I'm not sure where the data comes from

src/modules/intraangle/gui/intraanglewidget.h
```diff
@@ -0,0 +1,47 @@
...
+    void updateControls(const Flags<ModuleWidget::UpdateFlags> &updateFlags = {}) override;
+
+    /*
+     * Widgets / Functions
+     */
+    private:
+    // Set data targets in graphs
+    void setGraphDataTargets(IntraAngleModule *module);
+
+    private slots:
```
rprospero, line 46: I don't believe we need a slots section if we have no slots.

src/modules/intradistance/functions.cpp
rprospero, line 1: I might be a bit dense here, but do we need this file?  It seems like we're not adding any actual code.

rprospero (approved):
I had a couple of questions, but I'm happy to approve this as long as someone has answers.

src/modules/intradistance/functions.cpp
trisyoungs: Nope! This is in my intra-rdf PR that this one is based off.  I'll get rid of it there.

PR #1337, opened 2023-02-24
1334 Ignored Enumerations

### LINKED ISSUE #1334: Enumerations provided with values in keywords are not honoured
**Describe the bug**
Loading an input file specifying value/enum keywords (e.g. the `Density` keyword of the `AddProcedureNode`) correctly reads the value but shows the incorrect enum option in the GUI.

trisyoungs:
This PR fixes an issue with EnumOptions attached to keywords, specifically `NodeValueEnumOptionsKeyword`. While everything works fine from the CLI, displaying and editing the enum via `NodeValueEnumOptionsKeywordWidget` in the GUI fails to display or change the enum value. The underlying cause of this was the fact that the current enumeration was stored in both the keyword as part of its `data_` as well as the `EnumOptions` object used in the keyword widget.

To clarify things, storage of the current enum within `EnumOptions` has been removed, making it a pure read-only utility class for manipulating enums and their associated keywords. The major side effect of this is the partial refactoring of `FileAndFormat` which depended heavily on the now-removed enum storage.

Closes #1334.

### DIFF
--- a/src/base/enumoptions.h
+++ b/src/base/enumoptions.h
@@ -16,35 +16,16 @@ template <class E> class EnumOptions : public EnumOptionsBase
 {
     public:
     EnumOptions() = default;
-    EnumOptions(std::string_view name, std::vector<EnumOption<E>> options) : name_(name), options_(options)
-    {
-        currentOption_ = options_.cbegin();
-    }
-    EnumOptions(std::string_view name, std::vector<EnumOption<E>> options, E currentOption) : name_(name), options_(options)
-    {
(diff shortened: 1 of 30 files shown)


rprospero (approved):
I really like separating out the index from the enumeration.  When I start merging the TOML branch back onto main, it's going to make a bunch of merge conflicts, but the resulting code will be so much cleaner.

trisyoungs:
> When I start merging the TOML branch back onto main, it's going to make a bunch of merge conflicts...

Yeah - sorry about that! :smile:

PR #1339, opened 2023-02-25
[TEST] Move to conan 2.0

trisyoungs:
This PR moves us to Conan 2. Progress so far:

1) Conan2 CMake integration is "better" but forces us back to a static `conanfile.txt`.
2) All deps are enabled in this file, even `gtest` and `benchmark` which aren't necessarily required.
3) Windows build of `gtest` fails because it needs MSVC 19.44, while the runners have 19.14 because we had to downgrade them.


trisyoungs:
Superseded by #2237.

PR #1345, opened 2023-03-01
1295 Complete Release CI

trisyoungs:
This PR mainly updates our release workflow following testing during the push of v1.0.0.  Aside from some minor corrections here and there the biggest addition is the `update-release-info` action, which automates the process of updating the release version in the main `develop` branch via the opening of a specific PR. You can see an example of such a PR at https://github.com/disorderedmaterials/dissolve/pull/1347. Previously, this was achieved by directly pushing to `develop` in the CI, but which is now prohibited by our branch protection rules.

Works towards #1295. While not completely closing #1295, it does put us completely on parity with the functionality of the existing Azure pipelines, so this PR removes it.


DanNixon (approved):
:+1:

PR #1350, opened 2023-03-03
Web docs styling

trisyoungs:
This PR's primary purpose is to create better "visual separation" between the three versions of documentation. The relevant version number, as well as a `Legacy` or `Development` prefix, is now displayed, and the top banner colour is changed from the default deep blue.

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -28,7 +28,9 @@ jobs:
     - name: Build Website
       uses: "./.github/workflows/website"
       with:
-        continuous: ${{ github.ref_name == 'develop' }}
+        isDevelopment: ${{ github.ref_name == 'develop' }}
+        isRelease: ${{ startsWith(github.ref_name, 'release') }}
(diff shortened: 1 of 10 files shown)


.github/workflows/web.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     - name: Build Website
       uses: "./.github/workflows/website"
       with:
-        isDevelopment: ${{ github.ref_name == 'develop' }}
+        isDevelopment: ${{ github.ref_name == 'develop' || github.event_name == 'pull_request' }}
```
DanNixon, line 31: Was this added for testing or should it still be included?

trisyoungs: As it turns out this was added for testing the CI, but makes a lot of sense to leave it in.  Since our standard workflow means that PRs are only ever related to the `develop` branch, making the PR CI tests equivalent to a push on `develop` felt like the right thing to do.

PR #1354, opened 2023-03-14
added deprecated keyword for UseHalfCellWidth

Danbr4d:
Change from release/1.0.1 added to develop that adds a deprecated value to useHalfCellRange in the module g(r), that was deleted when adding optional variables


trisyoungs (approved):
:+1:

PR #1362, opened 2023-03-15
1358 FileAndFormat Reading

### LINKED ISSUE #1358: FileAndFormat keyword fails with a warning when it should raise an error
**Describe the bug**
Passing the wrong arguments to a `FileAndFormatKeyword` does not raise the correct type of error. The standard expectation of the keyword is:
```
MyExportKeyword  <format>  <filename>
...
EndMyExportKeyword
```
...but if one simply does this:
```
MyExportKeyword  True
```
...then the keyword correctly complains that `True` isn't a valid format, but Dissolve carries on with a warning when it should error.

trisyoungs:
This PR addresses #1358, making the reading of `FileAndFormat` data easier to warn/fail on errors as appropriate, rather than just throwing a warning no matter how bad the error and allowing Dissolve to carry on.

Closes #1358.

### DIFF
--- a/src/classes/data1dstore.cpp
+++ b/src/classes/data1dstore.cpp
@@ -18,7 +18,7 @@ bool Data1DStore::addData(std::string_view dataName, LineParser &parser, int sta
     data.setTag(dataName);
 
     // Read the file / format
-    if (!format.read(parser, startArg, endKeyword, coreData))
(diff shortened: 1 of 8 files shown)


rprospero (approved):
I'll always support replacing a boolean with a named enum!

PR #1363, opened 2023-03-17
Add Forcefield Term Dialog via QML

rprospero:
This is a rewrite of the Add Forcefield Terms dialog.  The following tasks need to be complete before it can be merged.

- [x] Agree that QML is something we want in our code base
- [x] Move to Qt 6.4
- [x] Add QML into application resource
- [x] More extensive unit testing of dialog
- [x] Test on Mac and PC (and another linux machine)
- [x] Rework Master Terms portion of dialog to use TreeView


trisyoungs:
@rprospero I've finally had a cursory look at this PR and there's a lot to like! The QML approach definitely has potential to clean up a lot of code, particularly in the wizard domain (make your own jokes here). Auto-determining which forcefields apply to the species is really nice.

I have an issue with the way the new dialog looks, however, and I suppose I blame QtQuick for this. There seem to be missing borders on lists / trees, overlap of controls with their containers, and the text on the buttons is in a really bad colour scheme! From an aesthetic perspective that's a bit of a blocker for me.

One other thing - when auto-determining the relevant forcefields it would be wise to enforce quietness on the `Messenger` class, otherwise when one returns to the GUI there are several hundred error messages captured!

image
image

rprospero:
I've added a border and background to the `IconListView`, which should hopefully clean up the appearance a bit.  I'm looking into why your button colours aren't the same as mine.

border

rprospero:
@trisyoungs I've added extra padding around the radio buttons and linked the buttons text colour to the main text colour, so it might be worth another look on your machine to see if the issues have been handled.

trisyoungs:
Looking better!  From my own offline experimentation I can offer the following:

1. Need to change the default `QGuiApplication::font()` pointSize to 10 for consistency with the rest of the UI.
2. Sprinkling `anchors.margins: 4;` around the places gives the "correct" spacing between controls, but is a pain to add everywhere.
3. Adding `padding: 2;` into the `Text` delegate for the `ListView` gives a result similar to the standard UI controls.

I think padding and margins, and borders on some controls, is where the final work is at.  Is it possible to apply defaults to all relevant controls automatically, rather than have to do it manually?

rprospero:
@trisyoungs I've merged in the changes to the icon paths.  If you can take another pass over the dialog to check the aesthetics, it may be ready to merge in.

trisyoungs:
@rprospero Definitely closer to the mark, but I still find several rough edges to the layout that catch my eye. This would be a useful exercise for me that I'd be happy to take on to get some proper QML experience under my belt, if that works for you?

We could merge this PR in now as it stands since it still offers additional functionality, and I can fixup the widgets etc. in a follow-up PR?

rprospero:
@trisyoungs I would support merging this in for now and making an issue for the GUI changes.  That issue could then be a blocker for the next release, ensuring that users never see the rougher GUI.

I'm mainly interested in getting this branch merged in because it's already had three painful merge conflicts and I have little faith that another two weeks won't give us a fourth.

trisyoungs:
@rprospero Agreed. I'll approve this now and add the update as a separate issue in the 1.4 milestone.

trisyoungs (approved):

PR #1365, opened 2023-03-17
Fix layer menu again

trisyoungs:
Another quick PR to address a hard crash caused by using the wrong module names in the Layer menu items. Also fixes a couple of uses of the old module names in error messages.


src/gui/menu_layer.cpp
rprospero, line 1: This raises the question: Should we create an enum for modules?  This would catch this kind of bug at compile time.  It would also enable `ModuleRegistry` to use an array that accesses each module by index instead of performing string comparisons through a vector.  On the other hand, this is exactly the kind of code that will *never* be a bottle neck.

rprospero (approved):

PR #1361, opened 2023-03-15
1295 GitHub Actions Environment

trisyoungs:
This PR puts all environment definitions, principally required package version numbers, in one sub-action, rather than duplicating them at the top of all four main workflow files (`cache.yml`, `continuous.yml`, `pr.yml`, and `release.yml`). The existing `get-version` action is combined into the new `setup` action and removed.

Closes #1295.

PR #1366, opened 2023-03-17
Move official Qt version to 6.4.2

rprospero:
This delegates building the custom qt to the qt-idaaas repo.  This will both help with caching and also make it easier to upgrade Qt versions.  At the moment, this moves us to Qt 6.4, which enables certain QML features that we will want (e.g. TreeView support).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,12 +13,6 @@ on:
     - 'web/**'
     - 'README.md'
 
-env:
-  pythonVersion: 3.9
-  qtVersion: 6.4.2
-  nixVersion: 2.4
-  nixOSVersion: 22.11
-
 jobs:
 
   QC:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ on:
 
 env:
   pythonVersion: 3.9
-  qtVersion: 6.2.2
+  qtVersion: 6.4.2
   nixVersion: 2.4
   nixOSVersion: 22.11
 
--- a/flake.nix
+++ b/flake.nix
@@ -6,9 +6,10 @@
     nixGL-src.flake = false;
     weggli.url = "github:googleprojectzero/weggli";
     weggli.flake = false;
+    qt-idaaas.url = "github:disorderedmaterials/qt-idaaas";
   };
-  outputs =
-    { self, nixpkgs, outdated, flake-utils, bundlers, nixGL-src, weggli }:
+  outputs = { self, nixpkgs, outdated, flake-utils, bundlers, nixGL-src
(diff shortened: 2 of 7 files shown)


flake.nix
```diff
@@ -159,7 +159,7 @@
               ccls
               cmake-format
               cmake-language-server
-              conan-1.58.0
```
trisyoungs, line 162: Conan2 does work on our Linux builds, but do we want to stick to 1.58 here for purposes of consistency? Just a thought.

trisyoungs (approved):
A small deletion request (to fix a mistake of my own) and one comment for consideration, but otherwise :+1:.

rprospero:
Those lines pull from the Nix repo, not the Ubuntu one, so it doesn’t pull in Conan 2.  Honestly, I think it pulls in Conan 1.49, but it’s never been an issue since the dev shell pulls all the dependencies anyway.  Come November, I’ll have to put in a one liner to use the old Conan, since Nix will switch to 2.0 then, but we might have switched Dissolve by that point.On 29 Mar 2023, at 09:53, Tristan Youngs ***@***.***> wrote:﻿
@trisyoungs approved this pull request.

A small deletion request (to fix a mistake of my own) and one comment for consideration, but otherwise 👍.

In .github/workflows/release.yml:
>  env:
   pythonVersion: 3.9
-  qtVersion: 6.2.2
+  qtVersion: 6.4.2
   nixVersion: 2.4
   nixOSVersion: 22.11

Actually, this whole env: section was obfuscated by #1361, and so can be removed (I clearly forgot to do so!).



In flake.nix:
> @@ -159,7 +159,7 @@
               ccls
               cmake-format
               cmake-language-server
-              conan-1.58.0

Conan2 does work on our Linux builds, but do we want to stick to 1.58 here for purposes of consistency? Just a thought.

flake.nix
rprospero: Those lines pull from the Nix repo, not the Ubuntu one, so it doesn’t pull in Conan 2.  Honestly, I think it pulls in Conan 1.49, but it’s never been an issue since the dev shell pulls all the dependencies anyway.  Come November, I’ll have to put in a one liner to use the old Conan, since Nix will switch to 2.0 then, but we might have switched Dissolve by then.

.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 env:
   pythonVersion: 3.9
-  qtVersion: 6.2.2
+  qtVersion: 6.4.2
   nixVersion: 2.4
   nixOSVersion: 22.11
```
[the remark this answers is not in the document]
rprospero: I've cleaned out the env section

PR #1381, opened 2023-03-31
calculateexpression node for expressions as quantities

### LINKED ISSUE #1369: Expressions as quantities
Related to #1368, we need the ability to introduce `Expression`s as quantities to use in the `CollectND` nodes. As I see it there are two possible approaches to this, outlined below.  Thoughts please!

## Approach 1: Add Expression-specific Keywords to CollectND nodes
For example, alongside the `QuantityX` keyword of the `Collect1D` node it would also have an `ExpressionX` keyword which takes an expression. We would have to have separate keywords for reasons of both speed and simplicity (the `QuantityX` keyword takes a node reference which would be difficult to incorporate alongside `Expression` parsing).

### Benefits
- Localised expressions within `CollectND` nodes
- No additional node types required

### Drawbacks
- Multiple keywords that are mutually-exclusive
- Mutual exclusivity must be checked for at processing time before the procedure is executed.

## Approach 2: Separate CalculateExpression node
We could implement a `CalculateExpression` node inheriting from `CalculateProcedureNode`, which evaluates one or more `Expression` objects and which can then be references by the existing `CollectND` node `QuantityXYZ` keywords (since it is just another `CalculateProcedureNode`).

### Benefits
- No additional keywords or modification required to existing nodes
- Can evaluate multiple expressions in order to generate 1-, 2-, or 3- (or more) deep quantity collections.

### Drawbacks
- Expressions are not local to the place where they are used.

Danbr4d:
Closes #1369 Added in the calculate expression node


Danbr4d:
A bit unsure on what the actual value/term for the keyword is in calculatedistanceexpression.cpp

trisyoungs (commented):
So a couple of initial, sweeping comments:

1. The name of the new node should be `CalculateExpression` - we just copied the `CalculateDistance` node as a template, and this new node has nothing to do with distance!
2. The `nSitesRequired()` function needs to return zero (is this the default if that particular function is not overriden?) since we no long work with nor care about calculating anything from `Site`s.
3. The single keyword can be called "Expression" and with description "Expression to evaluate"
4. The `execute()` method should then just return the `NodeValue`, I think its `asDouble()`.

src/procedure/nodes/calculateexpression.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+// Return dimensionality of calculated observable
+int CalculateExpression::dimensionality() const { return 1; }
+
+/*
+ * Execute
+ */
+
+// Execute node
+double CalculateExpression::execute() { return (asDouble()); }
```
trisyoungs, line 33: Once you've added your `NodeValue` member variable (and assuming that you call it something like `expression_`) then here you will return the evaluation of the expression:
```suggestion
double CalculateExpression::execute() { return expression_.asDouble(); }
```

trisyoungs:
Ah - I now realise how this passed the CI before: the new source files haven't been added to `src/procedure/nodes/CMakeLists.txt`!  That will highlight one remaining error...

trisyoungs:
@Danbr4d I've added in a placeholder unit test and also made a few other changes that I've documented in individual commits.  Hopefully the reasons behind each change are clear, but let me know if not!

Moving on, next steps are:
1. Rename the new node to `CalculateExpressionProcedureNode` to be consistent with the others.
2. The node's `execute()` function actually needs to store the result in the `value_` variable (a member of `CalculateProcedureNodeBase`). Check out the other `Calcualte*` nodes for examples.

unit/procedure/calculateexpression.cpp
```diff
@@ -0,0 +1,31 @@
...
+    auto expressionNode = procedure.createRootNode<CalculateExpressionProcedureNode>("X");
+    EXPECT_TRUE(procedure.rootSequence().check());
+
+    // Simple number
+    expressionNode->keywords().set("Expression", NodeValue(4));
+    expressionNode->execute();
+    EXPECT_DOUBLE_EQ(expressionNode->value(0), 4.0);
+
+    // Expresssions
+    expressionNode->keywords().set("Expression", NodeValue(3.8 * sin(1.2)));
```
trisyoungs, line 26: Subtle thing going on here - as-written this evaluates the code for `3.8 * sin(1.2)` and passes it to the `NodeValue` constructor. That's totally valid, but what we really need to be testing is the ability for Dissolve's "dynamic expression parsing" (let's call it that...) to reproduce a coded result. So, all that's missing is for the expression to be contained within double quotes when passing it to the `NodeValue` constructor. That way, it is interpreted as an equation that must be parsed and reconstructed internally as an `Expression`.  Basically, Dissolve has a built-in calculator!

trisyoungs (approved):
One more semantic thing, then good to go!

PR #1382, opened 2023-04-05
EPSR sampling and range

trisyoungs:
This PR addresses two issues with the EPSR module:
1) Coefficient smoothing is now performed periodically during the Monte Carlo parameter minimisation - this is more in-line with the original EPSR and promotes consistency between what is fitted and what is used.
2) Potential generation was imperfect since we mistakenly set up the function coefficients according to the QMin value rather than the "zero limit".  This appears to be the cause of low-Q drift / separation observed in some simulations.

### DIFF
--- a/src/math/gaussfit.cpp
+++ b/src/math/gaussfit.cpp
@@ -336,74 +336,8 @@ double GaussFit::sweepFitA(FunctionSpace::SpaceType space, double xMin, int samp
 double GaussFit::constructReciprocal(double rMin, double rMax, int nGaussians, double sigmaQ, int nIterations,
                                      double initialStepSize, std::optional<int> degreeOfSmoothing)
 {
-    // Clear any existing data
-    x_.clear();
-    A_.clear();
-    fwhm_.clear();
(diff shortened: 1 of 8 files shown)


src/math/gaussfit.cpp
```diff
@@ -333,86 +334,21 @@ double GaussFit::sweepFitA(FunctionSpace::SpaceType space, double xMin, int samp
...
-    if (reFitAtEnd)
-        sweepFitA(FunctionSpace::ReciprocalSpace, rMin);
-
-    // Regenerate approximation and calculate percentage error of fit
-    generateApproximation(FunctionSpace::ReciprocalSpace);
-    currentError_ = Error::percent(referenceData_, approximateData_);
-
-    return currentError_;
+    return constructReciprocal(rMin, rMax, std::vector<double>(nGaussians, 0.0), sigmaQ, nIterations, initialStepSize,
+                               degreeOfSmoothing);
```
rprospero, line 340: Always nice to get rid of duplicate code

rprospero (approved):

Issue #1398, opened 2023-04-11
Epic / Energy Overlays


trisyoungs:
### Focus
At present (v1.0, v0.9.X) Dissolve contains a full forcefield energy / force "engine" which is used by any module requiring evaluation of such quantities for one or more configurations. This includes `AtomShake`, `MolShake`, `IntraShake`, and `MD` at present. The `EnergyKernel` and `ForceKernel` classes handle the calculation of the total energy (or that of a specific molecule) in a configuration according to the forcefield terms defined in `AtomType`s/`PairPotential`s, `MasterTerm`s, and within individual `Species`.

It is proposed that this be extended to include arbitrary additional energy / force terms, permitting restraints, local potentials, and field potentials to be applied, but allowing the modules requiring energy / force calculation to operate blind without knowledge of these additional potentials (as is currently the case for normal forcefield terms).

### Benefits
Implementing such a mechanism will allow:
- Point restraints : atoms tethered by harmonic springs to specific locations (e.g. crystallographic sites)
- Geometric potentials such as cylindrical "exclusion" zones (useful for generating porous silicas such as MCM-41) or spherical ones (useful for excluding molecules from specific pores in framework species, for instance.
- Local coordination number potentials, dependent on surrounding atom densities
- Molecule potentials - generic potentials acting between molecules rather than atoms

### Implementation Notes

- Maintaining speed and efficiency will be important, since we are modifying some core routines of Dissolve.
- In particular, having an efficient link between individual atoms and the potentials they are "involved in" requires some thought.
- Potentially, one way of achieving this is to have a vector sized to the number of atoms in the system and which relates each individual atom to a list of potentials it is involved with (maintained as a separate vector of interaction definitions).
- This way, the total energy arising from the additional potentials can be calculated by traversing the vector of interaction definitions (which will reference any involved atoms) while the total energy of an atom or molecule can be determined by using the vector of potentials associated to individual atoms.

### Tasks
- [x] Preparation (#1401)
- [x] #1409
- [x] #1397
- [x] #1399

#### Realisation
- [x] #66

trisyoungs:
Closing with all tasks complete.

PR #1401, opened 2023-04-12
Kernel Tidying (1398 pt 1)

trisyoungs:
This PR performs some tidying of the current energy and force kernels ahead of work towards #1398.

Clarification of the various energy & force functions is made throughout, and new system tests are added (hence the high load of added lines!).


src/classes/forcekernel.cpp
```diff
@@ -165,11 +174,44 @@ void ForceKernel::forces(const Cell *centralCell, const Cell *otherCell, bool ap
...
+{
+    assert(cellArray_);
+    auto &cellArray = cellArray_->get();
+
+    // Create a ForceKernel
+    auto combinableForces = createCombinableForces(f);
+
+    // Set start/stride for parallel loop
+    auto start = processPool_.interleavedLoopStart(strategy);
+    auto stride = processPool_.interleavedLoopStride(strategy);
```
rprospero, line 195: If we do one day get rid of MPI, I think that would mean that we could also get rid of a bunch of logic for determining ranges.

rprospero (approved):

src/classes/forcekernel.cpp
trisyoungs: Yep, and no more passing `ProcessPool`s and associated strategies around.

PR #1406, opened 2023-04-17
Functions Code Duplication

trisyoungs:
A tidy-up PR which, once I'd spotted it, I couldn't leave it alone.  Reduces code duplication in classes providing functional form data by introducing suitable general functions in `DissolveSys`.

### DIFF
--- a/src/base/sysfunc.cpp
+++ b/src/base/sysfunc.cpp
@@ -366,6 +366,23 @@ std::string DissolveSys::doubleChars(const std::string_view s, const std::string
 }
 
 /*
+ * String Vectors
+ */
+
+// Return nth string from the given vector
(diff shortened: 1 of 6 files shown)


src/base/sysfunc.cpp
```diff
@@ -365,6 +365,23 @@ std::string DissolveSys::doubleChars(const std::string_view s, const std::string
     return result;
 }
 
+/*
+ * String Vectors
+ */
+
+// Return nth string from the given vector
+std::string DissolveSys::stringAt(const std::vector<std::string> &ss, int n) { return (n < 0 || n >= ss.size()) ? "" : ss[n]; }
```
rprospero, line 373: I was quite surprised to find that there is not an existing function in the standard that provides this functionality, but I cannot find one.

rprospero (approved):

PR #1407, opened 2023-04-18
Fix argument pass-through on singularity

rprospero:
This fixes the issue with CLI arguments not being passed to dissolve in the container.


trisyoungs (approved):
:+1:

Issue #1387, opened 2023-04-06
Lost parameters when changing torsion form to CosN


trisyoungs:
**Describe the bug**
In the torsions tables, changing the functional form to `CosN`, or selecting an existing interaction of that form and leaving it unedited as `CosN`, loses all but the first defined k value.

PR #1409, opened 2023-04-19
InteractionPotential keyword

trisyoungs:
This PR adds a new keyword and widget for the `InteractionPotential` template class.

Works towards #1398.


trisyoungs:
@rprospero - Have pushed another commit to fix a minor error caused by #1387 where the use of `EnumOptions<T>::validNArgs()` by `InteractionPotential::setForm()` would spew an error out, but which wasn't an error as such (just the negative result of a check).

rprospero (approved):
Honestly, having the `quiet` options seems like a good idea in general for the valid functions.

PR #1410, opened 2023-04-20
1398 Kernel Extension (#1398 pt 3)

trisyoungs:
Having attempted to introduce energy and force calculation for arbitrary external potentials, it became clear that a bit more refactoring of the existing `EnergyKernel` and `ForceKernel` classes from which all other kernels will derive was necessary.

Briefly, the two kernels now funnel requests for total energy/force calculation through a single function in order to ensure that the external potential cannot be "missed" by calling one of the (previously) subordinate functions, e.g. for pair potential terms alone.  This clears up a lot of ambiguity, and makes the whole process clean and controllable through a `Flags<T>` based argument.

Stub virtuals for external energy / force factors are also added.


rprospero (approved):
This feels so much clearer than what we had before.

PR #1412, opened 2023-04-24
1306 clear module data

### LINKED ISSUE #1306: Clear all module data from specific layer
Add an option to the layer tab context menu (or somewhere else appropriate) to allow clearing of module data for all the modules within that layer.  Useful for occasions when a layer further down the chain remained activated by mistake while equilibrating / preparing a system, for instance.

Danbr4d:
Closes #1306. Adds in function to clear all module data from a layer

### DIFF
--- a/src/gui/gui_funcs.cpp
+++ b/src/gui/gui_funcs.cpp
@@ -28,7 +28,7 @@ DissolveWindow::DissolveWindow(Dissolve &dissolve)
     ui_.setupUi(this);
 
     // Set up the main tabs widget
-    ui_.MainTabs->setUp(this);
+    ui_.MainTabs->setUp();
 
     // Connect signals to thread controller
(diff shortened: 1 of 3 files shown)


Danbr4d:
I tried only changing setup() back but that gave the same error until I also changed reconcile() back. 
When using clearmoduledata() at the moment there is a crash as well

trisyoungs (approved):
:+1:

trisyoungs:
This works just fine, but after actually using it I think we should have a confirmation dialog to make sure that's what the user wants to do.  Should also have one in the original module-specific "Clear module data" option as well, in hindsight.

You can more-or-less copy the code from `src/gui/menu_simulation.cpp` starting on line 62 and change the message.

trisyoungs (approved):
:+1:

PR #1413, opened 2023-04-27
Access Node Data

trisyoungs:
### For Discussion!

This PR proposes a mechanism for to access various node data from `Expression`s.  The proposed approach is to use local parameters on any node that needs to expose numeric data following the existing `ParametersNode`, and storing values as `ExpressionVariable`s. This uses a lot of existing infrastructure, and is very easily implemented.

Since we demand that nodes (where required) have unique names, we can enforce a naming style in the form of `NodeName.Parameter` to also ensure uniqueness of exposed parameters. This requires nodes to react to name changes since this obviously alters the names of the data as well, hence the `ProcedureNode::setName()` function is now virtual.

The input file attached here provides a minimal example of a coordination-number calculation, and which raises at least one follow-on task in that the processed data has centre-bin values along the x axis, while now we want to retain the pure histogram bin values.
[cn.txt](https://github.com/disorderedmaterials/dissolve/files/11343967/cn.txt)

The PR also tidies up some issues with the `CalculateExpressionNode` which were not noticed at the time of it being implemented, purely because it had not been tested in a useful context!


trisyoungs:
There will be a few follow-on commits to this PR before its ready to merge in - while everything worked as it should with the test input file, doing the same programmatically was rather broken!

PR #1414, opened 2023-04-28
No Atom Array Indexing (#1398 pt 5)

trisyoungs:
This PR removes the need for storing atom array / vector indices in the `Atom` class, and removes the need for `AtomChangeToken`. At the time of creation it was something of an "optional" branch, but has since become more of a "yep, we should do this", at least in my eyes!

At present all possible use cases for Dissolve are for "static" configurations where the contents are created at the very beginning of the simulation and then do not mutate in terms of atom ordering or count. Thus, with this perfectly valid assumption we don't need to worry about changing atom indices and the subsequent need to update all atoms. Instead we can just reconstruct all pointer vectors within molecules at specific, known places in the code to ensure our data relationships all line up. Removal of molecules / atoms obviously needs to be treated carefully, but this particular aspect is resolved (fixed) in upcoming PR #1423.

The worst effect of removing atom array indices from `Atom`s was in the force calculation routines which always need to reference a global force vector via the atom index. However, by storing the root atom offset in the main vector within each `Molecule`, and using the fact that a `std::vector` is contiguous in memory, a little pointer arithmetic in the `Molecule` class allows the global index of a given `Atom` to be easily determined.

Performance-wise, most of this work had a positive effect with an anecdotal 10% speed improvement in the total force calculation benchmark for medium sized molecules.  Once the additional index calculation was included this decreased to around 5%, but still a slight net positive from other simplifications its now possible to make (nothing got worse, at least!).


src/classes/molecule.cpp
```diff
@@ -21,34 +21,37 @@ const Species *Molecule::species() const { return species_; }
...
-void Molecule::updateAtoms(std::vector<Atom> &source)
+// Update local atom pointers from main vector
+void Molecule::updateAtoms(std::vector<Atom> &mainAtoms, int offset)
 {
-    if (!atoms_.empty() && atoms_[0] != &source[atomIndices_[0]])
-        std::transform(atomIndices_.begin(), atomIndices_.end(), atoms_.begin(),
-                       [&source](const auto idx) { return &source[idx]; });
+    globalAtomOffset_ = offset;
+    for (auto &i : atoms_)
+        i = &mainAtoms[offset++];
```
rprospero, line 46:
```suggestion
    std::iota(atoms_.begin(), atoms_.end(), &mainAtoms[globalAtomOffset_]);
```

I'll chalk this up as being a matter of style.  It's a bit more explicit in our intention and doesn't rely on references or postfix operator precedence.  On the other hand, it also requires knowledge of iterators and the `iota` algorithm, so it may be a lateral move in readability.

src/modules/md/process.cpp
```diff
@@ -86,8 +86,9 @@ bool MDModule::process(Dissolve &dissolve, const ProcessPool &procPool)
             if (std::find(restrictToSpecies_.begin(), restrictToSpecies_.end(), mol->species()) != restrictToSpecies_.end())
             {
                 targetMolecules.push_back(mol.get());
+                auto offset = mol->globalAtomOffset();
                 for (const auto &i : mol->atoms())
-                    free[i->arrayIndex()] = 1;
+                    free[offset++] = 1;
```
rprospero, line 91:
``` c++
std::fill(free.begin()+offset, free.begin()+offset+mol->atoms().size(), 1);
```

Another possible algorithm replacement for the `for` loop.  GitHub won't let me suggest changes to lines that haven't been changed.

rprospero (approved):
I'll miss my AtomChangeToken class, but this is obviously the better way to handle things.  I've made a couple of algorithm suggestions, but I'll let you decide if they're clearer or more obfuscated that what you already wrote.

src/classes/molecule.cpp
trisyoungs: This is such an obvious use of `std::iota` I'm slightly ashamed I didn't think of it myself!

src/modules/md/process.cpp
trisyoungs: :+1:

PR #1416, opened 2023-05-02
Set up of histogramcn

### LINKED ISSUE #216: Histogram of coordination numbers (CalculateCN)
CalculateCN presently calculates and outputs the total average coordination number for each range, but it would be instructive / nice to have a histogram of the actual coordination numbers found about the central molecule.

Danbr4d:
Closes #216. Set up new module to get histogram of coordination numbers.


Danbr4d:
Just an initial push of the copied siterdf module and rename.

src/modules/histogramcn/functions.cpp
trisyoungs, line 1: I feel like this whole file is a bit redundant since we most likely will bin all of the functions here...

src/modules/histogramcn/gui/histogramcnwidget.h
```diff
@@ -0,0 +1,38 @@
...
+    HistogramCNModuleWidget(QWidget *parent, HistogramCNModule *module, Dissolve &dissolve);
+
+    /*
+     * UI
+     */
+    private:
+    // Main form declaration
+    Ui::HistogramCNModuleWidget ui_;
+    // DataViewer contained within this widget
+    DataViewer *rdfGraph_;
```
trisyoungs, line 33: Needs renaming to something like:
```suggestion
    DataViewer *cnGraph_;
```

src/modules/histogramcn/gui/histogramcnwidget.ui
```diff
@@ -0,0 +1,252 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>SiteRDFModuleWidget</class>
```
trisyoungs, line 3: Need to do a bit of search/replace in the ui file to change `SiteRDF` to `HistogramCN` - can do this outside of Qt designer  and use a straight text editor instead.

src/modules/histogramcn/gui/histogramcnwidget_funcs.cpp
```diff
@@ -0,0 +1,68 @@
...
+    ui_.setupUi(this);
+
+    // Set up RDF graph
+    rdfGraph_ = ui_.RDFPlotWidget->dataViewer();
+
+    auto &view = rdfGraph_->view();
+    view.setViewType(View::FlatXYView);
+    view.axes().setTitle(0, "\\it{r}, \\sym{angstrom}");
+    view.axes().setMax(0, 10.0);
+    view.axes().setTitle(1, "g(r)");
```
trisyoungs, line 23: Labelling needs to reflect coordination numbers now.

src/modules/widgetproducer.cpp
```diff
@@ -30,6 +30,8 @@
 #include "modules/geomopt/gui/geomoptwidget.h"
 #include "modules/gr/gr.h"
 #include "modules/gr/gui/grwidget.h"
+#include "modules/histogramcn/histogramcn.h"
+#include "modules/histogramcn/gyu/histogramcnwidget.h
```
trisyoungs, line 34:
```suggestion
#include "modules/histogramcn/gui/histogramcnwidget.h
```

trisyoungs (commented):
Yep, looking good so far.  The main class and GUI widget need a bunch of `SiteRDF`-specific variables and controls stripping out - I assume that was your next step anyway!

src/modules/histogramcn/gui/histogramcnwidget_funcs.cpp
Danbr4d: In QT designer I go rid of the RDF graph plot as I thought we would just add in a new graph plot when needed instead of altering the one we already have, should I add it back in and change the labels instead though?

CMakeSettings.json
trisyoungs, line 1: This file shouldn't be included in the PR / commit since it's a Visual Studio build file specific to you.

Danbr4d:
Causing a crash currently when histogram module is ran. Think it has something to do with line 47 in process.cpp

src/math/integerhistogram1d.cpp
```diff
@@ -127,10 +130,13 @@ long int IntegerHistogram1D::nBinned() const { return nBinned_; }
 void IntegerHistogram1D::accumulate()
 {
     for (auto &[key, value] : raw_)
-        averages_[key] += (double)raw_[key];
+        averages_.try_emplace(key, zeroCounter_).first->second += (double)raw_[key];
```
trisyoungs, line 133: This was quite an interesting problem.  Originally here we just killed the `bins_` map (emptied it completely) so we had no "memory" of what bins we had previously used.  This plays havoc when we start to accumulate an average since it is critical information to know that a bin remained empty for a given pass of the calculation. Hence, we just zero the counts now and retain the `bin_` map.

Furthermore, this did not accumulate the same number of data points for each bin, since we only create the `SampledDouble` in the `averages_` array after a bin is populated for the first time it.  If it isn't populated until 10 passes in, that bin will have 9 less data points (all zero) than other bins.  Hence, we now have a `zeroCounter_` tracking the number of passes of accumulation made, and inserting this value into the map when an `averages_` bin is first used.

src/modules/registry.cpp
```diff
@@ -66,6 +67,8 @@ ModuleRegistry::ModuleRegistry()
     registerProducer<GeometryOptimisationModule>(ModuleTypes::GeometryOptimisation,
                                                  "Optimise geometry with respect to energy (minimise)", "Optimisation");
     registerProducer<GRModule>(ModuleTypes::GR, "Calculate partial and total g(r)", "Correlation Functions");
+    registerProducer<HistogramCNModule>(ModuleTypes::HistogramCN, "Produce a histogram of Coordination numbers of an atom",
```
trisyoungs, line 70:
```suggestion
    registerProducer<HistogramCNModule>(ModuleTypes::HistogramCN, "Produce a histogram of coordination numbers between sites",
```

src/procedure/nodes/process1d.cpp
```diff
@@ -83,12 +98,18 @@ OptionalReferenceWrapper<ProcedureNodeSequence> Process1DProcedureNode::branch()
...
 bool Process1DProcedureNode::prepare(const ProcedureContext &procedureContext)
 {
-    if (!sourceData_)
-        return Messenger::error("No source Collect1D node set in '{}'.\n", name());
-
-    normalisationBranch_.prepare(procedureContext);
+    if (sourceData_ && sourceIntegerData_)
+    {
+        return Messenger::error("Too many arguments passed to process1d\n");
+    }
```
trisyoungs, line 104:
```suggestion
        return Messenger::error("Specify either SourceData or SourceIntegerData, not both.\n");
```

src/procedure/nodes/process1d.cpp
```diff
@@ -83,12 +98,18 @@ OptionalReferenceWrapper<ProcedureNodeSequence> Process1DProcedureNode::branch()
...
+    else if (sourceData_ || sourceIntegerData_)
+    {
+        normalisationBranch_.prepare(procedureContext);
 
-    return true;
+        return true;
+    }
+    else
+        return Messenger::error("No source Collect1D or IntegerCollect1D node set in '{}'.\n", name());
```
trisyoungs, line 112:
```suggestion
        return Messenger::error("No source data node set in '{}'.\n", name());
```

src/procedure/nodes/process1d.cpp
```diff
@@ -83,12 +98,18 @@ OptionalReferenceWrapper<ProcedureNodeSequence> Process1DProcedureNode::branch()
 // Prepare any necessary data, ready for execution
 bool Process1DProcedureNode::prepare(const ProcedureContext &procedureContext)
 {
-    if (!sourceData_)
-        return Messenger::error("No source Collect1D node set in '{}'.\n", name());
-
-    normalisationBranch_.prepare(procedureContext);
+    if (sourceData_ && sourceIntegerData_)
```
trisyoungs, line 101: I like the logic here! :+1:

src/procedure/nodes/process1d.cpp
```diff
@@ -101,10 +122,20 @@ bool Process1DProcedureNode::finalise(const ProcedureContext &procedureContext)
...
+        else
+            data = sourceData_->accumulatedData();
+    }
+    else if (sourceIntegerData_)
+    {
+        if (instantaneous_)
+            data = sourceIntegerData_->data();
+        else
+            data = sourceIntegerData_->accumulatedData();
+    }
```
trisyoungs, line 138: Can simplify to:
```suggestion
    if (instantaneous_)
        data = sourceData_ ? sourceData_->data() : sourceIntegerData_->data();
    else
        data = sourceData_ ? sourceData_->accumulatedData() : sourceIntegerData_->accumulatedData();
```

trisyoungs (commented):
Looking good. Will add some more system tests.

src/math/integerhistogram1d.cpp
```diff
@@ -127,10 +130,13 @@ long int IntegerHistogram1D::nBinned() const { return nBinned_; }
...
 
     // Update accumulated data
     updateAccumulatedData();
+
+    // Accumulate zero counter
+    zeroCounter_ += 0.0;
```
rprospero, line 139: Could the comment here be a bit more verbose?  This line of code feels intuitively feels like a no-op.  That's obviously not the case, but it might be worth explaining in a bit more detail.

rprospero (approved):
One bit of clarification that I would appreciate, but things otherwise look good.

src/math/integerhistogram1d.cpp
trisyoungs: Done!

trisyoungs (approved):

PR #1422, opened 2023-05-06
Targeted Potentials (#1398 pt 6)

trisyoungs:
This PR adds in "targeted" potentials to complement the recently-added global potentials. A targeted potential affects a specific subset of atoms, or even a single atom.

Works towards #1398.

### DIFF
--- a/src/classes/atom.h
+++ b/src/classes/atom.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "classes/speciesatom.h"
+#include "kernels/potentials/base.h"
 #include "templates/vector3.h"
 #include <memory>
 #include <vector>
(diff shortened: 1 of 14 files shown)


src/classes/atom.h
```diff
@@ -94,4 +95,19 @@ class Atom
     public:
     // Return scaling type and factors (electrostatic, van der Waals) to employ with specified Atom
     SpeciesAtom::ScaledInteractionDefinition scaling(const Atom *j) const;
+
+    /*
+     * Targeted External Potentials
+     */
+    private:
+    // Vector of targeted potentials affecting this atom
+    std::vector<const ExternalPotential *> targetedPotentials_;
```
rprospero, line 104:
```suggestion
    std::vector<std::reference_wrappr<const ExternalPotential>> targetedPotentials_;
```

I'm offering the suggestion because we've used this pattern at other places in the code, but I'm not convinced that it offers significant benefits.  We're now at the point where *ownership* is always denoted with a `shared_ptr` or `unique_ptr`.  Using a reference makes it difficult to pass a `nullptr`, but not impossible, so we're not really getting any guarantees from the compiler.

rprospero (approved):
Everything looks good.  I've included one suggestion about references versus pointers, but I'm very ambivalent about it.

src/classes/atom.h
trisyoungs: But are those other uses referencing an actual object in an vector, rather than a `shared_ptr`- or `unique-ptr`-managed one?  To me it feels like storing a raw pointer to an object whose existence is managed by `shared_ptr` or `unique_ptr` gives a clearer picture of the intent / architecture. I love a reference where possible, but here it feels like a long-term reference to a managed pointer object is a bit misleading.

PR #1431, opened 2023-05-23
set up of IntegerHistogram1D

### LINKED ISSUE #1429: Fully integer-binned histogram class
For simple counted statistical quantities (i.e. populations etc.) we require an analogous class to `Histogram1D` which handles integer binning and outliers properly.

## Workload
- [x] Base class off `Histogram1D` - remove unnecessary functions (e.g. `allSum()`)
- [ ] Define integer limits (and bin width, or enforce this to `1`?)
- [ ] Handle outliers to defined range in separate bins

Danbr4d:
Closes #1429. For simple counted statistical quantities (i.e. populations etc.) we require an analogous class to Histogram1D which handles integer binning and outliers properly.

### DIFF
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -34,7 +34,7 @@ void IntegerHistogram1D::clear()
 // Update accumulated data
 void IntegerHistogram1D::updateAccumulatedData()
 {
-    std::pair<Data1D, std::optional<int>> average = createDisplayData();
+    auto average = createDisplayData();
     accumulatedData_ = average.first;
     auto minBin = average.second.value_or(0);
     // Poke bin values and errors into array
@@ -46,8 +46,8 @@ void IntegerHistogram1D::updateAccumulatedData()
     }
 }
 
-// Create dsiplay data
-const std::pair<Data1D, std::optional<int>> IntegerHistogram1D::createDisplayData()
+// Create dsiplay data object covering extents of current bins
+const std::pair<Data1D, int> IntegerHistogram1D::createDisplayData()
 {
 
     // Get limiting key values
@@ -89,16 +89,12 @@ void IntegerHistogram1D::initialise(std::optional<int> xMin, std::optional<int> 
 
     minimum_ = xMin;
     maximum_ = xMax;
-
-    raw_.clear();
-    averages_.clear();
 }
 
 // Zero histogram bins
 void IntegerHistogram1D::zeroBins()
 {
     raw_.clear();
-    averages_.clear();
     nBinned_ = 0;
     nMissed_ = 0;
 }
@@ -131,7 +127,7 @@ long int IntegerHistogram1D::nBinned() const { return nBinned_; }
 void IntegerHistogram1D::accumulate()
 {
     for (auto &[key, value] : raw_)
-        averages_[key] += (SampledDouble)raw_[key];
+        averages_[key] += (double)raw_[key];
 
     // Update accumulated data
     updateAccumulatedData();
@@ -143,7 +139,7 @@ Data1D IntegerHistogram1D::data() const
     Data1D result = accumulatedData_;
     for (auto &[key, value] : raw_)
     {
-        result.values()[(key)] = value;
+        result.values()[key] = value;
     }
     return result;
 }
@@ -175,10 +171,11 @@ bool IntegerHistogram1D::deserialise(LineParser &parser)
 
     auto nBins = parser.argli(0);
 
-    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
-        return false;
     for (int n = 0; n <= nBins; ++n)
+
     {
+        if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+            return false;
         auto bin = parser.argi(0);
         raw_[bin] = parser.argli(1);
         averages_[bin].deserialise(parser);
@@ -202,12 +199,9 @@ bool IntegerHistogram1D::serialise(LineParser &parser) const
     {
         if (!parser.writeLineF("{} {} \n", key, value))
             return false;
-    }
-
-    for (auto &[key, value] : averages_)
-    {
-        if (!value.serialise(parser))
+        if (!averages_[key].serialise(parser))
             return false;
     }
+
     return true;
 }
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -36,7 +36,7 @@ void IntegerHistogram1D::updateAccumulatedData()
 {
     auto average = createDisplayData();
     accumulatedData_ = average.first;
-    auto minBin = average.second.value_or(0);
+    auto minBin = average.second;
     // Poke bin values and errors into array
     for (const auto &[key, value] : averages_)
     {
@@ -79,7 +79,7 @@ const std::pair<Data1D, int> IntegerHistogram1D::createDisplayData()
     for (auto n = 0; n < expectedNBins; ++n)
         data.xAxis(n) = x++;
 
-    return std::make_pair(data, expectedMinimum);
+    return std::make_pair(data, expectedMinimum.value());
 }
 
 // Initialise with specified bin range
@@ -199,7 +199,7 @@ bool IntegerHistogram1D::serialise(LineParser &parser) const
     {
         if (!parser.writeLineF("{} {} \n", key, value))
             return false;
-        if (!averages_[key].serialise(parser))
+        if (!averages_.at(key).serialise(parser))
             return false;
     }
 
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -130,8 +130,8 @@ long int IntegerHistogram1D::nBinned() const { return nBinned_; }
 // Accumulate current histogram bins into averages
 void IntegerHistogram1D::accumulate()
 {
-    for (auto &[key, value] : averages_)
-        value += (SampledDouble)raw_[key];
+    for (auto &[key, value] : raw_)
+        averages_[key] += (SampledDouble)raw_[key];
 
     // Update accumulated data
     updateAccumulatedData();
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -74,11 +74,9 @@ void IntegerHistogram1D::initialise(std::optional<int> xMin, std::optional<int> 
 {
     clear();
 
-    // Store axis limits and binWidth, and then clamp as necessary
     minimum_ = xMin;
     maximum_ = xMax;
 
-    // Create the arrays
     bins_.clear();
 }
 
@@ -90,10 +88,10 @@ void IntegerHistogram1D::zeroBins()
     nMissed_ = 0;
 }
 
-// Return minimum value for data (hard left-edge of first bin)
+// Return minimum value for data
 std::optional<int> IntegerHistogram1D::minimum() const { return minimum_; }
 
-// Return maximum value for data (hard right-edge of last bin, adjusted to match bin width if necessary)
+// Return maximum value for data
 std::optional<int> IntegerHistogram1D::maximum() const { return maximum_; }
 
 // Bin specified value, returning success
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -19,9 +19,9 @@ IntegerHistogram1D::IntegerHistogram1D(const IntegerHistogram1D &source) { (*thi
 // Clear Data
 void IntegerHistogram1D::clear()
 {
-    minimum_ = 0.0;
-    maximum_ = 0.0;
-    binWidth_ = 0.0;
+    minimum_ = 0;
+    maximum_ = 0;
+    binWidth_ = 1;
     nBins_ = 0;
     nBinned_ = 0;
     nMissed_ = 0;
@@ -50,7 +50,7 @@ void IntegerHistogram1D::updateAccumulatedData()
 }
 
 // Initialise with specified bin range
-void IntegerHistogram1D::initialise(double xMin, double xMax, double binWidth)
+void IntegerHistogram1D::initialise(int xMin, int xMax, int binWidth)
 {
     clear();
 
@@ -76,12 +76,11 @@ void IntegerHistogram1D::zeroBins()
 }
 
 // Set up supplied axis
-void IntegerHistogram1D::setUpAxis(double axisMin, double &axisMax, double binWidth, int &nBins,
-                                   std::vector<double> &binCentres)
+void IntegerHistogram1D::setUpAxis(int axisMin, int &axisMax, int binWidth, int &nBins, std::vector<int> &binCentres)
 {
     // Min, max, and bin width should be set to requested values initially
     // We will clamp the maximum to the nearest bin boundary (not less than the supplied axisMax)
-    double range = axisMax - axisMin;
+    int range = axisMax - axisMin;
     nBins = int(range / binWidth);
     if ((axisMin + nBins * binWidth) < axisMax)
     {
@@ -92,25 +91,25 @@ void IntegerHistogram1D::setUpAxis(double axisMin, double &axisMax, double binWi
     // Create centre-bin array
     binCentres.clear();
     binCentres.resize(nBins);
-    double centre = axisMin + binWidth * 0.5;
+    int centre = axisMin + binWidth * 0.5;
     for (auto n = 0; n < nBins; ++n, centre += binWidth)
         binCentres[n] = centre;
 }
 
 // Return minimum value for data (hard left-edge of first bin)
-double IntegerHistogram1D::minimum() const { return minimum_; }
+int IntegerHistogram1D::minimum() const { return minimum_; }
 
 // Return maximum value for data (hard right-edge of last bin, adjusted to match bin width if necessary)
-double IntegerHistogram1D::maximum() const { return maximum_; }
+int IntegerHistogram1D::maximum() const { return maximum_; }
 
 // Return bin width
-double IntegerHistogram1D::binWidth() const { return binWidth_; }
+int IntegerHistogram1D::binWidth() const { return binWidth_; }
 
 // Return number of bins
 int IntegerHistogram1D::nBins() const { return nBins_; }
 
 // Bin specified value, returning success
-bool IntegerHistogram1D::bin(double x)
+bool IntegerHistogram1D::bin(int x)
 {
     // Calculate target bin
     auto bin = int((x - minimum_) / binWidth_);
@@ -135,14 +134,14 @@ long int IntegerHistogram1D::nBinned() const { return nBinned_; }
 void IntegerHistogram1D::accumulate()
 {
     for (auto n = 0; n < nBins_; ++n)
-        averages_[n] += double(bins_[n]);
+        averages_[n] += int(bins_[n]);
 
     // Update accumulated data
     updateAccumulatedData();
 }
 
 // Return Array of x centre-bin values
-const std::vector<double> &IntegerHistogram1D::binCentres() const { return binCentres_; }
+const std::vector<int> &IntegerHistogram1D::binCentres() const { return binCentres_; }
 
 // Return histogram data
 std::vector<long int> &IntegerHistogram1D::bins() { return bins_; }
@@ -243,19 +242,5 @@ bool IntegerHistogram1D::serialise(LineParser &parser) const
         if (!averages_.at(n).serialise(parser))
             return false;
 
-    return true;
-}
-
-/*
- * Parallel Comms
- */
-
-// Sum histogram data onto all processes
-bool IntegerHistogram1D::allSum(const ProcessPool &procPool, OptionalReferenceWrapper<Timer> commsTimer)
-{
-#ifdef PARALLEL
-    if (!procPool.allSum(bins_.data(), nBins_, ProcessPool::PoolProcessesCommunicator, commsTimer))
-        return false;
-#endif
     return true;
 }
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -20,8 +20,6 @@ IntegerHistogram1D::IntegerHistogram1D(const IntegerHistogram1D &source) { (*thi
 // Clear Data
 void IntegerHistogram1D::clear()
 {
-    minimum_ = 0;
-    maximum_ = std::nullopt;
     nBinned_ = 0;
     nMissed_ = 0;
     bins_.clear();
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -28,6 +28,7 @@ void IntegerHistogram1D::clear()
     bins_.clear();
     binCentres_.clear();
     averages_.clear();
+    histMap_.clear();
 }
 
 /*
@@ -173,6 +174,14 @@ Data1D IntegerHistogram1D::data() const
 
 // Return accumulated (averaged) data
 const Data1D &IntegerHistogram1D::accumulatedData() const { return accumulatedData_; }
+
+bool IntegerHistogram1D::insertIntoMap(int x)
+{
+    auto bin = int((x - minimum_));
+
+    ++histMap_[bin];
+    ++nBinned_;
+}
 
 /*
  * Operators
--- a/src/math/integerhistogram1d.cpp
+++ b/src/math/integerhistogram1d.cpp
@@ -23,7 +23,8 @@ void IntegerHistogram1D::clear()
 {
     nBinned_ = 0;
     nMissed_ = 0;
-    bins_.clear();
+    raw_.clear();
+    averages_.clear();
 }
 
 /*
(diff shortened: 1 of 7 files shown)


src/math/integerhistogram1d.h
```diff
@@ -0,0 +1,101 @@
...
+    // Maximum value for data (hard right-edge of last bin, adjusted to match bin width if necessary)
+    int maximum_;
+    // Bin width
+    int binWidth_;
+    // Number of bins
+    int nBins_;
+    // Histogram bins
+    std::vector<long int> bins_;
+    // Array of bin centres
+    std::vector<int> binCentres_;
```
trisyoungs, line 35: All of these member variables can be deleted since we don't want / need them any more.

```suggestion
```

src/math/integerhistogram1d.h
```diff
@@ -0,0 +1,101 @@
...
+    void clear();
+
+    /*
+     * Histogram Data
+     */
+    private:
+    // Minimum value for data (hard left-edge of first bin)
+    int minimum_;
+    // Maximum value for data (hard right-edge of last bin, adjusted to match bin width if necessary)
+    int maximum_;
```
trisyoungs, line 27: I'm inclined to keep some sort of limits in since we might want to restrict against binning "silly" values. I would suggest that we change them to `std::optional`s to make sure their intent is clear, but keep the zero as the minimum limit since preventing negative numbers seems sensible (but perhaps not always desired!).  I've changed the comments too.
```suggestion
    // Minimum value allowed for binning
    std::optional<int> minimum_{0};
    // Maximum value allowed for binning
    std::optional<int> maximum_;
```

src/math/integerhistogram1d.h
```diff
@@ -0,0 +1,101 @@
...
+    // Accumulated averages
+    std::vector<SampledDouble> averages_;
+    // Number of values binned over all bins
+    long int nBinned_;
+    // Number of points missed (out of bin range)
+    long int nMissed_;
+    // Accumulated data
+    Data1D accumulatedData_;
+    // key is the bin and value is the count
+    std::map<int, long int> histMap_;
```
trisyoungs, line 45: As mentioned above, we need to tie together the "instantaneous" binned values and their running averages, which was previously managed by two `std::vector`s (`bins_` and `averages_`) of the same size. My suggestion would be to use a `std::pair`.  Also renamed it to be `bins_` since we can use it with the old one removed.
```suggestion
    // Map of histogram bin values to instantaneous counts and accumulated averages
    std::map<int, std::pair<long int, SampledDouble>> bins_;
```

src/math/integerhistogram1d.h
```diff
@@ -0,0 +1,101 @@
...
+    std::vector<long int> &bins();
+    const std::vector<long int> &bins() const;
+    // Add source histogram data into local array
+    void add(IntegerHistogram1D &other, int factor = 1);
+    // Return current data
+    Data1D data() const;
+    // Return accumulated (averaged) data
+    const Data1D &accumulatedData() const;
+    // insert value into map
+    bool insertIntoMap(int x);
```
trisyoungs, line 84: Probably best to rewrite the `bin(double x)` to take an `int` rather than having a separate function for it.

Danbr4d:
Commented out some functions as they need editing to work with new parameters but will add them back in iteratively to check they work properly

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// Return number of values binned over all bins
+long int IntegerHistogram1D::nBinned() const { return nBinned_; }
+
+// Accumulate current histogram bins into averages
+void IntegerHistogram1D::accumulate()
+{
+    for (auto &[key, value] : bins_)
+        averages_[value.first] += bins_[value.first].first;
```
Danbr4d, line 112: Having issues with this operator as "more than one operand matches it in SampledDouble"

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if (!averages_[value.first].deserialise(parser))
+            return false;
+
+    return true;
+}
+
+// Write data through specified LineParser
+bool IntegerHistogram1D::serialise(LineParser &parser) const
+{
+    if (!parser.writeLineF("{} {} {}\n", minimum_, maximum_, 1))
```
Danbr4d, line 202: I think the optinalInt for maximum_ and minimum_ are causing problems here as well

trisyoungs: We'll need to write the information out a bit differently, I think, since both limits are optional.  I'd suggest something like this:
```suggestion
    if (!parser.writeLineF("{} {} {} {}\n", Dissolve::btoa(minimum_), minimum_.value_or(0), Dissolve::btoa(maximum_), maximum_.value_or(0)))
```

This would write out a line like this (for the default values of `minimum_ = 0` and `maximum_` as undefined:
```
True  0  False  0
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,211 @@
...
+bool IntegerHistogram1D::bin(int x)
+{
+    auto bin = x;
+    if (bins_.find(bin) == bins_.end())
+        bins_[bin] = std::make_pair(1, averages_);
+    else
+    {
+        bins_[bin].first++;
+        bins_[bin].second = averages_;
+    }
```
trisyoungs, line 101: Couple of things here.  First, the independent `averages_` vector is no longer needed, since we're tying the average value to the bin.  Second, the pair should be `int, SampledDouble` rather than a vector for the latter.  Third, you should be able to simplify the code here by using the `std::map` feature that returns a new value if the accessed key isn't currently present:
```suggestion
bins_[x].first++;
```
Glorious, huh? :smile:

Also need to do the check on min/max limits here before we do the binning.

src/math/integerhistogram1d.cpp
trisyoungs: Hmmm.  Not sure why that's happening, but once `averages_` is removed the loop would look something like this (I think!):
```suggestion
        value.second += value.first;
```
Basically, just summing in the current counts (first) to the associated average (second)

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    // Store bin centres and accumulated averages in the object
+    for (auto n = 0; n < bins_.size(); ++n)
+    {
+        accumulatedData_.xAxis(n) = n;
+    }
+}
+
+// Initialise with specified bin range
+void IntegerHistogram1D::initialise(int xMin, int xMax, int binWidth)
```
trisyoungs, line 48: This function needs a bit of modifiication - it doesn't need the `binWidth` argument any more, and the min/max arguments should be `std::optional<int>`s.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,202 @@
...
+    // Min, max, and bin width should be set to requested values initially
+    // We will clamp the maximum to the nearest bin boundary (not less than the supplied axisMax)
+    int range = axisMax - axisMin;
+    nBins = range;
+    if ((axisMin + nBins * binWidth) < axisMax)
+    {
+        ++nBins;
+        axisMax = axisMin + nBins * binWidth;
+    }
+}
```
trisyoungs, line 80: This function I think is redundant now - we don't currently use it, and the function it performs needs to be essentially rewritten in `updateAccumulatedData()`.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,202 @@
...
+void IntegerHistogram1D::updateAccumulatedData()
+{
+    // Set up arrays
+    accumulatedData_.initialise(bins_.size(), true);
+
+    // Store bin centres and accumulated averages in the object
+    for (auto n = 0; n < bins_.size(); ++n)
+    {
+        accumulatedData_.xAxis(n) = n;
+    }
```
trisyoungs, line 44: This is where things aren't as straightforward as for the well-defined, Histogram1D case. In the worst case we have no idea what the limits of the data are, and so need to work them out from the data contained in the  `std::map` `bins_`. Once the limits are established the `Data1D` object (`accumulatedData_`) can be initialised to the correct size, and then the `SampledDouble` values contained in `bins_` "re-binned" into the `Data1D`.  Hopefully this makes sense!  If you look at the unit test I just added you'll see the first part of this (the limit determination) written out already which you can steal.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,202 @@
...
+    clear();
+}
+
+IntegerHistogram1D::IntegerHistogram1D(const IntegerHistogram1D &source) { (*this) = source; }
+
+// Clear Data
+void IntegerHistogram1D::clear()
+{
+    minimum_ = 0;
+    maximum_ = 0;
```
trisyoungs, line 24: Need to be careful here, since this will reset the limits to an unusable range. 
```suggestion
    minimum_ = 0;
    maximum_ = std::nullopt;
```

src/math/integerhistogram1d.cpp
Danbr4d: I've noticed that this function is meant to be used in the `initialise` but isn't, do I add it back in there and keep the `setUpAxis()` function or should I still move it to `updateaccumulatedData()`?

trisyoungs: As written this is only useful for static axes with defined min/max, but our histogram is never guaranteed to have unchanging min/max bin values.  You've already been accounting for that with the expectedMinimum and maximum stuff, so I would just bin this code entirely.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,218 @@
...
+Data1D IntegerHistogram1D::data() const
+{
+    Data1D result = accumulatedData_;
+    for (auto &[key, value] : bins_)
+    {
+        result.values().push_back(value.first);
+    }
+    /*std::copy(bins_.begin(), bins_.end(), result.values().begin());*/
+    return result;
+}
```
trisyoungs, line 152: I'd have to check where exactly this gets used in the code, but as-written the function copies `accumulatedData_` (which may not yet exist...) and then _appends_ values to the back of the array, rather than setting them.  The actual function here would be quite similar to the `updateAccumulatedData()`, so we should probably consider taking the guts of that function out into a separate private function `const Data1D::createtDisplayData()` which does all the min/max determination and sets up the x-axis with the bin values.  Then the `data()` and `updateAccumulatedData()` functions can call that and just set the value/error data differently.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,218 @@
...
+//
+//     IntegerHistogram1D ret = *this;
+//
+//     std::transform(other.bins_.cbegin(), other.bins_.cend(), ret.bins_.cbegin(), ret.bins_.begin(), std::plus<>());
+//
+//     ret.nBinned_ = this->nBinned_ + other.nBinned_;
+//     ret.nMissed_ = this->nMissed_ + other.nMissed_;
+//
+//     return ret;
+// }
```
trisyoungs, line 183: Let's delete these fully for now - no point in worrying about writing them until we need them.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// Write data through specified LineParser
+bool IntegerHistogram1D::serialise(LineParser &parser) const
+{
+    if (!parser.writeLineF("{} {} {} {}\n", DissolveSys::btoa(minimum_.has_value()), minimum_.value_or(0),
+                           DissolveSys::btoa(maximum_.has_value()), maximum_.value_or(0)))
+        return false;
+    if (!parser.writeLineF("{}  {}\n", nBinned_, nMissed_))
+        return false;
```
trisyoungs, line 213: The `serialise` function needs to write out the `bins_` data.  It will need to write out the number of keys in the `bins_` map on the next line, followed by the key, count, and SampledDouble data for each bin in the map (one set of bin data per line).

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,218 @@
...
+ */
+
+// Read data through specified LineParser
+bool IntegerHistogram1D::deserialise(LineParser &parser)
+{
+    clear();
+
+    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+        return false;
+    initialise(parser.argd(0), parser.argd(1));
```
trisyoungs, line 196: This needs adjusting to reflect format of the limits written in the `serialise()` function.

trisyoungs (commented):
I've fixed up a few things here - see if all the commits make sense!  Have extended the unit test and now they all pass.

Made some comments on the final remaining pieces of work (mostly boring I/O stuff, but necessary!).

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,221 @@
...
+        auto n = key - minBin;
+        accumulatedData_.value(n) = value.second.value();
+        accumulatedData_.error(n) = value.second.stDev();
+    }
+}
+
+// Create dsiplay data
+const Data1D IntegerHistogram1D::createDisplayData()
+{
+    Data1D data;
```
trisyoungs, line 50: Just for "data locality" I would delete this from here and move it further down to where it is first used.
```suggestion
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,221 @@
...
+        expectedMinimum = expectedMaximum;
+    else if (!expectedMaximum)
+        expectedMaximum = expectedMinimum;
+
+    auto expectedNBins = (*expectedMaximum - *expectedMinimum) + 1;
+
+    // Set up data
+    data.initialise(expectedNBins, true);
+    auto x = *expectedMinimum;
+    for (auto n = 0; n < expectedNBins; n++)
```
trisyoungs, line 76: Being super-picky, but a prefix increment (`++n`) requires less CPU operations than a postfix increment (`n++`), and either could be used here.
```suggestion
    for (auto n = 0; n < expectedNBins; ++n)
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,221 @@
...
+void IntegerHistogram1D::add(IntegerHistogram1D &other, int factor)
+{
+    if (bins_.size() != other.bins_.size())
+    {
+        Messenger::print("BAD_USAGE - Can't add IntegerHistogram1D data since arrays are not the same size ({} vs {}).\n",
+                         bins_.size(), other.bins_.size());
+        return;
+    }
+    for (auto &[key, value] : bins_)
+
```
trisyoungs, line 147: Just moving the blank line.
```suggestion

    for (auto &[key, value] : bins_)
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,221 @@
...
+    nMissed_ = parser.argli(1);
+
+    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+        return false;
+
+    auto nBins = parser.argli(0);
+
+    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+        return false;
+    for (int n; n <= nBins; n++)
```
trisyoungs, line 191: Your start value is undefined here which is a bit dangerous! Loop starts at zero so can use the condition `n < nBins`. Plus `auto` and the prefix `++`.
```suggestion
    for (auto n = 0; n < nBins; ++n)
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,221 @@
...
+    {
+        bins_[parser.argli(0)].first = parser.argli(1);
+        bins_[parser.argli(0)].second.deserialise(parser);
```
trisyoungs, line 194: Yes, that's the way to do it!  I'd make a small correction just to be on the safe side as well as avoiding doing the string-to-int conversion twice:
```suggestion
        auto bin = parser.argi(0);
        bins_[bin].first = parser.argli(1);
        bins_[bin].second.deserialise(parser);
```

trisyoungs (changes requested):
Pretty much all good - one minor issue and a few picky corrections to make, then I think you can re-request a review from @rprospero.

src/math/integerhistogram1d.h
```diff
@@ -0,0 +1,76 @@
...
+    /*
+     * Histogram Data
+     */
+    private:
+    // Minimum value for binning
+    std::optional<int> minimum_{0};
+    // Maximum value for data (hard right-edge of last bin, adjusted to match bin width if necessary)
+    std::optional<int> maximum_;
+    // Number of values binned over all bins
+    long int nBinned_;
```
rprospero, line 30: I might have missed something, but is `nBinned_` ever not zero?  It doesn't seem to get updated when new elements are added to the histogram?  I'm not quite sure that I'm understanding properly, but the comment makes it sound like a copy of `bins_.size()`

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+// Add source histogram data into local array
+void IntegerHistogram1D::add(IntegerHistogram1D &other, int factor)
+{
+    if (bins_.size() != other.bins_.size())
+    {
+        Messenger::print("BAD_USAGE - Can't add IntegerHistogram1D data since arrays are not the same size ({} vs {}).\n",
+                         bins_.size(), other.bins_.size());
+        return;
+    }
```
rprospero, line 146: Is this a significant restriction?  We're using maps, not arrays, so we can add them and get a union of the data.  Conversely, two histograms might have the same number of bins, but different bin values, and this function will only update the bins that are in the `this` array, ignoring any bins that it didn't have.

rprospero (commented):
I've got a few questions below on bits that I didn't quite understand, along with some suggestions.

src/math/mathfunc.cpp
```diff
@@ -89,4 +89,7 @@ double sgn(double a, double signOf) { return signOf >= 0.0 ? fabs(a) : -fabs(a);
 // Return the cyclic permutation of the integer 'i', span 3
 int cp3(int i) { return (i % 3); }
 
+// Nearest Integer Function
+int nInt(double x) { return (int)(x >= 0.0 ? (x + 0.5) : (x - 0.5)); }
+
```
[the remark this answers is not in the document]
trisyoungs: My concern with that was that you still have to cast to an `int`, and if the result of `std::lround` is, for example, `2.999999999997` then we'll convert potentially down to `int` `2`.  However, just saw that a function `std::lrint` exists and which we should use instead of implementing our own `nInt` function.

src/math/integerhistogram1d.h
trisyoungs: Good catch - there should be a `++nBinned_` after `bins_[x].first++;` on line 116.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    for (auto &[key, value] : bins_)
+
+        bins_[value.first].first += other.bins_[value.first].first * factor;
```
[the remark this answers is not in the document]
trisyoungs: Another good catch!  And fair comment on the unit test - I'll add one once the other changes are made.

src/math/integerhistogram1d.h
```diff
@@ -0,0 +1,76 @@
...
+    // Number of points missed (out of bin range)
+    long int nMissed_;
+    // Accumulated data
+    Data1D accumulatedData_;
+    // Map of histogram bin values to instantaneous counts and accumulated averages
+    std::map<int, std::pair<long int, SampledDouble>> bins_;
```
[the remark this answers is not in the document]
trisyoungs: That would be more in line with the implementations in `Histogram1D` etc., but here we elected to go with the `std::pair` to tie the counts and averages together, mostly to ensure parity between the two quantities.

src/math/integerhistogram1d.cpp
trisyoungs: This is just old code that hasn't been updated yet - we'll add it to the To-Do!

Danbr4d: So is a check needed instead to see if the same bin values are present before adding them and if not it needs to add in the missing ones too?

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/*
+ * Data
+ */
+
+// Update accumulated data
+void IntegerHistogram1D::updateAccumulatedData()
+{
+    std::pair<Data1D, std::optional<int>> average = createDisplayData();
```
trisyoungs, line 37: For cleanliness this can be `auto`:
```suggestion
    auto average = createDisplayData();
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+    for (const auto &[key, value] : averages_)
+    {
+        auto n = key - minBin;
+        accumulatedData_.value(n) = value.value();
+        accumulatedData_.error(n) = value.stDev();
+    }
+}
+
+// Create dsiplay data
+const std::pair<Data1D, std::optional<int>> IntegerHistogram1D::createDisplayData()
```
trisyoungs, line 50: We explicitly test for the optional limits and set them to zero (or the other limit if only one bin exists) so this function can return just `std::pair<Data1D, int>`.

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+// Initialise with specified bin range
+void IntegerHistogram1D::initialise(std::optional<int> xMin, std::optional<int> xMax)
+{
+    clear();
+
+    minimum_ = xMin;
+    maximum_ = xMax;
+
+    raw_.clear();
+    averages_.clear();
```
trisyoungs, line 94: Can delete these since it's done in `clear()`.
```suggestion

```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Zero histogram bins
+void IntegerHistogram1D::zeroBins()
+{
+    raw_.clear();
+    averages_.clear();
```
trisyoungs, line 101: Don't want to clear the averages here - they're long lived over many accumulations - so just the bin counts.
```suggestion
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Return number of values binned over all bins
+long int IntegerHistogram1D::nBinned() const { return nBinned_; }
+
+// Accumulate current histogram bins into averages
+void IntegerHistogram1D::accumulate()
+{
+    for (auto &[key, value] : raw_)
+        averages_[key] += (SampledDouble)raw_[key];
```
trisyoungs, line 134: Small thing, but we can convert to `double` here instead of `SampledDouble` to keep things simple.
```suggestion
        averages_[key] += (double)raw_[key];
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+    updateAccumulatedData();
+}
+
+// Return current data
+Data1D IntegerHistogram1D::data() const
+{
+    Data1D result = accumulatedData_;
+    for (auto &[key, value] : raw_)
+    {
+        result.values()[(key)] = value;
```
trisyoungs, line 146: Few unnecessary brackets.
```suggestion
        result.values()[key] = value;
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+        return false;
+    for (int n = 0; n <= nBins; ++n)
+    {
+        auto bin = parser.argi(0);
+        raw_[bin] = parser.argli(1);
+        averages_[bin].deserialise(parser);
+    }
+    return true;
```
trisyoungs, line 186: We don't cover this (de)serialise yet with a system test - we will when the Histogram module comes along - but this definitely wouldn't have been happy!
```suggestion
    for (auto n = 0; n < nBins; ++n)
    {
        if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
            return false;

        auto bin = parser.argi(0);
        raw_[bin] = parser.argli(1);
        averages_[bin].deserialise(parser);
    }
    
    return true;
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+    {
+        if (!parser.writeLineF("{} {} \n", key, value))
+            return false;
+    }
+
+    for (auto &[key, value] : averages_)
+    {
+        if (!value.serialise(parser))
+            return false;
+    }
```
trisyoungs, line 211: Need to write the bin counts and averages as pairs, per bin:
```suggestion
    {
        if (!parser.writeLineF("{} {} \n", key, value))
            return false;
        if (!averages_[key].serialise(parser))
            return false;
    }
```

src/math/integerhistogram1d.h
```diff
@@ -0,0 +1,72 @@
...
+    // Accumulated data
+    Data1D accumulatedData_;
+    // Map of histogram bin values to instantaneous counts and accumulated averages
+    std::map<int, long int> raw_;
+    std::map<int, SampledDouble> averages_;
+
+    private:
+    // Update accumulated data
+    void updateAccumulatedData();
+    const std::pair<Data1D, std::optional<int>> createDisplayData();
```
trisyoungs, line 42: Just need a comment for the function
```suggestion
    // Create display data object covering extents of current bins
    const std::pair<Data1D, std::optional<int>> createDisplayData();
```

src/math/integerhistogram1d.cpp
```diff
@@ -0,0 +1,213 @@
...
+    // Poke bin values and errors into array
+    for (const auto &[key, value] : averages_)
+    {
+        auto n = key - minBin;
+        accumulatedData_.value(n) = value.value();
+        accumulatedData_.error(n) = value.stDev();
+    }
+}
+
+// Create dsiplay data
```
trisyoungs, line 49: Bit more detail:
```suggestion
// Create display data object covering extents of current bins
```

src/math/integerhistogram1d.cpp
Danbr4d: So then will the return for the function `return expectedMinimum.value()`?

trisyoungs: Yep, exactly that.

rprospero (approved):

trisyoungs (approved):

PR #1440, opened 2023-05-31
1419 Region from Expression

### LINKED ISSUE #1063: Access Parameters from OperateExpression
The `OperateExpressionProcedureNode` highlights an issue with accessibility of useful variables from two sources. The `OperateExpressionProcedureNode` contains an `Expression` member and a set of local `ExpressionVariable`s that are set according to the data context (e.g. operating on a `Data1D` sets `x`, `value`, and `xDelta`) and which are passed to the `Expression` as the source of "externalVariables". This is fine, but updating the `CalculateAngleModule` presented a case where introducing a variable located in a `ParametersNode` would be extremely useful and allow better control over the module's action.  However, this is not presently possible since only a single source (vector) of "externalVariables" can be passed to `Expression`.

### LINKED ISSUE #1419: Generate region from Expression
For maximum flexibility it would be great to have a region defined by an arbitrary expression. That expression would need access to the x, y, z coords of voxel centres.

trisyoungs:
This PR set out to implement the ability to generate a custom region from a provided mathematical expression. This is indeed the final result, but several other enhancements and modifications were made along the way, namely:

- Build on #1413 and allow `Expression`s to have their own local set of variables as well as access to `Procedure`-wide parameters / variables from other nodes.
- Make all trigonometric functions work in radians rather than auto-convert to/from degrees for consistency with C++ rather than user expectation.
- Add several new functions to the `Expression` lexer - `pi()`, `twopi()`, `toRad(x)`, and `toDeg(x)`.

The new `CustomRegion` node makes use of all of the above.

Closes #1419.
Closes #1063.

### DIFF
--- a/src/expression/ExpressionVisitor.cpp
+++ b/src/expression/ExpressionVisitor.cpp
@@ -159,7 +159,7 @@ antlrcpp::Any ExpressionVisitor::visitVariable(ExpressionParser::VariableContext
     {
         // Do we have any external variables available?
         if (!externalVariables_)
-            throw(ExpressionExceptions::ExpressionSyntaxException(fmt::format(
+            throw(ExpressionExceptions::ExpressionSemanticException(fmt::format(
                 "Variable '{}' does not exist in this context (there are no variables defined).\n", ctx->Name()->getText())));
 
         // Does the named variable exist?
@@ -167,7 +167,7 @@ antlrcpp::Any ExpressionVisitor::visitVariable(ExpressionParser::VariableContext
         it = std::find_if(extVars.begin(), extVars.end(),
                           [ctx](auto var) { return DissolveSys::sameString(var->name(), ctx->Name()->getText()); });
         if (it == extVars.end())
-            throw(ExpressionExceptions::ExpressionSyntaxException(
+            throw(ExpressionExceptions::ExpressionSemanticException(
                 fmt::format("Variable '{}' does not exist in this context.\n", ctx->Name()->getText())));
     }
 
--- a/src/expression/ExpressionVisitor.cpp
+++ b/src/expression/ExpressionVisitor.cpp
@@ -27,9 +27,11 @@ std::shared_ptr<ExpressionNode> ExpressionVisitor::currentContext() const
 // Construct description within supplied object, from given tree
 void ExpressionVisitor::create(
     Expression &expr, ExpressionParser::ExpressionContext *tree,
+    const std::vector<std::shared_ptr<ExpressionVariable>> &localVariables,
     OptionalReferenceWrapper<const std::vector<std::shared_ptr<ExpressionVariable>>> externalVariables)
 {
     expression_ = &expr;
+    localVariables_ = localVariables;
     externalVariables_ = externalVariables;
 
     // Add the definition's root node as the topmost context in the stack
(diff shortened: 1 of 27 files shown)


jws-1 (commented):
Looks good to me, and I agree with conforming to radians for consistency. I left a couple of stylistic comments, although they might not be in-line with the kind of standards we want to follow here (which I am still familiarising myself with). I did notice that there is not a node icon for the `CustomRegion` node, as seen in the below screenshot.

image

src/expression/ExpressionVisitor.cpp
```diff
@@ -149,19 +151,27 @@ antlrcpp::Any ExpressionVisitor::visitFunction(ExpressionParser::FunctionContext
...
                            [ctx](auto var) { return DissolveSys::sameString(var->name(), ctx->Name()->getText()); });
-    if (it == vars.end())
-        throw(ExpressionExceptions::ExpressionSyntaxException(
-            fmt::format("Variable '{}' does not exist in this context.\n", ctx->Name()->getText())));
-
+    if (it == localVars.end())
+    {
+        // Do we have any external variables available?
+        if (!externalVariables_)
+            throw(ExpressionExceptions::ExpressionSyntaxException(fmt::format(
```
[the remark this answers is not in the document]
trisyoungs: That would be more appropriate, yes.  I confess I was a bit lazy here!  I'll change this to `ExpressionSemanticException` as suggested.

trisyoungs:
>

trisyoungs:
@jws-1 - I need to rebase onto `develop` to fix the merge conflicts, so will add the appropriate icon at the same time.

src/procedure/nodes/customregion.cpp
```diff
@@ -0,0 +1,45 @@
...
+    auto rFrac = cfg->box()->getFractional(r);
+    xFrac_->setValue(rFrac.x);
+    yFrac_->setValue(rFrac.y);
+    zFrac_->setValue(rFrac.z);
+
+    // Assess expression
+    auto x = expression_.asDouble();
+
+    return (x >= minimumValue_ && x <= maximumValue_);
+}
```
rprospero, line 45: So between this PR and the Array3DIterator, I've had an idea.  It's out of scope for this PR, but might it be worth adding a `validVoxels` method to `RegionProcedureNodeBase`?  The idea would be to pass in the size of our voxel grid and return an Array3D<bool> where the valid voxels are set to true.  The default implementation wouldn't be any different than what we're doing right now (i.e. iterating over each voxel and calling `isVoxelValid`), but certain classes could implement optimized versions.  For example, the rectangular region can easily iterate through the known good positions instead of examining every voxel.  This could be a boon if we had a 1000³ voxel and only examining a 10³ region.  In the worst case, it wouldn't be slower than what we already have.

On the other hand, if we expect most regions to be at least half the voxel, then there's no real benefit.

rprospero (approved):
I have a possible suggestion for another Issue, but otherwise everything looks good.

src/procedure/nodes/customregion.cpp
trisyoungs: So my considered answer is "Maybe?". While this would indeed localise the work loop and allow some pretty aggressive throwing out of areas of voxels for (to-be-implemented) regions such as rectangular ones ("Slabs", I would call them!) it doesn't do much for the general case.

trisyoungs: @rprospero  and @jws-1 - Another thought that springs to mind, slightly unrelated to this PR, is that calling the above `isVoxelValid()` function from the upcoming Array3D threaded loop is going to absolutely blow Dissolve up since each call will try to set the variables for the equation.  Needs some thought on how this can be parallelised safely - may need to construct temporary objects of the target region type (i.e. `CustomRegionProcedureNode` so that each thread has its own object with its own local variables. We can then recombine the results afterwards. So, in a bigger context @rprospero 's suggestion above might be a useful one!

PR #1443, opened 2023-06-07
[DRAFT]: Array3D Iterator

### LINKED ISSUE #1426: TripletIterator for Array3D
To solve #883 we need an iterator which operates over all elements of an `Arrray3D`. We already have a `PairIterator` class (`src/classes/pair_iter.h`) and which is used by the `for_each_pair` algorithms (`src/templates/algorithms.h`).

Once we have that we can parallelise processing of 3D data (as required by #883).

jws-1:
Draft PR that aims to close #1426. Not to be merged until the issue causing segmentation faults when enabling parallel processing is fixed.


rprospero:
@jws-1 Looks like a good start.  I'll do a full review later on, but there's two big extensions we can add to Array3D on this PR.

1) Allow an `Array3DIterator` to be passed into the `[]` operator for accessing values in the array.
2) Modify the `begin`, `end`, `cbegin`, `cend` members to return Array3DIterator instead of a vector iterator.  It might also be useful to drop the `[]` operator that acts on vector iterators.

jws-1:
> 2\. Modify the `begin`, `end`, `cbegin`, `cend` members to return Array3DIterator instead of a vector iterator.  It might also be useful to drop the `[]` operator that acts on vector iterators

Might be a silly question, but returning an `Array3DIterator` instead of a vector iterator would mean that the iterator is over the indices rather than the values, how would I deal with these implications throughout? For instance with something like `std::min_element`?

jws-1:
> @jws-1 Looks like a good start. I'll do a full review later on, but there's two big extensions we can add to Array3D on this PR.
> 
>     1. Allow an `Array3DIterator` to be passed into the `[]` operator for accessing values in the array.
> 
>     2. Modify the `begin`, `end`, `cbegin`, `cend` members to return Array3DIterator instead of a vector iterator.  It might also be useful to drop the `[]` operator that acts on vector iterators.



> @jws-1 Looks like a good start. I'll do a full review later on, but there's two big extensions we can add to Array3D on this PR.
> 
>     1. Allow an `Array3DIterator` to be passed into the `[]` operator for accessing values in the array.
> 
>     2. Modify the `begin`, `end`, `cbegin`, `cend` members to return Array3DIterator instead of a vector iterator.  It might also be useful to drop the `[]` operator that acts on vector iterators.

@rprospero, I believe this is ready for another look. I implemented 1 - I decided to allow the usage of `{x,y,z}` and `Array3DIterator` for the `[]` operator. I also implemented `begin_indices`, `end_indices`, `cbegin_indices` and `cend_indices` methods to allow iterating over the indices of an `Array3D`.

src/templates/array3d.h
```diff
@@ -104,21 +115,43 @@ template <class A> class Array3D
...
+    {
+        auto [x, y] = index;
+        return {Array3DIterator(nX_, nY_, nZ_, x * y), Array3DIterator(nX_, nY_, nZ_, x * y * nZ_)};
+    }
+    std::pair<const Array3DIterator, const Array3DIterator> index_range(std::tuple<int, int> index) const
+    {
+        auto [x, y] = index;
+        return {Array3DIterator(nX_, nY_, nZ_, x * y), Array3DIterator(nX_, nY_, nZ_, x * y * nZ_)};
+    }
+
```
rprospero, line 154: Again, I think we can have a const member that will return a pair of mutable values.  See my comments below for more.

rprospero (changes requested):
Some more suggestions before we merge in.  As we saw this morning, not all of my ideas are good ones, so don't hesitate to ask if something I'm suggesting doesn't make sense.

src/templates/array3d.h
```diff
@@ -94,6 +96,15 @@ template <class A> class Array3D
...
     }
+    A &operator[](Array3DIterator it)
+    {
+        auto [x, y, z] = *it;
+        assert(x >= 0 && x < nX_);
+        assert(y >= 0 && y < nY_);
+        assert(z >= 0 && z < nZ_);
+
+        return array_[sliceOffsets_[x] + y * nZ_ + z];
+    }
```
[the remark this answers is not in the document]
jws-1: Yes, that's much better!

src/templates/array3d.h
```diff
@@ -148,6 +181,14 @@ template <class A> class Array3D
...
     // End of the underlying array
     typename std::vector<A>::const_iterator end() const { return array_.cend(); }
+    // Beginning of indices
+    Array3DIterator beginIndices() { return Array3DIterator(nX_, nY_, nZ_, 0); }
+    // End of indices
+    Array3DIterator endIndices() { return Array3DIterator(nX_, nY_, nZ_, nX_ * nY_ * nZ_); }
+    // Beginning of indices
+    const Array3DIterator beginIndices() const { return Array3DIterator(nX_, nY_, nZ_, 0); }
+    // End of indices
+    const Array3DIterator endIndices() const { return Array3DIterator(nX_, nY_, nZ_, nX_ * nY_ * nZ_); }
```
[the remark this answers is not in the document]
jws-1: I think you're right - I'll remove the non-const members and have the const members return mutable values.

jws-1:
> Some more suggestions before we merge in. As we saw this morning, not all of my ideas are good ones, so don't hesitate to ask if something I'm suggesting doesn't make sense.

OK, I've followed your suggestions - although I'm a bit uncertain of my updated implementation of the `++` operator for the `Array3DIterator`, so I'd appreciate your thoughts on that!

rprospero:
Your implementation looks correct.  That's how I would have written it.  Plus, it's still passing the tests, so that's another good sign that everything is okay.

One thing that does come up, is that you added the iterator to the lambda on the `for_each_triplet`, but you didn't use it in the array access, instead recreating the tuple.  Is there a reason for this?  I would have found using the iterator as being more clear, but I also know that my views on "clear" can be highly idiosyncratic, so I'm happy to hear a difference.

jws-1:
> Your implementation looks correct. That's how I would have written it. Plus, it's still passing the tests, so that's another good sign that everything is okay.
> 
> One thing that does come up, is that you added the iterator to the lambda on the `for_each_triplet`, but you didn't use it in the array access, instead recreating the tuple. Is there a reason for this? I would have found using the iterator as being more clear, but I also know that my views on "clear" can be highly idiosyncratic, so I'm happy to hear a difference.

Ah, I had just forgotten to use it! Sorted now, and it makes sense to use it too.

src/classes/region.h
```diff
@@ -44,14 +47,20 @@ class Region
...
-                        Vec3<int>(x, y, z),
-                        voxelCheckFunction(cfg, box_->getReal({(x + 0.5) * voxelSizeFrac_.x, (y + 0.5) * voxelSizeFrac_.y,
-                                                               (z + 0.5) * voxelSizeFrac_.z}))};
 
+        // Setup iterator for voxel map
+        Array3DIterator iterator(nVoxels_.x, nVoxels_.y, nVoxels_.z);
+
+        // Iterate voxels in parallel
+        dissolve::for_each_triplet(
+            ParallelPolicies::par, iterator.begin(), iterator.end(),
```
rprospero, line 56:
```suggestion
        // Iterate voxels in parallel
        dissolve::for_each_triplet(
            ParallelPolicies::par, voxelMap.beginIndices(), voxelMap.endIndices(),
```

Make our intentions a bit more clear

src/templates/array3d.h
```diff
@@ -104,21 +108,30 @@ template <class A> class Array3D
...
         return {begin, begin + nZ_};
     }
+
+    // Return index range for a given x and y value
+    std::pair<Array3DIterator, Array3DIterator> index_range(std::tuple<int, int> index) const
+    {
+        auto [x, y] = index;
+        return {Array3DIterator(nX_, nY_, nZ_, x * y), Array3DIterator(nX_, nY_, nZ_, x * y * nZ_)};
+    }
+
```
rprospero, line 134: Do we ever use `index_range`?  If not, it might be worth dropping the function entirely, unless there's a plan for it.

rprospero (changes requested):
Two more small comments and then we're hopefully done.

src/templates/array3d.h
jws-1: We do not, however I thought it may be useful "in some way, at some point" - happy to drop it though, if you can't foresee a use case for it.

rprospero: Eventually, we'll move to C++20, which includes official ranges support.  However, the ranges `concept` requires that the method would be part of `Array3D`.  While I agree that it will be useful in the future, we'd wind up needing to rewrite it when we finally got a chance to use it.

rprospero (approved):

Issue #1449, opened 2023-06-09
Epic / Species Charge Handling


trisyoungs:
### Focus
Addresses several UX issues with charges on `Species`, allowing simulations to be coaxed into running more easily.

### Tasks
- [x] #1446
- [x] #1447
- [x] #1448
- [x] #1453
- [x] #1454

trisyoungs:
Closed by #1450 and #1464.

PR #1450, opened 2023-06-12
Species charge handling - Part 1

### LINKED ISSUE #1446: Copy AtomType charges to species atoms
A useful feature would be the ability to copy charges from the `AtomType`s assigned to species atoms, on to the `SpeciesAtom`s themselves.  This is a destructive operation, so a dialog making sure the user is sure is a must!

### LINKED ISSUE #1447: Set AtomType charges from current SpeciesAtom charges
One possible nice thing to do would be to take the current charges assigned to a `Species` atoms and form an average charge for each assigned `AtomType`.  There is no guarantee that the charges for a given `AtomType` will be close to each other and form a nice average, so a "Test" option to work out what the new `AtomType` charges would be, along with the standard deviation (can use `SampledDouble` for this), and print those out to messages without actually changing anything.

Again, this is a destructive op so a double-check dialog is a must!

### LINKED ISSUE #1448: Species charge "smoothing"
A relatively common occurrence when importing species forcefields from LigParGen or from QM calculations is that the charges assigned to species atoms have lots of decimal places. These typically sum almost to zero (or +1, or -1) but not quite, and this is typically exacerbated into a significant non-zero total system charge for a configuration, which Dissolve will of course refuse to run with.

There is no solution in the code to handle this, so we should offer two options to the user:
1) Sum the species charges and re-scale them so as to give a specific target charge (0, -1, +1) 
2) Reduce the number of significant figures in the current charges, rounding off accordingly.

These two options go hand-in-hand in some respects, and one may wish to do 2) followed by 1) in order to give the target charge.

jws-1:
Initial PR regarding species charge handling (#1449).
Added the following menu items to the Species menu:
- Reduce Sig. Figs. in Charges - this is the first step towards smoothing charges (#1448), but additionally we need to add the option to scale and sum charges to 0/-1/+1.
- Copy Atom Charges from Atom Types - copies the charges from the defined Atom Types to the current species. Closes #1446.
- Set Atom Type Charges From Species - averages the charges over the atom types in the current species whilst offering the option to "test" by outputting the proposed changes to Messages, along with their standard deviations. Closes #1447.

Update:
- Additional menu item added that allows scaling charges so they sum to an arbitrary non-zero value. Possible closes #1448?


src/gui/menu_species.cpp
```diff
@@ -341,6 +346,84 @@ void DissolveWindow::on_SpeciesSetChargesInSelectionAction_triggered(bool checke
...
+                charges[atomType] += atom.charge();
+            }
+        }
+    }
+
+    QMessageBox msgBox(this);
+    msgBox.setWindowTitle("Copy Charges from Atom Types");
+    msgBox.setText("This will replace the species charges "
+                   "with those of the corresponding Atom Types.\n\n"
+                   "This cannot be undone! Proceed?");
```
trisyoungs, line 405:
```suggestion
    msgBox.setWindowTitle("Set Atom Type Charges from Species Atoms");
    msgBox.setText("This will replace the current atom type charges "
                   "with the average charge of relevant species atoms.\n\n"
                   "This cannot be undone! Proceed?");
```

src/gui/menu_species.cpp
```diff
@@ -341,6 +346,84 @@ void DissolveWindow::on_SpeciesSetChargesInSelectionAction_triggered(bool checke
...
+    for (auto &atomType : atomTypes)
+    {
+        for (auto &atom : species->atoms())
+        {
+            if (atom.atomType() == atomType)
+            {
+                charges[atomType] += atom.charge();
+            }
+        }
+    }
```
trisyoungs, line 399: I think you can just simplify this to:
```suggestion
    for (auto &atom : species->atoms())
        charges[atom.atomType()] += atom.charge();
```

src/gui/menu_species.cpp
```diff
@@ -341,6 +346,84 @@ void DissolveWindow::on_SpeciesSetChargesInSelectionAction_triggered(bool checke
...
+    else if (result != QMessageBox::No)
+    {
+        for (auto const &atomType : atomTypes)
+        {
+            Messenger::print("{}: {} -> {} \u00b1 {}", atomType->name(), atomType->charge(), charges[atomType].value(),
+                             charges[atomType].stDev());
+        }
+    }
+}
```
trisyoungs, line 425: Here, I would exit immediately if `No`, then always print out the charge map, then change the actual charges if `Yes`.

src/gui/menu_species.cpp
```diff
@@ -341,6 +346,84 @@ void DissolveWindow::on_SpeciesSetChargesInSelectionAction_triggered(bool checke
...
+    if (result == QMessageBox::Yes)
+    {
+        for (auto const &x : charges)
+            x.first->setCharge(x.second.value());
+    }
+    else if (result != QMessageBox::No)
+    {
+        for (auto const &atomType : atomTypes)
+        {
+            Messenger::print("{}: {} -> {} \u00b1 {}", atomType->name(), atomType->charge(), charges[atomType].value(),
```
trisyoungs, line 421: May also be nice to include the number of data contributing to each average, just to give full information to the user.

src/gui/menu_species.cpp
```diff
@@ -363,3 +446,24 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
...
+{
+    // Get the current Species (if a SpeciesTab is selected)
+    auto species = ui_.MainTabs->currentSpecies();
+    if (!species)
+        return;
+
+    auto ok = false;
+    auto significantFigures =
+        QInputDialog::getInt(this, "Reduce Signficant Figures in Charges",
+                             "Enter the number of significant figures to use for all atoms", 1, 1, 100, 1, &ok);
```
trisyoungs, line 460: A nicer default:
```suggestion
                             "Enter the number of significant figures to use for all atoms", 3, 1, 100, 1, &ok);
```

src/gui/menu_species.cpp
```diff
@@ -363,3 +446,24 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
...
+    if (!ok)
+        return;
+
+    for (auto &atom : species->atoms())
+    {
+        double rounded = std::round(atom.charge() * std::pow(10, significantFigures)) / std::pow(10, significantFigures);
+        atom.setCharge(rounded);
+    }
+}
```
trisyoungs, line 469: The temporary variable is a bit redundant:
```suggestion
    for (auto &atom : species->atoms())
        atom.setCharge(std::round(atom.charge() * std::pow(10, significantFigures)) / std::pow(10, significantFigures));
```

src/gui/menu_species.cpp
trisyoungs, line 421: Would be nice also to have a header line before the charge map gets printed, just to clarify what it is that we're seeing!

trisyoungs (changes requested):
A few algorithmic comments, but otherwise very nice indeed.

src/gui/menu_species.cpp
jws-1: Yes, much cleaner!

src/gui/menu_species.cpp
```diff
@@ -341,6 +346,82 @@ void DissolveWindow::on_SpeciesSetChargesInSelectionAction_triggered(bool checke
...
+        {
+            for (auto &atomType : atomTypes)
+            {
+                if (atomType == atom.atomType())
+                {
+                    atom.setCharge(atomType->charge());
+                    break;
+                }
+            }
+        }
```
trisyoungs, line 374: Just spotted this one.  Here's another double loop that can be reduced right back to a single one since there's no need to look up the atom type, we can just use it (*if* the atom has one set):
```suggestion
        for (auto &atom : species->atoms())
            if (atom.atomType())
                atom.setCharge(atomType->charge());
```

src/gui/menu_species.cpp
```diff
@@ -363,3 +444,60 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
...
+    for (auto &atom : species->atoms())
+        sum += atom.charge();
+
+    double scaleFactor = scaleTarget / sum;
+    for (auto &atom : species->atoms())
+        atom.setCharge(atom.charge() * scaleFactor);
+
+    setModified();
+
+    fullUpdate();
```
trisyoungs, line 480: This function is fundamentally the same as the existing one above, save for a slightly different scale factor.  In order not to confuse the user with two very similar (menu) functions perhaps we should consider combining the two into the one dialog with "Scale", "Scale To", and "Cancel" buttons?

trisyoungs (changes requested):

src/gui/gui.ui
```diff
@@ -722,6 +725,16 @@
...
    </property>
   </action>
+  <action name="SpeciesCopyChargesFromAtomTypesAction">
+   <property name="text">
+    <string>Copy Atom C&amp;harges from Atom Types...</string>
+   </property>
+  </action>
+  <action name="SpeciesSetAtomTypeChargesFromSpeciesAction">
+   <property name="text">
+    <string> Set Atom Type C&amp;harges from Species...</string>
```
trisyoungs, line 735: These menu items have the same shortcut!

trisyoungs (approved):
One small correction to make, then I think we should get this merged in.  Having just used the new features in earnest there are some obvious other things to do which are closely related, so I'll update the epic.

src/gui/gui.ui
jws-1: Oops! Sorted.

Issue #1455, opened 2023-06-14
(SDF) Orientational constraints


tfheaden:
**Describe the feature / issue**
For SDF calculations: The ability to select relative orientation of the 1st and 2nd molecules that are included in the calculations. e.g. for benzene calculate SDFs for when normal to plane is within 0-10 degrees or 80-90 degrees

PR #1462, opened 2023-06-19
Node Contexts Tidy

trisyoungs:
As part of work towards #1455 this PR tidies up "context handling" of procedure nodes.  Essentially, #1455 requires some new functionality to be introduced to the complement of `ProcedureNode`s in the form of `if`-style nodes, which are essentially usable in any context. Introducing that, and handling the checks in a nice way, warranted a bit of a refactor.

### DIFF
--- a/src/procedure/nodes/node.cpp
+++ b/src/procedure/nodes/node.cpp
@@ -80,7 +80,7 @@ EnumOptions<ProcedureNode::NodeContext> ProcedureNode::nodeContexts()
 
 ProcedureNode::ProcedureNode(ProcedureNode::NodeType nodeType, std::vector<NodeContext> relevantContexts,
                              ProcedureNode::NodeClass classType)
-    : class_(classType), type_(nodeType), relevantContexts_(std::move(relevantContexts))
+    : type_(nodeType), relevantContexts_(std::move(relevantContexts)), class_(classType)
 {
 }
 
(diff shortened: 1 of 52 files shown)


src/procedure/nodes/node.cpp
```diff
@@ -74,11 +73,14 @@ EnumOptions<ProcedureNode::NodeContext> ProcedureNode::nodeContexts()
...
+                                                                   {ProcedureNode::OperateContext, "Operate"},
+                                                                   {ProcedureNode::OperateContext, "Any"},
+                                                                   {ProcedureNode::ParentProcedureContext, "ParentProcedure"}});
 }
 
-ProcedureNode::ProcedureNode(ProcedureNode::NodeType nodeType, ProcedureNode::NodeClass classType)
-    : class_(classType), type_(nodeType)
+ProcedureNode::ProcedureNode(ProcedureNode::NodeType nodeType, std::vector<NodeContext> relevantContexts,
+                             ProcedureNode::NodeClass classType)
+    : class_(classType), type_(nodeType), relevantContexts_(std::move(relevantContexts))
```
jws-1, line 83: I think I'd prefer if the ordering of the parameters matched the ordering of the member initialisation, although I assume there is a reason why it doesn't!

jws-1, line 83: I think I'd prefer if the ordering of the parameters matched the ordering of the member initialisations, but I'm sure there is a good reason that it doesn't!

jws-1 (approved):
I like it - the changes seem reasonable to me, from what I can understand! The negative diff is nice to see, also. I had one stylistic comment/question, but ultimately it's up to you what you do with it!

src/procedure/nodes/node.cpp
trisyoungs: Nope - just the order the members were defined in the header.  However, I agree that they should reflect the ctor ordering, so have reordered things accordingly.

PR #1464, opened 2023-06-19
Species Charge Handling Part 2

### LINKED ISSUE #1453: Reduce significant figures in atom type charges
In an analogous manner to recent work on species charges, the ability to reduce the significant figures on atom type charges themselves would also be highly useful.

### LINKED ISSUE #1454: Species charge trimming
While another issue refers to species charge "smoothing", that is really just scaling, but of course we can't easily scale to zero.

However, there will always be occasions where a species is _nearly_ zero (or _nearly +1, -1 etc.) and we want to "smooth" the species atom charges to that value.  In this instance, "smoothing" can be considered as the subtraction of the total charge difference divided by _N_, the number of atoms in the species, from each species atom.

jws-1:
This PR should finish of the Species Charge Handling Epic (#1449).
- Similarly to the functionality available in the `Species` menu, the option to reduce the number of significant figures in the charges in the `Atom Types` has been added to the `Simulation` menu. Closes #1453.
- The option to "smooth" the sum of the charges in a `Species` to a given value has been added to the `Species` menu. Closes #1454.

Side note: while #1454 names the latter functionality as "trimming", I felt like "smoothing" was more suiting (and we also haven't used that terminology yet). Although I'm happy to change it, if you don't agree.

### DIFF
--- a/src/gui/menu_simulation.cpp
+++ b/src/gui/menu_simulation.cpp
@@ -73,3 +73,24 @@ void DissolveWindow::on_SimulationClearAdditionalPotentialsAction_triggered(bool
 }
 
 void DissolveWindow::on_SimulationClearModuleDataAction_triggered(bool checked) { clearModuleData(); }
+
+void DissolveWindow::on_SimulationReduceChargesSigFigsAction_triggered(bool checked)
+{
+    // Get Atom Types
+    auto atomTypes = dissolve().coreData().atomTypes();
+
+    auto ok = false;
+    auto significantFigures =
+        QInputDialog::getInt(this, "Reduce Signficant Figures in Charges",
+                             "Enter the number of significant figures to use for all atom types", 3, 1, 100, 1, &ok);
+    if (!ok)
+        return;
+
+    for (auto &atomType : atomTypes)
+        atomType->setCharge(std::round(atomType->charge() * std::pow(10, significantFigures)) / std::pow(10, significantFigures));
+
+    setModified();
+
+    fullUpdate();
+}
+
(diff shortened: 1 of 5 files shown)


src/gui/menu_species.cpp
```diff
@@ -450,6 +451,31 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
...
+    auto targetSum = QInputDialog::getDouble(this, "Smooth atom charges", "Enter the target sum to smooth atom charges to.",
+                                             0.0, -100.0, 100.0, 5, &ok);
+
+    if (!ok)
+        return;
+
+    double sum = 0.0, shiftVal = 0.0;
+    for (auto &atom : species->atoms())
+        sum += atom.charge();
+    shiftVal = (sum - targetSum) / species->nAtoms();
```
trisyoungs, line 471: I don't think this works for negative total charges - I blindly tried to smooth the charge of a `-1` species to zero, and ended up with a charge of `-2`!

src/gui/menu_simulation.cpp
```diff
@@ -73,3 +73,24 @@ void DissolveWindow::on_SimulationClearAdditionalPotentialsAction_triggered(bool
 }
 
 void DissolveWindow::on_SimulationClearModuleDataAction_triggered(bool checked) { clearModuleData(); }
+
+void DissolveWindow::on_SimulationReduceChargesSigFigsAction_triggered(bool checked)
```
trisyoungs, line 77: Can't test this as the relevant item doesn't appear in the menu!

trisyoungs (changes requested):

src/gui/menu_simulation.cpp
trisyoungs: So hey, guess what - I can see it too! Not sure now I didn't have it earlier today - probably an out-of-date branch (somehow) but it's all good now.  I would suggest one minor change, and that is to move the new menu option above the two "Clear..." options, and add a separator between them.

trisyoungs (approved):
:+1:

Issue #1463, opened 2023-06-19
Optimise Generation/Usage of Advanced Sites during Simulation


jws-1:
**Describe the feature / issue**
Performance of site generation during simulation for advanced sites (`Dynamic`, `Fragment`) may not scale well to larger molecules or frameworks. This can be optimised, as mentioned in #1461.

The atom ordering in a `Species` cannot be changed once it has been defined. Therefore, where advanced sites (`Dynamic`, `Fragment`) are defined, occurrences of those sites within the `Species` can be pre-determined, and utilised during simulation (where currently they are created on-the-fly by the `SiteStack`). This could improve performance for more-complex species, such as larger molecules and frameworks. A first step towards this (in the case of `Fragments`) could be to store unique fragments within the `Site` itself, since once defined, these will not change.

jws-1:
Closed by #1465.

PR #1465, opened 2023-06-20
Fragments Part 2

jws-1:
This PR follows on from #1461, with the aim of optimising site generation as described in #1463. This is a draft PR, as I am seeking some intermediate feedback to make sure I'm on the right track with my design decisions!


src/classes/speciessite.cpp
```diff
@@ -324,6 +331,84 @@ const std::vector<std::shared_ptr<AtomType>> &SpeciesSite::atomTypes() const { r
 // Return fragment definition
 const NETADefinition &SpeciesSite::fragment() const { return fragment_; }
 
+bool SpeciesSite::generateUniqueSites()
+{
+    sitesOriginAtoms_.clear();
+    xAxisAtoms_.clear();
+    yAxisAtoms_.clear();
```
trisyoungs, line 338:
```suggestion
    sitesXAxisAtoms_.clear();
    sitesYAxisAtoms_.clear();
```

src/classes/speciessite.cpp
```diff
@@ -403,105 +488,74 @@ std::vector<std::shared_ptr<Site>> SpeciesSite::createFromParent() const
...
         return sites;
     }
     else if (type_ == SiteType::Fragment)
     {
         std::vector<std::shared_ptr<Site>> sites;
-        double mass;
-        std::vector<std::vector<int>> matchedIndices;
-        for (auto &i : parent_->atoms())
+        
+        for (int i = 0; i < nSites(); ++i)
```
trisyoungs, line 499:
```suggestion
        for (auto i = 0; i < nSites(); ++i)
```

src/classes/speciessite.cpp
```diff
@@ -403,105 +488,74 @@ std::vector<std::shared_ptr<Site>> SpeciesSite::createFromParent() const
...
-                matchedIndices.push_back(std::move(matchedAtomIndices));
-
-                // Identifiers which label origin, x and y axis atoms.
-                auto identifiers = fragment_.matchedPath(&i).identifiers();
-
-                // Compute the origin.
-                auto originAtoms = identifiers["origin"];
-
-                if (originMassWeighted_)
+                double massNorm = 0.0;
```
trisyoungs, line 506:
```suggestion
                auto massNorm = 0.0;
```

src/classes/speciessite.h
```diff
@@ -154,6 +154,26 @@ class SpeciesSite : public Serialisable
     public:
     const NETADefinition &fragment() const;
 
+    /*
+     * Advanced Sites
+     */
+    private:
+    // Species atoms whose average position is the oriin of the site
```
trisyoungs, line 161:
```suggestion
    // Species atoms whose average position is the origin of the site
```

jws-1:
> The `createFromParent()` function is now doing a lot of duplicate work, isn't it? The vectors of origin, x, and y-atoms we now have save most of this effort, and you just need to check whether the x and y vectors are the same size as the origin to know whether axes are present, no? You could also include the generation of the `Static` site in the `generateUniqueSites()` function which would simplify the usage elsewhere (unless I'm missing a good reason why we can't do this?)

Yes, we could generate `Static` sites there too. I should be able to remove the if-else logic in `createFromParent`, making it similar to how sites are constructed in `SiteStack`.
I had one design question. Currently I have members specifically for `Static` sites: `originAtoms_`, `xAxisAtoms_` and `yAxisAtoms_` which each store a single vector of `SpeciesAtom*`; as a `Static` site definition produces a single site. However, for the "advanced" sites I have members `siteOriginAtoms_`, `siteXAxisAtoms_`, and `siteYAxisAtoms_`, which are nested vectors of `SpeciesAtom*`, as there can be multiple sites according to the definition. I am tempted to drop the members specific to `Static` sites, and store their information as the first (and only) elements in the aforementioned vectors - however this would mean we need to use `siteOriginAtoms_.at(0)` etc. everywhere - what do you think?

trisyoungs:
I guess it boils down to a question of data separation - as currently written there is a clear separation between the definitions of all site types and the generated vector of "unique" sites (which may be only one).  By re-using the vectors of vectors as you suggest we blur those lines a bit, and perhaps obfuscate our intentions? There will also need to be, as you say, lots of `at(0)` or `front()` calls on those vectors.  On balance perhaps sticking with the clear separation is the better option.  However, I would strongly suggest that we might rename some member variables to clearly delineate between definition and generation.  As this class has been extended it's absorbed a lot of new information, so perhaps renaming `originAtoms_` to `staticOrigin_`, and `elements_` to `dynamicElements_`, for example, helps to clarify everything that's going on here.

src/classes/sitestack.cpp
```diff
@@ -57,329 +58,94 @@ Vec3<double> SiteStack::centreOfMass(const Molecule &mol, const Box *box, const
...
-    sites_.reserve(siteIndices.size() * spPop);
+    // Set new index and clear old arrays
+    configurationIndex_ = configuration_->contentsVersion();
+    sites_.clear();
+    orientedSites_.clear();
 
-    // Get Molecule array from Configuration and search for the target Species
-    for (const auto &molecule : configuration_->molecules())
+    std::vector<std::vector<const SpeciesAtom *>> originAtoms, xAxisAtoms, yAxisAtoms;
+    if (site->type() == SpeciesSite::SiteType::Static)
```
trisyoungs, line 80: We can remove the need to check for `SiteType::Static` if we just make `SpeciesSite::generateUniqueSites()` copy the `static???Atoms_` vectors into the `site???Atoms_` ones.

src/classes/sitestack.cpp
```diff
@@ -57,329 +58,94 @@ Vec3<double> SiteStack::centreOfMass(const Molecule &mol, const Box *box, const
...
+                y = box->minimumVector(origin, centreOfGeometry(*molecule, box, staticYAxisAtomIndices));
+                y.orthogonalise(x);
+                y.normalise();
 
-    // Set new index and clear old arrays
-    configurationIndex_ = configuration_->contentsVersion();
-    sites_.clear();
-    orientedSites_.clear();
+                // Calculate z vector from cross product of x and y
+                z = x * y;
```
trisyoungs, line 140:
```suggestion
```

src/classes/sitestack.cpp
```diff
@@ -57,329 +58,94 @@ Vec3<double> SiteStack::centreOfMass(const Molecule &mol, const Box *box, const
...
     for (const auto &molecule : configuration_->molecules())
     {
         if (molecule->species() != targetSpecies)
             continue;
 
-        auto &atoms = molecule->atoms();
-
-        // Loop over site indices
-        for (auto id : siteIndices)
+        for (int i = 0; i < site->nSites(); ++i)
```
trisyoungs, line 111:
```suggestion
        for (auto i = 0; i < site->nSites(); ++i)
```

src/classes/speciessite.cpp
```diff
@@ -41,11 +41,12 @@ int SpeciesSite::version() const { return version_; }
...
-    xAxisAtoms_.clear();
-    yAxisAtoms_.clear();
-    elements_.clear();
-    atomTypes_.clear();
+    staticOriginAtoms_.clear();
+    staticXAxisAtoms_.clear();
+    staticYAxisAtoms_.clear();
+    dynamicElements_.clear();
+    dynamicAtomTypes_.clear();
+    fragment_ = NETADefinition{};
```
trisyoungs, line 49: :+1:

src/classes/speciessite.cpp
```diff
@@ -272,159 +273,91 @@ std::vector<int> SpeciesSite::yAxisAtomIndices() const
...
-                v += parent_->atom(index).r();
-            v /= yAxisIndices.size();
-
-            // Get vector from site origin, normalise it, and orthogonalise
-            auto y = v - origin;
-            y.orthogonalise(x);
-            y.normalise();
+    sitesOriginAtoms_.clear();
+    sitesXAxisAtoms_.clear();
+    sitesYAxisAtoms_.clear();
```
trisyoungs, line 338: As mentioned, if we just copy the vectors over here...

src/classes/speciessite.cpp
```diff
@@ -439,73 +372,103 @@ std::vector<std::shared_ptr<Site>> SpeciesSite::createFromParent() const
...
+    }
+}
 
-                    auto xAxisAtoms = identifiers["x"];
-                    auto yAxisAtoms = identifiers["y"];
+// Return number of unique sites
+const int SpeciesSite::nSites() const { return type_ == SiteType::Static ? 1 : sitesOriginAtoms_.size(); }
 
-                    Vec3<double> v;
+const std::vector<std::vector<const SpeciesAtom *>> &SpeciesSite::sitesOriginAtoms() const { return sitesOriginAtoms_; }
```
trisyoungs, line 396: Need commented function description here and the others
```suggestion
// Return atoms representing unique site origins
const std::vector<std::vector<const SpeciesAtom *>> &SpeciesSite::sitesOriginAtoms() const { return sitesOriginAtoms_; }
```

src/classes/speciessite.cpp
```diff
@@ -439,73 +372,103 @@ std::vector<std::shared_ptr<Site>> SpeciesSite::createFromParent() const
...
+        originAtoms.push_back(staticOriginAtoms());
+        xAxisAtoms.push_back(staticXAxisAtoms());
+        yAxisAtoms.push_back(staticYAxisAtoms());
+    }
+    else
+    {
+        originAtoms = sitesOriginAtoms();
+        xAxisAtoms = sitesXAxisAtoms();
+        yAxisAtoms = sitesYAxisAtoms();
+    }
```
trisyoungs, line 423: Can remove most of this and just refer to `sitesOriginAtoms_` etc. once the `Static` atom vectors are added.

src/classes/speciessite.h
```diff
@@ -154,6 +154,26 @@ class SpeciesSite : public Serialisable
...
+    std::vector<std::vector<const SpeciesAtom *>> sitesXAxisAtoms_;
+    // Species atom(s) that indicate the y axis with the origin, after orthogonalisation
+    std::vector<std::vector<const SpeciesAtom *>> sitesYAxisAtoms_;
+
+    public:
+    // Generate unique sites
+    bool generateUniqueSites();
+    // Number of unique sites
+    const int nSites() const;
+    const std::vector<std::vector<const SpeciesAtom *>> &sitesOriginAtoms() const;
```
trisyoungs, line 173: Function comments.
```suggestion
    // Return atoms representing unique site origins
    const std::vector<std::vector<const SpeciesAtom *>> &sitesOriginAtoms() const;
```

src/classes/speciessite.h
```diff
@@ -154,6 +154,26 @@ class SpeciesSite : public Serialisable
...
+    /*
+     * Advanced Sites
+     */
+    private:
+    // Species atoms whose average position is the origin of the site
+    std::vector<std::vector<const SpeciesAtom *>> sitesOriginAtoms_;
+    // Species atom(s) that indicate the x axis with the origin
+    std::vector<std::vector<const SpeciesAtom *>> sitesXAxisAtoms_;
+    // Species atom(s) that indicate the y axis with the origin, after orthogonalisation
+    std::vector<std::vector<const SpeciesAtom *>> sitesYAxisAtoms_;
```
trisyoungs, line 166: Can we save ourselves a job and store the `int` atom indices rather than `const SpeciesAtom *`? In `SiteStack` we end up converting these to atom index `int`s anyway, so it feels like we may as well optimise and do it here instead.

trisyoungs (commented):
Nearly there.  Of course, going through it again I see more things we can change! :)

src/classes/sitestack.cpp
jws-1: Ah, yes. That is a good idea!

src/classes/speciessite.cpp
```diff
@@ -272,240 +276,212 @@ std::vector<int> SpeciesSite::yAxisAtomIndices() const
...
+    sitesYAxisAtomsIndices_.clear();
+
+    switch (type_)
+    {
+        case (SiteType::Static):
+        {
+            sitesOriginAtomsIndices_.push_back(staticOriginAtomIndices());
+            sitesXAxisAtomsIndices_.push_back(staticXAxisAtomIndices());
+            sitesYAxisAtomsIndices_.push_back(staticYAxisAtomIndices());
+        }
```
trisyoungs, line 351: Stylistic convention for Dissolve is to not add unnecessary scope with curly braces if not required, but you deffo need a `break;`!
```suggestion
            sitesOriginAtomsIndices_.push_back(staticOriginAtomIndices());
            sitesXAxisAtomsIndices_.push_back(staticXAxisAtomIndices());
            sitesYAxisAtomsIndices_.push_back(staticYAxisAtomIndices());
            break;
```

src/classes/speciessite.cpp
```diff
@@ -272,240 +276,212 @@ std::vector<int> SpeciesSite::yAxisAtomIndices() const
...
+            }
+            return true;
+        }
+        default:
+        {
+            Messenger::print("Can't generate unique sites for site '{}'.", name());
+            return false;
+        }
+    }
+}
```
trisyoungs, line 414:
```suggestion

    return true;
}
```

src/classes/speciessite.cpp
```diff
@@ -272,240 +276,212 @@ std::vector<int> SpeciesSite::yAxisAtomIndices() const
...
+        {
+            for (auto &i : parent_->atoms())
+            {
+                // Valid element or atom type?
+                if ((std::find(dynamicElements_.begin(), dynamicElements_.end(), i.Z()) != dynamicElements_.end()) ||
+                    std::find(dynamicAtomTypes_.begin(), dynamicAtomTypes_.end(), i.atomType()) != dynamicAtomTypes_.end())
+                    sitesOriginAtomsIndices_.push_back({i.index()});
+            }
+            return true;
+        }
```
trisyoungs, line 362:
```suggestion
            for (auto &i : parent_->atoms())
            {
                // Valid element or atom type?
                if ((std::find(dynamicElements_.begin(), dynamicElements_.end(), i.Z()) != dynamicElements_.end()) ||
                    std::find(dynamicAtomTypes_.begin(), dynamicAtomTypes_.end(), i.atomType()) != dynamicAtomTypes_.end())
                    sitesOriginAtomsIndices_.push_back({i.index()});
            }
            break;
```

src/classes/speciessite.cpp
```diff
@@ -272,240 +276,212 @@ std::vector<int> SpeciesSite::yAxisAtomIndices() const
...
+                        std::transform(identifiers["x"].begin(), identifiers["x"].end(), sitesXAxisAtomsIndices_.back().begin(),
+                                       [](const auto &atom) { return atom->index(); });
+                        // Determine y axis atoms.
+                        sitesYAxisAtomsIndices_.emplace_back(identifiers["y"].size());
+                        std::transform(identifiers["y"].begin(), identifiers["y"].end(), sitesYAxisAtomsIndices_.back().begin(),
+                                       [](const auto &atom) { return atom->index(); });
+                    }
+                }
+            }
+            return true;
```
trisyoungs, line 406:
```suggestion
            break;
```

src/classes/speciessite.cpp
```diff
@@ -272,240 +276,212 @@ std::vector<int> SpeciesSite::yAxisAtomIndices() const
...
+        }
+        default:
+        {
+            Messenger::print("Can't generate unique sites for site '{}'.", name());
+            return false;
+        }
```
trisyoungs, line 412: This is probably important so we should show a proper error:
```suggestion
            return Messenger::error("Can't generate unique sites for site '{}'.", name());
```

src/classes/speciessite.cpp
```diff
@@ -272,240 +276,212 @@ std::vector<int> SpeciesSite::yAxisAtomIndices() const
...
+                if ((std::find(dynamicElements_.begin(), dynamicElements_.end(), i.Z()) != dynamicElements_.end()) ||
+                    std::find(dynamicAtomTypes_.begin(), dynamicAtomTypes_.end(), i.atomType()) != dynamicAtomTypes_.end())
+                    sitesOriginAtomsIndices_.push_back({i.index()});
+            }
+            return true;
+        }
+        case (SiteType::Fragment):
+        {
+            std::vector<std::vector<int>> matchedIndices;
+            for (auto &i : parent_->atoms())
```
trisyoungs, line 366: Now, this is interesting.  So you're obviously storing the complete set of match atom indices in order to perform unique fragment checking - for that, you need a temporary and so need to indulge in a scope in this `case` statement. However, the complete set of matched atoms is actually an integral part of the fragment site, and which should be stored for future use.  If we're doing that, then this implies another member variable (`std::vector<std::vector<int>> sitesAllAtomsIndices_`) which means you can remove the scope and the local var.

Why the need to store all the atoms in the fragment, you ask? Well, when it comes to actually manipulating the fragment (i.e. rotating those benzene rings) we need to know all the atoms to rotate / translate.  This modification could be left until that functionality is required in `Fragments Part 3+N` if you think it makes sense to do the overhaul at the point of usage implementation.

trisyoungs (approved):
Some small stylistic things, and a modification that can be left to a future PR, bot otherwise looks good to me. :+1:

src/classes/speciessite.cpp
jws-1: Oops!

src/classes/speciessite.cpp
jws-1: Agreed.

src/classes/speciessite.cpp
jws-1: Ah. Yes, in the bigger picture we should definitely be storing all the atoms in the fragment. I'm tempted to leave it for a subsequent PR, as we don't require all of the atoms for the functionality here/

Issue #1247, opened 2022-10-05
[Epic] Fragments


trisyoungs:
### Focus
Implement the ability to detect bound fragments within `Species` for the purposes of subsequent use as analysis sites or as targets for manipulation in `Configuration` generation.

The relevant data boils down to a vector of relevant `Molecule` atoms and an associated `Site` or `OrientedSite`.

### Tasks (NETA Paths)
Captured in #1249.
- [x] Better "or" handling with brackets
- [x] Enable returning of matched atoms from `NETADefinition::match()`.
- [x] Use `std::optional` for node parameters which may have a default "no action" value
- [x] Remove `NETARootNode` since it is now just `NETANode` with no useful overrides.

### Tasks (Fragment Implementation)
- [x] Exclude previously-matched atoms from subsequent `NETADefinition::match()` calls (use selection in `Species`, or operate on separate vector data?)
- [x] Incorporate into `SelectNode` as a new target type.

### Realisation
At this point the ability to label atoms with particular tags exists in `NETA`, and is tested by `unit/classes/neta.cpp` but is not used in earnest anywhere. To enable the actual use of `Fragments` we need to:

- [x] Extend `SpeciesSite` to contain a new `SiteType::Fragment`, with fragment-specific `NETADefinition::fragment_` member, and update I/O to read in NETA string from the input file.
- [x] Extend `SiteStack` to generate sites from a `SiteType::Fragment` (see line 210 in `src/classes/sitestack.cpp`)
- [x] #1463
- [x] Test - reproduce water RDF system test using fragment sites.
- [x] Unit Test - generate orientable site on "complex" test species and perform some manipulations on the molecule to ensure we're getting expected results. A benzene molecule placed at the origin in the XY plane is a good example - we can test that we only generate a single site, and that any rotation of the fragment about the z-axis maintains all the Z-coordinates of the atoms as zero. Molecule information for this and the next task can be added to `unit/data/species.h` for common use.
- [x] Unit Test - More specific example at the origin - fluorobenzene - so that we can generate a single "known" fragment with all axes in well-defined directions. Perform similar tests to above.
- [x] Allow editing / display checking in GUI.
- [x] Integrate into TOML I/O (pending initial TOML integration work).

PR #1467, opened 2023-06-21
Fragments Part 3

jws-1:
This small PR continues work towards #1247, building on work done in #1461 and #1465 - and thus should be merged into `fragments` branch (corresponding to #1461) if/when approved.

Implements GUI functionality for defining/editing fragment sites - an indicator is placed adjacent to the NETA description line edit that indicates the validity of the description.


trisyoungs (commented):
Looks good - one small addition to make.

I'll address the system test stuff at the same time as some UI testing before approving.

trisyoungs (approved):

PR #1469, opened 2023-06-22
Nix - LLVM v13

trisyoungs:
Quick PR to update the version of LLVM installed by nix to 13 so we have a  `clang-format` version consistent with the CI.


jws-1 (approved):
:ok_hand:

Issue #72, opened 2019-04-19
Q atoms


trisyoungs:
Reproduce 'Q' atoms to offer consistency with EPSR. How can they be introduced consistently with the code (AtomTypeList etc.)

Or, don't - how often are they used?  Would potential gradients / geometries be more useful? For instance, a defined cylindrical gradient could be used to reproduce the formation of cylindrical pores in MCM-41.

jws-1:
Closed by #1473 & #1476.

PR #1473, opened 2023-06-26
Finite range simple global potentials

jws-1:
This PR implements a new `SimpleGlobalPotential` of LJ form, and works towards #72.


trisyoungs (changes requested):
So for a small PR there's a lot to unpack as it turns out!  First off, the easy one - the "SoftSphere" stuff I had added in can just be removed since it was just my original tinkering with implementing a general N-M potential, but was a bit of a waste of time!

Second - we now have the "Harmonic" and "LJ" potentials side-by-side as "simple" potentials. Trouble with that it that you can use the "LJ" form as an option in the `SimpleRestraintPotential` node, which causes havoc and crashed Dissolve because you're basically forcing atoms to overlap. 

So, what I think we need to do is have a separate class specifically for the restraint case - perhaps `SimpleRestraintPotentialFunction` - which contains only the "Harmonic" option (for now).  The `SimplePotentialFunction` case can have both LJ and Harmonic as it is now.

src/kernels/potentials/simple.cpp
```diff
@@ -10,15 +10,16 @@
...
+        "SimplePotentialFunction",
+        {{SimplePotentialFunctions::Form::Harmonic, "Harmonic", 1}, {SimplePotentialFunctions::Form::LJSphere, "LJSphere", 2}});
 }
 
 // Return parameters for specified form
 const std::vector<std::string> &SimplePotentialFunctions::parameters(Form form)
 {
     static std::map<SimplePotentialFunctions::Form, std::vector<std::string>> params_ = {
-        {SimplePotentialFunctions::Form::Harmonic, {"k"}}};
+        {SimplePotentialFunctions::Form::Harmonic, {"k"}}, {SimplePotentialFunctions::Form::LJSphere, {"epsilon", "sigma"}}};
```
trisyoungs, line 22:
```suggestion
        {SimplePotentialFunctions::Form::Harmonic, {"k"}}, {SimplePotentialFunctions::Form::LJ, {"epsilon", "sigma"}}};
```

src/kernels/potentials/simple.cpp
```diff
@@ -72,6 +73,14 @@ double SimplePotential::energy(const Atom &i, const Box *box) const
     {
         case (SimplePotentialFunctions::Form::Harmonic):
             return 0.5 * interactionPotential_.parameters()[0] * box->minimumDistanceSquared(i.r(), origin_);
+        case (SimplePotentialFunctions::Form::LJSphere):
```
trisyoungs, line 76:
```suggestion
        case (SimplePotentialFunctions::Form::LJ):
```

src/kernels/potentials/simplerestraint.cpp
trisyoungs, line 1: I want to suggest that this file is actually called just `restraint.cpp` (and `restraint.h`) since it's hard to imagine other classes of non-simple restraint.

src/kernels/potentials/simplerestraint.h
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#pragma once
+
+#include "classes/interactionpotential.h"
+#include "kernels/potentials/base.h"
+
+// SimpleRestraintPotential functional forms
+class SimpleRestraintPotentialFunctions
```
trisyoungs, line 10: And here too, sorry!
```suggestion
// RestraintPotential functional forms
class RestraintPotentialFunctions
```

src/kernels/potentials/simplerestraint.h
```diff
@@ -0,0 +1,58 @@
...
+    // Return parameters for specified form
+    static const std::vector<std::string> &parameters(Form form);
+    // Return nth parameter for the given form
+    static std::string parameter(Form form, int n);
+    // Return index of parameter in the given form
+    static std::optional<int> parameterIndex(Form form, std::string_view name);
+};
+
+// Simple Restraint Potential
+class SimpleRestraintPotential : public ExternalPotential
```
trisyoungs, line 28:
```suggestion
// Restraint Potential
class RestraintPotential : public ExternalPotential
```

src/procedure/nodes/simplerestraintpotential.cpp
trisyoungs, line 1: Rename to `restraintpotential.cpp`.

trisyoungs (changes requested):
Sorry - really annoying renaming changes requested.  Otherwise all looks tickety-boo!

src/kernels/potentials/restraint.h
```diff
@@ -0,0 +1,58 @@
...
+class RestraintPotential : public ExternalPotential
+{
+    public:
+    RestraintPotential();
+    ~RestraintPotential() = default;
+
+    /*
+     * Definition
+     */
+
```
trisyoungs, line 37:
```suggestion
```

trisyoungs (approved):
One more extraneous blank line, then good to go! :+1:

PR #1475, opened 2023-06-27
Fix SDF

trisyoungs:
This PR addresses an issue with the SDF module whereby the calculation was not correctly formed, and the surface was not properly constructed about the central system of axes.

To go into a patch release immediately.


src/math/matrix3.cpp
```diff
@@ -133,6 +133,13 @@ Matrix3 &Matrix3::transpose()
     return A;
 }
 
+// Transform the supplied vector by the transpose of the current matrix
+Vec3<double> Matrix3::transposeMultiply(const Vec3<double> &v) const
+{
+    return {v.x * matrix_[0] + v.y * matrix_[1] + v.z * matrix_[2], v.x * matrix_[3] + v.y * matrix_[4] + v.z * matrix_[5],
+            v.x * matrix_[6] + v.y * matrix_[7] + v.z * matrix_[8]};
+}
```
jws-1, line 141: My linear algebra is not quite up to scratch, so I might be wrong - but I don't see a transpose here!

src/modules/datatest/process.cpp
```diff
@@ -83,6 +83,25 @@ bool DataTestModule::process(Dissolve &dissolve, const ProcessPool &procPool)
...
+            return Messenger::error("No data with tag '{}' exists.\n", referenceData.tag());
+        const Data3D &data = *optData;
+        Messenger::print("Located reference data '{}'.\n", referenceData.tag());
+
+        // Generate the error estimate and compare against the threshold value
+        auto error = Error::error(errorType_, data.values().linearArray(), referenceData.values().linearArray(), true);
+        Messenger::print("Target data '{}' has error of {:7.3f} with calculated data and is {} (threshold is {:6.3e})\n\n",
+                         referenceData.tag(), error, isnan(error) || error > threshold_ ? "NOT OK" : "OK", threshold_);
+        if (isnan(error) || error > threshold_)
+            return false;
```
jws-1, line 102: Since we check `isnan(error) || error > threshold_` twice, it might be better to store the result in a variable

jws-1 (commented):
Looks good. A couple of a stylistic comments and one thing I wasn't 100% sure about.

src/io/import/data3d_pdens.cpp
```diff
@@ -0,0 +1,61 @@
...
+     * Line 4 : loop order (e.g. 'zyx')
+     * Line 5+: data (N = gridx*gridy*gridz)
+     */
+
+    // Get array dimensioos
+    if (parser.getArgsDelim() != LineParser::Success)
+        return false;
+    auto N = parser.argi(0);
+    auto iMin = parser.arg3i(3);
+    auto iMax = parser.arg3i(6);
```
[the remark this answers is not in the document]
trisyoungs: Nope!

src/math/matrix3.cpp
trisyoungs: If you compare this to `operator*`  on line 58 you'll see that the access on the `matrix_` array has had rows and columns swapped (from the original column-major storage pattern). I have also refactored that function to utilise the same simple inline construction as in the new transpose function.

src/modules/datatest/process.cpp
trisyoungs: Yep, we definitely should.

src/math/matrix3.cpp
jws-1: Aha!

jws-1 (approved):

PR #1479, opened 2023-06-30
Add benzene data and tests

trisyoungs:
This PR adds in a basic benzene simulation for testing purposes. At present only energy terms are tested, with the intention to calculate and add in structural data as and when other modules require it.

As it turns out, there was no "benzene-specific" bug in the code to speak of - just my continual (and repeated) erroneous interpretation of a DL_POLY warning message.


jws-1 (approved):
Looks good to me 👍

Issue #1415, opened 2023-04-28
Code Tidy 2023


trisyoungs:
**Describe the feature / issue**
For readability it is well worth renaming many of our source files so they are not wholly lowercase names - for multi-part names something like `interactionPotential.h` is significantly more readable than `interactionpotential.h`.

At the same time, we could consider moving the `Configuration` and `Species` classes to their own separate source directory, since these two objects are quite significant pieces of the puzzle, and `src/classes` is getting rather large.

### Tasks
- [x] Use lowerCamelCase for all source files
- [x] ~Move `Configuration` class to its own directory~
- [x] ~Move `Species` class to its own directory~
- [x] Handle `SpeciesSite`, `SiteStack`, and `RemoveFragment` classes after merge of #1478.
- [x] ~Rename ANTLR4 files (currently uppercamelcase)~ Not a good idea, since this would enforce related class names into lowerCamelCase against the rest of the code's convention.
- [x] ~Query use of `src/gui/delegates/customcombodelegate.h`, and rename/remove as necessary.~ Still in use.
- [x] ~variantpointer.h still exists as a template?~ Still in use by two classes. Captured in #964
- [x] `io`
- [x] `ff`

PR #1482, opened 2023-07-03
1415 Code Tidy 2023 - Part 1

trisyoungs:
This PR makes the first arduous step towards code tidying 2023, renaming the vast majority of files to conform to lowerCamelCase.

Future PRs will address remaining reorganisation and restructuring in (much!) smaller PRs.

Works towards #1415.


jws-1 (approved):
For such a massive PR, the review didn't end up taking very long! Everything seems OK so far. I see what you mean now regarding the `class_something` files. I wonder if we should consider having a separate directory for such classes, and then dropping the underscore notation? Just something to think about.

trisyoungs:
It's on the to-do list as a "trial" PR to follow!

PR #1483, opened 2023-07-03
Random functions for Expression

jws-1:
This small PR adds a couple of new functions for Expressions, that generate random numbers:
- `randomi(N)` generates a random integer in the inclusive range `(0, N)`.
- `random()` generates a number in the inclusive range `(0,1)`.

### DIFF
--- a/src/expression/function.cpp
+++ b/src/expression/function.cpp
@@ -17,15 +17,14 @@ EnumOptions<ExpressionFunctionNode::InternalFunction> ExpressionFunctionNode::in
                                                                                       {LnFunction, "ln", 1},
                                                                                       {LogFunction, "log", 1},
                                                                                       {PiFunction, "pi"},
-                                                                                      {RandIntFunction, "randint"},
-                                                                                      {RandIntRangeFunction, "randintrange", 1},
+                                                                                      {RandIntRangeFunction, "randomi", 1},
                                                                                       {SinFunction, "sin", 1},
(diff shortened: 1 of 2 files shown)


src/expression/function.cpp
```diff
@@ -15,13 +16,16 @@ EnumOptions<ExpressionFunctionNode::InternalFunction> ExpressionFunctionNode::in
                                                                                       {ExpFunction, "exp", 1},
                                                                                       {LnFunction, "ln", 1},
                                                                                       {LogFunction, "log", 1},
+                                                                                      {PiFunction, "pi"},
+                                                                                      {RandIntFunction, "randint"},
```
trisyoungs, line 20: I would probably lose this one as the least useful of the three - end users probably don't need the INT_MAX version!

src/expression/function.cpp
```diff
@@ -15,13 +16,16 @@ EnumOptions<ExpressionFunctionNode::InternalFunction> ExpressionFunctionNode::in
...
+                                                                                      {RandIntRangeFunction, "randintrange", 1},
```
trisyoungs, line 21: Maybe `randomi` is a more concise name, and we can change `rand` to `random` for consistency? This also mostly sidesteps the capitalisation of `randIntRange` to be consistent with `toDeg` and `toRad`!

trisyoungs (approved):
Works beautifully and I now have randomised ring rotations!  Couple of small tweaks, then good to go.

PR #1485, opened 2023-07-04
Remove nodes

### LINKED ISSUE #1420: Can't remove nodes from procedures
**Describe the bug**
It is currently not possible to remove nodes from procedures in the GUI.

That's it!

jws-1:
This PR adds the ability to remove nodes from a procedure. Closes #1420.

### DIFF
--- a/src/gui/models/procedureModel.cpp
+++ b/src/gui/models/procedureModel.cpp
@@ -475,6 +475,6 @@ bool ProcedureModel::removeRows(int row, int count, const QModelIndex &parent)
         scope->get().removeNode(data(index(row + i, 0), Qt::UserRole).value<std::shared_ptr<ProcedureNode>>());
     endRemoveRows();
 
-    emit(dataChanged(QModelIndex(), QModelIndex());
+    emit(dataChanged(QModelIndex(), QModelIndex()));
     return true;
 }
--- a/src/gui/models/procedureModel.cpp
+++ b/src/gui/models/procedureModel.cpp
@@ -5,6 +5,7 @@
 #include "gui/models/procedureModelMimeData.h"
 #include "procedure/nodes/registry.h"
 #include "procedure/procedure.h"
+#include "qabstractitemmodel.h"
 #include <QIODevice>
 #include <QIcon>
 #include <QMimeData>
@@ -458,6 +459,22 @@ bool ProcedureModel::insertRows(int row, int count, const QModelIndex &parent)
     for (auto i = 0; i < count; ++i)
         scope->get().insertEmpty(row + i);
     endInsertRows();
+
+    return true;
+}
+
+bool ProcedureModel::removeRows(int row, int count, const QModelIndex &parent)
+{
+    if (!procedure_)
+        return false;
+
+    // Get the scope associated to the parent index
+    auto scope = getScope(parent);
+
+    beginRemoveRows(parent, row, row + count - 1);
+    for (auto i = 0; i < count; ++i)
+        scope->get().removeNode(data(index(row + i, 0), Qt::UserRole).value<std::shared_ptr<ProcedureNode>>());
+    endRemoveRows();
 
     return true;
 }
--- a/src/gui/models/procedureModel.cpp
+++ b/src/gui/models/procedureModel.cpp
@@ -476,5 +476,6 @@ bool ProcedureModel::removeRows(int row, int count, const QModelIndex &parent)
         scope->get().removeNode(data(index(row + i, 0), Qt::UserRole).value<std::shared_ptr<ProcedureNode>>());
     endRemoveRows();
 
+    emit(dataChanged(QModelIndex(), QModelIndex());
     return true;
 }
--- a/src/gui/models/procedureModel.cpp
+++ b/src/gui/models/procedureModel.cpp
@@ -5,7 +5,6 @@
 #include "gui/models/procedureModelMimeData.h"
 #include "procedure/nodes/registry.h"
 #include "procedure/procedure.h"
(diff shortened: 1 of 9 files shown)


src/gui/procedureWidgetFuncs.cpp
```diff
@@ -175,6 +175,18 @@ void ProcedureWidget::on_AvailableNodesTree_doubleClicked(const QModelIndex &ind
...
 
+void ProcedureWidget::on_DeleteNodeButton_clicked(bool checked)
+{
+    auto selected = ui_.NodesTree->selectionModel()->selectedIndexes();
+    if (selected.size() > 0)
+    {
+        auto node = procedureModel_.data(selected.at(0), Qt::UserRole).value<std::shared_ptr<ProcedureNode>>();
+        assert(node);
+        procedure_.value().get().removeNode(node);
+        updateNodeTree();
```
trisyoungs, line 186: Ah, you see what you've done here is opened up a nasty can of worms for yourself! Unfortunately it's bad practice to remove the node from the procedure directly like this, since our widget is guided by a model (`ProcedureModel` in `src/gui/models/procedureModel.cpp` - already loving the new naming scheme!). That means that to ensure correctness we need to delete the node via the model itself so it understands what has changed. This would also emit `dataChanged` so that other keywords in the procedure are updated following (potentially) the removal of a node that one or more of them referenced.

src/gui/procedureWidgetFuncs.cpp
```diff
@@ -175,6 +175,18 @@ void ProcedureWidget::on_AvailableNodesTree_doubleClicked(const QModelIndex &ind
...
+    }
```
trisyoungs, line 187: As a related issue, once a node has been deleted the controls are still displayed for it, and no node ends up being selected in the tree.

trisyoungs (changes requested):
Good start, but it's time to revisit (and re-despise) Qt's Model-View framework I'm afraid!

src/gui/procedureWidgetFuncs.cpp
jws-1: Of course.. Luckily I have plenty of experience with models!

src/gui/models/procedureModel.cpp
```diff
@@ -5,6 +5,7 @@
 #include "gui/models/procedureModelMimeData.h"
 #include "procedure/nodes/registry.h"
 #include "procedure/procedure.h"
+#include "qabstractitemmodel.h"
```
trisyoungs, line 8: Not sure why this `#include` is here?  I don't need it on my system, and it's also a system header being included in quotes rather than `<>`. VS to blame?

src/gui/procedureWidget.h
```diff
@@ -60,6 +60,7 @@ class ProcedureWidget : public QWidget
     void updateNodeTree();
     void on_NodesTree_customContextMenuRequested(const QPoint &pos);
     void on_AvailableNodesTree_doubleClicked(const QModelIndex &index);
+    void on_DeleteNodeButton_clicked(bool checked);
```
trisyoungs, line 63: Bug :lady_beetle: report!  If I click the delete node button when the list is empty (or presumably when no node is selected) I get a segfault.  May be nicer UX to disable the button if no node is selected.

src/gui/models/procedureModel.cpp
```diff
@@ -461,3 +462,19 @@ bool ProcedureModel::insertRows(int row, int count, const QModelIndex &parent)
...
+        return false;
+
+    // Get the scope associated to the parent index
+    auto scope = getScope(parent);
+
+    beginRemoveRows(parent, row, row + count - 1);
+    for (auto i = 0; i < count; ++i)
+        scope->get().removeNode(data(index(row + i, 0), Qt::UserRole).value<std::shared_ptr<ProcedureNode>>());
+    endRemoveRows();
+
```
trisyoungs, line 478: Need to emit `dataChanged` otherwise we don't properly update other things.
```suggestion

    emit(dataChanged(QModelIndex(), QModelIndex());
```

Furthermore, need to call `updateControls()` in `ProcedureWidget::procedureDataChanged()` (line 134) since the relationships between data are validated by the function, but the incorrect values are still displayed in the current control widget because we don't update it.

src/gui/models/procedureModel.cpp
jws-1: That'd be my neovim LSP config to blame!

trisyoungs: Why am I only just hearing about `neovim`? :disappointed:

trisyoungs (approved):
Looks good, works nicely!  I'm obliged to mention for the general record that the inability to undo a node deletion (or addition, or move, or keyword change...) is a bit of an issue. So, while we are not using TOML for input file I/O immediately, it is in our interest to get that up to scratch and compiled as part of the `develop` stream so we can leverage it for undo/redo actions like this.

Aaaaaanyway, that's obviously not in scope for this PR, but I just wanted to write it down!

jws-1:
> Looks good, works nicely! I'm obliged to mention for the general record that the inability to undo a node deletion (or addition, or move, or keyword change...) is a bit of an issue. So, while we are not using TOML for input file I/O immediately, it is in our interest to get that up to scratch and compiled as part of the `develop` stream so we can leverage it for undo/redo actions like this.
> 
> Aaaaaanyway, that's obviously not in scope for this PR, but I just wanted to write it down!

Yes, I thought it might be an issue. I was tempted to include a dialog to validate the user's decision - but I decided it might be a bit annoying!

PR #1491, opened 2023-07-05
Node-ify Size Factor

### LINKED ISSUE #1395: Node-ify SizeFactor control
SizeFactor control is currently enforced by controls in the Configuration tab, but would be more appropriate as a simple `ProcedureNode`.

jws-1:
This PR moves the Size Factor for configurations from being a GUI control to a node: `SizeFactor`. Closes #1395.

### DIFF
--- a/src/procedure/nodes/registry.cpp
+++ b/src/procedure/nodes/registry.cpp
@@ -56,7 +56,7 @@ ProcedureNodeRegistry::ProcedureNodeRegistry()
     registerProducer<CopyProcedureNode>(ProcedureNode::NodeType::Copy, "Copy the contents of a configuration", "Build");
     registerProducer<RemoveProcedureNode>(ProcedureNode::NodeType::Remove, "Remove molecules from a configuration", "Build");
     registerProducer<SizeFactorProcedureNode>(ProcedureNode::NodeType::SizeFactor,
-                                              "Request a size factor to apply to configuration", "Build");
+                                              "Scale a configuration's contents with a size factor", "Build");
     registerProducer<TransmuteProcedureNode>(ProcedureNode::NodeType::Transmute,
                                              "Turn molecules from one species type into another", "Build");
 
--- a/src/procedure/nodes/registry.cpp
+++ b/src/procedure/nodes/registry.cpp
@@ -55,7 +55,8 @@ ProcedureNodeRegistry::ProcedureNodeRegistry()
                                                   "Generate coordinate sets for a species", "Build");
     registerProducer<CopyProcedureNode>(ProcedureNode::NodeType::Copy, "Copy the contents of a configuration", "Build");
     registerProducer<RemoveProcedureNode>(ProcedureNode::NodeType::Remove, "Remove molecules from a configuration", "Build");
-    registerProducer<SizeFactorProcedureNode>(ProcedureNode::NodeType::SizeFactor, "Request a size factor to apply to configuration", "Build");
+    registerProducer<SizeFactorProcedureNode>(ProcedureNode::NodeType::SizeFactor,
+                                              "Request a size factor to apply to configuration", "Build");
     registerProducer<TransmuteProcedureNode>(ProcedureNode::NodeType::Transmute,
                                              "Turn molecules from one species type into another", "Build");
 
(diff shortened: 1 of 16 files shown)


src/procedure/nodes/sizeFactor.h
```diff
@@ -0,0 +1,32 @@
...
+     */
+    public:
+    // Return whether a name for the node must be provided
+    bool mustBeNamed() const override;
+
+    /*
+     * Node Data
+     */
+    private:
+    NodeValue sizeFactor_{1.0};
```
trisyoungs, line 24: I think we can default to `10.0` here since adding the node into the procedure implies you want to do the scaling, and `10.0` is what most people will go for.
```suggestion
    NodeValue sizeFactor_{10.0};
```

src/procedure/nodes/sizeFactor.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+/*
+ * Execute
+ */
+
+// Execute node
+bool SizeFactorProcedureNode::execute(const ProcedureContext &procedureContext)
+{
+    auto *cfg = procedureContext.configuration();
+    cfg->setRequestedSizeFactor(sizeFactor_.asDouble());
```
trisyoungs, line 29: It feels like we should actually do the physical scaling here as well.  Otherwise the procedure appears to "not work" as the box isn't scaled at the beginning.

jws-1: Would this mean we no longer need to keep track of the `requestedSizeFactor_` and `appliedSizeFactor_` outside of the node?

jws-1: I'm guessing that we still need to keep the `appliedSizeFactor_`?

trisyoungs: No, they still need to be stored in the `Configuration` since it handles the resizing itself (see `main/simulation.cpp`, line 277). The node just needs to set the initial value and perform the first scaling - the reduction of the size factor happens as the main simulation proceeds so the node is "out of use" by then.

src/procedure/nodes/sizeFactor.cpp
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "procedure/nodes/sizeFactor.h"
+#include "classes/configuration.h"
+#include "keywords/nodeValue.h"
+#include "procedure/nodes/node.h"
+SizeFactorProcedureNode::SizeFactorProcedureNode()
```
trisyoungs, line 8: Arrgggh - it's the Blank Line Police again! :smile:
```suggestion

SizeFactorProcedureNode::SizeFactorProcedureNode()
```

src/procedure/nodes/sizeFactor.cpp
```diff
@@ -0,0 +1,31 @@
...
+    : ProcedureNode(ProcedureNode::NodeType::SizeFactor, {ProcedureNode::GenerationContext})
+{
+    keywords_.add<NodeValueKeyword>("SizeFactor", "Size factor for scaling", sizeFactor_, this);
```
trisyoungs, line 11:
```suggestion
    keywords_.add<NodeValueKeyword>("SizeFactor", "Size factor scaling to apply to configuration", sizeFactor_, this);
```

src/procedure/nodes/registry.cpp
```diff
@@ -54,6 +55,8 @@ ProcedureNodeRegistry::ProcedureNodeRegistry()
                                                   "Generate coordinate sets for a species", "Build");
     registerProducer<CopyProcedureNode>(ProcedureNode::NodeType::Copy, "Copy the contents of a configuration", "Build");
     registerProducer<RemoveProcedureNode>(ProcedureNode::NodeType::Remove, "Remove molecules from a configuration", "Build");
+    registerProducer<SizeFactorProcedureNode>(ProcedureNode::NodeType::SizeFactor,
+                                              "Request a size factor to apply to configuration", "Build");
```
trisyoungs, line 59:
```suggestion
                                              "Scale a configuration's contents with a size factor", "Build");
```

trisyoungs (approved):
More picky text / minor stuff, then good to go. :+1:

PR #1492, opened 2023-07-05
Node-ify Temperature

### LINKED ISSUE #146: Configuration temperature control (new Module)
Allow control over Configuration temperature as a function of iteration, providing schedules for heating / cooling.

New module - TemperatureControlModule?

jws-1:
This PR moves the Temperature for configurations from being a GUI control to a node: `Temperature`. For now, we still read `TemperatureKeywords`, but with a deprecation warning. Closes #146.

### DIFF
--- a/src/gui/addConfigurationDialogFuncs.cpp
+++ b/src/gui/addConfigurationDialogFuncs.cpp
@@ -182,7 +182,7 @@ void AddConfigurationDialog::finalise()
     auto &generator = newConfiguration->generator();
 
     // Add Temperature node, by default
-    generator.createRootNode<TemperatureProcedureNode>("Temperature");
+    generator.createRootNode<TemperatureProcedureNode>({});
 
     std::shared_ptr<GeneralRegionProcedureNode> regionNode;
 
--- a/src/gui/addConfigurationDialogFuncs.cpp
+++ b/src/gui/addConfigurationDialogFuncs.cpp
@@ -180,7 +180,10 @@ void AddConfigurationDialog::finalise()
     // Create the Configuration and a suitable generator
     auto *newConfiguration = dissolve_.addConfiguration();
     auto &generator = newConfiguration->generator();
+
+    // Add Temperature node, by default
     generator.createRootNode<TemperatureProcedureNode>("Temperature");
+
     std::shared_ptr<GeneralRegionProcedureNode> regionNode;
 
     // Add the framework species if present
--- a/src/gui/addConfigurationDialogFuncs.cpp
+++ b/src/gui/addConfigurationDialogFuncs.cpp
@@ -10,6 +10,7 @@
 #include "procedure/nodes/coordinateSets.h"
 #include "procedure/nodes/generalRegion.h"
 #include "procedure/nodes/parameters.h"
+#include "procedure/nodes/temperature.h"
 #include <QInputDialog>
 #include <QMessageBox>
 
@@ -179,6 +180,7 @@ void AddConfigurationDialog::finalise()
     // Create the Configuration and a suitable generator
     auto *newConfiguration = dissolve_.addConfiguration();
     auto &generator = newConfiguration->generator();
+    generator.createRootNode<TemperatureProcedureNode>("Temperature");
     std::shared_ptr<GeneralRegionProcedureNode> regionNode;
 
     // Add the framework species if present
--- a/src/gui/gui.ui
+++ b/src/gui/gui.ui
@@ -806,7 +806,7 @@
   </action>
   <action name="ConfigurationAdjustTemperatureAction">
    <property name="text">
-    <string>Adjust Temperature...</string>
+    <string>Adjust &amp;Temperature...</string>
    </property>
   </action>
  </widget>
--- a/src/gui/gui.ui
+++ b/src/gui/gui.ui
@@ -270,6 +270,8 @@
     <addaction name="ConfigurationDeleteAction"/>
     <addaction name="separator"/>
     <addaction name="ConfigurationExportToMenu"/>
+    <addaction name="separator"/>
+    <addaction name="ConfigurationAdjustTemperatureAction"/>
    </widget>
    <widget class="QMenu" name="LayerMenu">
     <property name="font">
@@ -800,6 +802,11 @@
    </property>
    <property name="text">
     <string>Create...</string>
+   </property>
+  </action>
+  <action name="ConfigurationAdjustTemperatureAction">
+   <property name="text">
+    <string>Adjust Temperature...</string>
    </property>
   </action>
  </widget>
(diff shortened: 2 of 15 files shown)


src/main/keywords_configuration.cpp
```diff
@@ -77,6 +77,7 @@ bool ConfigurationBlock::parse(LineParser &parser, Dissolve *dissolve, Configura
                 cfg->setRequestedSizeFactor(parser.argd(1));
                 break;
             case (ConfigurationBlock::TemperatureKeyword):
+                Messenger::warn("The TemperatureKeyword will be deprecated in future versions.\n");
```
trisyoungs, line 80:
```suggestion
                Messenger::warn("The 'Temperature' keyword will be deprecated in a future version.\n");
```

src/procedure/nodes/temperature.cpp
```diff
@@ -0,0 +1,28 @@
...
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "procedure/nodes/temperature.h"
+#include "classes/configuration.h"
+#include "keywords/nodeValue.h"
+#include "procedure/nodes/node.h"
+TemperatureProcedureNode::TemperatureProcedureNode()
+    : ProcedureNode(ProcedureNode::NodeType::Temperature, {ProcedureNode::GenerationContext})
+{
+    keywords_.add<NodeValueKeyword>("Temperature", "Temperature in K", temperature_, this);
```
trisyoungs, line 11:
```suggestion
    keywords_.add<NodeValueKeyword>("Temperature", "Temperature (K)", temperature_, this);
```

src/procedure/nodes/temperature.cpp
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/*
+ * Identity
+ */
+bool TemperatureProcedureNode::mustBeNamed() const { return false; }
```
trisyoungs, line 17:
```suggestion

bool TemperatureProcedureNode::mustBeNamed() const { return false; }
```

src/procedure/nodes/temperature.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+/*
+ * Execute
+ */
+// Execute node
```
trisyoungs, line 22:
```suggestion

// Execute node
```

trisyoungs (changes requested):
Looks good, but I think there's a bit more useful work we should do so as not to unnecessarily confuse our users!

First, from the main menu when a user creates a new `Configuration` we need to add our new `TemperatureNode` at the top of the procedure - see `gui/addConfigurationDialogFuncs.cpp`. It should go in at around line 183.

Second, I think we should add a new menu item to allow the temperature of an existing configuration to be set. We shouldn't force the user to regenerate the config from the beginning if they need to tweak the temperature!

src/gui/addConfigurationDialogFuncs.cpp
```diff
@@ -179,6 +180,10 @@ void AddConfigurationDialog::finalise()
     // Create the Configuration and a suitable generator
     auto *newConfiguration = dissolve_.addConfiguration();
     auto &generator = newConfiguration->generator();
+
+    // Add Temperature node, by default
+    generator.createRootNode<TemperatureProcedureNode>("Temperature");
```
trisyoungs, line 185: I'd just default this one so it gets the standard node naming
```suggestion
    generator.createRootNode<TemperatureProcedureNode>({});
```

src/gui/gui.ui
```diff
@@ -802,6 +804,11 @@
     <string>Create...</string>
    </property>
   </action>
+  <action name="ConfigurationAdjustTemperatureAction">
+   <property name="text">
+    <string>Adjust Temperature...</string>
```
trisyoungs, line 809:
```suggestion
    <string>Adjust &amp;Temperature...</string>
```

src/gui/menu_configuration.cpp
```diff
@@ -136,3 +137,24 @@ void DissolveWindow::on_ConfigurationExportToXYZAction_triggered(bool checked)
...
+{
+    // Get the currently-displayed Configuration
+    auto *cfg = ui_.MainTabs->currentConfiguration();
+    if (!cfg)
+        return;
+
+    auto ok = false;
+    auto temperature = QInputDialog::getDouble(this, "Set temperature for configuration",
+                                               "Enter the temperature (K) to apply to this configuration", cfg->temperature(),
+                                               -10000.0, 10000.0, 3, &ok);
```
trisyoungs, line 151: Well, that's breaking the laws of physics!  :smiley:
```suggestion
                                               0.0, 1000000.0, 3, &ok);
```

src/procedure/nodes/temperature.cpp
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "procedure/nodes/temperature.h"
+#include "classes/configuration.h"
+#include "keywords/nodeValue.h"
+#include "procedure/nodes/node.h"
+TemperatureProcedureNode::TemperatureProcedureNode()
```
trisyoungs, line 8: All hail the BLP!
```suggestion

TemperatureProcedureNode::TemperatureProcedureNode()
```

src/procedure/nodes/temperature.cpp
```diff
@@ -0,0 +1,30 @@
...
+    : ProcedureNode(ProcedureNode::NodeType::Temperature, {ProcedureNode::GenerationContext})
+{
+    keywords_.add<NodeValueKeyword>("Temperature", "Temperature (K)", temperature_, this);
+}
+
+/*
+ * Identity
+ */
+
+bool TemperatureProcedureNode::mustBeNamed() const { return false; }
```
trisyoungs, line 18: Convention.
```suggestion
// Return whether a name for the node must be provided
bool TemperatureProcedureNode::mustBeNamed() const { return false; }
```

src/procedure/nodes/temperature.h
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#pragma once
+
+#include "keywords/nodeValue.h"
+#include "procedure/nodes/node.h"
+class TemperatureProcedureNode : public ProcedureNode
```
trisyoungs, line 8:
```suggestion

class TemperatureProcedureNode : public ProcedureNode
```

trisyoungs (approved):
Again, just minor stuff. :+1:

We now don't display the temperature on the `ConfigurationTab`, but as long as we're anticipating the tab redesign to be done in the near future no problem.  If not, then adding a label in during the interim is the way to go.

src/gui/menu_configuration.cpp
jws-1: Just keeping you on your toes!

trisyoungs:
Agreed.  We'll leave this as-is for now, and work the GUI reorganisation to address the issue.

trisyoungs (approved):

PR #1500, opened 2023-07-06
Reorganise GUI

jws-1:
This PR reorganises the `SpeciesTab` and the `ConfigurationTab` making them less cluttered, more aesthetically pleasing and more immediately informative with the ultimate aim of greatly improving user experience.


trisyoungs (commented):
OK, for the `SpeciesTab` I think the toolbox is a good solution, if a touch on the ugly side (not your fault!).  For the `ConfigurationTab` its nice and clean, but I'm going to vote against the actual `QStatusBar` for two reasons. First, I always disliked having to programmatically create all the content in them, rather than doing it with the UI. Second, I don't think it has the control we need or want. Actually, the third one not necessarily related to the `QStatusBar` is the little form that pops up when you hover over the box icon - on my laptop (Wayland) you have to click on it to get rid of it!

I've mocked up what I have in my head using sunken `QFrames` and disabled `QToolButton`s as the icon holders.  The box lengths angles can easily be encapsulated in  a simple `QToolTip` for the label. The density units should be a combo box, but you get the idea. The final two numbers are the number of molecules and number of atoms.

image

jws-1:
> 

Well.. that's definitely much better looking than what I had in mind!

trisyoungs (commented):
First batch of comments, but this is all heading in the right direction.  I had a thought about the organisation of the `SpeciesTab`, trying to remove the horrid `QToolBox` and make it a bit more consistent with the `ConfigurationTab`:
image

Here we're using the same ideas for the top information bar as in the `ConfigurationTab`, meaning that we can remove the "Box" toolbox on the left, and I've ditched the `QToolBox` entirely for an arrangement of buttons and frames which can be hidden as the buttons are selected.  Just a thought!

src/gui/configurationTab.h
```diff
@@ -67,17 +67,17 @@ class ConfigurationTab : public QWidget, public MainTab
     public:
     // Update controls in tab
     void updateControls() override;
-    // Prevent editing within tab
+    // Prevent editing of the generator
     void preventEditing() override;
-    // Allow editing within tab
+    // Allow editing of the generator
```
trisyoungs, line 72: These updated comments might be accurate, but the functions to prevent/allow editing are still general across all widget types, so I would stick to "within tab".

src/gui/configurationTabFuncs.cpp
```diff
@@ -108,35 +108,40 @@ void ConfigurationTab::updateControls()
...
-    ui_.CurrentBoxCLabel->setText(QString::number(box->axisLengths().z));
-    ui_.CurrentBoxAlphaLabel->setText(QString::number(box->axisAngles().x));
-    ui_.CurrentBoxBetaLabel->setText(QString::number(box->axisAngles().y));
-    ui_.CurrentBoxGammaLabel->setText(QString::number(box->axisAngles().z));
+    QString boxInfo = QString("<b>A:</b>  %1<br>").arg(box->axisLengths().x);
+    boxInfo += QString("<b>B:</b>  %1<br>").arg(box->axisLengths().y);
+    boxInfo += QString("<b>C:</b>  %1<br>").arg(box->axisLengths().z);
+    boxInfo += QString("<b>%1:</b>  %2<br>").arg(QString::fromUtf8("\u03B1")).arg(box->axisAngles().x);
+    boxInfo += QString("<b>%1:</b>  %2<br>").arg(QString::fromUtf8("\u03B2")).arg(box->axisAngles().y);
+    boxInfo += QString("<b>%1:</b>  %2").arg(QString::fromUtf8("\u03B3")).arg(box->axisAngles().z);
```
trisyoungs, line 122: Gotta have some units! Angstroms for the lengths (&#8491;) and degrees for the angles. Space between number and Angstrom, and none between number and degrees, otherwise I'll get grumpy. :)

src/gui/geometryDialog.ui
trisyoungs, line 1: The font in the dialog is the default size (12 pt) while we reduce that for Dissolve.  It's annoying that widgets, especially dialogs and `QMenu`s, don't seem to inherit their parent's font, so you'll have to grab it and set it manually when you create it.

src/gui/geometryDialogFuncs.cpp
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#pragma once
+
```
trisyoungs, line 5:
```suggestion
```

src/gui/speciesTab.h
```diff
@@ -150,4 +147,6 @@ class SpeciesTab : public QWidget, public MainTab
     public slots:
     // Update sites tab
     void updateSitesTab();
+    // Button group toggled
+    void buttonGroupToggled(QAbstractButton *button, bool checked);
```
trisyoungs, line 151: Strictly speaking this should be a `private slot`.

src/gui/speciesTab.ui
trisyoungs, line 1: I actually think this layout works really well.  One issue I had with it was that the buttons and frame contents seemed to blend into one continuous mess, so I made a couple of small tweaks to get the result in the pic below:

- Boldified the button titles
- Placed each button / frame pair in its own `QWidget` container with zero margins and zero spacing
- Reduced the spacing in the overall vertical container from `6` to `4` in line with the rest of Dissolve.

image

src/gui/speciesTabFuncs.cpp
```diff
@@ -149,24 +120,20 @@ void SpeciesTab::updateControls()
...
-    ui_.ViewerWidget->postRedisplay();
+    // Current Box
+    const auto *box = species_->box();
+    ui_.CurrentBoxTypeLabel->setText(QString::fromStdString(std::string(Box::boxTypes().keyword(box->type()))));
+    QString boxInfo = QString("<b>A:</b>  %1<br>").arg(box->axisLengths().x);
+    boxInfo += QString("<b>B:</b>  %1<br>").arg(box->axisLengths().y);
+    boxInfo += QString("<b>C:</b>  %1<br>").arg(box->axisLengths().z);
+    boxInfo += QString("<b>%1:</b>  %2<br>").arg(QString::fromUtf8("\u03B1")).arg(box->axisAngles().x);
+    boxInfo += QString("<b>%1:</b>  %2<br>").arg(QString::fromUtf8("\u03B2")).arg(box->axisAngles().y);
+    boxInfo += QString("<b>%1:</b>  %2").arg(QString::fromUtf8("\u03B3")).arg(box->axisAngles().z);
```
trisyoungs, line 131: Units.

src/gui/speciesViewer_input.cpp
```diff
@@ -164,6 +193,85 @@ void SpeciesViewer::contextMenuRequested(QPoint pos)
...
+            elements.push_back(i->Z());
+        site->setDynamicElements(elements);
+        setSite(site);
+        postRedisplay();
+        emit(siteCreatedAndShown());
+    }
+    else if (actionMap[selectedAction] == "CreateDynamicAtomTypes")
+    {
+        auto *site =
+            species_->addSite(joinStrings(species_->selectedAtoms(), " ", [](const auto &i) { return i->atomType()->name(); }));
```
trisyoungs, line 239: Prefer underscores to spaces for generated names in Dissolve.
```suggestion
            species_->addSite(joinStrings(species_->selectedAtoms(), "_", [](const auto &i) { return i->atomType()->name(); }));
```

src/gui/configurationTab.ui
trisyoungs, line 1: For consistency  we should adopt the same left-hand "custom tab" stack for the `ConfigurationTab` , even though its just the Generator and that might get moved at some point.

src/gui/configurationTabFuncs.cpp
```diff
@@ -108,35 +108,40 @@ void ConfigurationTab::updateControls()
...
+    ui_.CurrentBoxFrame->setToolTip(boxInfo);
     updateDensityLabel();
 
+    // Populations
+    ui_.AtomPopulationLabel->setText(QString::number(configuration_->nAtoms()));
+    ui_.MoleculePopulationLabel->setText(QString::number(configuration_->nMolecules()));
```
trisyoungs, line 128: I think a static tooltip on these two frames explicitly stating "Number of molecules in the configuration" and "Number of atoms in the configuration" would be useful.

src/gui/speciesTab.ui
trisyoungs, line 1: Also, no reason why we shouldn't have a "Number of atoms in the species" frame on our little status bar.  Also, actually, one displaying the empirical formula (as is done currently in the bottom of the viewer) would be nice. I _also_ wonder whether the total charges should be put up there as well for better visibility.

trisyoungs, line 1: Keeping the geometry terms in a separate dialog I think is definitely the way to go, but perhaps a summary of the number of each term type within the Atoms frame might be nice.

trisyoungs, line 1: In the Atoms table we can reclaim a bit of useful space by calling the first to columns `El` and `AT`, and we should call resizeColumnsToContents() to get a minimised version on show.  Furthermore, we really don't need to display the element name, and we should just do the element symbol instead.

trisyoungs (changes requested):
Round 2!  This is working really well, I think, and the new speciestab layout feels pretty natural to use. I haven't fully played around with all the context menu options, but the ones I did try worked just as I expected them to which was nice!

Gonna need some new icons at some point.

Have made some more suggestions for improvements in the comments, but :+1: :+1:

trisyoungs (approved):
I've updated the density icon and added one (supposed to) represent the periodic table for the empirical formula so we don't overuse the other icon. If we can add a tooltip explaining what the empirical formula frame is (i.e. "Empirical formula of the species") then I reckon we're done!

PR #1501, opened 2023-07-06
Empty Expressions

### LINKED ISSUE #1487: Empty Expressions
**Describe the bug**
An empty expression (often) causes a segmentation fault.

**To Reproduce**
1. Add a node to a procedure that uses a `NodeValueKeyword` (e.g. `CustomRegion`), i.e. allows an expression to be entered.
2. Empty the text edit of the keyword and press enter.
3. Segmentation fault.

**Expected behaviour**
Not to segfault, should instead indicate that the expression is invalid.

**Operating System:**
 - OS: Linux
 - OS version / flavour: Ubuntu 20.04

jws-1:
This slightly-hacky PR prevents segmentation faults occurring when expression strings in nodes are empty. Closes #1487.

### DIFF
--- a/src/base/sysFunc.cpp
+++ b/src/base/sysFunc.cpp
@@ -295,7 +295,7 @@ bool DissolveSys::isNumber(std::string_view text, bool &isFloatingPoint)
                 break;
         }
     }
-    
+
     isFloatingPoint &= nDigits > 0;
     return nDigits > 0;
 }
--- a/src/base/sysFunc.cpp
+++ b/src/base/sysFunc.cpp
@@ -251,6 +251,7 @@ bool DissolveSys::isNumber(std::string_view text, bool &isFloatingPoint)
     isFloatingPoint = false;
 
     std::optional<int> exponentIndex;
+    auto nDigits = 0;
 
     const auto length = text.size();
     for (auto n = 0; n < length; ++n)
@@ -290,11 +291,12 @@ bool DissolveSys::isNumber(std::string_view text, bool &isFloatingPoint)
                 // If not a digit, return false
                 if (!isdigit(c))
                     return false;
+                ++nDigits;
                 break;
         }
     }
 
-    return true;
+    return nDigits > 0;
 }
 
 // Replace all occurrences of search string with replace string
--- a/src/base/sysFunc.cpp
+++ b/src/base/sysFunc.cpp
@@ -295,7 +295,8 @@ bool DissolveSys::isNumber(std::string_view text, bool &isFloatingPoint)
                 break;
         }
     }
-
+    
+    isFloatingPoint &= nDigits > 0;
     return nDigits > 0;
 }
 
--- a/src/procedure/nodes/sequence.cpp
+++ b/src/procedure/nodes/sequence.cpp
@@ -401,6 +401,11 @@ bool ProcedureNodeSequence::check() const
         // Check node branch if present
         if (node->branch() && !node->branch()->get().check())
             return false;
+
+        // Check validity of NodeValue-based keywords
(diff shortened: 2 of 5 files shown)


trisyoungs:
Ignoring the swathe of broken unit tests for now (:grin:) I can also segfault Dissolve as simply as adding in a rogue backtick on to an expression, for example adding one on to the end of `rho` in my `AddNode` density keyword (heh - overly descriptive: try getting a backtick to display in code in a comment!) I get a `throw` of "Failed to evaluate expression ''.".  This shows up correctly as a red cross in the GUI, but is blindly ignored when you run the procedure.  So this is a bigger problem than perhaps anticipated.

I think the neatest solution would be to extend the `ProcedureNodeSequence::check()` to check any `NodeValue`-based keywords on each node for validity, calling the `check()` function before the procedure is executed.

jws-1:
> Ignoring the swathe of broken unit tests for now (grin) I can also segfault Dissolve as simply as adding in a rogue backtick on to an expression, for example adding one on to the end of `rho` in my `AddNode` density keyword (heh - overly descriptive: try getting a backtick to display in code in a comment!) I get a `throw` of "Failed to evaluate expression ''.". This shows up correctly as a red cross in the GUI, but is blindly ignored when you run the procedure. So this is a bigger problem than perhaps anticipated.
> 
> I think the neatest solution would be to extend the `ProcedureNodeSequence::check()` to check any `NodeValue`-based keywords on each node for validity, calling the `check()` function before the procedure is executed.

While that does seem to solve the `backtick` problem, it doesn't solve the issue of empty expressions and expressions with periods and no numbers. Both issues occur as `isNumber` returns true, and then `std::stod` and `std::stoi` throw `std::invalid_argument`.

src/procedure/nodes/sequence.cpp
```diff
@@ -401,6 +401,11 @@ bool ProcedureNodeSequence::check() const
         // Check node branch if present
         if (node->branch() && !node->branch()->get().check())
             return false;
+
+        // Check validity of NodeValue-based keywords
+        for (auto &kwd : node->keywords().allOfType<NodeValue>())
+            if (!kwd->isValid())
+                return false;
```
trisyoungs, line 408: Just adding a couple of other `NodeValue`-based classes, and to note that these must reference the keyword class, not the datatype.  Code not tested, BTW!
```suggestion

        // Check validity of NodeValue-based keywords
        for (auto &kwd : node->keywords().allOfType<NodeValueKeyword>())
            if (!kwd->data().isValid())
                return false;
        for (auto &kwd : node->keywords().allOfType<Vec3NodeValueKeyword>())
            if (!kwd->data().x.isValid() || !kwd->data().y.isValid() || !kwd->data().z.isValid())
                return false;
        for (auto &kwd : node->keywords().allOfType<NodeValueEnumOptionsBaseKeyword>())
            if (!kwd->value().isValid())
                return false;
```

trisyoungs (changes requested):

src/procedure/nodes/sequence.cpp
jws-1: Oops.. still familiarising myself with the `NodeValue` stuff.

src/base/sysFunc.cpp
```diff
@@ -290,11 +291,13 @@ bool DissolveSys::isNumber(std::string_view text, bool &isFloatingPoint)
                 // If not a digit, return false
                 if (!isdigit(c))
                     return false;
+                ++nDigits;
                 break;
         }
     }
 
-    return true;
+    isFloatingPoint &= nDigits > 0;
```
trisyoungs, line 299: Bit witchcraft!

trisyoungs (approved):
:+1:

PR #1507, opened 2023-07-10
Site Crashes

### LINKED ISSUE #1504: Crashing when creating static/fragment sites
When creating a `Static` or `Fragment` site, crashes occur. In the case of `Static` sites, this occurs if you create a new site and then set X-axis atoms. With `Fragment` sites, this occurs if you have a description without any origin tags (such as the empty string!). This does seem to be GUI related issue.

jws-1:
Small PR to fix some *dumb* mistakes causing crashes when adding/creating `Static`/`Fragment` sites. Closes #1504.

### DIFF
--- a/src/classes/speciesSite.cpp
+++ b/src/classes/speciesSite.cpp
@@ -70,7 +70,7 @@ SpeciesSite::SiteType SpeciesSite::type() const { return type_; }
 // Return whether the site has defined axes sites
 bool SpeciesSite::hasAxes() const
 {
-    return (type_ == SiteType::Static && !(staticXAxisAtoms_.empty() && staticYAxisAtoms_.empty())) ||
+    return (type_ == SiteType::Static && !(staticXAxisAtoms_.empty() || staticYAxisAtoms_.empty())) ||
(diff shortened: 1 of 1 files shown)


src/classes/speciesSite.cpp
```diff
@@ -70,7 +70,7 @@ SpeciesSite::SiteType SpeciesSite::type() const { return type_; }
 // Return whether the site has defined axes sites
 bool SpeciesSite::hasAxes() const
 {
-    return (type_ == SiteType::Static && !(staticXAxisAtoms_.empty() && staticYAxisAtoms_.empty())) ||
+    return (type_ == SiteType::Static && !(staticXAxisAtoms_.empty() || staticYAxisAtoms_.empty())) ||
```
trisyoungs, line 73: I did look at this logic first time around, but also didn't quite process it right!

trisyoungs (approved):

src/classes/speciesSite.cpp
jws-1: Yep... not sure what happened with either of us there!

PR #1508, opened 2023-07-10
UI Fixes 07/23

### LINKED ISSUE #1494: Crash when adding configuration
**Describe the bug**
Using the Add Configuration Wizard gives a crash when trying to create a mixture of atoms with ‘fixed geometry and size’ and adjusting the density.

On IDAaaS.

### LINKED ISSUE #1496: Load restart file menu option - bad UX
**Describe the bug**
Choosing "File->Load restart file" from the main menu pops up a dialog regarding clearing the module data, which is partially accurate, but it also suggests that configuration data will remain intact which is definitely not the case.

The process should be checked and reworked as appropriate.

### LINKED ISSUE #1497: Load recent doesn't handle missing files well
**Describe the bug**
From a fresh start of Dissolve, opening a recent file from the menu which no longer exists causes an error to be printed to the messages, but of course we can't see it because the messages tab isn't available yet. Need a dialog or similar.

### LINKED ISSUE #1498: Messages not cleared when opening new file
**Describe the bug**
Opening a recent file does not clear the messages buffed, meaning that any old stuff (including errors) are still present and detected in the status bar.

trisyoungs:
This PR addresses a handful of small and simple UI fixes.

Closes #1498.
Closes #1497.
Closes #1496.
Closes #1494.


jws-1 (approved):
Everything looks good to me.

PR #1509, opened 2023-07-10
LowerCamelCase for Unit Tests

trisyoungs:
This PR renames several files in the unit tests dir to lowerCamelCase, and which was missed in the original two PRs.


unit/classes/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
...
 dissolve_unit_test(elements.cpp)
-dissolve_unit_test(enumoptions.cpp)
+dissolve_unit_test(enumOptions.cpp)
 dissolve_unit_test(flags.cpp)
-dissolve_unit_test(genericlist.cpp)
+dissolve_unit_test(genericList.cpp)
 dissolve_unit_test(interactionPotential.cpp)
-dissolve_unit_test(neutronweights.cpp)
+dissolve_unit_test(neutronWeights.cpp)
 dissolve_unit_test(spacegroups.cpp)
```
jws-1, line 9: Should this also be renamed appropriately?

unit/gui/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
 set(TEST_GUI TRUE)
 
-dissolve_unit_test(datamanager.cpp)
+dissolve_unit_test(dataManager.cpp)
 dissolve_unit_test(forcefieldTab.cpp)
 dissolve_unit_test(isotopologeSetModel.cpp exchangeable)
```
jws-1, line 5: While we're here, should this file be named `isotopologueSetModel.cpp` instead?

jws-1 (commented):
Everything looks good. I think one file got missed, and we could rename another within this PR.

unit/classes/CMakeLists.txt
trisyoungs: Was halfway through writing why this should in fact remain as `spacegroups.cpp` and how long ago there was a PR to rename everything to be one word, but that is precisely the opposite to what actually happened!  Will change accordingly.

jws-1 (approved):

PR #1513, opened 2023-07-10
Removing advertisement from custom 404 page

rhinoella:
Removed the "You can learn how to make a 404 page like this in [Custom 404 Pages](https://gohugo.io/templates/404/)." promotion.

image


trisyoungs (approved):
:+1:

PR #1519, opened 2023-07-11
Messages Hyperlink

### LINKED ISSUE #1266: Link to Messages tab from status bar
When an error is encountered in the setup of Dissolve prior to running a simulation in the GUI, the status bar displays the number of errors and says to check the messages tab.  Having a hyperlink to go to the tab would be nice!

jws-1:
Small PR which adds a hyperlink to the `Messages` tab in the status label, when it suggests that it should be viewed. Closes #1266.

Currently you can right click on the hyperlink and `Copy link`, which could be deceiving to users - if necessary, we can install a simple `eventFilter` that ignores right clicking on the link.

### DIFF
--- a/src/gui/guiFuncs.cpp
+++ b/src/gui/guiFuncs.cpp
@@ -432,7 +432,6 @@ void DissolveWindow::clearMessages()
     Messenger::clearErrorCounts();
 }
 
-// Open the messages window
 void DissolveWindow::statusLabelLinkClicked(const QString &link)
 {
     if (DissolveSys::sameString(link.toStdString(), "Messages"))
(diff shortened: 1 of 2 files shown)


src/gui/guiFuncs.cpp
```diff
@@ -429,3 +431,10 @@ void DissolveWindow::clearMessages()
     ui_.MainTabs->messagesTab()->clearMessages();
     Messenger::clearErrorCounts();
 }
+
+// Open the messages window
```
trisyoungs, line 435: Probably bin this comment since the name of the slot should capture the gist of things - we don't tend to comment the individual slots in the code for that reason.
```suggestion
```

trisyoungs (approved):
Works nice.  I don't think I'd worry about the right-click for now. Let's let somebody complain first!

PR #1531, opened 2023-07-12
Empty Potential Functions

### LINKED ISSUE #1525: Empty Potential Functions
When adding a `Potential` node (e.g. `DirectionalGlobalPotential`, `RestraintPotential`, `SimpleGlobalPotential`) sensible default parameters are not shown in the associated line edit, leaving it empty. Therefore, if you add one of these nodes and then save straight away, no parameters are written out, which means errors when loading the same file back in.

jws-1:
Simply updates the `InteractionPotential` class to call `setForm` in it's constructor (which constructs the `parameters_` vector), rather than just assigning the form. Closes #1525.

### DIFF
--- a/src/classes/interactionPotential.h
+++ b/src/classes/interactionPotential.h
@@ -14,7 +14,7 @@
 template <class Functions> class InteractionPotential
 {
     public:
-    explicit InteractionPotential(typename Functions::Form form) { setForm(form); };
+    InteractionPotential(typename Functions::Form form) { setForm(form); };
     virtual ~InteractionPotential() = default;
     InteractionPotential(const InteractionPotential &source) = default;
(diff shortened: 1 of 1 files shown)


src/classes/interactionPotential.h
```diff
@@ -14,7 +14,7 @@
 template <class Functions> class InteractionPotential
 {
     public:
-    explicit InteractionPotential(typename Functions::Form form) : form_(form){};
+    InteractionPotential(typename Functions::Form form) { setForm(form); };
```
trisyoungs, line 17: Why does this no longer need to b marked `explicit`? It has a single non-default argument, so must be marked as such to expressly avoid unintentional implicit conversion.  At least that's what my CLion says!

trisyoungs (approved):
:+1:

PR #1532, opened 2023-07-12
1528 Input File Error Handling

### LINKED ISSUE #1528: Make input file loading less fussy
We currently error and bail out of reading our custom-format input file, which causes no end of drama for users when mistakes happen, or something doesn't get written correctly etc.

We should just carry on reading the input file as best as is possible, and just print error messages along the way.

trisyoungs:
This PR significantly "softens" the assumed severity of errors encountered while reading our (current) input file format. Wherever possible errors are simply reported, but parsing continues, and should greatly reduce the amount of bad UX when loading in a slightly older input file.

Closes #1528.

### DIFF
--- a/src/classes/species_io.cpp
+++ b/src/classes/species_io.cpp
@@ -539,10 +539,6 @@ bool Species::read(LineParser &parser, CoreData &coreData)
                 break;
         }
 
-        // Error encountered?
-        if (error)
-            break;
-
         // End of block?
         if (blockDone)
             break;
--- a/src/classes/species_io.cpp
+++ b/src/classes/species_io.cpp
@@ -96,11 +96,6 @@ bool Species::read(LineParser &parser, CoreData &coreData)
                 }
                 else
                     a = addAngle(parser.argi(1) - 1, parser.argi(2) - 1, parser.argi(3) - 1);
-                if (!a)
-                {
-                    error = true;
-                    break;
-                }
 
                 /*
                  * If only the indices were given, create an angle without a specified functional form (a
@@ -191,11 +186,6 @@ bool Species::read(LineParser &parser, CoreData &coreData)
                 }
                 else
                     b = addBond(parser.argi(1) - 1, parser.argi(2) - 1);
-                if (!b)
-                {
-                    error = true;
-                    break;
-                }
 
                 /*
                  * If only the indices were given, create a bond without a specified functional form (a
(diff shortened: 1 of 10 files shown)


jws-1 (changes requested):
I like where this is going, and it works nicely! However, I noticed that when parsing a `Procedure` if a keyword within a Node has an invalid number of parameters, the error propagates through the rest of the sequence. The rest of the sequence will be parsed correctly, but the error messages suggest otherwise.

image

trisyoungs:
Good spot.  Have modified files accordingly to prevent all these spurious errors!

src/procedure/nodes/sequence.cpp
```diff
@@ -511,10 +519,13 @@ bool ProcedureNodeSequence::deserialise(LineParser &parser, const CoreData &core
...
         if (!newNode->deserialise(parser, coreData))
-            return Messenger::error("Failed to read node sequence.\n");
+        {
+            Messenger::error("Failed to read node sequence.\n");
+            errorsEncountered = true;
+        }
     }
 
-    return true;
+    return !errorsEncountered;
```
jws-1, line 528:
```suggestion
    return (!errorsEncountered);
```
For consistency

src/procedure/nodes/node.cpp
```diff
@@ -296,10 +297,10 @@ bool ProcedureNode::deserialise(LineParser &parser, const CoreData &coreData)
         else if (result == KeywordBase::ParseResult::Deprecated)
             Messenger::warn("The '{}' keyword is deprecated and will be removed in a future version.\n", parser.argsv(0));
         else if (result == KeywordBase::ParseResult::Failed)
-            return Messenger::error("Failed to parse keyword '{}'.\n", parser.argsv(0));
+            errorsEncountered = true;
     }
 
-    return true;
+    return !errorsEncountered;
```
jws-1, line 303:
```suggestion
    return (!errorsEncountered);
```
Consistency

src/classes/species_io.cpp
```diff
@@ -73,10 +73,16 @@ bool Species::read(LineParser &parser, CoreData &coreData)
 
         // Do we recognise this keyword and, if so, do we have an appropriate number of arguments?
         if (!keywords().isValid(parser.argsv(0)))
-            return keywords().errorAndPrintValid(parser.argsv(0));
+        {
+            keywords().errorAndPrintValid(parser.argsv(0));
+            continue;
+        }
```
jws-1, line 79: Should we also set `errorsEncountered = true` here?

jws-1 (approved):
Looks good to me now. However, I wonder if we should also set `errorsEncountered` when an invalid keyword is given - for example see my comment on `species_io.cpp`

src/classes/species_io.cpp
trisyoungs: Yes, and in a bunch of other similar places.

PR #1533, opened 2023-07-12
Fix Expression Normalisations

### LINKED ISSUE #1521: (Angle) Overflow issue?
**Describe the bug**
The `Angle` module appears to have an overflow issue, with the normalised histogram sometimes appearing as:
image

trisyoungs:
This PR addresses a fairly critical bug affecting the `Angle`, `IntraAngle`, `DAngle`, and `AxisAngle` modules (see the theme here?) whereby the normalisation of the calculated data was wrong. This became broken in #1440 when the trigonometric functions implemented in `Expression` were modified to take an argument in radians rather than user-friendly degrees.

Closes #1521, which is not an overflow issue as suggested, but is related to this.


jws-1 (approved):
:+1:

PR #1536, opened 2023-07-14
Toml fixes

rprospero:
Fixes a couple of issues in the TOML serialisation

1) Replace the serialisation stub for Data3DStore with actual methods
2) Don't export null modules
3) Don't serialise deprecate keywords

We're still at 35 failing tests, but two tests now fail at runtime instead of crashing during parsing.

### DIFF
--- a/src/classes/data3DStore.cpp
+++ b/src/classes/data3DStore.cpp
@@ -44,7 +44,6 @@ OptionalReferenceWrapper<const Data3D> Data3DStore::data(std::string_view name) 
 // Return vector of all data
 const std::vector<std::shared_ptr<std::pair<Data3D, Data3DImportFileFormat>>> &Data3DStore::data() const { return data_; }
 
-
 // Express as a serialisable value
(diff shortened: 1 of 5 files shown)


src/classes/data3DStore.cpp
```diff
@@ -43,3 +43,27 @@ OptionalReferenceWrapper<const Data3D> Data3DStore::data(std::string_view name)
 
 // Return vector of all data
 const std::vector<std::shared_ptr<std::pair<Data3D, Data3DImportFileFormat>>> &Data3DStore::data() const { return data_; }
+
+// Express as a serialisable value
+SerialisedValue Data3DStore::serialise() const
+{
+
```
trisyoungs, line 50:
```suggestion
```

trisyoungs (approved):
:+1:
One rogue blank line, then all good.  In fact I had managed to get an hour to go through the TOML, and made exactly the same changes to `Data3D`-related stuff, but your PR beat me to the finish line!

PR #1537, opened 2023-07-17
Minor fixes

### LINKED ISSUE #1502: Mass Weighted Fragment Site Origin
Currently there is no option in the GUI to enable/disable determining the origin of a `FragmentSite` by mass weighting. Therefore, the origin currently always defaults to the centre of geometry.

### LINKED ISSUE #1526: Configuration information not updated on creation
When creating a new configuration, "Configuration Information" is defaulted and not updated according to the created configuration.

### LINKED ISSUE #1535: ModuleVector keyword allows same module to be added multiple times
**Describe the bug**
The `ModuleVectorKeyword` allows the same `Module` target to be added multiple times to the vector. This is wrong! We should warn about the duplicate, but not add it.

jws-1:
This PR fixes a few small issues.
- Made the "Weight origin by mass" option available to all site types (even though it does nothing for `Dynamic` sites, which we could disable it for). Closes #1502.
- Update the Configuration "status bar" on creation. Closes #1526.
- Prevent the same `Module` from being added to a `ModuleVector` more than once. Closes #1535.

### DIFF
--- a/src/keywords/moduleVector.cpp
+++ b/src/keywords/moduleVector.cpp
@@ -59,10 +59,8 @@ bool ModuleVectorKeyword::deserialise(LineParser &parser, int startArg, const Co
                                     parser.argsv(n), ModuleTypes::moduleType(module->type()), name(),
                                     joinStrings(moduleTypes_));
         if (!data_.empty() && std::find(data_.cbegin(), data_.cend(), module) != data_.cend())
-        {
-            Messenger::warn("Module '{}' has already been added.\n", parser.argsv(n));
-            continue;
-        }
+            Messenger::warn("Module '{}' has already been added to keyword '{}'.\n", parser.argsv(n), name());
+        else
         data_.emplace_back(module);
     }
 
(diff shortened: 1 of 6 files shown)


src/keywords/moduleVector.cpp
```diff
@@ -58,7 +58,11 @@ bool ModuleVectorKeyword::deserialise(LineParser &parser, int startArg, const Co
             return Messenger::error("Module '{}' is of type '{}', and is not relevant to keyword '{}' (allowed types = {}).\n",
                                     parser.argsv(n), ModuleTypes::moduleType(module->type()), name(),
                                     joinStrings(moduleTypes_));
-
+        if (!data_.empty() && std::find(data_.cbegin(), data_.cend(), module) != data_.cend())
+        {
+            Messenger::warn("Module '{}' has already been added.\n", parser.argsv(n));
+            continue;
+        }
```
trisyoungs, line 65: Just a suggestion!
```suggestion
            Messenger::warn("Module '{}' has already been added to keyword '{}'.\n", parser.argsv(n), name());
        else
```

trisyoungs (approved):
:+1:
Just one small suggestion.

PR #1539, opened 2023-07-18
917 fixing module timing

rhinoella:
## Changes made:
- Module `process()` functions return an enum instead of a bool
    - Enum has three types: `failed`, `success` and `notExecuted`
    - `notExecuted` is returned when the process legitimately fails (e.g process is not needed)
- Process time is only added to if `success` is returned
- Fixed small issue: when Simulation > Step is used, after running the step the ETA is set to `??:??:??`, now it goes back to `--:--:--`

### DIFF
--- a/src/main/simulation.cpp
+++ b/src/main/simulation.cpp
@@ -304,7 +304,7 @@ bool Dissolve::iterate(int nIterations)
 
                 Messenger::heading("{} ({})", ModuleTypes::moduleType(module->type()), module->name());
 
-                if (!module->executeProcessing(*this, worldPool()))
+                if (module->executeProcessing(*this, worldPool()) == failed)
                     return Messenger::error("Module '{}' experienced problems. Exiting now.\n", module->name());
             }
         }
--- a/src/main/simulation.cpp
+++ b/src/main/simulation.cpp
@@ -304,9 +304,9 @@ bool Dissolve::iterate(int nIterations)
 
                 Messenger::heading("{} ({})", ModuleTypes::moduleType(module->type()), module->name());
 
-                if (module->executeProcessing(*this, worldPool()))
-                    if(module->executeProcessing(*this, worldPool()) == 2) Messenger::print("aaa");
+                if (!module->executeProcessing(*this, worldPool()))
                     return Messenger::error("Module '{}' experienced problems. Exiting now.\n", module->name());
+
             }
         }
 
--- a/src/main/simulation.cpp
+++ b/src/main/simulation.cpp
@@ -304,7 +304,7 @@ bool Dissolve::iterate(int nIterations)
 
                 Messenger::heading("{} ({})", ModuleTypes::moduleType(module->type()), module->name());
 
-                if (!module->executeProcessing(*this, worldPool()))
+                if (module->executeProcessing(*this, worldPool()) == ExecutionResult::Failed)
                     return Messenger::error("Module '{}' experienced problems. Exiting now.\n", module->name());
             }
         }
--- a/src/main/simulation.cpp
+++ b/src/main/simulation.cpp
@@ -305,7 +305,7 @@ bool Dissolve::iterate(int nIterations)
                 Messenger::heading("{} ({})", ModuleTypes::moduleType(module->type()), module->name());
 
                 if (module->executeProcessing(*this, worldPool()))
-                    Messenger::print(module->executeProcessing(*this, worldPool()));
+                    if(module->executeProcessing(*this, worldPool()) == 2) Messenger::print("aaa");
                     return Messenger::error("Module '{}' experienced problems. Exiting now.\n", module->name());
             }
         }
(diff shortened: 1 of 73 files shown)


src/module/module.cpp
```diff
@@ -151,7 +151,7 @@ bool Module::isDisabled() const { return !enabled_; }
  */
 
 // Run main processing
-bool Module::process(Dissolve &dissolve, const ProcessPool &procPool) { return false; }
+enum Module::executionResult Module::process(Dissolve &dissolve, const ProcessPool &procPool) { return failed; }
```
trisyoungs, line 154: You don't / shouldn't have to specify `enum Module::executionResult` here, just `Module::executionResult` - in fact, doing so adds a bit of clutter to the code.  Thankfully, a simple search/replace will get you there!

src/module/module.h
```diff
@@ -107,6 +107,13 @@ class Module : public Serialisable<const CoreData &>
     int frequency_;
     // Whether the Module is enabled
     bool enabled_;
+    // Process exit information
+    enum executionResult
+    {
+        failed,
+        success,
+        notExecuted
+    };
```
trisyoungs, line 116: Got my picky hat on today!  Our enum-naming convention is to capitalise the first letter as well. Also we can make this an `enum class` which enforces scope and type better.
```suggestion
    // Module execution result
    enum class ExecutionResult
    {
        Failed,
        Success,
        NotExecuted
    };
```

src/modules/accumulate/process.cpp
```diff
@@ -22,11 +22,14 @@ EnumOptions<AccumulateModule::TargetPartialSet> AccumulateModule::targetPartialS
...
 // Run main processing
-bool AccumulateModule::process(Dissolve &dissolve, const ProcessPool &procPool)
+enum Module::executionResult AccumulateModule::process(Dissolve &dissolve, const ProcessPool &procPool)
 {
     // Get the modules and decide on the PartialSet data name we're looking for
     if (targetModules_.empty())
-        return Messenger::error("No target modules set.");
+    {
+        Messenger::error("No target modules set.");
+        return failed;
```
trisyoungs, line 31: I don't recall whether using `enum class` will enforce this, but to me this makes the intent clearer:
```suggestion
        return ExecutionResult::Failed;
```

trisyoungs (changes requested):
A little bit of renaming and I think we're good!

src/main/simulation.cpp
```diff
@@ -304,7 +304,7 @@ bool Dissolve::iterate(int nIterations)
 
                 Messenger::heading("{} ({})", ModuleTypes::moduleType(module->type()), module->name());
 
-                if (!module->executeProcessing(*this, worldPool()))
+                if (static_cast<int>(module->executeProcessing(*this, worldPool())) == 0)
```
trisyoungs, line 307: Ick!  Let's not do this as it really hides intent, and kinda undoes all the niceness of `enum class`.  All we need to do is make the enum `public` in the `Module` class.
```suggestion
                if (module->executeProcessing(*this, worldPool()) == Module::ExecutionResult::Failed)
```

src/module/module.cpp
```diff
@@ -191,7 +191,16 @@ bool Module::executeProcessing(Dissolve &dissolve, const ProcessPool &procPool)
...
+
+    if (result == ExecutionResult::Success)
+    {
+        processTimes_ += timer.secondsElapsed();
+    }
+    else
+    {
+        // If process is not run, do not add to the timing
+        processTimes_ += 0;
+    }
```
trisyoungs, line 203: Do we actually want to accumulate a zero here?  This will add to the statistics and effectively reduce the average time it takes for the module to run.  If we don't add anything then the `processTimes_` will reflect some sort of upper maximum time for the module.

trisyoungs, line 203:
```suggestion
        processTimes_ += timer.secondsElapsed();
```

src/module/module.h
```diff
@@ -107,6 +107,13 @@ class Module : public Serialisable<const CoreData &>
...
+    // Process exit information
+    enum class ExecutionResult
+    {
+        Failed,
+        Success,
+        NotExecuted
+    };
 
     public:
     // Set frequency at which to run Module (relative to layer execution count)
```
trisyoungs, line 119:
```suggestion

    public:
    // Process exit information
    enum class ExecutionResult
    {
        Failed,
        Success,
        NotExecuted
    };
    // Set frequency at which to run Module (relative to layer execution count)
```

trisyoungs (changes requested):
Looks a lot nicer - just need to tidy away that nasty `static_cast`!

src/module/module.cpp
rhinoella: This was done because if processTimes_ was left empty, the ETA would flash on and off as it would encounter module processes returning `NotExecuted`, but now I will tackle that problem in the GUI code rather than the module timing!

trisyoungs (approved):
:+1:

Issue #276, opened 2020-05-04
CIF Import


trisyoungs:
**Describe the feature / issue**
Allow the import of crystallographic data files, e.g. cifs, in the GUI to permit construction of suitable configurations for subsequent simulation.

The process should allow individual species types within the crystal to be correctly recognised as copies of the same underlying molecule, and replicated accordingly.

PR #1549, opened 2023-07-20
Molecular CIF Import

jws-1:
This PR implements the ability to import CIFs that have a molecular structure. Distinct species are extracted and created, as well as a Configuration that represents the initial Crystal structure.

This PR works towards #276.


src/gui/importCIFDialogFuncs.cpp
```diff
@@ -509,6 +547,151 @@ void ImportCIFDialog::on_MoietyNETARemoveFragmentsCheck_clicked(bool checked)
...
+        neta.flags() += NETADefinition::NETAFlags::IncludeRootElement;
+
+        // Find a sp definition, if one exists.
+        auto nDistinctMatches = 0;
+        idx = 0;
+        while (nDistinctMatches != 1)
+        {
+            if (idx >= tempSpecies->nAtoms())
+                return false;
+            neta.create(&tempSpecies->atom(idx++), 128);
```
trisyoungs, line 607: This definitely feels like the maximum path length argument to `create()` should be something like a `std::optional<int>`, still defaulting to `1`, but which we can pass `std::nullopt` to here in order to specify no max path length.  That'll need modifications elsewhere, of course!

src/gui/importCIFDialogFuncs.cpp
```diff
@@ -509,6 +547,151 @@ void ImportCIFDialog::on_MoietyNETARemoveFragmentsCheck_clicked(bool checked)
...
+                                     [&](int value)
+                                     { return std::find(fragment.begin(), fragment.end(), value) != fragment.end(); }),
+                      indices.end());
+
+        // Choose starting index of the next fragment
+        idx = *std::min_element(indices.begin(), indices.end());
+    }
+
+    // Construct sp NETA definitions from the fragments
+    // By sp, we mean that the definition produces a single match,
```
trisyoungs, line 580: Not sure I'm getting the "sp" lingo here!

src/neta/neta.h
```diff
@@ -19,6 +20,13 @@ class NETADefinition
     NETADefinition(const SpeciesAtom *i, int maxDepth = 1);
     ~NETADefinition() = default;
 
+    public:
+    enum NETAFlags
```
trisyoungs, line 24: Bit more of an explicit name, since they are only relevant to creation.
```suggestion
    // NETA Definition Creation Flags
    enum NETACreationFlags
```

src/neta/neta.h
```diff
@@ -19,6 +20,13 @@ class NETADefinition
...
+    {
+        MatchHydrogens,
```
trisyoungs, line 26: Perhaps `ExplicitHydrogens` instead?

src/neta/neta.h
```diff
@@ -51,6 +61,9 @@ class NETADefinition
     void addIdentifier(std::string identifier);
     // Return identifiers from definition
     const std::set<std::string> &identifiers() const;
+    // Return flags
+    const Flags<NETAFlags> &flags() const;
+    Flags<NETAFlags> &flags();
```
trisyoungs, line 66: Since these are specific to creating NETA definitions I think we don't need accessors like this - a better alternative might be to pass them to the `create()` function as an optional argument?

src/neta/neta.h
```diff
@@ -31,6 +39,8 @@ class NETADefinition
     bool valid_;
     // Identifiers (#)
     std::set<std::string> identifiers_;
+    // Flags
+    Flags<NETAFlags> flags_;
```
trisyoungs, line 43: If you pass flags in to `create()` as suggest below, this can be removed and will boil down to a straight variable pass-through to `netaString()` in `NETADefinition::create()`.

src/neta/neta.cpp
```diff
@@ -95,12 +95,15 @@ bool NETADefinition::create(std::string_view definition, const Forcefield *assoc
 }
 
 // Recursively create a NETA string for the specified atom
-std::string netaString(const SpeciesAtom *i, int currentDepth, const int maxDepth, std::vector<const SpeciesAtom *> &path)
+std::string netaString(const SpeciesAtom *i, int currentDepth, const int maxDepth, std::vector<const SpeciesAtom *> &path,
+                       Flags<NETADefinition::NETAFlags> flags = Flags<NETADefinition::NETAFlags>())
```
trisyoungs, line 99:
```suggestion
                       const Flags<NETADefinition::NETAFlags> &flags = {})
```

src/neta/hydrogenCount.cpp
```diff
@@ -28,12 +29,23 @@ void NETAHydrogenCountNode::set(ComparisonOperator op, int value)
 // Evaluate the node and return its score
 int NETAHydrogenCountNode::score(const SpeciesAtom *i, NETAMatchedGroup &matchPath) const
 {
+    auto newMatchPath = matchPath;
     if (!value_)
+    {
```
trisyoungs, line 34: Personally, I'm not a fan of scoping braces when there's a single line of code, since it turns it into three lines of code!
```suggestion
```

src/neta/hydrogenCount.cpp
```diff
@@ -28,12 +29,23 @@ void NETAHydrogenCountNode::set(ComparisonOperator op, int value)
...
         return NETANode::NoMatch;
-
-    // Count number of hydrogens attached to this atom
-    auto nH = std::count_if(i->bonds().begin(), i->bonds().end(),
-                            [i](const SpeciesBond &bond) { return bond.partner(i)->Z() == Elements::H; });
-
+    }
```
trisyoungs, line 36:
```suggestion
```

src/gui/importCIFDialogFuncs.cpp
jws-1: Oops.... replace all!

src/neta/neta.h
jws-1: The (now) `ExplicitHydrogens` flag is relevant to matching and the `IncludeRootElement` flag is relevant to creation, so I'll create two separate enums.

src/gui/importCIFDialogFuncs.cpp
```diff
@@ -184,27 +192,72 @@ bool ImportCIFDialog::prepareForPreviousPage(int currentIndex)
...
+            auto base = sp->name();
+            std::string uniqueSuffix{base};
+            if (!generator.nodes().empty())
+            {
+                // Start from the last root node
+                auto root = generator.nodes().back();
+                auto suffix = 0;
+
+                // We use 'CoordinateSets' here, because in this instance we are working with (CoordinateSet, Add) pairs
+                while (generator.rootSequence().nodeInScope(root, fmt::format("CoordinateSets_{}", uniqueSuffix)) != nullptr)
```
trisyoungs: Naming-wise, perhaps instead of calling the nodes `CoordinateSets_x` we should call them `SymmetryCopies_x` to make things nice and clear?

src/io/import/cifClasses.h
```diff
@@ -111,3 +115,56 @@ class CIFAssembly
...
+    /*
+     * Information
+     */
+    public:
+    // Return the output species
+    const Species *species() const;
+    // Return the NETA definition string that uniquely describes the reference instance
+    const std::string netaString() const;
+    // Return the reference instance
+    const std::vector<int> referenceInstance() const;
```
trisyoungs:
```suggestion
    const std::vector<int> &referenceInstance() const;
```

src/io/import/cifClasses.h
```diff
@@ -111,3 +115,56 @@ class CIFAssembly
...
+    // Return all found instances
+    const std::vector<std::vector<int>> instances() const;
```
trisyoungs:
```suggestion
    const std::vector<std::vector<int>> &instances() const;
```

src/io/import/cifClasses.h
```diff
@@ -111,3 +115,56 @@ class CIFAssembly
...
+    // Return the coordinates corresponding to the instances
+    const std::vector<std::vector<Vec3<double>>> coordinates() const;
```
trisyoungs:
```suggestion
    const std::vector<std::vector<Vec3<double>>> &coordinates() const;
```

src/gui/importCIFDialogFuncs.cpp
```diff
@@ -184,27 +192,72 @@ bool ImportCIFDialog::prepareForPreviousPage(int currentIndex)
...
-        // Remove the unit cell and any cell-crossing bonds
-        sp->removePeriodicBonds();
-        sp->updateIntramolecularTerms();
-        sp->removeBox();
+        // Create a Configuration
+        auto *cfg = dissolve_.addConfiguration();
+        cfg->setName(cifImporter_.chemicalFormula());
+        auto cellLengths = cifImporter_.getCellLengths();
+        auto cellAngles = cifImporter_.getCellAngles();
+        cfg->createBoxAndCells(cellLengths.value(), cellAngles.value(), false, 1.0);
```
trisyoungs: This needs to go in as an explicit `Box` node in the generator.

trisyoungs (commented):
Looking very nice indeed. Some minor comments in the review.

I also don't get a properly unfolded Species:
image

...and the generated configuration shows broken bonds, but this is an issue with Cells rather than your code.  Once you run a proper simulation (which I have done :smile:) everything snaps back into place.

src/gui/importCIFDialogFuncs.cpp
```diff
@@ -509,6 +569,48 @@ void ImportCIFDialog::on_MoietyNETARemoveFragmentsCheck_clicked(bool checked)
...
+    // Clear any existing species
+    cifSpecies_.clear();
+
+    std::vector<int> indices(cleanedSpecies_->nAtoms());
+    std::iota(indices.begin(), indices.end(), 0);
+
+    auto idx = 0;
+
+    // Find all CIFSpecies, and their instances
+    while (!indices.empty())
```
trisyoungs: Let's move this declaration tighter to the `while` loop so its intended place of use is a little clearer.
```suggestion
    // Find all CIFSpecies, and their instances
    auto idx = 0;
    while (!indices.empty())
```

src/io/import/cifClasses.h
```diff
@@ -111,3 +115,56 @@ class CIFAssembly
...
+    // Found copies/instances
+    std::vector<std::vector<int>> instances_;
+    // Coordinates corresponding to the instances
+    std::vector<std::vector<Vec3<double>>> coordinates_;
+    // Does the reference instance contain symmetry?
+    bool hasSymmetry_{false};
+
+    /*
+     * Information
+     */
```
trisyoungs:
```suggestion
```

src/io/import/cifClasses.h
```diff
@@ -111,3 +115,56 @@ class CIFAssembly
...
+
+// CIF Species
+class CIFSpecies
+{
+
+    public:
+    CIFSpecies(Species *spRef, Species *sp, std::vector<int> referenceInstance);
+    ~CIFSpecies() = default;
+
+    private:
```
trisyoungs, line 130:
```suggestion
    /*
     * Information
     */
    private:
```

trisyoungs (approved):
Two very minor suggestions, but I think this is ready to go in.

As it stands it still generates the original unit cell even when a supercell is requested, but this is one to tackle in the Epic (I'll add that now, along with the unit testing if it isn't already there).

PR #1555, opened 2023-07-22
1554 Dissolve CoreData

### LINKED ISSUE #1554: Clean up interaction between main classes
The main `Dissolve` class uses a `CoreData` object to store all valuable simulation data - having this level of detachment allows a lot of flexibility when it comes to many workflows in the code, since a temporary `CoreData` object can be created in order to test out / create / fix / modify simulation objects independently of the currently defined user data.

However, the `Dissolve` class currently provides a lot of pass-through functions offering access to its `CoreData` object (which is already accessible) and which cause clutter without offering anything truly useful.  A cleanup on this should be attempted in order to remove unnecessary accessors etc.

trisyoungs:
This PR removes a bunch of clutter from the main `Dissolve` class, cleansing it of the pass-through functions which give other objects access to its `CoreData`. That object has an accessor anyway, so it didn't save much!

Closes #1554.


src/main/keywords_pairPotentials.cpp
```diff
@@ -31,6 +31,7 @@ bool PairPotentialsBlock::parse(LineParser &parser, Dissolve *dissolve)
     std::shared_ptr<AtomType> at1;
     auto blockDone = false, errorsEncountered = false;
     Elements::Element Z;
+    auto &coreData = dissolve->coreData();
```
jws-1, line 34: Do we need this?

src/main/keywords_pairPotentials.cpp
```diff
@@ -94,13 +95,13 @@ bool PairPotentialsBlock::parse(LineParser &parser, Dissolve *dissolve)
                 }
 
                 // Find / create AtomType and check element...
-                at1 = dissolve->findAtomType(parser.argsv(1));
+                at1 = coreData.findAtomType(parser.argsv(1));
```
jws-1, line 98:
```suggestion
                at1 = dissolve->coreData().findAtomType(parser.argsv(1));
```
See above comment.

src/main/keywords_pairPotentials.cpp
```diff
@@ -94,13 +95,13 @@ bool PairPotentialsBlock::parse(LineParser &parser, Dissolve *dissolve)
...
                 if (!at1)
                 {
                     Messenger::warn("Unknown atom type '{}' referenced in PairPotentials block - creating "
                                     "it now...\n",
                                     parser.argsv(1));
-                    at1 = dissolve->addAtomType(Z);
+                    at1 = coreData.addAtomType(Z);
```
jws-1, line 104:
```suggestion
                    at1 = dissolve->coreData().addAtomType(Z);
```
See above.

jws-1 (approved):
Looks great to me! Just one minor comment.

src/main/keywords_pairPotentials.cpp
trisyoungs: Well, technically "No", but I tend to get into the habit of grabbing oft-used references in order to prevent long chains of `.` and `->`. In this case it isn't really a big issue, but elsewhere it can promote a lot of code readability IMHO!

jws-1 (approved):
:+1:

src/classes/coreData.cpp
```diff
@@ -327,6 +330,145 @@ Species *CoreData::findSpecies(std::string_view name) const
     }
 }
 
+// Copy AtomType, creating a new one if necessary
+void CoreData::copyAtomType(const SpeciesAtom *sourceAtom, SpeciesAtom *destAtom)
```
rprospero: Could the parameters be references, instead of pointers?  As near as I can tell, this is only ever called in the AddForcefieldDialog, where we just wind up casting the values to pointers before passing them to the function.

Still just a little paranoid about raw pointers.

trisyoungs: @rprospero Indeed they can.

rprospero (approved):

PR #1556, opened 2023-07-22
1543 Keyword Section Header Styling

trisyoungs:
This PR slightly adjusts the styling of our keyword widgets, making the header easier to read, and indenting keyword names to differentiate between sections more clearly by putting emphasis on the section headers.

image

### DIFF
--- a/src/gui/keywordWidgets/widgetFuncs.cpp
+++ b/src/gui/keywordWidgets/widgetFuncs.cpp
@@ -44,6 +44,7 @@ void KeywordsWidget::setUp(KeywordStore::KeywordStoreIndexInfo keywordIndexInfo,
         if (!sectionName.empty())
         {
             auto *sectionLabel = new SectionHeaderWidget(QString::fromStdString(std::string(sectionName)));
+            sectionLabel->setContentsMargins(0, 15, 0, 0);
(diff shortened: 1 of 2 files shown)


src/gui/keywordWidgets/widgetFuncs.cpp
rhinoella, line 1: `sectionLabel->setContentsMargins(0, 15, 0, 0);`

Increase margins in between sections:

image

trisyoungs: @rhinoella See, this works and looks really nice, so we should do it, but because I'm particularly fussy I'd want to avoid adding it to the first section header...

rhinoella: Changing it to:
`if (row != 0) sectionLabel->setContentsMargins(0, 15, 0, 0);`
Fixes that!

trisyoungs: > Changing it to: `if (row != 0) sectionLabel->setContentsMargins(0, 15, 0, 0);` Fixes that!

Yep!  Push it!

rhinoella (approved):

PR #1557, opened 2023-07-22
TOML Throws

trisyoungs:
Just a small PR to capture a minor modification I made while testing the TOML and forgot about!

### DIFF
--- a/src/base/geometry.h
+++ b/src/base/geometry.h
@@ -56,8 +56,8 @@ template <> struct from<Geometry::GeometryType>
         else if (typeString == "torsion")
             return Geometry::GeometryType::TorsionType;
         else
-            throw(std::runtime_error(
-                fmt::format("Unhandled geometry type '{}' - can't convert from TOML value.\n", std::string(typeString))));
+            throw toml::syntax_error(
+                fmt::format("Unhandled geometry type '{}' - can't convert from TOML value.\n", std::string(typeString)), node.location());
     }
 };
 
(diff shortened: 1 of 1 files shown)


src/base/geometry.h
```diff
@@ -52,8 +53,11 @@ template <> struct from<Geometry::GeometryType>
             return Geometry::GeometryType::AngleType;
         else if (typeString == "distance")
             return Geometry::GeometryType::DistanceType;
-        else
+        else if (typeString == "torsion")
             return Geometry::GeometryType::TorsionType;
+        else
+            throw(std::runtime_error(
+                fmt::format("Unhandled geometry type '{}' - can't convert from TOML value.\n", std::string(typeString))));
```
rprospero:
```suggestion
            throw toml::syntax_error(
                fmt::format("Unhandled geometry type '{}' - can't convert from TOML value.\n", std::string(typeString)), node.location());
```

Using the `toml::syntax_error` exception means that the error message will include the line number where the problem occurs.  Makes life much easier for the user debugging the input file.

src/base/geometry.h
```diff
@@ -69,6 +73,8 @@ template <> struct into<Geometry::GeometryType>
                 return "distance";
             case Geometry::GeometryType::TorsionType:
                 return "torsion";
+            default:
+                throw(std::runtime_error("Unhandled geometry type - can't convert to TOML value.\n"));
```
rprospero:
```suggestion
                throw toml::syntax_error("Unhandled geometry type - can't convert to TOML value.\n", node.location());
```

Same

src/base/geometry.h
trisyoungs: This is something I now realise from working on the (de)serialise for my new keyword type!

rprospero (approved):

PR #1561, opened 2023-07-22
Displaying elapsed time when running

### LINKED ISSUE #1262: Time elapsed instead of time remaining
When running for a certain number of steps we have a label showing an estimate of the remaining time.  When "just running" (i.e. with no specific number of steps) - we could display the time elapsed.

rhinoella:
### Elapsed time is now shown when running
image
### Main Changes:
- `iterationTimer_` (timer for the main loop) used to start/stop inside the `Dissolve::iterate()` function which is also iterated itself, meaning it was measuring the time for each iteration and not the total runtime
  - The starting/stopping of `iterationTimer_` is moved to `DissolveThreadWorker::beginIterating()` which calls `iterate()` in a loop
  - Start/stop timer is called through signals/slots (not sure if this is the ideal way to do this, but I could not think of another way)

Closes #1262


rhinoella:
Also please let me know which format for displaying the time is preferred: "x hours, x minutes and x seconds" or "xx : xx : xx"

src/base/timer.cpp
```diff
@@ -93,13 +94,17 @@ std::string Timer::timeString(double seconds)
...
     else if (minutes != 0)
-        return fmt::format("{} minutes and {:0.1f} seconds", minutes, seconds);
+        return fmt::format("{} minute{} and {:0.1f} seconds", minutes, plural(minutes), seconds);
     else
         return fmt::format("{:0.1f} seconds", seconds);
 }
 
+// Returns 's' if value is greater than 1
+std::string Timer::plural(int value) { return (value == 1 ? "" : "s"); }
+
```
trisyoungs: This is a nice little function, but is so general that it should be available everywhere.  I suggest moving it to the `DissolveSys` namespace in `src/base/sysFunc.h`.

src/base/timer.cpp
```diff
@@ -18,24 +18,25 @@ Timer::Timer(bool immediateStart)
...
 {
     int n = duration.count();
-    auto hours = n / 3600;
+    int hours = n / 3600;
     n %= 3600;
-    auto minutes = n / 60;
+    int minutes = n / 60;
     n %= 60;
-    double seconds = duration.count() - hours * 3600 - minutes * 60;
+    int seconds = duration.count() - hours * 3600 - minutes * 60;
```
trisyoungs: Curious to know why you removed the `auto` and went to `int`?  Given two `int` arguments, the `/` operator returns an `int` and so `auto` is indeed valid here.

trisyoungs (changes requested):
OK, so first off I think I prefer the X hours, Y minutes and Z seconds format, so let's stick with that. Second, I think it would be less confusing, and more easily implemented, to not try and use the current `Dissolve::iterationTimer_` and implement a new `elapsedTimer_` in the `DissolveWindow` class itself.  That way we don't have to worry about signals/slots for communication (which was the correct way, BTW).

rhinoella:
I've made the requested changes and also kept `iterationTimer_` in the `printTiming()` function because I realised that `iterationTimer` reports the total elapsed time for people using Dissolve in the command-line, as in that case it does not loop over `iterate()`. 

### Changes:

- Moved `elapsedTimer_` to DissolveWindow class
- `plural()` has been moved from Timer class to DissolveSys
- Edited some of the Timer functions to work with `elapsedTimer` (should not affect other timers)
- Changed ETA when using 'Run For ...' to be in the format "x minutes and x seconds" instead of "--:--:--" for consistency

src/base/sysFunc.cpp
```diff
@@ -233,6 +233,12 @@ std::string DissolveSys::niceName(std::string_view original)
     return s;
 }
 
+// Returns 's' if value is greater than 1
+std::string DissolveSys::plural(int value) { return (value == 1 ? "" : "s"); }
+
+// Returns 's' if value is greater than 1
+std::string DissolveSys::plural(double value) { return (0.5 < value && value < 1.5 ? "" : "s"); }
+
```
trisyoungs: OK so this is interesting, but for a very annoying reason!  I want to say that the `double` overload is unnecessary, or at least a bit redundant, since when expressing time in string form as we are here, seconds is only singular when its exactly `1`.  Otherwise we're writing things like`1 minute and 0.5 seconds`, otherwise we'd have to write `1 minute and 0.5 of a second` which is just nasty.  I think I would suggest keeping just the `int` version, and making the string output always `seconds`.

src/base/timer.cpp
```diff
@@ -84,29 +73,43 @@ double Timer::secondsElapsed() const
...
+                           DissolveSys::plural(minutes), seconds, DissolveSys::plural(seconds));
     else if (minutes != 0)
-        return fmt::format("{} minutes and {:0.1f} seconds", minutes, seconds);
+        return fmt::format("{} minute{} and {:0.0f} second{}", minutes, DissolveSys::plural(minutes), seconds,
+                           DissolveSys::plural(seconds));
+    else if (truncate)
+        return fmt::format("{:0.0f} second{}", seconds, DissolveSys::plural(seconds));
     else
-        return fmt::format("{:0.1f} seconds", seconds);
+        return fmt::format("{:0.1f} second{}", seconds, DissolveSys::plural(seconds));
```
trisyoungs: Suggesting to keep this as `seconds` for all values.

src/base/timer.cpp
```diff
@@ -84,29 +73,43 @@ double Timer::secondsElapsed() const
         return totalTime_.count();
 }
 
-// Return time string for number of seconds provided
-std::string Timer::timeString(double seconds)
+// Return time string based on provided duration in seconds
+std::string Timer::timeString(std::chrono::duration<double> duration, bool truncate)
```
trisyoungs: There's a lot of code duplication between this function and `timeString(double seconds)` - I would suggest moving the code here (which is more flexible because of the `truncate` argument) and moving it to the other function, and just calling it from here as you do with `elapsedTimeString`.

src/gui/gui.h
```diff
@@ -105,6 +108,7 @@ class DissolveWindow : public QMainWindow
     void updateWhileRunning(int iterationsRemaining);
     // Clear the messages window
     void clearMessages();
+    // Start run timer
```
trisyoungs:
```suggestion
```

src/gui/guiFuncs.cpp
```diff
@@ -48,9 +48,9 @@ DissolveWindow::DissolveWindow(Dissolve &dissolve)
     addStatusBarIcon(":/control/icons/control_step.svg")->setToolTip("Current step / iteration number");
     iterationLabel_ = addStatusBarLabel("00000");
     iterationLabel_->setToolTip("Current step / iteration number");
-    addStatusBarIcon(":/general/icons/general_clock.svg")->setToolTip("Time remaining to completion");
+    addStatusBarIcon(":/general/icons/general_clock.svg")->setToolTip("Timer");
```
trisyoungs: "Timer" is not a particularly useful tooltip, and given the text label accompanying the time string values I think we probably don't need one here or on the `etaLabel_` any more.

src/gui/guiFuncs.cpp
```diff
@@ -48,9 +48,9 @@ DissolveWindow::DissolveWindow(Dissolve &dissolve)
...
     etaLabel_ = addStatusBarLabel("--:--:--");
```
trisyoungs: This placeholder (HH:MM:SS) doesn't represent our format any more.

src/base/timer.h
```diff
@@ -21,12 +21,12 @@ class Timer
     std::chrono::duration<double> totalTime_;
     // Whether the timer is running or not
     bool running_{false};
-
-    private:
-    // Return time string based on provided duration in seconds
-    std::string timeString(std::chrono::duration<double> duration);
+    // Whether the timer has started before
```
trisyoungs: Is this meant to mean "Whether the timer is currently running"?

src/gui/gui.h
```diff
@@ -69,10 +70,12 @@ class DissolveWindow : public QMainWindow
     QLabel *restartFileIndicator_;
     // Label for iteration number
     QLabel *iterationLabel_;
-    // Label for simulation ETA (when using RunFor)
+    // Label for simulation timer (ETA if RunFor, time elapsed if Run)
     QLabel *etaLabel_;
```
trisyoungs: I think this warrants a name change, given the new purpose of the label.
```suggestion
    QLabel *timerLabel_;
```

src/gui/guiFuncs.cpp
```diff
@@ -237,6 +237,10 @@ bool DissolveWindow::loadInputFile(std::string_view inputFile, bool handleRestar
         fullUpdate();
     }
 
+    // Empty timer text
+    elapsedTimer_.zero();
+    etaLabel_->setText("------------");
```
trisyoungs: Instead of dashes, perhaps we mirror the main status indicator and say "Idle" or "Not running"? Regardless, since we set the text of the `etaLabel_` to some common string in at least three places it's probably worth adding that string as a `static constexpr` member in the main gui header, and using that instead.  Then if we want to change it there's a single source.

trisyoungs (changes requested):
Looking good, but have made some suggestions for improvement / simplification.

src/base/timer.h
rhinoella: It is whether or not the timer has been started before , it is essentially there because the elapsed timer text is set in the `updateStatusBar()` function and it needs to know whether or not the `elapsedTimer_` is running (`started_ == true`), and that means it should be displaying time elapsed. As we are calling this in the `DissolveWindow` class, it doesn't know whether or not `dissolveIterating_` is true due to the user hitting 'Run' or 'Run For' (would instead need to display ETA).

rhinoella:
### Changes
- Removed overloaded `plural()` function
- Merged the two `timeString()` functions together
- Made default timer label text a `constexpr`

src/gui/gui.h
```diff
@@ -69,10 +70,14 @@ class DissolveWindow : public QMainWindow
     QLabel *restartFileIndicator_;
     // Label for iteration number
     QLabel *iterationLabel_;
-    // Label for simulation ETA (when using RunFor)
-    QLabel *etaLabel_;
+    // Label for simulation timer (ETA if RunFor, time elapsed if Run)
+    QLabel *timerLabel_;
+    // Timer label default text
+    static constexpr const char *timerText_ = "Idle";
```
trisyoungs: I appreciate that this is the only reasonable way of defining our default text in the header, but it's a bit "archaic C", so I'd suggest we move it to the `guiFuncs.cpp` file since that's the only place its used.
```suggestion
```

In between the `#include`s and the constructor:
```
// Default timer text
const QString defaultTimerText{"Idle"};
```

src/gui/gui.h
```diff
@@ -69,10 +70,14 @@ class DissolveWindow : public QMainWindow
...
     // General status indicator and label
     QLabel *statusIndicator_, *statusLabel_;
+    // Main loop iteration timer
+    Timer elapsedTimer_ = Timer(false);
```
trisyoungs: Prefer brace initialisation:
```suggestion
    // Main loop iteration timer
    Timer elapsedTimer_{Timer(false)};
```

src/gui/guiFuncs.cpp
```diff
@@ -48,9 +48,8 @@ DissolveWindow::DissolveWindow(Dissolve &dissolve)
     addStatusBarIcon(":/control/icons/control_step.svg")->setToolTip("Current step / iteration number");
     iterationLabel_ = addStatusBarLabel("00000");
     iterationLabel_->setToolTip("Current step / iteration number");
-    addStatusBarIcon(":/general/icons/general_clock.svg")->setToolTip("Time remaining to completion");
-    etaLabel_ = addStatusBarLabel("--:--:--");
-    etaLabel_->setToolTip("Time remaining to completion");
+    addStatusBarIcon(":/general/icons/general_clock.svg");
+    timerLabel_ = addStatusBarLabel(timerText_);
```
trisyoungs:
```suggestion
    timerLabel_ = addStatusBarLabel(defaultTimerText);
```

src/gui/guiFuncs.cpp
```diff
@@ -237,6 +236,10 @@ bool DissolveWindow::loadInputFile(std::string_view inputFile, bool handleRestar
         fullUpdate();
     }
 
+    // Empty timer text
+    elapsedTimer_.zero();
+    timerLabel_->setText(timerText_);
```
trisyoungs:
```suggestion
    timerLabel_->setText(defaultTimerText);
```

src/gui/guiFuncs.cpp
```diff
@@ -306,6 +309,9 @@ void DissolveWindow::updateStatusBar()
     {
         statusLabel_->setText("Running (ESC to stop)");
         statusIndicator_->setPixmap(QPixmap(":/control/icons/control_play.svg"));
+        // timerLabel_ reports elapsed time
+        timerLabel_->setText(
+            QString::fromStdString(elapsedTimer_.hasStarted() ? elapsedTimer_.elapsedTimeString(true) : timerText_));
```
trisyoungs:
```suggestion
        timerLabel_->setText(elapsedTimer_.hasStarted() ? 
            QString::fromStdString(elapsedTimer_.elapsedTimeString(true)) : defaultTimerText);
```

src/gui/guiFuncs.cpp
```diff
@@ -401,18 +407,23 @@ void DissolveWindow::fullUpdate()
 // Update while running
 void DissolveWindow::updateWhileRunning(int iterationsRemaining)
 {
+
```
trisyoungs, line 414:
```suggestion
```

src/gui/guiFuncs.cpp
```diff
@@ -401,18 +407,23 @@ void DissolveWindow::fullUpdate()
...
     refreshing_ = true;
 
     // Set current iteration number
     iterationLabel_->setText(QStringLiteral("%1").arg(dissolve_.iteration(), 6, 10, QLatin1Char('0')));
 
+    // Text is set to time elapsed if iterating indefinitely
+    if (iterationsRemaining == -1)
+    {
+        timerLabel_->setText(QString::fromStdString(elapsedTimer_.elapsedTimeString(true)));
+    }
```
trisyoungs: For one-liners after an `if` I tend to omit the braces. Why take up three lines when you only need one!
```suggestion
    if (iterationsRemaining == -1)
        timerLabel_->setText(QString::fromStdString(elapsedTimer_.elapsedTimeString(true)));
```

src/gui/guiFuncs.cpp
```diff
@@ -401,18 +407,23 @@ void DissolveWindow::fullUpdate()
...
     // Set ETA text if we can
-    if (iterationsRemaining == -1 || iterationsRemaining == 0)
-        etaLabel_->setText("--:--:--");
     else
     {
         auto estimatedTime = dissolve_.estimateRequiredTime(iterationsRemaining);
-        etaLabel_->setText(estimatedTime ? QString::fromStdString(Timer::etaString(estimatedTime.value())) : "??:??:??");
+        timerLabel_->setText(estimatedTime ? QString::fromStdString(Timer::timeString(estimatedTime.value(), false))
+                                           : timerText_);
```
trisyoungs:
```suggestion
        timerLabel_->setText(estimatedTime ? QString::fromStdString(Timer::timeString(estimatedTime.value(), false))
                                           : defaultTimerText);
```

trisyoungs (commented):
Looking pretty good!  Few minor suggestions here and there.

I wonder if we should prefix the time strings with either "Time Elapsed: " or "Time Remaining: ", depending on the context, to make things clear?

I also notice that  after running normally (i.e. not for a set number of steps) the time elapsed remains shown, rather than reverting to the default text.

rhinoella:
image

### Comments:
- When stopping an infinite run, the timer label remains displaying the time elapsed
- When a run of specified iterations finishes, the timer label displays the default text ('Idle')

src/gui/guiFuncs.cpp
```diff
@@ -307,6 +313,10 @@ void DissolveWindow::updateStatusBar()
         statusLabel_->setText("Running (ESC to stop)");
         statusIndicator_->setPixmap(QPixmap(":/general/icons/play.svg"));
     }
+    else if (!dissolveIterating_ && elapsedTimer_.secondsElapsed() != 0.0)
```
trisyoungs: I always find exact comparison of `double` values a bit dubious when used in logic statements, so suggesting a greater-than instead.
```suggestion
    else if (!dissolveIterating_ && elapsedTimer_.secondsElapsed() > 0.0)
```

trisyoungs (approved):
One teeny-tiny comment, but this all looks good and works really nicely.

PR #1566, opened 2023-07-24
1563 Module Double Vector

### LINKED ISSUE #1563: Vector of Modules plus Double
Specifying some kind of "weight" to a target module is not currently possible in a nice way. We should introduce a new keyword to enable a vector of `Module`/`double` pairs to be specified in our input.

trisyoungs:
This PR implements a new keyword type based on a `std::vector` of `std::pair<Module *, double>`, with the intent to allow weighting factors to be applied to modules in specific cases.

image

Closes #1563.


src/keywords/weightedModuleVector.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if (it != data_.end())
+        data_.erase(it);
+}
+
+// Express as a serialisable value
+SerialisedValue WeightedModuleVectorKeyword::serialise() const
+{
+    return fromVector(data_,
+                      [](const auto &item) -> SerialisedValue {
+                          return {{"target", item.first->name()}, {"weight", item.second}};
```
rprospero, line 89: 👍

rprospero (approved):
Everything looks good.  I like the TableModel

PR #1569, opened 2023-07-25
1548 Calculate errors over a specific range

### LINKED ISSUE #1548: (EPSR) Calculate errors over specific ranges
While the specific use case here is for the `EPSRModule`, the ability to calculate r-factors (errors) over specific ranges would be useful when assessing fit quality, and particularly when comparing across simulations.

As a first step towards this the main `error(ErrorType ...)` function in `src/math/error.h` could be overridden to accept a range over which the error is to be calculated. This can then construct a suitably "trimmed" `Data1D` object for the "B" data which was passed - since all the error functions use the "B" data to assess the calculation limits, this is probably the easiest way to achieve what we want.  Should capture the general operation of the error calculation, including ranges, in a new unit test with some artificial data.

Secondly, implementing the capability in EPSR could be as simple as introducing a new keyword through which we can define _additional_ ranges to calculate error over (we assume we will always calculated the total R-factor over the whole range). A new type of keyword acting on a `std::vector<Range>` will be required, as will a suitable GUI widget (at some point).  Displaying the different results in the GUI is something to be discussed once implemented!

rhinoella:
Will allow error calculations to happen in a specified range.

Closes #1548

### DIFF
--- a/src/math/error.cpp
+++ b/src/math/error.cpp
@@ -18,9 +18,9 @@ EnumOptions<ErrorType> errorTypes()
                                                 {MAAPEError, "MAAPE"},
                                                 {MAPEError, "MAPE"},
                                                 {PercentError, "Percent"},
-                                                {ASEError, "Absolute Squared"},
+                                                {ASEError, "ASE"},
                                                 {RFactorError, "RFactor"},
-                                                {EuclideanError, "Euclidean Distance"}});
+                                                {EuclideanError, "Euclidean"}});
 }
 
 std::string errorReportString(ErrorReport errorReport)
@@ -45,6 +45,8 @@ ErrorReport error(ErrorType errorType, const Data1D &A, const Data1D &B, Optiona
         return mape(A, B, range);
     else if (errorType == PercentError)
         return percent(A, B, range);
+    else if (errorType == ASEError)
+        return ase(A, B, range);
     else if (errorType == RFactorError)
         return rFactor(A, B, range);
     else if (errorType == EuclideanError)
@@ -79,6 +81,8 @@ ErrorReport error(ErrorType errorType, const std::vector<double> &vecA, const st
         return mape(A, B);
     else if (errorType == PercentError)
         return percent(A, B);
+    else if (errorType == ASEError)
+        return ase(A, B, range);
     else if (errorType == RFactorError)
         return rFactor(A, B);
     else if (errorType == EuclideanError)
--- a/src/math/error.cpp
+++ b/src/math/error.cpp
@@ -18,39 +18,45 @@ EnumOptions<ErrorType> errorTypes()
                                                 {MAAPEError, "MAAPE"},
                                                 {MAPEError, "MAPE"},
                                                 {PercentError, "Percent"},
+                                                {ASEError, "Absolute Squared"},
                                                 {RFactorError, "RFactor"},
-                                                {EuclideanError, "Euclidean"}});
+                                                {EuclideanError, "Euclidean Distance"}});
+}
+
+std::string errorReportString(ErrorReport errorReport)
+{
+    return fmt::format("{} between datasets is {:15.9e} over {:15.9e} < x < {:15.9e} ({} points).\n",
+                       errorTypes().keyword(errorReport.errorType), errorReport.value, errorReport.firstX, errorReport.lastX,
+                       errorReport.nPointsConsidered);
 }
 
 // Return error of specified type between supplied data
-double error(ErrorType errorType, const Data1D &A, const Data1D &B, bool quiet, OptionalReferenceWrapper<Range> range)
+ErrorReport error(ErrorType errorType, const Data1D &A, const Data1D &B, OptionalReferenceWrapper<Range> range)
 {
-
     // if range is unset
     auto rangeMin = range ? range->get().minimum() : B.xAxis().front();
     auto rangeMax = range ? range->get().maximum() : B.xAxis().back();
 
     if (errorType == RMSEError)
-        return rmse(A, B, quiet, range);
+        return rmse(A, B, range);
     else if (errorType == MAAPEError)
-        return maape(A, B, quiet, range);
+        return maape(A, B, range);
     else if (errorType == MAPEError)
-        return mape(A, B, quiet, range);
+        return mape(A, B, range);
     else if (errorType == PercentError)
-        return percent(A, B, quiet, range);
+        return percent(A, B, range);
     else if (errorType == RFactorError)
-        return rFactor(A, B, quiet, range);
+        return rFactor(A, B, range);
     else if (errorType == EuclideanError)
-        return euclidean(A, B, quiet, range);
+        return euclidean(A, B, range);
 
     throw(
         std::runtime_error(fmt::format("Error type {} is not accounted for! Take the developer's Kolkata privileges away...\n",
                                        errorTypes().keyword(errorType))));
-    return 0.0;
 }
 
 // Return error of specified type between supplied double vectors
-double error(ErrorType errorType, const std::vector<double> &vecA, const std::vector<double> &vecB, bool quiet)
+ErrorReport error(ErrorType errorType, const std::vector<double> &vecA, const std::vector<double> &vecB)
 {
     // Size check
     assert(vecA.size() == vecB.size());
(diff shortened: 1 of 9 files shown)


rhinoella:
### Changes
- All error functions now accept a range parameter
- Range has a default object (does not need to be specified)

src/math/error.cpp
```diff
@@ -23,20 +23,27 @@ EnumOptions<ErrorType> errorTypes()
 }
 
 // Return error of specified type between supplied data
-double error(ErrorType errorType, const Data1D &A, const Data1D &B, bool quiet)
+double error(ErrorType errorType, const Data1D &A, const Data1D &B, bool quiet, Range range)
```
trisyoungs: Two things to say about the functions in general: 1) I think the `Range` argument should legitimately come before the `bool quiet` flag in terms of importance, and; 2) if we use a `std::optional<Range>` this gives us a better ability to check whether a `Range` has actually been supplied, rather than comparing the `minimum()` and `maximum()` for equality. This isn't the only "invalid" range condition, so I think we have to leave that as the responsibility of the user / coder.

src/math/error.cpp
```diff
@@ -23,20 +23,27 @@ EnumOptions<ErrorType> errorTypes()
...
 {
+
+    // if range is unset
+    if (range.maximum() == range.minimum())
+    {
+        range.set(B.xAxis().front(), B.xAxis().back());
+    }
```
trisyoungs: With `std::optional<Range>` as our argument, we could do some checks here, but we would have to do the same check for every function since they're generally accessible. So, I would remove the check here and modify the checks in the individual functions...
```suggestion
```

src/math/error.cpp
```diff
@@ -45,7 +52,7 @@ double error(ErrorType errorType, const Data1D &A, const Data1D &B, bool quiet)
 }
 
 // Return error of specified type between supplied double vectors
-double error(ErrorType errorType, const std::vector<double> &vecA, const std::vector<double> &vecB, bool quiet)
+double error(ErrorType errorType, const std::vector<double> &vecA, const std::vector<double> &vecB, bool quiet, Range range)
```
trisyoungs: We should leave this particular function be as a "non-Rangable" error check - since we're artificially creating the `Data1D` anyway an applied `Range` is kinda meaningless.  This function was introduced to handle a few specific test cases, so is specialised use anyway.

src/math/error.cpp
```diff
@@ -82,23 +95,30 @@ double error(ErrorType errorType, const std::vector<double> &vecA, const std::ve
  */
 
 // Return RMSE between supplied data
-double rmse(const Data1D &A, const Data1D &B, bool quiet)
+double rmse(const Data1D &A, const Data1D &B, bool quiet, Range range)
```
trisyoungs: Switch order of `bool quiet` and `std::optional<Range> range` as suggested.

src/math/error.cpp
```diff
@@ -82,23 +95,30 @@ double error(ErrorType errorType, const std::vector<double> &vecA, const std::ve
...
 {
     // First, generate interpolation of data B
     Interpolator interpolatedB(B);
 
+    // if range is unset
+    if (range.maximum() == range.minimum())
+    {
+        range.set(B.xAxis().front(), B.xAxis().back());
+    }
```
trisyoungs: What I suggest we do here, given our `std::optional<Range>` argument, is establish the relevant limits in some temporary vars, and check those vars in the main loop:
```suggestion
    // Set calculation range
    auto rangeMin = range ? range.minimum() : B.xAxis().front();
    auto rangeMax = range ? range.maximum() : B.xAxis().back();
```

src/math/error.cpp
```diff
@@ -82,23 +95,30 @@ double error(ErrorType errorType, const std::vector<double> &vecA, const std::ve
...
     {
         // Is our x value lower than the lowest x value of the reference data?
-        if (x < B.xAxis().front())
+        if (x < range.minimum())
             continue;
 
         // Is our x value higher than the last x value of the reference data?
-        if (x > B.xAxis().back())
+        if (x > range.maximum())
             break;
```
trisyoungs: This can then be something like:
```
            // Check range
            if (x < rangeMin)
                continue;
            if (x > rangeMax)
                break;
```

src/math/error.h
```diff
@@ -31,21 +32,22 @@ EnumOptions<Error::ErrorType> errorTypes();
...
-// Return error of specified type between supplied double vectors
-double error(ErrorType errorType, const std::vector<double> &vecA, const std::vector<double> &vecB, bool quiet = false);
+// Return error of specified type between supplied data, allows a range parameter to be specified
+double error(ErrorType errorType, const Data1D &A, const Data1D &B, bool quiet = false, Range range = Range());
+// Return error of specified type between supplied double vectors, allows a range parameter to be specified
+double error(ErrorType errorType, const std::vector<double> &vecA, const std::vector<double> &vecB, bool quiet = false,
+             Range range = Range());
 // Return RMSE between supplied data
-double rmse(const Data1D &A, const Data1D &B, bool quiet = false);
+double rmse(const Data1D &A, const Data1D &B, bool quiet = false, Range range = Range());
```
trisyoungs:
```suggestion
double rmse(const Data1D &A, const Data1D &B, std::optional<Range> range = {}, bool quiet = false);
```
etc.

trisyoungs (changes requested):
Good case for a `std::optional` I think! See comments below.

rhinoella:
### Changes
- Range parameter is now an `OptionalReferenceWrapper<Range>`
- Removed `bool quiet` parameter (not Error's job to print output now)
- All functions now return an `ErrorReport` struct that contains all the information needed to print an error report
- New error function `ase()` (and corresponding error type `ASEError`) returns absolute squared error (this used to happen in the `percent()` function, if sum of y values is zero it returned ase instead to avoid dividing by zero)
  - This is one of the reasons I chose to have the functions return a struct, so that the caller of `percent()` would be able to know that they are getting ase error instead of percent error

src/math/error.h
```diff
@@ -20,32 +22,46 @@ enum ErrorType
...
+    ASEError,
     RFactorError,
     EuclideanError,
     nErrorCalculationTypes
 };
+// Error information struct
+struct ErrorReport
+{
+    ErrorType errorType;
+    double value;
```
trisyoungs: Semantics here, but do you think this should be called `double error` rather than `double value`?

src/math/error.h
```diff
@@ -20,32 +22,46 @@ enum ErrorType
...
+// Return error report formatted to string
+std::string errorReportString(ErrorReport errorReport);
 
 /*
  * Data1D
  */
 
-// Return error of specified type between supplied data
-double error(ErrorType errorType, const Data1D &A, const Data1D &B, bool quiet = false);
+// Return error of specified type between supplied data, allows a range parameter to be specified
```
trisyoungs:
```suggestion
// Return error of specified type between supplied data over all points or the range given
```

src/modules/gr/functions.cpp
```diff
@@ -574,47 +574,56 @@ bool GRModule::testReferencePartials(PartialSet &setA, PartialSet &setB, double
 {
     // Get a copy of the AtomTypeList to work from
     auto atomTypes = setA.atomTypeMix();
-    double error;
+    Error::ErrorReport error;
```
trisyoungs: I think its safer / neater if this is moved inside the loop.
```suggestion
```

src/modules/gr/functions.cpp
```diff
@@ -574,47 +574,56 @@ bool GRModule::testReferencePartials(PartialSet &setA, PartialSet &setB, double
...
 
     for_each_pair_early(
         atomTypes.begin(), atomTypes.end(),
         [&](int n, const AtomTypeData &typeI, int m, const AtomTypeData &typeJ) -> EarlyReturn<bool>
         {
             // Full partial
             error = Error::percent(setA.partial(n, m), setB.partial(n, m));
```
trisyoungs:
```suggestion
            auto error = Error::percent(setA.partial(n, m), setB.partial(n, m));
```

src/math/error.h
```diff
@@ -20,32 +22,46 @@ enum ErrorType
...
+    ErrorType errorType;
+    double value;
+    double firstX;
+    double lastX;
+    int nPointsConsidered;
+};
 // Return enum options for ErrorType
 EnumOptions<Error::ErrorType> errorTypes();
+// Return error report formatted to string
+std::string errorReportString(ErrorReport errorReport);
```
trisyoungs: Only a simple `struct` I know, but you may as well make this a `const &` to ensure it doesn't get modified in the function.
```suggestion
std::string errorReportString(const ErrorReport &errorReport);
```

trisyoungs (approved):
Couple of minor bits, but this is nice.   I like the little `ErrorReport` object - that's a neat thing.

Issue #88, opened 2019-08-06
Control Layer


trisyoungs:
An (optional) Control Layer can manage the execution of processing layers, enabling regeneration of configurations, minimisation, refinement, testing, fitting of parameters, rather than just running through them one-by-one. With Procedure definitions for configuration creation, this could be the best route to allow searching through this type of parameter space. Can dump a restart file at each 'good' or 'best' point, rather than every 'n' iterations of the main loop.

Control Layer could be present all the time, but hidden, with the default option being just to continually iterate over the layers.

jws-1:
Closing as superseded by #1652.

PR #1570, opened 2023-07-25
Layer Keyword

jws-1:
This PR implements a `Layer` keyword and an associated widget, for usage in nodes. Also moves `processingLayers_` (and the relevant member functions) to `CoreData`. The addition of this keyword does not have any implications currently, in terms of functionality, but it will be useful for #88 (and currently #1565)..

image


src/keywords/layer.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Read values from a serialisable value
+void LayerKeyword::deserialise(const SerialisedValue &node, const CoreData &coreData)
+{
+    data_ = coreData.findProcessingLayer(std::string_view(std::string(node.as_string())));
+}
+
+// Express as a serialisable value
+SerialisedValue LayerKeyword::serialise() const { return data_->name(); }
```
trisyoungs: Need to check for the validity of `data_` before accessing it, just to be on the safe side.

src/keywords/layer.cpp
```diff
@@ -0,0 +1,50 @@
...
+    if (!data_)
+        return true;
+
+    return parser.writeLineF("{}{}  '{}'\n", prefix, keywordName, data_->name());
+}
+
+// Read values from a serialisable value
+void LayerKeyword::deserialise(const SerialisedValue &node, const CoreData &coreData)
+{
+    data_ = coreData.findProcessingLayer(std::string_view(std::string(node.as_string())));
```
trisyoungs: Need to check for the layer's existence and `throw` if it isn't there (see e.g. `ModuleKeyword` equivalent).

src/main/layers.cpp
```diff
@@ -7,62 +7,15 @@
...
 // Run the set-up stages of all modules in all layers
 bool Dissolve::setUpProcessingLayerModules()
 {
     auto setUpResult = true;
-    for (auto &layer : processingLayers_)
+    for (auto &layer : coreData_.processingLayers())
         if (!layer->setUpAll(*this, worldPool()))
             setUpResult = false;
     return setUpResult;
 }
```
trisyoungs: Kinda feel like this should be part of `CoreData` too since it only operates on data local to it, but of course we pass in the `Dissolve` object by necessity.  Worth capturing the test of whether `Module`s really need `Dissolve` and can cope with just `CoreData` as a separate issue, and flag this function to be moved if so.

src/gui/keywordWidgets/layerFuncs.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+/*
+ * Update
+ */
+void LayerKeywordWidget::updateLayersCombo()
+{
+    refreshing_ = true;
+
+    moduleLayersModel_.setData(coreData_.processingLayers());
+    ui_.LayerCombo->setCurrentIndex(-1);
```
trisyoungs: Need to set the combo to the correct index based on the current data, rather than "no item":
```suggestion
    // Set current index based on keyword data
    auto it = std::find(coreData_.processingLayers().begin(), coreData_.processingLayers().end(), keyword_->data());
    ui_.LayerCombo->setCurrentIndex(it == coreData_.processingLayers().end() ? -1 : it - coreData_.processingLayers().begin());
```
Something like this, anyway!

trisyoungs (changes requested):
Yep, I think moving the layers to `CoreData` feels very much like the right thing to do.  Have made a few notes on implementation of the keyword itself, and a bug or two to squash.

src/classes/coreData.cpp
```diff
@@ -528,6 +531,64 @@ Configuration *CoreData::findConfigurationByNiceName(std::string_view name) cons
...
+
+// Add new processing layer
+ModuleLayer *CoreData::addProcessingLayer() { return processingLayers_.emplace_back(std::make_unique<ModuleLayer>()).get(); }
+
+// Remove processing layer
+void CoreData::removeProcessingLayer(ModuleLayer *layer)
+{
+    if (!layer)
+        return;
+
```
trisyoungs, line 546: Don't we need this in here?
```suggestion

    // Remove any references to the Modules in the layer before we delete it
    for (auto &module : layer->modules())
        removeReferencesTo(module.get());
        
```

src/classes/coreData.cpp
```diff
@@ -528,6 +531,64 @@ Configuration *CoreData::findConfigurationByNiceName(std::string_view name) cons
...
+        return Messenger::error("Already own ModuleLayer '{}', so nothing to do.\n", layer->name());
+
+    processingLayers_.emplace_back(layer);
+
+    return true;
+}
+
+// Return number of defined processing layers
+int CoreData::nProcessingLayers() const { return processingLayers_.size(); }
+
```
trisyoungs: I believe these are both unused, old functions so can be removed.
```suggestion
```

src/classes/coreData.cpp
```diff
@@ -528,6 +531,64 @@ Configuration *CoreData::findConfigurationByNiceName(std::string_view name) cons
...
+// Return list of defined processing layers
```
trisyoungs:
```suggestion
// Return current processing layers
```

src/classes/coreData.h
```diff
@@ -186,6 +188,30 @@ class CoreData
...
+    // Add new processing layer
+    ModuleLayer *addProcessingLayer();
+    // Remove processing layer
+    void removeProcessingLayer(ModuleLayer *layer);
+    // Find named processing layer
+    ModuleLayer *findProcessingLayer(std::string_view name) const;
+    // Own the specified processing layer
+    bool ownProcessingLayer(ModuleLayer *layer);
+    // Return number of defined processing layers
+    int nProcessingLayers() const;
```
trisyoungs: I believe these are both unused, old functions so can be removed.
```suggestion
```

src/classes/coreData.h
```diff
@@ -186,6 +188,30 @@ class CoreData
...
+    // Return list of defined processing layers
```
trisyoungs:
```suggestion
    // Return current processing layers
```

src/keywords/layer.cpp
```diff
@@ -0,0 +1,59 @@
...
+/*
+ * Arguments
+ */
+
+// Deserialise from supplied LineParser, starting at given argument offset
+bool LayerKeyword::deserialise(LineParser &parser, int startArg, const CoreData &coreData)
+{
+    data_ = coreData.findProcessingLayer(parser.argsv(startArg));
+    if (!data_)
+        return Messenger::error("Error setting Layer - no Layer named '{}' exists.\n", parser.argsv(startArg));
```
trisyoungs: For consistency.
```suggestion
        return Messenger::error("Layer '{}' given to keyword {} doesn't exist.\n", parser.argsv(startArg), KeywordBase::name());
```

src/main/dissolve.h
```diff
@@ -109,26 +109,10 @@ class Dissolve : public Serialisable<>
      * Layers
      */
```
trisyoungs: This needs renaming now:
```suggestion
     * Processing Module Data
     */
```

src/main/layers.cpp
trisyoungs, line 1: This file now only contains a single function and is the "wrong" name, so I think we should move that function to `dissolve.cpp` under the renamed heading ("Processing Module Data") and delete this file.

trisyoungs (changes requested):
Another iteration required here, but mostly just to tidy a few things I spotted / realised.

src/classes/coreData.cpp
jws-1: Yes, we probably do!

trisyoungs (approved):

PR #1571, opened 2023-07-25
Fix KeywordStore::get

trisyoungs:
A very brief PR to fix an issue with the `get()` template functions in `KeywordStore`.  The return type was `OptionalReferenceWrapper<D>`, but this is a wholly unsuitable type to be returning, e.g. `double` data and many others.

The return type is modified to be a plain `std::optional`, and given the sparing use of the function in the code we don't need to modify the target data through its reference anyway.

### DIFF
--- a/src/keywords/store.h
+++ b/src/keywords/store.h
@@ -200,7 +200,7 @@ class KeywordStore
     int getInt(std::string_view name) const;
 
     // Get specified keyword data, casting as necessary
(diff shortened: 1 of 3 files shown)


rprospero (approved):
I feel that Dissolve has largely outgrown OptionalReferenceWrapper and I'm always happy for PRs that get rid of it.

And, yes, I appreciate the irony.

PR #1572, opened 2023-07-25
Fix EPSR test data

trisyoungs:
This PR fixes up the EPSR test simulation data supplied with Dissolve, and which previously had one incorrect dataset (that for D2O, which was in fact a 25:75 mix of H2O;D2O).  This didn't cause any major (or even minor) issues with the testing since the numerical methods still worked, but I would rather the data be correct and what it alludes to be!


jws-1 (approved):
:+1:

PR #1573, opened 2023-07-25
Space Separated Quoted Strings in CIFs

### LINKED ISSUE #1551: Allow space-separated-values in CIFs
Some CIFs have quoted space-separated-values rather than being separated by new-lines, currently we don't account for this and thus when trying to load such files Dissolve crashes.

For illustration, the following causes a crash:
```
...
_audit_author_name
_audit_author_address
'Name'  'Address'
...
```

jws-1:
Small PR to fix a bug where CIFs are not parsed when (single and double) quoted strings are space-separated on the same line. This bug is apparently due to the fact that the `*` operator is greedy, and meant that the closing `'`/`"` were included in the match, leading to: "'x' 'y'" being parsed as "x' 'y".

Closes #1551.

### DIFF
--- a/src/io/import/CIFImportLexer.g4
+++ b/src/io/import/CIFImportLexer.g4
@@ -71,6 +71,6 @@ UnsignedInteger:    DIGIT+;
 
 // Character Strings
 UnquotedString:     OrdinaryChar NonBlankChar*;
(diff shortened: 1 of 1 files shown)


trisyoungs (approved):
This makes sense in terms of the lexer!
:+1:

PR #1574, opened 2023-07-26
Filter Node Tree

### LINKED ISSUE #1424: Filter node tree by context
We can easily filter the available nodes tree by the current procedure context to which it is attached. Thus, for a Configuration generator, for example, we would only show the relevant nodes for a Generate context.

jws-1:
This PR implements the ability to filter Node trees using the parents context. The filtering is done by maintaining a vector of categories for each context - I'm not at all sure about these categories, so these likely need to be modified before merging this in.

Closes #1424.

### DIFF
--- a/src/gui/models/nodePaletteFilterProxy.cpp
+++ b/src/gui/models/nodePaletteFilterProxy.cpp
@@ -5,7 +5,7 @@
 #include "procedure/nodes/registry.h"
 
 // Set allowed categories
-void NodePaletteFilterProxy::setContext(const ProcedureNode::NodeContext context)
+void NodePaletteFilterProxy::setContext(ProcedureNode::NodeContext context)
 {
     context_ = context;
     invalidateFilter();
--- a/src/gui/models/nodePaletteFilterProxy.cpp
+++ b/src/gui/models/nodePaletteFilterProxy.cpp
@@ -4,12 +4,10 @@
 #include "gui/models/nodePaletteFilterProxy.h"
 #include "procedure/nodes/registry.h"
 
-NodePaletteFilterProxy::NodePaletteFilterProxy(const std::vector<std::string>& categories) : categories_(categories) {}
-
 // Set allowed categories
-void NodePaletteFilterProxy::setCategories(const std::vector<std::string>& categories)
+void NodePaletteFilterProxy::setContext(const ProcedureNode::NodeContext context)
 {
-    categories_ = categories;
+    context_ = context;
     invalidateFilter();
 }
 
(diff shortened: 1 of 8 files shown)


src/procedure/nodes/node.cpp
```diff
@@ -81,7 +81,7 @@ EnumOptions<ProcedureNode::NodeContext> ProcedureNode::nodeContexts()
                                                                    {ProcedureNode::AnalysisContext, "Analysis"},
                                                                    {ProcedureNode::GenerationContext, "Generation"},
                                                                    {ProcedureNode::OperateContext, "Operate"},
-                                                                   {ProcedureNode::OperateContext, "Any"},
+                                                                   {ProcedureNode::AnyContext, "Any"},
```
trisyoungs, line 84: :+1:

src/gui/models/nodePaletteFilterProxy.cpp
```diff
@@ -0,0 +1,35 @@
...
+        {ProcedureNode::NodeContext::AnalysisContext, {"Data"}},
+        {ProcedureNode::NodeContext::GenerationContext,
+         {"Calculate", "Build", "Calculate", "Data", "General", "Operate", "Pick", "Potentials", "Sites"}},
+        {ProcedureNode::NodeContext::OperateContext,
+         {"Calculate", "Data", "General", "Operate", "Pick", "Potentials", "Sites"}},
+    };
+
+    const auto &contextCategoryList = contextCategories.at(context_);
+    auto category = std::next(ProcedureNodeRegistry::categoryMap().begin(), parent.isValid() ? parent.row() : row)->first;
+    return std::find(contextCategoryList.begin(), contextCategoryList.end(), category) != contextCategoryList.end();
```
trisyoungs: I'm finding it hard to decide whether this is they way we should do it or not!  Given that, for a given context, I think there are either very few, or very many, exclusions to make, perhaps it might be better to structure it something like this:
```suggestion
    auto category = std::next(ProcedureNodeRegistry::categoryMap().begin(), parent.isValid() ? parent.row() : row)->first;
    
    switch (context_)
    {
        case (ProcedureNode::NodeContext::AnyContext):
            return true;
        case (ProcedureNode::NodeContext::NoContext):
            return false;
        case (ProcedureNode::NodeContext::OperateContext):
            return category == "Operate";
        case (ProcedureNode::NodeContext::GenerateContext):
            return category != "Operate";
        case (ProcedureNode::NodeContext::AnalysisContext):
            return category != "Build" && category != "Potentials" && category != "Operate";
        default:
            throw(.....);
    }
```

trisyoungs (commented):
Tricky one this!  Have made a suggested change to your category selection method - see what you think.

src/gui/models/nodePaletteFilterProxy.cpp
jws-1: That's certainly nicer than trying to maintain quite a large map - much less likely to miss something out!

trisyoungs (approved):

PR #1577, opened 2023-07-26
deletion seclectConfigurationWidget and listWidgetUpdater

Danbr4d:
Make the SelectConfigurationWidget use a model for its list of Configurations rather than the ancient and ready-to-be-removed ListWidgetUpdater class.


trisyoungs (approved):
:+1:

PR #1579, opened 2023-07-27
Show Applied Size Factor

### LINKED ISSUE #1578: Display current SizeFactor in GUI
With the move of sizefactor control to a node, there is now no way of knowing what the current sizefactor applied to a configuration box actually is.

Perhaps we can have another frame on our status bar, but one which is only shown when we have an applied size factor (i.e. it's != 1.0). It could also use a warning symbol next to the number, since it's worth highlighting to the user that a size factor is currently in force.

jws-1:
This PR updates the Configuration tab to show the applied size factor, if it is not equal to the default value (of 1.0). Only blocker here is that I couldn't decide on a suitable icon to pair with the frame.

Closes #1578.

### DIFF
--- a/src/classes/configuration.h
+++ b/src/classes/configuration.h
@@ -154,11 +154,12 @@ class Configuration : public Serialisable<const CoreData &>
      * Periodic Box and Cells
      */
     private:
+    // Default size factor for Box
     static constexpr double defaultSizeFactor_ = 1.0;
     // Requested size factor for Box
-    double requestedSizeFactor_{defaultSizeFactor_};
+    std::optional<double> requestedSizeFactor_;
     // Size factor currently applied to Box / Cells
-    double appliedSizeFactor_{defaultSizeFactor_};
+    std::optional<double> appliedSizeFactor_;
     // Periodic Box
     std::unique_ptr<Box> box_{nullptr};
     // Requested side length for individual Cell
@@ -182,12 +183,14 @@ class Configuration : public Serialisable<const CoreData &>
     const Box *box() const;
     // Scale Box lengths (and associated Cells) by specified factors
     void scaleBox(Vec3<double> scaleFactors);
+    // Return default size factor for Box
+    static double defaultSizeFactor();
     // Set requested size factor for Box
     void setRequestedSizeFactor(double factor);
     // Return requested size factor for Box
-    double requestedSizeFactor() const;
+    std::optional<double> requestedSizeFactor() const;
     // Return last size factor applied to Box / Cells
-    double appliedSizeFactor() const;
+    std::optional<double> appliedSizeFactor() const;
     // Set requested side length for individual Cell
     void setRequestedCellDivisionLength(double a);
     // Return requested side length for individual Cell
--- a/src/classes/configuration.h
+++ b/src/classes/configuration.h
@@ -183,8 +183,6 @@ class Configuration : public Serialisable<const CoreData &>
     const Box *box() const;
     // Scale Box lengths (and associated Cells) by specified factors
     void scaleBox(Vec3<double> scaleFactors);
-    // Return default size factor for Box
-    static double defaultSizeFactor();
     // Set requested size factor for Box
     void setRequestedSizeFactor(double factor);
     // Return requested size factor for Box
--- a/src/classes/configuration_box.cpp
+++ b/src/classes/configuration_box.cpp
@@ -65,9 +65,6 @@ void Configuration::scaleBox(Vec3<double> scaleFactors)
     cells_.scale(scaleFactors);
 }
 
-// Return default size factor for Box
-double Configuration::defaultSizeFactor() { return defaultSizeFactor_; }
-
 // Set requested size factor for Box
 void Configuration::setRequestedSizeFactor(double factor) { requestedSizeFactor_ = factor; }
 
--- a/src/classes/configuration_box.cpp
+++ b/src/classes/configuration_box.cpp
@@ -96,7 +96,8 @@ void Configuration::applySizeFactor(const ProcessPool &procPool, const Potential
     while (true)
     {
         // Calculate ratio between current and applied size factors for use later on
-        const auto sizeFactorRatio = requestedSizeFactor_.value_or(defaultSizeFactor_) / appliedSizeFactor_.value_or(defaultSizeFactor_);
+        const auto sizeFactorRatio =
+            requestedSizeFactor_.value_or(defaultSizeFactor_) / appliedSizeFactor_.value_or(defaultSizeFactor_);
 
         // Check current vs applied size factors (via the ratio) - if unequal, perform scaling and set the new applied
         // size factor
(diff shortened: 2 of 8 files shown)


src/gui/configurationTabFuncs.cpp
```diff
@@ -127,6 +130,10 @@ void ConfigurationTab::updateControls()
     ui_.AtomPopulationLabel->setText(QString::number(configuration_->nAtoms()));
     ui_.MoleculePopulationLabel->setText(QString::number(configuration_->nMolecules()));
 
+    // Applied size factor
+    ui_.SizeFactorLabel->setText(QString::number(configuration_->appliedSizeFactor()));
+    ui_.SizeFactorFrame->setVisible(configuration_->appliedSizeFactor() != 1.0);
```
trisyoungs: Comparing real numbers always feels a bit "risky" because of rounding.  I wonder if this is a good opportunity to change both `appliedSizeFactor_` and `requestedSizeFactor_` member variables to be `std::optional<double>`s? Then, things are absolutely clear, and we avoid checking for `1.0` as an effective "magic number".

trisyoungs (approved):
Yep, exactly this!  I'll address the icon in an upcoming PR with a bunch of others which are currently missing / need updating.

Left a comment on the actual size factor member variables for you to muse on, but I'll approve this PR.

src/classes/configuration.h
```diff
@@ -182,12 +183,14 @@ class Configuration : public Serialisable<const CoreData &>
     const Box *box() const;
     // Scale Box lengths (and associated Cells) by specified factors
     void scaleBox(Vec3<double> scaleFactors);
+    // Return default size factor for Box
+    static double defaultSizeFactor();
```
trisyoungs: I think we don't need this..
```suggestion
```

src/classes/configuration_box.cpp
```diff
@@ -65,14 +65,17 @@ void Configuration::scaleBox(Vec3<double> scaleFactors)
     cells_.scale(scaleFactors);
 }
 
+// Return default size factor for Box
+double Configuration::defaultSizeFactor() { return defaultSizeFactor_; }
+
```
trisyoungs:
```suggestion
```

src/classes/configuration_box.cpp
```diff
@@ -93,15 +96,17 @@ void Configuration::applySizeFactor(const ProcessPool &procPool, const Potential
     while (true)
     {
         // Calculate ratio between current and applied size factors for use later on
-        const auto sizeFactorRatio = requestedSizeFactor_ / appliedSizeFactor_;
+        const auto sizeFactorRatio =
+            requestedSizeFactor_.value_or(defaultSizeFactor_) / appliedSizeFactor_.value_or(defaultSizeFactor_);
```
trisyoungs: It feels like we should tidy this up and do both `value_or` retrievals into a couple of temporary variables (`requestedSF` and `appliedSF` perhaps) before the `while` loop.

src/classes/configuration_box.cpp
```diff
@@ -125,18 +130,20 @@ void Configuration::applySizeFactor(const ProcessPool &procPool, const Potential
...
         else if (EnergyModule::interMolecularEnergy(procPool, this, potentialMap) <= 0.0)
         {
-            requestedSizeFactor_ *= reductionFactor;
-            if (requestedSizeFactor_ < 1.0)
+            requestedSizeFactor_ = requestedSizeFactor_.value_or(defaultSizeFactor_) * reductionFactor;
+            if (requestedSizeFactor_.value_or(defaultSizeFactor_) < 1.0)
                 requestedSizeFactor_ = 1.0;
-            Messenger::print("Intermolecular energy is zero or negative, so reducing SizeFactor to {}\n", requestedSizeFactor_);
+            Messenger::print("Intermolecular energy is zero or negative, so reducing SizeFactor to {}\n",
+                             requestedSizeFactor_.value_or(defaultSizeFactor_));
```
trisyoungs: It definitely has a value here as well as in a few lines above, so:
```suggestion
                             *requestedSizeFactor_);
```

src/classes/configuration_box.cpp
```diff
@@ -125,18 +130,20 @@ void Configuration::applySizeFactor(const ProcessPool &procPool, const Potential
...
         }
         else
         {
-            Messenger::print("Intermolecular energy is positive, so SizeFactor remains at {}\n", requestedSizeFactor_);
+            Messenger::print("Intermolecular energy is positive, so SizeFactor remains at {}\n",
+                             requestedSizeFactor_.value_or(defaultSizeFactor_));
             break;
         }
```
trisyoungs, line 148: At the end of the function, if `appliedSizeFactor_` is now (within error) equal to `1.0` we can set it to `std::nullopt`.

src/classes/configuration_io.cpp
```diff
@@ -156,7 +157,8 @@ bool Configuration::deserialise(LineParser &parser, const CoreData &coreData, do
     atomTypes_.finalise();
 
     // Scale box and cells according to the applied size factor
-    scaleBox({appliedSizeFactor_, appliedSizeFactor_, appliedSizeFactor_});
+    scaleBox({appliedSizeFactor_.value_or(defaultSizeFactor_), appliedSizeFactor_.value_or(defaultSizeFactor_),
```
trisyoungs:
```suggestion
    auto appliedSF = appliedSizeFactor_.value_or(defaultSizeFactor_);
    scaleBox({appliedSF, appliedSF, appliedSF});
```

src/module/layer.cpp
```diff
@@ -84,7 +84,10 @@ bool ModuleLayer::canRun(GenericList &processingModuleData) const
         // Check that Configurations have unmodified size factor
         if (std::any_of(cfgs.begin(), cfgs.end(),
                         [](const auto *cfg)
-                        { return std::abs(cfg->appliedSizeFactor() - 1.0) > 2 * std::numeric_limits<double>::epsilon(); }))
+                        {
+                            return std::abs(cfg->appliedSizeFactor().value_or(Configuration::defaultSizeFactor()) - 1.0) >
+                                   2 * std::numeric_limits<double>::epsilon();
```
trisyoungs:
```suggestion
                            return cfg->appliedSizeFactor();
```

trisyoungs (changes requested):
Yes, I think optionals were the way to go, but let's use their full power! :)

src/gui/configurationTabFuncs.cpp
```diff
@@ -127,6 +130,10 @@ void ConfigurationTab::updateControls()
     ui_.AtomPopulationLabel->setText(QString::number(configuration_->nAtoms()));
     ui_.MoleculePopulationLabel->setText(QString::number(configuration_->nMolecules()));
 
+    // Applied size factor
+    ui_.SizeFactorLabel->setText(QString::number(configuration_->appliedSizeFactor().value_or(0.0)));
```
trisyoungs: Picky, I know, but the `0.0` doesn't represent a valid size factor (even if it isn't being shown!)
```suggestion
    ui_.SizeFactorLabel->setText(configuration_->appliedSizeFactor() ? QString::number(*configuration_->appliedSizeFactor() : "N/A"));
```

trisyoungs (approved):
One tiny and almost irrelevant suggestion, then good to go! :+1:

src/gui/configurationTabFuncs.cpp
jws-1: I was getting ready to defend this, but I think your option is better.

PR #1581, opened 2023-07-27
Broken Layer Context Menu

### LINKED ISSUE #1580: Layer context menu is broken
**Describe the bug**
The tab right-click / context menu looks to be badly broken, since clicking on _any_ tab gives the options to enable / disable layers. Also, the target layer of the click seems to suffer an off-by-one error, since the one to the right is affected...

jws-1:
Small PR fixing issues with the tab context menu.

Closes #1580.


trisyoungs (approved):
:+1:
And good catch on the tab array construction!

PR #1582, opened 2023-07-27
Fix TOML serialisation of ProcedureSequenceNode

rprospero:
This moves to using an array, instead of a table, for the export of ProcedureSequenceNode.  I'd originally went the map route because the file editing felt *slightly* more ergonomic, but the backing the table is an `unordered_map` and there's no good way of assuring that the sequence is loaded in the proper order.  Furthermore, moving to an array, instead of a table, turned out to be more ergonomic in practice (i.e. no need for the user to assign a name to every node).

This takes us from 35 failing tests down to 20.  Also, we now have zero segfaults.

### DIFF
--- a/src/classes/braggReflection.cpp
+++ b/src/classes/braggReflection.cpp
@@ -90,39 +90,3 @@ void BraggReflection::setHKL(int h, int k, int l) { hkl_.set(h, k, l); }
 
 // Return Miller indices of reflection
 const Vec3<int> &BraggReflection::hkl() const { return hkl_; }
-
-/*
- * Serialisation
- */
-
-// Read data through specified parser
-bool BraggReflection::deserialise(LineParser &parser)
-{
-    // Read index, Q centre, and number of contributing K-vectors
-    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
-        return false;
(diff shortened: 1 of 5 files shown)


src/classes/braggReflection.cpp
```diff
@@ -90,39 +90,3 @@ void BraggReflection::setHKL(int h, int k, int l) { hkl_.set(h, k, l); }
...
-    // Write index, Q centre, and number of contributing K-vectors
-    if (!parser.writeLineF("{} {} {} {} {} {}\n", index_, q_, nKVectors_, hkl_.x, hkl_.y, hkl_.z))
-        return false;
-
-    // Write intensities array
-    if (!GenericItemSerialiser::serialise<Array2D<double>>(intensities_, parser))
-        return false;
-
-    return true;
-}
```
trisyoungs: I'm afraid this is required - line 32 in `src/modules/bragg/functions.cpp` creates a `std::vector<BraggReflection>` item in the processing module data, and which is stored in the restart file.

trisyoungs (changes requested):
Looks good on the TOML, less so on the Bragg (see comment)!

trisyoungs (approved):

PR #1584, opened 2023-07-28
Q-Species module setup

Danbr4d:
Setup of a module that calculates the q-species of a network modifier


src/modules/registry.cpp
```diff
@@ -80,6 +81,7 @@ ModuleRegistry::ModuleRegistry()
     registerProducer<MDModule>(ModuleTypes::MD, "Evolve a Configuration using molecular dynamics", "Evolution");
     registerProducer<MolShakeModule>(ModuleTypes::MolShake, "Perform molecular Monte Carlo moves", "Evolution");
     registerProducer<NeutronSQModule>(ModuleTypes::NeutronSQ, "Calculate neutron-weighted S(Q)", "Correlation Functions");
+    registerProducer<QSpeciesModule>(ModuleTypes::QSpecies, "Calculate QSpecies of a network former", "Analysis");
```
trisyoungs, line 80:
```suggestion
    registerProducer<QSpeciesModule>(ModuleTypes::QSpecies, "Calculate Q species of a network former", "Analysis");
```

trisyoungs (commented):
Yeah, all looking sound so far. Just need that new node to progress things!

Danbr4d:
It would also be useful to calculate the percentage of bridging oxygens and non-bridging oxygens and free oxygens (not bonded to any network former). If we are checking what the oxygens are bonded to already how hard would this be to implement?

trisyoungs:
> It would also be useful to calculate the percentage of bridging oxygens and non-bridging oxygens and free oxygens (not bonded to any network former). If we are checking what the oxygens are bonded to already how hard would this be to implement?

That should be pretty trivial - you'd just need another `IntegerCollect1D` node binning the value of `NF.nSelected` for each oxygen, I think.

src/modules/qSpecies/qSpecies.cpp
```diff
@@ -0,0 +1,83 @@
...
+    normalisation.create<OperateNormaliseProcedureNode>({});
+
+    // Percentage of BO
+    auto percentBO = forEachBO.create<CalculateExpressionProcedureNode>({});
+    percentBO->setExpression("NF.nSelected");
+    auto collectBO = forEachBO.create<IntegerCollect1DProcedureNode>("NumberBO", percentBO, ProcedureNode::AnalysisContext);
+    auto processPercent = analyser_.createRootNode<Process1DProcedureNode>("PercentBO", collectQ);
+    auto &normalisationPercent = processNF->branch()->get();
+    normalisationPercent.create<OperateNormaliseProcedureNode>({});
+    /*
```
trisyoungs, line 60:
```suggestion

    /*
```

src/modules/qSpecies/qSpecies.cpp
```diff
@@ -0,0 +1,83 @@
...
+    // Get site index of NF
+    auto siteIndex = ifThen.create<CalculateExpressionProcedureNode>({});
+    siteIndex->setExpression("NF.siteIndex");
+    auto collectNFIndex = forEachNF.create<IntegerCollect1DProcedureNode>("NFBins", siteIndex, ProcedureNode::AnalysisContext);
+
+    // Get number of BO within NF
+    auto qIterator = analyser_.createRootNode<IterateData1DProcedureNode>("QIterator", collectNFIndex);
+    auto &forEachQIterator = qIterator->branch()->get();
+    auto value = forEachQIterator.create<CalculateExpressionProcedureNode>({});
+    value->setExpression("QIterator.value");
```
trisyoungs: We're grabbing the `value` here (which corresponds to the population of BO for each NF) but aren't binning it anywhere...

src/modules/qSpecies/qSpecies.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    // Collect and process BO
+    auto collectQ = forEachNF.create<IntegerCollect1DProcedureNode>("Bins", value, ProcedureNode::AnalysisContext);
```
trisyoungs: Ah - should this be in the bit above?
```suggestion
    auto collectQ = forEachQIterator.create<IntegerCollect1DProcedureNode>("Bins", value, ProcedureNode::AnalysisContext);
```

src/modules/qSpecies/qSpecies.cpp
```diff
@@ -0,0 +1,83 @@
...
+    auto &forEachNF = selectNF_->branch()->get();
+    auto valuesInRange = forEachNF.create<IfValueInRangeProcedureNode>({});
+    valuesInRange->keywords().set("Value", NodeValueProxy("NF.nSelected"));
+    valuesInRange->keywords().set("ValidRange", Range(1.9, 2.1));
+    auto &ifThen = valuesInRange->branch()->get();
+
+    // Get site index of NF
+    auto siteIndex = ifThen.create<CalculateExpressionProcedureNode>({});
+    siteIndex->setExpression("NF.siteIndex");
+    auto collectNFIndex = forEachNF.create<IntegerCollect1DProcedureNode>("NFBins", siteIndex, ProcedureNode::AnalysisContext);
```
trisyoungs: This should be created within the `ifThen` branch, no?
```suggestion
    auto collectNFIndex = ifThen.create<IntegerCollect1DProcedureNode>("NFBins", siteIndex, ProcedureNode::AnalysisContext);
```

src/modules/qSpecies/gui/qSpeciesWidgetFuncs.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+    refreshing_ = false;
+}
+
+// Update controls within widget
+void QSpeciesModuleWidget::updateControls(const Flags<ModuleWidget::UpdateFlags> &updateFlags)
+{
+
+    ui_.RegionAResultFrame->setText(
+        dissolve_.processingModuleData().valueOr("{}//Process1D//PercentBO", module_->name(), SampledDouble()));
```
trisyoungs: The `PercentBO` output is actually another `Data1D`, so you can't grab it as a `SampledDouble`.  Another plot widget will be needed if you want to display it.

trisyoungs (commented):
Looking all good.  There's just one thing stopping this from being merged in, and that's the display of the PercentBO data (see comment).

As a follow-up issue we should handle the proper consideration of all Q0 species, and add a real-world unit test using a single configuration from one of your phosphate glasses. I'll bung that in an issue, and we can work that up together once this is in.

src/modules/qSpecies/gui/qSpeciesWidgetFuncs.cpp
Danbr4d: So this will have to be another graph rather than a output at the bottom of the module display?

Danbr4d: After actually thinking about this, this can be caught in the new issue when we can differentiate between Q0 and free oxygens, I think It'll be more useful then

trisyoungs (approved):

PR #1587, opened 2023-07-28
The Great Big Icon Update 2023

trisyoungs:
This PR started off as a simple reorganisation to tidy the `src/gui/icons` directory, but turned into something draining, annoying, and tedious.... but ultimately highly worthwhile. Very little has changed programatically in the code - this is mostly just cosmetic.

Tasks performed:
- Reorganisation of the main `src/gui/icons` dir to remove the myriad name prefixes (e.g. `module_*`, `viewer_*`, etc.) - node and module icons are now in separate subdirectories, and all remaining prefixes have been removed.
- All filenames have been lowerCamelCased in line with the rest of the codebase - this necessitated some special functions to return the right filenames to look for when associating icons to nodes and modules.
- Around 95% of the icons have been resized to maximally fill their 100x100 px svg canvas, and icon sizes in the GUI have been reduced where necessary. On the whole, though, IMHO the GUI looks better for having slightly larger icons. 
- A bunch of missing icons have been added.
- Some icons have been redone entirely, or recoloured to make associations clearer (e.g. nodes related to 1D, 2D, and 3D data capture and processing)
- Duplicate icons have been removed, and some icons have been renamed to be more general.
- A handful of defunct or unused icons have been removed.


src/procedure/nodes/node.cpp
```diff
@@ -74,6 +74,14 @@ EnumOptions<ProcedureNode::NodeType> ProcedureNode::nodeTypes()
...
 }
 
+// Return the lowerCamelCase name of the node type provided
+std::string ProcedureNode::lccNodeType(NodeType nodeType)
+{
+    auto lccNodeName = ProcedureNode::nodeTypes().keyword(nodeType);
+    lccNodeName.front() = tolower(lccNodeName.front());
+    return lccNodeName;
+}
+
```
rprospero, line 84: There's also a bit in the TOML code that converts strings to lower camel case.  Maybe not for this PR, but we should but in an issue for factoring out that utility function.

rprospero (approved):

jws-1 (commented):
I'm really liking the look of things!

Couple of things to double check that you are aware of before we merge this in;
- The `IntraAngle` module doesn't yet have an icon.
- I still have the problem with the density/rho symbol (not sure how I hadn't noticed that it wasn't fixed before) - although it might be specific to me? See below:
image
- In the `NodeControlWidget`s, I feel like the icon looks like it's overlapping the border and could do with being slightly reduced in size, in my opinion! See below:
image
- There are still some icons that don't use lcc naming, e.g. `axes_on.svg`.

src/procedure/nodes/node.cpp
trisyoungs: Yep. Captured in #1603.

trisyoungs:
@jws-1 Curses to that density icon!  I absolutely changed it at some point, but it has since reverted back to the original "proper text".

To address your full list of stuff:
- Yeah, I had intended to add an icon for `IntraAngle` but then realised I wanted to have a bit of a go at all of the module icons. So that will probably come in a second (but much smaller) icon-related PR.
- Density icon is definitely fixed this time.
- Yes, agreed. Have made the Node control widget icon smaller.
- Weeeeeelll, I would tend to draw an analogy here with those classes which have their functions split up into separate files. Not all icons have an on/off or alternative state, so the underscore helps delineate the name from the "option".

jws-1 (approved):

rhinoella (approved):
Icons look really good!

PR #1588, opened 2023-07-31
Fix Rotate Fragment Unit Test

trisyoungs:
This PR addresses an issue with the rotate fragment unit test hanging - a missing `&` is the main cause! The test now runs, but fails, so needs a little bit of further work...


unit/procedure/rotateFragment.cpp
```diff
@@ -25,45 +26,67 @@ TEST(RotateTest, Benzene)
...
     auto select = procedure.createRootNode<SelectProcedureNode>("BenzeneSite", sites);
-    auto rotate = procedure.createRootNode<RotateFragmentProcedureNode>("RotateBenzene", select);
+    // Rotate Benzene around the Z-Axis of the site (which is defined at the COG, which is at the origin)
+    auto &forEachB = select->branch()->get();
+    auto rotate = forEachB.create<RotateFragmentProcedureNode>("RotateBenzene", select);
+    rotate->keywords().setEnumeration("Axis", OrientedSite::SiteAxis::ZAxis);
 
-    // Rotate four times around the x-axis of the site
-    for (auto x = 90.0; x <= 360.0; x += 90.0)
+    for (auto x = 180.0; x <= 360.0; x += 180.0)
```
trisyoungs: Does this loop, which tests only two rotation values, seem a little "coarse"? Would going from 90 to 360 in steps of 90 provide a more thorough test?

unit/procedure/rotateFragment.cpp
```diff
@@ -25,45 +26,67 @@ TEST(RotateTest, Benzene)
...
+        // Rotate by 'x' degrees
         rotate->keywords().set("Rotation", NodeValueProxy(x));
+
+        // Re-generate the configuration
         cfg->generate({ProcessPool(), dissolve});
 
-        for (const auto &[mol1, mol2] : zip(molecules_before, cfg->molecules()))
+        // Move the molecule to the origin
+        mol = cfg->molecule(0);
+        mol->setCentreOfGeometry(cfg->box(), {0., 0., 0.});
```
trisyoungs: Also, I think we should update the `benzeneSpecies()` to be properly centred at zero in the first place - there's no good reason why it shouldn't be, and it will simplify things here.

unit/procedure/rotateFragment.cpp
jws-1: I suppose it would!

jws-1: Sorted.

unit/procedure/rotateFragment.cpp
jws-1: Hmm. After double-checking, the `benzeneSpecies()` is already centred at the origin... There's something else going on here, I think!

jws-1 (approved):

PR #1590, opened 2023-07-31
IfValueInRange Node

trisyoungs:
This PR introduces a new node in what is likely to become a series of useful "logic" additions to procedure writing in Dissolve.  The `IfValueInRange` node checks a given value (as a `NodeValue` expression) against a defined `Range`, only executing its "Then" branch if the value is within the specified limits.

Some other notable changes were made in order to allow this PR - principally, a sequence node can inherit the context from the encompassing (owning) parent sequence.  This is useful / necessary when introducing fully generic nodes like this one, and was technically already a feature in the code but it turned out to be horribly broken!

### DIFF
--- a/src/procedure/nodes/ifValueInRange.cpp
+++ b/src/procedure/nodes/ifValueInRange.cpp
@@ -12,7 +12,7 @@
 
 IfValueInRangeProcedureNode::IfValueInRangeProcedureNode()
     : ProcedureNode(ProcedureNode::NodeType::IfValueInRange, {NodeContext::AnyContext}),
-      thenBranch_(NodeContext::ParentProcedureContext, *this, "Then")
+      thenBranch_(NodeContext::InheritContext, *this, "Then")
 {
     keywords_.setOrganisation("Options", "Condition");
     keywords_.add<NodeValueKeyword>("Value", "Value expression to test", value_, this);
--- a/src/procedure/nodes/ifValueInRange.cpp
+++ b/src/procedure/nodes/ifValueInRange.cpp
@@ -72,8 +72,7 @@ bool IfValueInRangeProcedureNode::execute(const ProcedureContext &procedureConte
 {
     // Evaluate and check
     if (valueRange_.contains(value_.asDouble()))
-        if (!thenBranch_.execute(procedureContext))
-            return false;
+        return thenBranch_.execute(procedureContext);
 
     return true;
 }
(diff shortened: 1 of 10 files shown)


src/procedure/nodes/ifValueInRange.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return thenBranch_.prepare(procedureContext);
+}
+
+// Execute node
+bool IfValueInRangeProcedureNode::execute(const ProcedureContext &procedureContext)
+{
+    // Evaluate and check
+    if (valueRange_.contains(value_.asDouble()))
+        if (!thenBranch_.execute(procedureContext))
+            return false;
```
jws-1:
```suggestion
        return thenBranch_.execute(procedureContext);
```

src/procedure/nodes/node.cpp
```diff
@@ -101,25 +102,17 @@ ProcedureNode::NodeType ProcedureNode::type() const { return type_; }
 // Return whether the supplied context is relevant for the current node
 bool ProcedureNode::isContextRelevant(NodeContext targetContext) const
 {
-    // If the node is suitable in Any context return immediately
-    if (std::find(relevantContexts_.begin(), relevantContexts_.end(), ProcedureNode::NodeContext::AnyContext) !=
-        relevantContexts_.end())
+    // Check for spurious context checks
+    if (targetContext == ProcedureNode::NodeContext::InheritContext)
+        throw(std::runtime_error(fmt::format("Attempted to context check node '{}' with an InheritedContext.\n")));
```
jws-1:
```suggestion
        throw(std::runtime_error(fmt::format("Attempted to context check node '{}' with an InheritedContext.\n", !name_.empty() ? name_ : nodeTypes().keyword(type_))));
```

Missing argument here. Although I'm not 100% convinced by this check, since when I add an `IfValueInRange` node and then try and add another node in the branch, this gets thrown.

jws-1 (changes requested):
Can't add a node to the branch as a `std::runtime_error` is thrown (see above comment)!

src/procedure/nodes/node.cpp
trisyoungs: Interesting.   I was only testing via an input file, not the GUi, so obviously missed something here!

jws-1 (approved):

PR #1591, opened 2023-07-31
Fixup HistogramCN

trisyoungs:
This PR focused on addressing some reports of output from `HistogramCN` "not being visible", or "not working properly". The observed behaviour was that the output graph would be blank, with coordinates showing as `-nan,-nan`. Inspecting the restart file did indeed show the presence of `-nan`s in the generated `Data1D`.

The root cause of this seemed to be related to our limit determination for the `Data1D` which did not always result in an `averages_` `Data1D` consistent with the `raw_` vector.  Removing the "assumed" limits passed to the `IntegerCollect1D` node and tidying up our determination of the data limits seems to resolve these issues.

### DIFF
--- a/src/math/integerHistogram1D.cpp
+++ b/src/math/integerHistogram1D.cpp
@@ -48,40 +48,26 @@ void IntegerHistogram1D::updateAccumulatedData()
     }
 }
 
-// Create dsiplay data object covering extents of current bins
+// Create display data object covering extents of current bins
 const std::pair<Data1D, int> IntegerHistogram1D::createDisplayData()
 {
+    // If we haven't binned anything yet, return now
+    if (raw_.empty())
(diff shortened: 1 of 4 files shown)


Danbr4d:
Would this be a good pull request to also add the ability to export the data from HistogramCN as that can't be currently done. I've also noticed that a lot of histograms of coordination numbers use bar charts, I'm not sure if this is something we could implement of if we should just let users output the data to plot how they want?

trisyoungs:
@Danbr4d Yep, that sounds sensible. If you want to you can add that and push it on to this PR. Otherwise I'll do it when I'm back in circulation.

Danbr4d (approved):
Looks good to me, added in the export and tested it too and seems to work fine.

PR #1593, opened 2023-08-01
Display Global Potentials

### LINKED ISSUE #1527: Configuration Potentials
It would be nice to have an additional "tab" on the Configuration page, which gives an overview of the global potentials that have been applied to the current Configuration.

jws-1:
This PR adds a new "sidebar tab" to the Configuration tab, which displays the applied global potentials, and their parameters.
This does require a new icon to be added, and therefore steps on the toes of #1587...

Closes #1527.


src/gui/models/globalPotentialModel.cpp
```diff
@@ -0,0 +1,85 @@
+#include "gui/models/globalPotentialModel.h"
+#include "kernels/potentials/directional.h"
+GlobalPotentialModel::GlobalPotentialModel(const std::vector<std::unique_ptr<ExternalPotential>> &globalPotentials)
```
trisyoungs: Wooop, wooop! That's the sound of the (blank line) police!
```suggestion

GlobalPotentialModel::GlobalPotentialModel(const std::vector<std::unique_ptr<ExternalPotential>> &globalPotentials)
```

trisyoungs (commented):
Looks good and works as it should.  My only real comment is on the choice to have a specific model for global potentials, with the presumption that one for targeted potentials will be added later.  The only difference between the two types is that one has target atoms (indices) and one does not. So, its probably neater to have a single `ExternalPotentialModel` to handle both, but use a filter proxy to hide the target column when the global potentials are the subjects for display.

src/gui/models/globalPotentialModel.cpp
jws-1: Agh!

jws-1:
OK - so now this shows the `Targeted Potentials` also. However, in the case of the `RestraintPotentialProcedureNode`, it currently only sets the target atom indices in the potential, and nothing else. So for now, the `Targets` column simply shows them indices. Also, it adds a targeted potential for each individual index. I think it'd be better if a single potential was added, but it might be best to leave this for another PR.

src/gui/models/externalPotentialModel.cpp
```diff
@@ -0,0 +1,90 @@
...
+                return QString::fromStdString(keyword(p->type()));
+            // Functional form
+            case (1):
+                return QString::fromStdString(p->formString());
+            // Parameters
+            case (2):
+                return QString::fromStdString(p->formParametersString());
+            // Target atom indices
+            case (3):
+                return QString::fromStdString(joinStrings(p->targetAtomIndices(), " "));
```
trisyoungs: The atom indices displayed need to be human-compatible (1...N) rather than zero indexed, so here you'll have to use the lambda variant of the `joinStrings` function.

trisyoungs (approved):
Looking good!  Just need to correct the display of the atom indices and we're good to go.

PR #1595, opened 2023-08-01
Run Layer Node

jws-1:
This PR achieves two things:
- A reference to `Dissolve` can be (optionally) added to a `ProcedureContext`.
- Implements a `RunLayer` node which is analogous to the "execution" part of the simulation hot-loop.

### DIFF
--- a/src/classes/configuration.cpp
+++ b/src/classes/configuration.cpp
@@ -76,7 +76,7 @@ bool Configuration::generate(const ProcedureContext &procedureContext)
     Messenger::print("\n");
 
     // Set-up Cells for the Box
-    cells_.generate(box_.get(), requestedCellDivisionLength_, context.dissolve().potentialMap().range());
+    cells_.generate(box_.get(), requestedCellDivisionLength_, context.potentialMap().range());
 
     // Make sure all objects know about each other
     updateObjectRelationships();
--- a/src/classes/configuration.cpp
+++ b/src/classes/configuration.cpp
@@ -10,6 +10,7 @@
 #include "classes/potentialMap.h"
 #include "classes/species.h"
 #include "modules/energy/energy.h"
+#include "main/dissolve.h"
 
 Configuration::Configuration() : generator_(ProcedureNode::GenerationContext, "Generator")
 {
@@ -75,7 +76,7 @@ bool Configuration::generate(const ProcedureContext &procedureContext)
     Messenger::print("\n");
 
     // Set-up Cells for the Box
-    cells_.generate(box_.get(), requestedCellDivisionLength_, context.potentialMap().range());
+    cells_.generate(box_.get(), requestedCellDivisionLength_, context.dissolve().potentialMap().range());
 
     // Make sure all objects know about each other
     updateObjectRelationships();
--- a/src/classes/configuration.cpp
+++ b/src/classes/configuration.cpp
@@ -9,8 +9,8 @@
 #include "classes/cell.h"
 #include "classes/potentialMap.h"
 #include "classes/species.h"
-#include "modules/energy/energy.h"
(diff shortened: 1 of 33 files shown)


trisyoungs (commented):
As discussed, I think it's important to implement the `RunLayerNode` much as you have done here in order to reproduce the standard workflow in the `1.X` series.  My feeling is that it will probably set the path for being eventually mutated into a `RunTaskNode` or similar later down the line.

src/procedure/nodes/runLayer.h
```diff
@@ -0,0 +1,33 @@
...
+
+    /*
+     * Identity
+     */
+    public:
+    // Return whether a name for the node must be provided
+    bool mustBeNamed() const override;
+
+    private:
+    const ModuleLayer *layer_{nullptr};
```
trisyoungs, line 24:
```suggestion
    // Target module layer to run
    const ModuleLayer *layer_{nullptr};
```

src/procedure/nodes/node.h
```diff
@@ -96,7 +97,8 @@ class ProcedureNode : public std::enable_shared_from_this<ProcedureNode>, public
         GenerationContext = 2,
         OperateContext = 4,
         AnyContext = 8,
-        InheritContext = 16
+        InheritContext = 16,
+        ControlContext = 32
```
trisyoungs: Because `ControlContext` is a "proper" context I think it should go after `OperateContext` rather than at the end.  As an aside, I'm inclined to say that we should use a `Flags` object to capture multiple contexts, rather than using the good old power of two magic numbers. But that's a separate PR for sure!

src/procedure/nodes/context.h
```diff
@@ -39,14 +38,16 @@ class ProcedureContext
     void setConfiguration(Configuration *cfg);
     // Return target Configuration
     Configuration *configuration() const;
-    // Set target data list and prefix
-    void setDataListAndPrefix(GenericList &list, std::string_view prefix);
+    // Set prefix
+    void setPrefix(std::string_view prefix);
```
trisyoungs: With the removal of the `GenericList` argument the intention of this function becomes a little vague - I'd suggest we rename it:
```suggestion
    // Set prefix for generated processing data
    void setProcessingDataPrefix(std::string_view prefix);
```

src/modules/intraAngle/process.cpp
```diff
@@ -28,9 +28,11 @@ Module::ExecutionResult IntraAngleModule::process(ModuleContext &moduleContext)
     calculateAngle_->keywords().set("Symmetric", symmetric_);
     collectABC_->keywords().set("RangeX", angleRange_);
 
+    // Execute the analysis
```
trisyoungs, line 31: Already got one! :)
```suggestion
```

src/classes/configuration.cpp
```diff
@@ -75,7 +76,7 @@ bool Configuration::generate(const ProcedureContext &procedureContext)
     Messenger::print("\n");
 
     // Set-up Cells for the Box
-    cells_.generate(box_.get(), requestedCellDivisionLength_, context.potentialMap().range());
+    cells_.generate(box_.get(), requestedCellDivisionLength_, context.dissolve().potentialMap().range());
```
trisyoungs: Not sure why this was changed, since you still have the `ProcedureContext::potentialMap()` function...

trisyoungs (approved):
Taking a `Dissolve&` actually cleans things up a bit which is nice.  Couple of small suggestions, then good to go.

src/classes/configuration.cpp
jws-1: Oops.. I had removed that function, and then changed my mind back..

PR #1596, opened 2023-08-01
Module contexts

jws-1:
This PR adds a `ModuleContext` associated with `Modules`, akin to the system in place for `ProcedureNodes`. This currently only contains a `ProcessPool` and an optional reference to `Dissolve`. This is motivated by #88.


trisyoungs (approved):
:+1: Looks good.

PR #1598, opened 2023-08-01
1597 Broken SizeFactor Reduction

### LINKED ISSUE #1597: SizeFactor does not reduce
**Describe the bug**
In the very latest development versions the applied size factor never reduces. In fact, the box gets scaled out of all proportion!

trisyoungs:
A two-liner to fix size factor reduction, which we broke in a previous PR.

Closes #1597.


jws-1 (approved):
Oops...

PR #1602, opened 2023-08-02
1601 EReq Never Changes

### LINKED ISSUE #1601: (EPSR) EReq never changes?
**Describe the bug**
In the current 1.4.0 version the `EPSR` module appears to run just fine but does not generate any empirical potential (ereq remains at zero).

trisyoungs:
This PR fixes an issue whereby the `EPSR` module never generated an ereq value, and so never introduced any potential and did any fitting.

This bug was introduced in #1569 - a rather important line was inadvertently deleted in the `PoissonFit` class, meaning that the fit function was never generated in the first place!

I've also made a little cosmetic change to the code to tidy up long chains of objects (i.e. `dissolve.coreData().atomTypes()`)

Closes #1601.


Danbr4d (approved):
Just pulled and checked and my EReq now increases with what I would expect

Issue #1576, opened 2023-07-26
Testing CIF import


trisyoungs:
**Describe the feature / issue**
Testing of CIF import is restricted to a simple parsing of CIF files as present.  With the ability to generate molecular CIFs now forthcoming, it is important to test the code behind that workflow thoroughly.  This likely necessitates a refactoring of the code to more clearly separate the GUI wizard from the CIF (classes) in order that the functionality of the latter can be tested properly.

PR #1607, opened 2023-08-03
Refactor CIF Import

jws-1:
This PR refactors the CIF importing, moving the bulk of the code from the dialog into `cifClasses`. A single class has been added `CIFSpecies` that now does the work.

I think this is moving in the right direction, but further work likely needs to be done - so a second pair of eyes would be useful!

Works towards #1576.


src/gui/importCIFDialogFuncs.cpp
```diff
@@ -618,47 +392,17 @@ bool ImportCIFDialog::detectUniqueSpecies()
...
-                    supercell->addAtom(i.Z(), i.r() + deltaR, 0.0, i.atomType());
-            }
-
-    // Bonding
-    if (ui_.CalculateBondingRadio->isChecked())
-        supercell->addMissingBonds();
-    else
-        applyCIFBonding(supercell);
+    cifSpecies_->createSupercellSpecies(repeat, ui_.CalculateBondingRadio->isChecked(),
+                                        ui_.BondingPreventMetallicCheck->isChecked());
```
trisyoungs: This feels like a place where we should move to a `Flags` object in order to pass various options into our functions.

src/gui/importCIFDialogFuncs.cpp
```diff
@@ -47,45 +47,14 @@ ImportCIFDialog::ImportCIFDialog(QWidget *parent, Dissolve &dissolve)
...
-            continue;
-
-        // Retrieve distance
-        auto r = cifImporter_.bondDistance(i.atomType()->name(), j.atomType()->name());
-        if (!r)
-            continue;
-        else if (fabs(box->minimumDistance(i.r(), j.r()) - r.value()) < 1.0e-2)
-            sp->addBond(&i, &j);
-    }
+    cifSpecies_ = new CIFSpecies(cifImporter_, temporaryCoreData_);
```
trisyoungs: There's probably a good reason, but why `new` this instead of just having it as a full member variable?

trisyoungs (commented):
First off, I really like the way this cleans up the wizard code and collects all of the CIF stuff together - clear separation, like it a lot.

I don't think `CIFSpecies` is an appropriate class name any longer since it does an awful lot more than that.  As a related broad comment I wonder about whether the functionality of `CIFSpecies` should be absorbed into `CIFImporter`, and that class renamed to something like `CIFHandler`? Having the import functionality as a separate thing doesn't offer a lot.

My final comment relates to unit testing, which is now one thing we'll be able to do once this refactor is complete.  I think it would be most wise to include the testing in this PR - I'm happy to work on that once the present changes have settled.

jws-1 (commented):
I like where things are going - I'm quite pleased with our new interface!

src/io/import/cif.h
```diff
@@ -79,4 +91,105 @@ class CIFImport
...
+    enum UpdateFlags
+    {
+        CleanMoietyRemoveAtomics,  /* Remove atoms of single moiety */
+        CleanMoietyRemoveWater,    /* Remove water molecules of single moiety  */
+        CleanMoietyRemoveNETA,     /* Remove single atoms by NETA definition  */
+        CleanRemoveBoundFragments, /* Remove entire fragments when using NETA definition  */
+        CalculateBonding,          /* Calculate bonding */
+        PreventMetallicBonding,    /* Prevent metallic bonding */
+        CreateSupermolecule
+    };
```
rprospero, line 109: This flag enum makes me unreasonably happy

src/io/import/cif.cpp
```diff
@@ -376,3 +402,569 @@ std::optional<double> CIFImport::bondDistance(std::string_view labelI, std::stri
...
+
+        // Set up the CIF species
+        auto *sp = coreData_.addSpecies();
+        sp->copyBasic(cleanedUnitCellSpecies_);
+
+        // Empty the species of all atoms, except those in the reference instance
+        std::vector<int> allIndices(sp->nAtoms());
+        std::iota(allIndices.begin(), allIndices.end(), 0);
+        std::vector<int> indicesToRemove;
+        std::set_difference(allIndices.begin(), allIndices.end(), fragment.begin(), fragment.end(),
```
rprospero, line 601: I'd never seen `set_difference` before, but it's the perfect tool for the job.

rprospero (approved):
This came together beautifully.  The only comments I have are praise.

jws-1:
Also happy with this!

PR #1608, opened 2023-08-03
Select Better

trisyoungs:
Here we focus primarily on the operation of the `SelectProcedureNode`, tidying a few things, using references where possible, and adding in some additional parameters. Those parameters are:

- `index` - the index of the site in the `ForEach` loop.
- `stackIndex` - the index of the current site in the original `SiteStack` from which it was selected.
- `siteIndex` - the index of the current site over all `SiteStack`s from all specified `SpeciesSite`s

Thus, if there are _N_ sites of a given type, the `index` for a given site will change based on how many of the _N_ are selected according to other criteria (i.e. within a certain distance, not on the same molecule etc.) while a given site's `stackIndex` will remain the same regardless of how many are selected, provided _N_ does not change (i.e. population of molecules in the `Configuration` is static).

Similarly, the `siteIndex` remains constant for a given site regardless of how many species sites are requested for selection. So, if three sites are provided having _N_, _M_, and _O_ sites respectively, then the `siteIndex` number goes from `1...(N+M+O)`, with sites from _N_ always being numbered `1...N`, sites from _M_ being numbered `N+1...N+M`, and sites from _O_ being numbered `N+M+1...N+M+O`.

### DIFF
--- a/src/procedure/nodes/select.cpp
+++ b/src/procedure/nodes/select.cpp
@@ -269,7 +269,7 @@ bool SelectProcedureNode::execute(const ProcedureContext &procedureContext)
                     continue;
             }
 
-            // All OK, so add site and index in its stack (1...N numbering)
+            // All OK, so add site, its global index among all site stacks, and index in its stack (1...N numbering)
             sites_.emplace_back(site, siteIndex, n + 1);
         }
     }
@@ -286,7 +286,7 @@ bool SelectProcedureNode::execute(const ProcedureContext &procedureContext)
         {
             currentSite_ = std::get<0>(siteInfo);
             siteIndexParameter_->setValue(std::get<1>(siteInfo));
-            stackIndexParameter_->setValue(std::get<1>(siteInfo));
+            stackIndexParameter_->setValue(std::get<2>(siteInfo));
             indexParameter_->setValue(index++);
 
             ++nCumulativeSites_;
--- a/src/procedure/nodes/select.cpp
+++ b/src/procedure/nodes/select.cpp
@@ -159,6 +159,9 @@ EnumOptions<SelectProcedureNode::SelectionPopulation> SelectProcedureNode::selec
                                 {SelectProcedureNode::SelectionPopulation::Average, "Average"}});
 }
 
+// Return vector of selected sites
+const std::vector<std::tuple<const Site &, int, int>> &SelectProcedureNode::sites() const { return sites_; }
+
 // Return the number of available sites in the current stack, if any
 int SelectProcedureNode::nSitesInStack() const { return sites_.size(); }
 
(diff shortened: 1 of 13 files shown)


src/procedure/nodes/sequence.cpp
```diff
@@ -348,12 +348,9 @@ std::vector<std::shared_ptr<ExpressionVariable>> ProcedureNodeSequence::paramete
...
     if (owner_)
     {
-        auto optOtherParams = owner_->get().parameters();
-        if (optOtherParams)
-        {
-            const std::vector<std::shared_ptr<ExpressionVariable>> otherParams = (*optOtherParams);
-            parameters.insert(parameters.end(), otherParams.begin(), otherParams.end());
-        }
+        auto otherParams = owner_->get().getParameters();
+
```
jws-1:
```suggestion
```

src/procedure/nodes/select.cpp
```diff
@@ -264,25 +260,31 @@ bool SelectProcedureNode::execute(const ProcedureContext &procedureContext)
...
+    currentSite_ = std::nullopt;
     if (!forEachBranch_.empty())
     {
-        for (currentSiteIndex_ = 0; currentSiteIndex_ < sites_.size(); ++currentSiteIndex_)
+        auto index = 1;
+        for (const auto &siteInfo : sites_)
         {
+            currentSite_ = std::get<0>(siteInfo);
+            siteIndexParameter_->setValue(std::get<1>(siteInfo));
+            stackIndexParameter_->setValue(std::get<1>(siteInfo));
```
jws-1: Should this be `2` instead?
```suggestion
            stackIndexParameter_->setValue(std::get<2>(siteInfo));
```

src/procedure/nodes/select.cpp
```diff
@@ -264,25 +260,31 @@ bool SelectProcedureNode::execute(const ProcedureContext &procedureContext)
...
-                r = procedureContext.configuration()->box()->minimumDistance(site.origin(), distanceRef->origin());
-                if (!inclusiveDistanceRange_.contains(r))
+                if (!inclusiveDistanceRange_.contains(
+                        procedureContext.configuration()->box()->minimumDistance(site.origin(), distanceRef->get().origin())))
                     continue;
             }
 
-            // All OK, so add site
-            sites_.push_back(&site);
+            // All OK, so add site and index in its stack (1...N numbering)
```
jws-1: Maybe not this exact wording, but...
```suggestion
            // All OK, so add site, its global index among all site stacks, and index in its stack (1...N numbering)
```

jws-1 (approved):
Mostly looks good to me, just a couple of minor comments to ponder on. I also wasn't sure how to functionally test this, but no previously existing functionality seems to have been broken.

src/procedure/nodes/select.cpp
trisyoungs: Oops!

src/procedure/nodes/sequence.cpp
trisyoungs: Touché.

jws-1 (approved):
Looks good and the new tests pass, so I'm approving this now.

PR #1609, opened 2023-08-03
Collect parameters when deserialising TOML node

rprospero:
The main thrust of this PR is to ensure that, when a NodeValue is deserialised, it has access to any variables that might be called in the NodeValue.  This involves pulling the parameters from the parent node.

This does lead to a land mine where `Vec3<NodeValue>.deserialise(node)` will throw an exception, since we cannot pass in the parameters.  The correct behaviour is to call `Vec3<NodeValue>.deserialise(node, params)` and the current solution is simply not to call the single argument version of the function.  I will add an issue about adding some more type machinery (i.e. a ```template <typename T> struct SerialisationContext<T>```) that will enable use to perform this in a type safe manner.  This may also simplify other issues later down the line.

As of this PR, every test will correctly parse, deserialise, and will correctly perform a round-trip on the resulting TOML.  The remaining issues are 15 tests where the TOML code produces a different calculated result than the original input file.


src/procedure/nodeValue.cpp
```diff
@@ -153,26 +153,23 @@ SerialisedValue NodeValue::serialise() const
...
             {
-                type_ = DoubleNodeValue;
-                valueD_ = arg;
+                set((double)arg);
             }
             else if constexpr (std::is_same_v<T, toml::string>)
             {
-                type_ = ExpressionNodeValue;
-                expression_.create(std::string_view(std::string(arg)), {});
+                set(std::string_view(std::string(arg)), params);
```
trisyoungs: Is the explicit conversion to `std::string_view` necessary here?

trisyoungs (approved):
I actually think this is a nice solution for now, but of course removing the type checking would be better in the long run.

:+1:

PR #1611, opened 2023-08-03
Fix GUI Module Contexts

jws-1:
This urgent PR fixes a bunch of instances where a reference to `Dissolve` was not passed to a `ModuleContext`, where it definitely should've been - this meant that the GUI crashes if you add a new "template" Layer, or modify a local keyword in a Module through the GUI...


rprospero (approved):

trisyoungs (approved):
:+1:

PR #1612, opened 2023-08-03
1568 EPSR DataSet Weighting

### LINKED ISSUE #1568: (EPSR) Allow weighting of datasets
Add in functionality to allow the weight of datasets (as they appear in the scattering matrix) to be modified, a la the original EPSR.

trisyoungs:
This PR implements a per-dataset weighting option into the `EPSR` module, replicating the `efilereq` weighting scheme implemented in the original EPSR. A new keyword `TargetWeight` makes use of the keyword introduced in #1563.

Here we also introduce a middle ground between unit & system tests, whereby a unit test loads a full input file and runs a "simulation" with the view to testing specific aspects of the process.

Closes #1568.


rhinoella (commented):
Looks great, just one small thing:
In the gui keyword widget, when the module is checked, the cell displaying the weight is not automatically refreshed/displayed.


Changing this line:
https://github.com/disorderedmaterials/dissolve/blob/3cf8256802e8457aeea5a6dd0fafa163cd27f64f/src/gui/models/weightedModuleModel.cpp#L114

To: 
`emit dataChanged(index, index.sibling(index.row(), 1));`

Solves this. I know this is more so a reference to #1566, but I didn't catch this until now!

trisyoungs:
It's a great catch! Changed according to your suggestion, with a slight modification to use `siblingAtColumn()` instead. Have also update the summary text for the widget to include the weight in parentheses.

rhinoella (approved):

PR #1613, opened 2023-08-03
Enforce ordering of modules in TOML

rprospero:
Just as with the nodes, the modules must be kept in an array, instead of a table, because the table will *not* maintain ordering.  This cleans up a couple of failing tests where data was referenced before it was loaded.


trisyoungs (approved):
:+1:

PR #1618, opened 2023-08-04
1617 Parallel Region Generation

### LINKED ISSUE #1617: Parallel Custom Region Generation Not Possible
**Describe the bug**
PR #1616 regresses the custom region generation to use a sequential loop over voxels rather than a parallel one, since threads attempt to set / access the local variables simultaneously leading to incorrect, unpredictable results (but not observed crashes to date).

To solve this, we clearly need to establish some sort of independence between the parallel threads. A simple solution could be to consider implementing a mechanism to creating a sort of "kernel" object for each thread containing a set of local variables and expression (if required) and / or the math to calculate validity of the voxel.

trisyoungs:
This PR enables region generation to be performed in parallel. The idea of a `VoxelKernel` is introduced - a small class which encapsulates the necessary objects and/or mathematics to be able to check voxel validity.  Essentially this just means farming out the existing member variables of the region procedure nodes, and moving the virtual `isVoxelValid()` function.

I have to say I quite enjoyed this one!  I would appreciate any and all comments on the methodology, the naming of the classes, and the introduction of a new combinable class, `CombinableFunctor` - since we don't sum or `finalise()` any data in this workflow, a new class seemed appropriate.

Closes #1617.

### DIFF
--- a/benchmark/region/region_benchmark.cpp
+++ b/benchmark/region/region_benchmark.cpp
@@ -7,6 +7,13 @@
 #include "common/problems.h"
 #include <cmath>
 
+class RandomVoxelKernel : public VoxelKernel
+{
+    public:
+    // Return whether voxel centred at supplied real coordinates is valid
+    bool isVoxelValid(const Configuration *cfg, const Vec3<double> &r) const override { return DissolveMath::random() > 0.5; }
+};
+
 template <ProblemType problem, Population population> static void BM_Region_Generate(benchmark::State &state)
 {
     Problem<problem, population> problemDef;
@@ -14,7 +21,7 @@ template <ProblemType problem, Population population> static void BM_Region_Gene
     for (auto _ : state)
     {
         Region region;
-        region.generate(cfg, 1.0, [&](const Configuration *cfg, Vec3<double> r) { return DissolveMath::random() > 0.5; });
+        region.generate(cfg, 1.0, []() { return std::make_shared<RandomVoxelKernel>(); });
     }
 }
 
--- a/src/classes/region.h
+++ b/src/classes/region.h
@@ -10,10 +10,19 @@
 #include "templates/array3D.h"
 #include "templates/parallelDefs.h"
 #include "templates/vector3.h"
+#include <functional>
 #include <vector>
 
 // Forward Declarations
 class Configuration;
+
+// Voxel Kernel
+class VoxelKernel
+{
+    public:
+    // Return whether voxel centred at supplied real coordinates is valid
+    virtual bool isVoxelValid(const Configuration *cfg, const Vec3<double> &r) const = 0;
+};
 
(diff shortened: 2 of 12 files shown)


benchmark/region/region_benchmark.cpp
```diff
@@ -7,14 +7,21 @@
 #include "common/problems.h"
 #include <cmath>
 
+class RandomVoxelKernel : public VoxelKernel
+{
+    public:
+    // Return whether voxel centred at supplied real coordinates is valid
+    bool isVoxelValid(const Configuration *cfg, const Vec3<double> &r) const override { return DissolveMath::random() > 0.5; }
+};
```
rprospero, line 15: Do we know that our random number generator can be called in parallel?  Most random number generators contain a signficant amount of state and I've encountered issues in past projects with that state being shared between all the threads.  This has both lead to each thread getting the exact same random numbers and to the state getting corrupted by two threads writing to it at the same time, causing corruption.

Granted, it doesn't matter that much, since this is just for the benchmark.

src/classes/region.h
```diff
@@ -8,13 +8,23 @@
...
 class Configuration;
 
+// Voxel Kernel
+class VoxelKernel
+{
+    public:
+    // Return whether voxel centred at supplied real coordinates is valid
+    virtual bool isVoxelValid(const Configuration *cfg, const Vec3<double> &r) const = 0;
+};
+
```
rprospero, line 27: I really like putting in a name for this, instead of a template.  Something that adds type safety *and* speeds up compile time is always a good plan.  That side, my first thought on the implementation would have been

```c++
using VoxelKernel = std::function(bool(const Configuration*, const Vec3<double>);
```

Is there a particular reason that you went with a class?  I'm not opposed to it, especially since I'm paranoid about how this will interact with future work on GPU parallelisation (I feel we may need to extend the class with a second method when that happens).

rprospero (approved):
I'm happy with how things look, thought I did have a couple of non-blocking concerns.

benchmark/region/region_benchmark.cpp
trisyoungs: It's a good point, and since it's the very standard library RNG, "No", it's probably not.  This does raise an interesting question on how we should handle this in a (very specific) case where a custom region uses our `Expression`'s `random()` function, but that is probably best left for another day.

src/classes/region.h
trisyoungs: Ultimately, the `class` is required to enable more complex regions (i.e. `CustomRegion`) to have somewhere to store their thread-local data (the `Expression` in that specific example).  If it were just a case of creating a "floating" `isVoxelValid()` function, then absolutely a `std::function` definition would be enough, but that of course doesn't account for the necessary thread-local data.

PR #1619, opened 2023-08-05
Warning Hunt August 2023

trisyoungs:
A simple PR to clean up a handful of warnings generated while compiling in Debug mode.

One change of note is related to TOML parsing of the `GeometryList`.


jws-1 (approved):
:+1:

rprospero:
@trisyoungs It looks like the `GeometryListKeyword` takes the type as part of the constructor, so the type is already known before we even begin deserialising the data.

My line of code from before was obviously the wrong thing to do, so I approve of the change.  However, I'm wondering if one of the following two approaches might be better:

1) Don't include the type info in the serialisation.  Since we always know the type before starting deserialisation, it's superfluous
2) Include the type info when serialising and throw a toml exception during deserialisation if the type doesn't match the expected type.

trisyoungs:
@rprospero  This is very true. In that case then we may as well just not bother serialising the type at all. Do you want to address that as a minor follow-on issue after this PR get's merged in?

PR #1622, opened 2023-08-08
iterateSelection procedure node setup

### LINKED ISSUE #1610: Iterate over existing Select sites
In some use cases we wish to `Select` a bunch of sites, but check how many we actually selected before iterating over them. There is no control over when the `ForEach` branch runs, so an additional node - perhaps `IterateSelection` - which would allow iteration of the sites selected in the `Select` node would be useful.

This work will need to be done on top of work performed in #1608.

## Work List
- [x] Create new node `IterateSelection` with a keyword to target a given `SelectProcedureNode`
- [x] Add accessor to `sites_` vector in existing `SelectProcedureNode`
- [x] Add a `ForEach` branch to the `IterateSelectionProcedureNode` with the same behaviour as that in `SelectProcedureNode`, looping over the `sites_` we get from the targeted `SelectProcedureNode` and setting the same parameters etc.

Danbr4d:
closes #1610 In some use cases we wish to Select a bunch of sites, but check how many we actually selected before iterating over them. There is no control over when the ForEach branch runs, so an additional node - perhaps IterateSelection - which would allow iteration of the sites selected in the Select node would be useful.

This work will need to be done on top of work performed in https://github.com/disorderedmaterials/dissolve/pull/1608.

### DIFF
--- a/src/procedure/nodes/iterateSelection.cpp
+++ b/src/procedure/nodes/iterateSelection.cpp
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "procedure/nodes/iterateSelection.h"
+#include "classes/configuration.h"
+#include "classes/coreData.h"
+#include "classes/siteReference.h"
+#include "classes/species.h"
+#include "expression/variable.h"
+#include "keywords/node.h"
+#include "keywords/nodeBranch.h"
+#include "keywords/nodeVector.h"
+#include "keywords/range.h"
+#include "keywords/speciesSiteVector.h"
+#include "procedure/nodes/sequence.h"
+#include <algorithm>
+
+IterateSelectionProcedureNode::IterateSelectionProcedureNode(std::vector<const SpeciesSite *> sites,
+                                                             ProcedureNode::NodeContext forEachContext)
+    : ProcedureNode(ProcedureNode::NodeType::Select, {ProcedureNode::AnalysisContext, ProcedureNode::GenerationContext}),
+      speciesSites_(std::move(sites)))
+{
+    // Keywords
+}
+
+/*
+ * Selection Targets
+ */
+
+// Return vector of sites to select
+std::vector<const SpeciesSite *> &IterateSelectionProcedureNode::speciesSites() { return speciesSites_; }
+
+/*
+ * Selected Sites
+ */
+
+/*
+ * Branch
+ */
+
+// Return the branch from this node (if it has one)
+OptionalReferenceWrapper<ProcedureNodeSequence> IterateSelectionProcedureNode::branch() { return forEachBranch_; }
+
+/*
+ * Execute
+ */
+
+// Prepare any necessary data, ready for execution
+bool IterateSelectionProcedureNode::prepare(const ProcedureContext &procedureContext)
+{
+    // Check for at least one site being defined
+    if (speciesSites_.empty())
+        return Messenger::error("No sites are defined in the Select node '{}'.\n", name());
+
+    // If one exists, prepare the ForEach branch nodes
+    if (!forEachBranch_.prepare(procedureContext))
+        return false;
+
+    return true;
+}
+
+// Execute node
+bool IterateSelectionProcedureNode::execute(const ProcedureContext &procedureContext)
+{
+    // Create our site vector
+    sites_.clear();
+
+    return true;
+}
--- a/src/procedure/nodes/iterateSelection.cpp
+++ b/src/procedure/nodes/iterateSelection.cpp
@@ -74,7 +74,24 @@ bool IterateSelectionProcedureNode::execute(const ProcedureContext &procedureCon
                 return false;
         }
     }
+    // Update parameters
+    nSelectedParameter_->setValue(int(sites.size()));
     return true;
 }
 
-bool IterateSelectionProcedureNode::finalise(const ProcedureContext &procedureContext) { return true; }
+bool IterateSelectionProcedureNode::finalise(const ProcedureContext &procedureContext)
+{
+    auto selectionSize = selection_->returnSites().size();
+    // If one exists, finalise the ForEach branch nodes
+    if (!forEachBranch_.finalise(procedureContext))
+        return false;
+
+    // Print out summary information
+    Messenger::print("Select - Site '{}': Number of selections made = {} (last contained {} sites).\n", name(), nSelections_,
+                     selectionSize);
+    Messenger::print("Select - Site '{}': Average number of sites selected per selection = {:.2f}.\n", name(),
+                     nSelections_ == 0 ? 0 : double(nCumulativeSites_) / nSelections_);
+    Messenger::print("Select - Site '{}': Cumulative number of sites selected = {}.\n", name(), nCumulativeSites_);
+
+    return true;
+}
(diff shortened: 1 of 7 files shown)


src/procedure/nodes/iterateSelection.cpp
```diff
@@ -0,0 +1,97 @@
...
+        return false;
+
+    return true;
+}
+
+// Execute node
+bool IterateSelectionProcedureNode::execute(const ProcedureContext &procedureContext)
+{
+    // If a ForEach branch has been defined, process it for each of our sites in turn. Otherwise, we're done.
+    auto sites = selection_->returnSites();
```
trisyoungs: Take a (`const`) reference here - we definitely don't want to copy it:
```suggestion
    const auto &sites = selection_->returnSites();
```

src/procedure/nodes/iterateSelection.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+            ++nCumulativeSites_;
+
+            // If the branch fails at any point, return failure here.  Otherwise, continue the loop
+            if (!forEachBranch_.execute(procedureContext))
+                return false;
+        }
+    }
+    // Update parameters
+    nSelectedParameter_->setValue(int(sites.size()));
```
trisyoungs, line 93: Bit of formatting.
```suggestion

    // Update parameters
    nSelectedParameter_->setValue(int(sites.size()));

```

src/procedure/nodes/iterateSelection.cpp
```diff
@@ -0,0 +1,97 @@
...
+    // If one exists, finalise the ForEach branch nodes
+    if (!forEachBranch_.finalise(procedureContext))
+        return false;
+
+    // Print out summary information
+    Messenger::print("Select - Site '{}': Number of selections made = {} (last contained {} sites).\n", name(), nSelections_,
+                     selectionSize);
+    Messenger::print("Select - Site '{}': Average number of sites selected per selection = {:.2f}.\n", name(),
+                     nSelections_ == 0 ? 0 : double(nCumulativeSites_) / nSelections_);
+    Messenger::print("Select - Site '{}': Cumulative number of sites selected = {}.\n", name(), nCumulativeSites_);
```
trisyoungs: Messages should reflect the name of the module, *and* be in square brackets by convention (this is of course a "bug" in the original `SelectProcedureNode`, if you'd like to correct that at the same time!):
```suggestion
    Messenger::print("[IterateSelection] - Site '{}': Number of selections made = {} (last contained {} sites).\n", name(), nSelections_,
                     selectionSize);
    Messenger::print("[IterateSelection] - Site '{}': Average number of sites selected per selection = {:.2f}.\n", name(),
                     nSelections_ == 0 ? 0 : double(nCumulativeSites_) / nSelections_);
    Messenger::print("[IterateSelection] - Site '{}': Cumulative number of sites selected = {}.\n", name(), nCumulativeSites_);
```

src/procedure/nodes/iterateSelection.h
```diff
@@ -0,0 +1,78 @@
...
+    // Selection to iterate over
+    std::shared_ptr<const SelectProcedureNode> selection_;
+
+    /*
+     * Selection Targets
+     */
+    private:
+    // Vector of sites to select
+    std::vector<const SpeciesSite *> speciesSites_;
+
```
trisyoungs: Not needed.
```suggestion
```

src/procedure/nodes/select.cpp
```diff
@@ -177,6 +177,9 @@ double SelectProcedureNode::nAvailableSitesAverage() const { return double(nAvai
 // Return current site
 OptionalReferenceWrapper<const Site> SelectProcedureNode::currentSite() const { return currentSite_; }
 
+// Return sites vector
+std::vector<std::tuple<const Site &, int, int>> SelectProcedureNode::returnSites() const { return sites_; }
+
```
trisyoungs: This can be removed since it's just a duplicate of the `site()` function above (that one also returns the `const &` that we want (i.e. read only, and we don't copy the whole thing when we ask for it).
```suggestion
```

src/procedure/nodes/select.h
```diff
@@ -131,6 +131,8 @@ class SelectProcedureNode : public ProcedureNode
     double nAvailableSitesAverage() const;
     // Return current site
     OptionalReferenceWrapper<const Site> currentSite() const;
+    // Return site vector
+    std::vector<std::tuple<const Site &, int, int>> returnSites() const;
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
Looks very good as far as I can see.  Some small formatting things and a suggestions to act on, then it can go in.

trisyoungs (changes requested):
Need to request changes, since I just realised you aren't creating the parameters in the constructor here - you can basically just copy the bits from the `SelectProcedureNode`...

trisyoungs:
And you'll also need the `setName()` overloaded function too...

src/procedure/nodes/iterateSelection.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // If one exists, finalise the ForEach branch nodes
+    if (!forEachBranch_.finalise(procedureContext))
+        return false;
+
+    // Print out summary information
+    Messenger::print("[Iterate Selection] - Site '{}': Number of selections made = {} (last contained {} sites).\n", name(),
+                     nSelections_, selectionSize);
+    Messenger::print("[Iterate Selection] - Site '{}': Average number of sites selected per selection = {:.2f}.\n", name(),
+                     nSelections_ == 0 ? 0 : double(nCumulativeSites_) / nSelections_);
+    Messenger::print("[Iterate Selection] - Site '{}': Cumulative number of sites selected = {}.\n", name(), nCumulativeSites_);
```
trisyoungs: Picky, but your node keyword is "IterateSelection", so that's what we should print out here:
```suggestion
    Messenger::print("[IterateSelection] - Site '{}': Number of selections made = {} (last contained {} sites).\n", name(),
                     nSelections_, selectionSize);
    Messenger::print("[IterateSelection] - Site '{}': Average number of sites selected per selection = {:.2f}.\n", name(),
                     nSelections_ == 0 ? 0 : double(nCumulativeSites_) / nSelections_);
    Messenger::print("[IterateSelection] - Site '{}': Cumulative number of sites selected = {}.\n", name(), nCumulativeSites_);
```

src/procedure/nodes/iterateSelection.h
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#pragma once
+
+#include "math/range.h"
+#include "procedure/nodes/node.h"
+#include "procedure/nodes/sequence.h"
+#include <memory>
+#include <set>
```
trisyoungs: Don't think this include is required.
```suggestion
```

src/procedure/nodes/iterateSelection.cpp
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+// Prepare any necessary data, ready for execution
+bool IterateSelectionProcedureNode::prepare(const ProcedureContext &procedureContext)
+{
+    // If one exists, prepare the ForEach branch nodes
+    if (!forEachBranch_.prepare(procedureContext))
+        return false;
+
+    return true;
```
trisyoungs: Can simplify this:
```suggestion
    return forEachBranch_.prepare(procedureContext);
```

trisyoungs (approved):
Couple of small bits, but otherwise looks good. :+1:

PR #1625, opened 2023-08-08
Created unit test for ranged error calculations

### LINKED ISSUE #1621: Unit test for error ranges
We should probably add a unit test to cover the `Range`-based functionality implemented in the `Errors` routines.

This could be as simple as generating a random dataset, calculating the error over the whole range, and then calculating the summed error over smaller, continuous ranges to compare against the total.  This of course relies on using an error metric which can be decomposed in this way, but we're really focussing on testing the ranges functionality.

rhinoella:
Closes #1621


trisyoungs:
@rhinoella - Hmm, one failing unit test.... :smile:

unit/math/error.cpp
```diff
@@ -0,0 +1,43 @@
...
+namespace UnitTest
+{
+
+const std::vector<Range> ranges = {Range(0, 4), Range(5, 22), Range(23, 30), Range(31, 33), Range(34, 50)};
+const Range totalRange{0, 50};
+
+Data1D basicData()
+{
+    Data1D d;
+    for (int x = totalRange.minimum(); x < totalRange.maximum(); x++)
```
trisyoungs:
```suggestion
    for (auto x = totalRange.minimum(); x < totalRange.maximum(); ++x)
```

trisyoungs: An algorithm works here but doesn't offer any further clarity to the code since you have to initialise the `Data1D` as you've written it. However, if you wanted to you could do this:
```
std::generate(d.values().begin(), d.values.end(), []() { return std::rand(); });
```

trisyoungs (approved):
:+1:

Issue #1626, opened 2023-08-09
Icon Update 1.4


trisyoungs:
**Describe the feature / issue**
A final icon update is required for the 1.4 release.

Tasks:
- [x] Harmonise module icons, in particular those for analysis modules.
- [x] Some new nodes need icons (`IfValueInRange` etc.)

PR #1627, opened 2023-08-09
1455 Orientational SDF

### LINKED ISSUE #1455: (SDF) Orientational constraints

trisyoungs:
### Builds off #1623

This PR adds in an `OrientedSDF` module, a variant of the `SDF` module which allows the selection of sites depending on the angle between specified axes.

No icon is added here, pending #1626.

Closes #1455.

### DIFF
--- a/src/modules/orientedSDF/orientedSDF.cpp
+++ b/src/modules/orientedSDF/orientedSDF.cpp
@@ -46,8 +46,7 @@ OrientedSDFModule::OrientedSDFModule() : Module(ModuleTypes::OrientedSDF), analy
 
     // -- -- -- Collect3D: 'OSDF'
     collectVector_ =
-        ifThen.create<Collect3DProcedureNode>({}, calcVector, ProcedureNode::AnalysisContext, rangeX_.x, rangeX_.y, rangeX_.z,
-                                              rangeY_.x, rangeY_.y, rangeY_.z, rangeZ_.x, rangeZ_.y, rangeZ_.z);
+        ifThen.create<Collect3DProcedureNode>({}, calcVector, ProcedureNode::AnalysisContext, rangeX_, rangeY_, rangeZ_);
 
     // Process3D
     processPosition_ = analyser_.createRootNode<Process3DProcedureNode>("SDF", collectVector_);
--- a/src/modules/orientedSDF/orientedSDF.cpp
+++ b/src/modules/orientedSDF/orientedSDF.cpp
@@ -71,7 +71,7 @@ OrientedSDFModule::OrientedSDFModule() : Module(ModuleTypes::OrientedSDF), analy
     keywords_.add<EnumOptionsKeyword<OrientedSite::SiteAxis>>("AxisA", "Axis to use from site A", calculateAxisAngle_->axis(0),
                                                               OrientedSite::siteAxis());
     keywords_.add<SpeciesSiteVectorKeyword>(
-        "SiteB", "Set the site(s) 'B' for which the distribution around the origin sites 'A' should be calculated",
+        "SiteB", "Set the site(s) 'B' for which the distribution around the origin site(s) 'A' should be calculated",
         selectB_->speciesSites(), selectB_->axesRequired());
     keywords_.add<EnumOptionsKeyword<OrientedSite::SiteAxis>>("AxisB", "Axis to use from site B", calculateAxisAngle_->axis(1),
                                                               OrientedSite::siteAxis());
--- a/src/modules/orientedSDF/orientedSDF.cpp
+++ b/src/modules/orientedSDF/orientedSDF.cpp
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "modules/orientedSDF/orientedSDF.h"
+#include "keywords/bool.h"
+#include "keywords/configuration.h"
+#include "keywords/fileAndFormat.h"
+#include "keywords/range.h"
+#include "keywords/speciesSiteVector.h"
+#include "keywords/vec3Double.h"
+#include "procedure/nodes/calculateAxisAngle.h"
+#include "procedure/nodes/calculateVector.h"
+#include "procedure/nodes/collect3D.h"
+#include "procedure/nodes/ifValueInRange.h"
+#include "procedure/nodes/operateGridNormalise.h"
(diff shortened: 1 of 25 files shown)


web/docs/userguide/modules/analysis/orientedsdf/_index.md
```diff
@@ -0,0 +1,48 @@
...
+|`RangeZ`|`double`<br/>`double`<br/>`double`| `-10.0`<br/>`10.0`<br/>`0.5` | Range (min, max, bin width) of volume over which to bin along Z.                           |
+
+### Control
+
+|Keyword|Arguments|Default| Description                                                                                                                                                                                                     |
+|:------|:--:|:-----:|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+|`ExcludeSameMolecule`|`bool`|`false`| Whether to exclude correlations between A and B sites on the same molecule.                                                                                                                                     |
+|`Symmetric`|`bool`|`false`| Whether to consider the angle as symmetric about 90, mapping all angles to the 0-90 range. For instance, if the target axes are perpendicular to a mirror plane in the mollecule, this should be set to `true`.|
```
jws-1:
```suggestion
|`Symmetric`|`bool`|`false`| Whether to consider the angle as symmetric about 90, mapping all angles to the 0-90 range. For instance, if the target axes are perpendicular to a mirror plane in the molecule, this should be set to `true`.|
```

src/modules/orientedSDF/process.cpp
```diff
@@ -0,0 +1,63 @@
...
+    collectVector_->keywords().set("RangeZ", rangeZ_);
+    if (excludeSameMolecule_)
+        selectB_->keywords().set("ExcludeSameMolecule", ConstNodeVector<SelectProcedureNode>{selectA_});
+    else
+        selectB_->keywords().set("ExcludeSameMolecule", ConstNodeVector<SelectProcedureNode>{});
+
+    // Execute the analysis
+    ProcedureContext context(moduleContext.processPool(), targetConfiguration_);
+    context.setDissolve(moduleContext.dissolve());
+    context.setProcessingDataPrefix(name());
```
jws-1, line 38: I assume this will utilise #1628?

src/modules/orientedSDF/orientedSDF.h
```diff
@@ -0,0 +1,66 @@
...
+#include "module/module.h"
+#include "procedure/procedure.h"
+
+// Forward Declarations
+class CalculateAxisAngleProcedureNode;
+class Collect3DProcedureNode;
+class IfValueInRangeProcedureNode;
+class Process3DProcedureNode;
+class SelectProcedureNode;
+class SpeciesSite;
```
jws-1:
```suggestion
```

src/modules/orientedSDF/orientedSDF.cpp
```diff
@@ -0,0 +1,98 @@
...
+    checkAxisValue_->keywords().set("Value", NodeValueProxy("ABangle.theta"));
+    auto &ifThen = checkAxisValue_->branch()->get();
+
+    // -- -- -- Calculate: 'v(B->A)'
+    auto calcVector = ifThen.create<CalculateVectorProcedureNode>({}, selectA_, selectB_, true);
+
+    // -- -- -- Collect3D: 'OSDF'
+    collectVector_ =
+        ifThen.create<Collect3DProcedureNode>({}, calcVector, ProcedureNode::AnalysisContext, rangeX_.x, rangeX_.y, rangeX_.z,
+                                              rangeY_.x, rangeY_.y, rangeY_.z, rangeZ_.x, rangeZ_.y, rangeZ_.z);
```
jws-1: Would be nicer if we could just pass three `Vec3<double>`s here, although not a necessity.

src/modules/orientedSDF/orientedSDF.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    keywords_.addTarget<ConfigurationKeyword>("Configuration", "Set target configuration for the module", targetConfiguration_);
+
+    keywords_.setOrganisation("Options", "Sites");
+    keywords_.add<SpeciesSiteVectorKeyword>("SiteA", "Set the site(s) 'A' which are to represent the origin of the SDF",
+                                            selectA_->speciesSites(), selectA_->axesRequired());
+    keywords_.add<EnumOptionsKeyword<OrientedSite::SiteAxis>>("AxisA", "Axis to use from site A", calculateAxisAngle_->axis(0),
+                                                              OrientedSite::siteAxis());
+    keywords_.add<SpeciesSiteVectorKeyword>(
+        "SiteB", "Set the site(s) 'B' for which the distribution around the origin sites 'A' should be calculated",
```
jws-1:
```suggestion
        "SiteB", "Set the site(s) 'B' for which the distribution around the origin site(s) 'A' should be calculated",
```

jws-1 (changes requested):
A few minor comments, waiting to review `orientedSDFWidgetFuncs.cpp` due to the widget not being displayed (as discussed).

src/modules/orientedSDF/process.cpp
trisyoungs: Yes - this will be retrofitted into #1628 before that PR gets merged.

src/modules/orientedSDF/orientedSDF.cpp
trisyoungs: I agree.  Not sure why we don't do that already, but there you go!  I'll change this before merge.

jws-1 (commented):
I can now see the output tab! One minor question before approving - after running, I can't see anything for the `SDF` within it, but I can see the respective reference molecules if I select them. Should I be seeing anything additional?

trisyoungs:
The lower cutoff you need to set on the 3D display is quite low - `1.0e-3` should show you some "halos" of density above and below the ring. The default "guess" by the module is nowhere near this!

jws-1:
> The lower cutoff you need to set on the 3D display is quite low - 1.0e-3 should show you some "halos" of density above and below the ring. The default "guess" by the module is nowhere near this!

image

Something like this?

trisyoungs:
@jws-1 More or less!  If you rotate the view so you're looking "edge on" to the benzene rather than face on as above, you should see density above and below the plane of the ring.

jws-1 (approved):

PR #1628, opened 2023-08-09
Procedure Context Fold

trisyoungs:
Thoughts welcome on this PR - I've replaced the old `ProcedureContext` constructors with templated ones taking a parameter pack, and then uses a fold expression to iterate over the passed objects and set them in the class.

This allows code to be neatened considerably, as well as permitting some nice inline brace-initialisation of contexts. The drawback is that the intent behind some arguments is assumed based on type. Specifically I'm referring to the `std::string_view` type which is assumed to be the module data prefix. I think I can live with this, but of course open to your comments on the matter.

I've also taken the liberty of tidying a few function names etc.  If we agree this is all good and approved, I'll wait until other PRs involving new modules (~#1627~ and #1584) have been merged in so I can retrofit them here first.

### DIFF
--- a/src/procedure/nodes/context.cpp
+++ b/src/procedure/nodes/context.cpp
@@ -5,55 +5,54 @@
 #include "main/dissolve.h"
 #include <stdexcept>
 
-ProcedureContext::ProcedureContext(const ProcessPool &procPool) : processPool_(procPool) {}
-
-ProcedureContext::ProcedureContext(const ProcessPool &procPool, Configuration *cfg)
-    : processPool_(procPool), configuration_(cfg)
+// Set reference to Dissolve
+void ProcedureContext::set(Dissolve &dissolve)
 {
+    dissolve_ = dissolve;
+    processPool_ = dissolve.worldPool();
 }
 
-ProcedureContext::ProcedureContext(const ProcessPool &procPool, Dissolve &dissolve)
-    : processPool_(procPool), dissolve_(dissolve)
-{
-}
+// Set available ProcessPool
+void ProcedureContext::set(const ProcessPool &procPool) { processPool_ = procPool; }
 
-// Return available process pool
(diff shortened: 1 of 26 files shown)


src/procedure/nodes/context.cpp
```diff
@@ -5,63 +5,61 @@
...
 
-ProcedureContext::ProcedureContext(const ProcessPool &procPool) : processPool_(procPool) {}
-
-ProcedureContext::ProcedureContext(const ProcessPool &procPool, Configuration *cfg)
-    : processPool_(procPool), configuration_(cfg)
+// Set reference to Dissolve
+void ProcedureContext::set(Dissolve &dissolve)
 {
+    dissolve_ = dissolve;
+    processPool_ = dissolve.worldPool();
```
jws-1, line 12: Is this always the case?

jws-1 (approved):
I very much like this, and definitely think it's in-line with the direction of things, I had just one comment to think about.

 We probably will want to do something similar for `ModuleContexts` later on. Also, I think(?) we had a conversation regarding having more than one `Configuration` associated with a context - could it be worthwhile to move to a `std::vector<Configuration*>`, but this might be better left for later on..

Lastly, it could be worthwhile to clean up the `unit/calculateExpression.cpp` to use braced-initialisations too.

src/procedure/nodes/context.cpp
trisyoungs: Probably-possibly!  But I reasoned that the most common case was taking it from the supplied `Dissolve` object.  If an alternative pool is required, that can be passed in as a separate argument in order to overwrite this one.

trisyoungs:
Ah, good spot - missed the ones in `unit/calculateExpression.cpp`. Regarding the `Configuration`s its easy enough change the behaviour to use a `std::vector`, but I left that for now because it wasn't (yet) required.

PR #1632, opened 2023-08-11
Setup of iteratedata1d

### LINKED ISSUE #1630: Iterating over collected data
Sometimes there is a need for an "intermediate" processing of `Collect`ed data in a `Procedure` - i.e. the collected data does not represent the final result, and must be further and non-trivially transformed in some way first.

To achieve that we should introduce something akin to an `IterateData1DProcedureNode` which, for now, can take a `Collect1D` or `IntegerCollect1D` node (in the same way that `Process1D` can).  The `ForEach` loop of the new node will iterate over points in the (averaged) `Data1D` which those nodes provide, setting the values into parameters for use by other nodes.

For example:

```
IntegerCollect1D  'MyData'
...
EndIntegerCollect1D

IterateData1D  'MyDataIterator'
  Data  'MyData'
  ForEach
    CalculateExpression  'MyValue'
      Expression  'MyDataIterator.value'
    EndCalculateExpression
    IntegerCollect1D  'NFBins'
      QuantityX  'MyValue'
    EndIntegerCollect1D
  EndForEach
EndSelect
```

Note that we need to follow the convention of the `DataND` classes when it comes to variables: one-dimensional data comprises an "x" axis with each point having an associated "value" (not "y").

Danbr4d:
Closes #1630 . Sometimes there is a need for an "intermediate" processing of Collected data in a Procedure - i.e. the collected data does not represent the final result, and must be further and non-trivially transformed in some way first.

To achieve that we should introduce something akin to an IterateData1DProcedureNode which, for now, can take a Collect1D or IntegerCollect1D node (in the same way that Process1D can). The ForEach loop of the new node will iterate over points in the (averaged) Data1D which those nodes provide, setting the values into parameters for use by other nodes.

Note that we need to follow the convention of the DataND classes when it comes to variables: one-dimensional data comprises an "x" axis with each point having an associated "value" (not "y").


src/procedure/nodes/iteratedata1d.cpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+        return Messenger::error("Specify either SourceData or SourceIntegerData, not both.\n");
+    }
+    else if (sourceData_ || sourceIntegerData_)
+    {
+        // If one exists, prepare the ForEach branch nodes
+        if (!forEachBranch_.prepare(procedureContext))
+            return false;
+        else
+            return true;
```
trisyoungs: Can simplify this:
```suggestion
        return forEachBranch_.prepare(procedureContext);
```

src/procedure/nodes/iteratedata1d.cpp
```diff
@@ -0,0 +1,112 @@
...
+            xParameter_->setValue(data.xAxis(index));
+            valueParameter_->setValue(data.value(index));
+
+            // If the branch fails at any point, return failure here.  Otherwise, continue the loop
+            if (!forEachBranch_.execute(procedureContext))
+                return false;
+        }
+        return true;
+    }
+    return true;
```
trisyoungs:
```suggestion

    return true;
```

src/procedure/nodes/iteratedata1d.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Finalise any necessary data after execution
+bool IterateData1DProcedureNode::finalise(const ProcedureContext &procedureContext)
+{
+    // If one exists, finalise the ForEach branch nodes
+    if (!forEachBranch_.finalise(procedureContext))
+        return false;
+    return true;
```
trisyoungs:
```suggestion
    return forEachBranch_.finalise(procedureContext);
```

src/procedure/nodes/iteratedata1d.h
```diff
@@ -0,0 +1,67 @@
...
+    ~IterateData1DProcedureNode() override = default;
+
+    private:
+    // Set up keywords for node
+    void setUpKeywords();
+
+    /*
+     * Data
+     */
+    // Whether to use only the current binned data of the histogram, rather than the accumulated average
```
trisyoungs:
```suggestion
    private:
    // Whether to use only the current binned data of the histogram, rather than the accumulated average
```

src/procedure/nodes/iteratedata1d.cpp
```diff
@@ -0,0 +1,112 @@
...
+        for (auto index : data.values())
+        {
+            xParameter_->setValue(data.xAxis(index));
+            valueParameter_->setValue(data.value(index));
+
+            // If the branch fails at any point, return failure here.  Otherwise, continue the loop
+            if (!forEachBranch_.execute(procedureContext))
+                return false;
+        }
+        return true;
```
trisyoungs: Don't strictly need this one since we will fall through to the one at the end of the function after the loop completes.
```suggestion
```

src/procedure/nodes/iteratedata1d.cpp
```diff
@@ -0,0 +1,112 @@
...
+        // Copy the averaged data from the associated Process1D node
+        if (instantaneous_)
+            data = sourceData_ ? sourceData_->data() : sourceIntegerData_->data();
+        else
+            data = sourceData_ ? sourceData_->accumulatedData() : sourceIntegerData_->accumulatedData();
+
+        for (auto index : data.values())
+        {
+            xParameter_->setValue(data.xAxis(index));
+            valueParameter_->setValue(data.value(index));
```
trisyoungs: This loop is actually wrong - it's iterating over the actual values of the data, rather than the indices of the vectors (that would have needed `data.nPoints()`. Anyway, we can use our nice `zip` function here to iterate simultaneously over the two parts:
```suggestion
        for (const auto &&[x, value] : zip(data.xAxis(), data.values()))
        {
            xParameter_->setValue(x);
            valueParameter_->setValue(value);
```

src/procedure/nodes/iteratedata1d.cpp
```diff
@@ -0,0 +1,112 @@
...
+void IterateData1DProcedureNode::setUpKeywords()
+{
+    keywords_.setOrganisation("Options", "Source");
+    keywords_.add<NodeKeyword<Collect1DProcedureNode>>("SourceData", "Collect1D node containing the histogram data to process",
+                                                       sourceData_, this, ProcedureNode::NodeType::Collect1D, false);
+    keywords_.add<NodeKeyword<IntegerCollect1DProcedureNode>>(
+        "SourceIntegerData", "IntegerCollect1D node containing the histogram data to process", sourceIntegerData_, this,
+        ProcedureNode::NodeType::IntegerCollect1D, false);
+    keywords_.addHidden<NodeBranchKeyword>("ForEach", "Branch to run on each site selected", forEachBranch_);
+}
```
trisyoungs: Need to create the parameters:
```suggestion

    xParameter_ = parameters_.emplace_back(std::make_shared<ExpressionVariable>("x"));
    valueParameter_ = parameters_.emplace_back(std::make_shared<ExpressionVariable>("value"));
}
```

trisyoungs (changes requested):
I think this is exactly what we're after.  Just small comments and suggestions again, and you also need to create the parameter variables  in the `setupKeywords()` function. You also need to reimplement the `setName()` virtual since otherwise your parameters won't get renamed if the node does (see `srrc/procedure/nodes/select.cpp`, line 61 for example).

src/procedure/nodes/iteratedata1d.h
```diff
@@ -0,0 +1,80 @@
...
+    void setName(std::string_view name) override;
+
+    /*
+     * Data
+     */
+    private:
+    bool instantaneous_{false};
+    // Collect1D node that we are processing
+    std::shared_ptr<const Collect1DProcedureNode> sourceData_;
+    std::shared_ptr<const IntegerCollect1DProcedureNode> sourceIntegerData_;
```
rprospero: We could make an abstract class `AbstractCollect1DProcedureNode` and have both `Collect1DProcedureNode` and `IntegerCollect1DProcedureNode` descend from it.  This would allow IterateData1D to have a single `sourceData_` pointer that could reference either type, depending on which concrete class it's pointing toward.  This would simplify a bunch of the logic in the prepare and execute functions.

rprospero (approved):
I'm approving this, but I've also included an improvement in the comments.  You can make those changes now or you can write an issue for it and we'll merge what we have.

trisyoungs (changes requested):
Sorry - also just noticed that the new files don't conform to lowerCamelCase - the source files should be `iterateData1D.h` and `iterateData1D.cpp`.

trisyoungs (approved):
:+1:

src/procedure/nodes/iteratedata1d.cpp
trisyoungs, line 1: @Danbr4d Still have the old files (before rename) included in the PR!

PR #1634, opened 2023-08-15
Fix TOML testing framework

rprospero:
Several of our failing TOML tests aren't problems with the TOML serialisation, but due to an issue with the testing framework.  We have to call into Dissolve twice and we've been having issues with cross talk between the two invocations.  The two objects are distinct, so there must be some static class member being set.  Either way, destroying the original CoreData and Dissolve causes clears these values out and allows the second pass to run properly.

### DIFF
--- a/unit/io/toml.cpp
+++ b/unit/io/toml.cpp
@@ -42,16 +42,16 @@ void runParse(std::filesystem::path input, int steps = 1)
         input.remove_filename();
         std::filesystem::current_path(input);
 
-	SerialisedValue toml;
-	bool knownGood;
-	{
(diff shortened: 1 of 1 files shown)


trisyoungs (approved):
Interesting. As far as I can see we still have two static singletons in the code that could be the source of the issues - one holding all currently-required `Keyword`s, and one holding all current `Module` instances.  However, these are not just blindly cleared by the code at any point, and objects are removed selectively only on their destruction, so this _shouldn't_ be an issue. Of course it we be nice to remove those static singletons for sure, so I'll add an issue in to investigate that.

PR #1636, opened 2023-08-16
Just Tests - Part 1

trisyoungs:
This PR aims to address the current mix of system and unit tests we have in our test suite, with the ultimate goals being:

- A consistent set of "just tests", whether they are unit, system, or somewhere in-between
- Test data is organised in a single directory and run under a single testing framework.
- Refactoring all existing system tests, which more-or-less encompass running a small workflow in order to generate data, and then testing this data against "known correct" values, over `gtest`.

It is obviously an enormous PR, so I would recommend reviewing focus on the new `tests/testData.h` file and the classes within, which adds useful helper functions for system-type tests, and reproduces much of the functionality of the `DataTest` module, and perhaps the `tests/modules/energy.cpp` file as an examplar of what has been performed.

Testing of the input files in the `examples/` dir have not been transferred to `gtest` since there isn't really a good way of doing so whilst maintaining a "nice" structure for the users.

As a result of this work, several modules and keyword types can be removed, along with many keywords in the user-facing modules. This will be addressed in #1645.

Finally, its worth nothing that the parallel MPI testing for system tests is removed in this PR as we move to unit-like tests. We don't currently compile an MPI version anyway since as-implemented it is probably heading into legacy status.


jws-1:
I like it - looks like writing new system tests should be nice and easy in the future!

unit/modules/avgMol/water.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+    public:
+    AvgMolModuleTest(){};
+
+    protected:
+    DissolveSystemTest systemTest;
+
+    void SetUp() override
+    {
+        EXPECT_NO_THROW(systemTest.setUp("../unit/modules/avgMol/water.txt"));
```
rhinoella:
```suggestion
        ASSERT_NO_THROW(systemTest.setUp("../unit/modules/avgMol/water.txt"));
```
Maybe it is better to `assert` here as it doesn't make sense for the test to continue if dissolve isn't set up?

rhinoella:
All looks good!

rhinoella (approved):

tests/testData.h
```diff
@@ -0,0 +1,555 @@
...
+        if (!optDataA)
+            throw(std::runtime_error(fmt::format("No data with tag '{}' exists.\n", tagA)));
+
+        auto optDataB = dissolve_.processingModuleData().searchBase<Data1DBase, Data1D, SampledData1D>(tagB);
+        if (!optDataB)
+            throw(std::runtime_error(fmt::format("No data with tag '{}' exists.\n", tagB)));
+
+        return checkData1D(optDataA->get(), tagA, optDataB->get(), tagB, tolerance, errorType);
+    }
+    // Test Data1D
```
jws-1:
```suggestion
    // Test Data3D
```

tests/testData.h
```diff
@@ -0,0 +1,555 @@
...
+    // Test SampledVector data
+    [[nodiscard]] bool checkSampledVector(std::string_view tag, const std::vector<double> &referenceData,
+                                          double tolerance = 5.0e-3,
+                                          Error::ErrorType errorType = Error::ErrorType::EuclideanError)
+    {
+        // Locate the target reference data
+        auto optData = dissolve_.processingModuleData().search<const SampledVector>(tag);
+        if (!optData)
+            throw(std::runtime_error(fmt::format("No data with tag '{}' exists.\n", tag)));
+        const SampledVector &data = optData->get();
```
jws-1: Any reason not to use `auto` here?
```suggestion
        const auto &data = optData->get();
```

tests/testData.h
```diff
@@ -0,0 +1,555 @@
...
+        ASSERT_TRUE(externalForces.importData(B));
+        checkVec3Vector(vec, B, tolerance);
+    }
+    // Test species atom type
+    void checkSpeciesAtomType(const Species *sp, int atomIndex, std::string_view atomTypeName)
+    {
+        ASSERT_TRUE(atomIndex >= 0 && atomIndex < sp->nAtoms());
+        auto &spAtom = sp->atom(atomIndex);
+        auto at = spAtom.atomType();
+        ASSERT_TRUE(at);
```
jws-1, line 289: I can understand the intention here, but it might be clearer to explicitly compare to `nullptr` - although it's up to you, ultimately!
```suggestion
        ASSERT_TRUE(at != nullptr);
```

tests/testData.h
```diff
@@ -0,0 +1,555 @@
...
+        EXPECT_EQ(at->name(), atomTypeName);
+    }
+    // Test interaction parameters
+    template <class Intra>
+    void checkIntramolecularTerms(const std::string &termInfo, const InteractionPotential<Intra> &expectedParams,
+                                  const InteractionPotential<Intra> &actualParams, double tolerance = 1.0e-6)
+    {
+        Messenger::print("Testing intramolecular interaction: {}...\n", termInfo);
```
jws-1, line 297: Do we need this?
```suggestion
```

flake.nix
```diff
@@ -75,7 +75,7 @@
                 type != "directory" || builtins.baseNameOf path
                 != ".azure-pipelines" || builtins.baseNameOf path != "web";
             };
-            patches = [ ./nix/patches/ctest.patch ];
+            patches = [ ];
```
rprospero:
```suggestion
```
Might as well drop the entire patches tag.

rprospero (changes requested):
I can't add a tag for this, but the `tests` preset in CMakePresets.json needs to be updated to the new `BUILD_TESTS` flag.  I'm taking a deeper look, but that needs handling before the PR can go in.

jws-1 (commented):
I really, really like this piece of work! A few very minor comments to consider, from the small portion of code that I reviewed! One thing to note - only 67% of tests pass on my machine, due to: `Failed to open file 'full.txt' for reading` (which I think I recall you mentioning previously)..

tests/testData.h
trisyoungs: If the tests all pass, then "No", but if a single parameter fails the double comparison, for instance, tracking in which term it was is much easier if we print this out prior to the checks!

tests/testData.h
trisyoungs: Our standards document forbids doing this implicit conversion for `int`s, but not for (shared) pointers. I'd be inclined to keep this as-is, personally.

rprospero (approved):

jws-1 (approved):

PR #1639, opened 2023-08-16
Overview Tab

jws-1:
This PR implements an initial `OverviewTab`, which currently acts as a navigation tab among the `Forcefield`, `Species` and `Configuration` tabs. This tab inherits from `MainTab`, but is constructed from QML.

To do:
- [x] `DissolveModel` as `QObject` that contains other relevant models as `Q_PROPERTIES`.
- [x] Display a `node` in the form of a `Rectangle` for Atom Types, each of the Master Terms, each Species and each Configuration.
- [x] Atom Types Summary: count
- [x] Master Terms Summaries: count for each
- [x] Species Summary: Count for each intramolecular term
- [x] Configuration Summary: Molecule and atom counts
- [x] Properly update views
- [x] Move away from embedded HTML. 

Currently this tab is not very pretty (or that useful!), but it's been a good exercise in QML and QML-C++ inter-operation, and lays the foundations for future work regarding #88. 

Future PRs will focus us on making the tab more aesthetic and providing more information.

The `OverviewTab` is hidden behind a flag defined in `mainTabsWidget.hui`.


src/gui/models/configurationModel.cpp
```diff
@@ -55,6 +65,10 @@ QVariant ConfigurationModel::data(const QModelIndex &index, int role) const
             }
         case (Qt::UserRole):
             return QVariant::fromValue(rawData(index));
+        case (Qt::UserRole + 1):
+            return QVariant::fromValue(rawData(index)->nMolecules());
+        case (Qt::UserRole + 2):
+            return QVariant::fromValue(rawData(index)->nAtoms());
```
trisyoungs: Since we're making heavier use of the roles now it would be well worth introducing our own enum with value starting at `Qt::UserRole` in order to make intention clear and remove the magic numbers.

src/gui/models/dissolveModel.h
```diff
@@ -0,0 +1,104 @@
...
+    // The number of master torsions
+    Q_PROPERTY(int nMasterTorsions READ nMasterTorsions NOTIFY modelsUpdated)
+    // The Master Improper Model
+    Q_PROPERTY(const MasterImproperModel *impropers READ impropers NOTIFY mastersChanged)
+    // The number of master impropers
+    Q_PROPERTY(int nMasterImpropers READ nMasterImpropers NOTIFY modelsUpdated)
+    // The Species Model
+    Q_PROPERTY(const SpeciesModel *species READ species NOTIFY speciesChanged)
+    // The Configuration Model
+    Q_PROPERTY(const ConfigurationModel *configurations READ configurations NOTIFY configurationsChanged)
```
trisyoungs: Again for purposes of clarity it feels like our return functions for the various models should include `Model` in the name - e.g. `configurationModel`. Otherwise, the casual "old school" reader of the code (i.e. me!) might interpret this as being the `std::vector` of current `Configuration`.s

src/gui/qml/Configuration.qml
trisyoungs, line 1: A general point regarding naming - as our qml directory gets more and more populated we need to consider a directory heirarchy in order to separate things into, in the present example, and `overview/` directory rather than renaming this file to `ConfigurationOverview.qml` (which more accurately reflects what it is).

src/gui/qml/Species.qml
```diff
@@ -0,0 +1,34 @@
...
+    property variant modelData
+
+    function generateSpeciesMarkup(index) {
+        const speciesData = model.data;
+        const speciesIndex = model.index(index, 0);
+        const speciesName = speciesData(speciesIndex, Qt.DisplayRole);
+        const bonds = speciesData(speciesIndex, Qt.UserRole + 1);
+        const angles = speciesData(speciesIndex, Qt.UserRole + 2);
+        const torsions = speciesData(speciesIndex, Qt.UserRole + 3);
+        const impropers = speciesData(speciesIndex, Qt.UserRole + 4);
```
trisyoungs: Another new enumeration.

trisyoungs (commented):
This is a great start - the QML implementation really does make it nice and easy to understand whats going on.

There's obviously a lot of future scope to think about here! Moving away from the HTML-style layout and using a compound widget sub-layout for instance, but of course that's all out of scope for this PR.  The point is that this clearly marks out the way forward to do all of that.

src/gui/models/atomTypeModel.cpp
```diff
@@ -9,6 +9,12 @@
 
 AtomTypeModel::AtomTypeModel(const CoreData &coreData) : coreData_(coreData) {}
 
+void AtomTypeModel::reset()
+{
+    beginResetModel();
+    endResetModel();
+}
+
```
rprospero, line 17: On the one hand, I hate this code.  The reset model calls should always wrap the changes to the model.  This sort of code is an indication that we're fighting the QML model, instead of working with it.

On the other hand, at the moment, we are *not* working with the QML model.  The rest of the GUI doesn't know about these models and will be making changes to the AtomTypes without sending any signals.  Eventually, we should get to the point where those are also properly modeled and everything that deals with AtomTypes will be passed a reference to the main AtomType model and we deal with proper method calls from there.  In the interim, this =reset= method *must* exist.

src/gui/models/configurationModel.h
```diff
@@ -21,11 +21,22 @@ class ConfigurationModel : public QAbstractListModel
...
     Configuration *rawData(const QModelIndex &index) const;
 
+    public:
+    enum class ConfigurationUserRole
+    {
+        RawData = Qt::UserRole,
+        MoleculesCount,
+        AtomsCount
+    };
+    Q_ENUM(ConfigurationUserRole);
```
rprospero, line 31: Good call on the enum here 👍

src/gui/overviewTabFuncs.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <QQmlContext>
+#include <QQuickItem>
+#include <QQuickWidget>
+
+OverviewTab::OverviewTab(DissolveWindow *dissolveWindow, Dissolve &dissolve, MainTabsWidget *parent, const QString title)
+    : MainTab(dissolveWindow, dissolve, parent, title, this)
+{
+    qmlRegisterType<DissolveModel>("Dissolve", 1, 0, "DissolveModel");
+    qmlRegisterType<SpeciesModel>("Dissolve", 1, 0, "SpeciesModel");
+    qmlRegisterType<ConfigurationModel>("Dissolve", 1, 0, "ConfigurationModel");
```
rprospero, line 15: As we move toward more QML, we should probably create a single method that registers all of the QML types when the program first loads.  I'm not sure if this is in scope for this PR, but it should at least be tagged as an issue.

src/gui/qml/ForcefieldView.qml
rprospero, line 1: Is this a load bearing empty file?  I've had encountered those before, but we might delete this if we don't need it.

src/gui/qml/overview/Angles.qml
```diff
@@ -0,0 +1,11 @@
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+
+Node {
+    property int nAngles
+
+    Text {
+        text: `Angles(${nAngles})`
+    }
```
rprospero, line 10: My temptation here is that we should pass in the `dissolveModel.masterAngleModel` and make the text

```qml
text: `Angles(${angleModel.rowCount()})`
```

This way, if we want to provide more angle information later on, we can simple edit the Angles file without needing to edit the calls sites.  This might also let us pull `nMasterAngles` out of the `DissolveModel` class and make it less of a [God Object Anti-Pattern](https://en.wikipedia.org/wiki/God_object).


All of the above as extends to the other Master terms and AtomTypes.

src/gui/qml/overview/Node.qml
rprospero, line 1: I like having the separate widget for the node container 👍

src/gui/models/dissolveModel.h
```diff
@@ -0,0 +1,104 @@
...
+    // The number of master angles
+    Q_PROPERTY(int nMasterAngles READ nMasterAngles NOTIFY modelsUpdated)
+    // The Master Torsion Model
+    Q_PROPERTY(const MasterTorsionModel *masterTorsionsModel READ masterTorsionsModel NOTIFY mastersChanged)
+    // The number of master torsions
+    Q_PROPERTY(int nMasterTorsions READ nMasterTorsions NOTIFY modelsUpdated)
+    // The Master Improper Model
+    Q_PROPERTY(const MasterImproperModel *masterImpropersModel READ masterImpropersModel NOTIFY mastersChanged)
+    // The number of master impropers
+    Q_PROPERTY(int nMasterImpropers READ nMasterImpropers NOTIFY modelsUpdated)
```
rprospero, line 36:
```suggestion
    Q_PROPERTY(AtomTypeModel *atomTypesModel READ atomTypesModel NOTIFY atomTypesChanged)
    // The Master Bond Model
    Q_PROPERTY(const MasterBondModel *masterBondsModel READ masterBondsModel NOTIFY mastersChanged)
    // The Master Angle Model
    Q_PROPERTY(const MasterAngleModel *masterAnglesModel READ masterAnglesModel NOTIFY mastersChanged)
    // The Master Torsion Model
    Q_PROPERTY(const MasterTorsionModel *masterTorsionsModel READ masterTorsionsModel NOTIFY mastersChanged)
    // The Master Improper Model
    Q_PROPERTY(const MasterImproperModel *masterImpropersModel READ masterImpropersModel NOTIFY mastersChanged)
```

I get into a discussion lower down on why we should get rid of the count properties.

rprospero (changes requested):
Most things look good.  I've got a couple of optional comments in there, but I do feel that we should clean up the passing of models and counts before we merge.

src/gui/qml/ForcefieldView.qml
jws-1: Oops. That was leftover from one of the earlier iterations. I'll get rid of this.

src/gui/overviewTabFuncs.cpp
jws-1: I agree with you - I'm also not too keen on it living inside this constructor.

src/gui/qml/overview/Angles.qml
jws-1: I agree with this in general. The only problem I see, and this might be due to me doing something wrong elsewhere, is that the view doesn't update with new rows, etc, automatically.

rprospero: You're right about that.  What we'll need to do is add a `rowCount` *property* to the `MasterTermModel` so that the binding works correctly.  However, this is growing beyond the scope of the PR.  Just add an issue for this and assign it to me.  Request another review and I'll mark this as ready to merge.

rprospero (approved):

src/gui/models/atomTypeModel.cpp
jws-1: Yeah, I felt like I was cheating when I wrote this. However, the way things currently are written this is a necessity..

PR #1640, opened 2023-08-17
Empty Expressions

### LINKED ISSUE #1638: Empty Expressions
`NodeValue`s can be set to empty expressions successfully (#1487), but then using them within a procedure (e.g. generating a `Configuration`) results in a `std::runtime_error` being thrown.

jws-1:
This super simple PR modifies expression evaluation to return a default `ExpressionValue` (0) on an empty expression. This prevents crashes from occurring when calling `NodeValue::asInteger/asDouble` with empty expressions, which particularly occurs when not using a `Configuration` of fixed density. Closes #1638.

### DIFF
--- a/src/expression/expression.cpp
+++ b/src/expression/expression.cpp
@@ -129,9 +129,6 @@ std::shared_ptr<ExpressionNode> Expression::rootNode() { return rootNode_; }
 // Execute expression
 std::optional<ExpressionValue> Expression::evaluate() const
 {
-    if (rootNode_)
-        return rootNode_->evaluate();
-
     // If there is no rootNode_ (i.e. expression) this is valid, and we just return an empty value
     if (!rootNode_)
         return ExpressionValue();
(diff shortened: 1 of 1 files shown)


src/expression/expression.cpp
```diff
@@ -132,7 +132,7 @@ std::optional<ExpressionValue> Expression::evaluate() const
     if (rootNode_)
         return rootNode_->evaluate();
 
-    return std::nullopt;
+    return ExpressionValue();
```
trisyoungs: For the purposes of being ultra-clear on what we're guarding against here, I'd perhaps suggest increasing the payload of this PR by a few hundred percent and writing (as the whole function):
```suggestion
    // If there is no rootNode_ (i.e. expression) this is valid, and we just return an empty value
    if (!rootNode_)
        return ExpressionValue();

    return rootNode_->evaluate();
```

trisyoungs (approved):
One suggestion on your one line! :smile:

Issue #1620, opened 2023-08-06
(DataTest) Data Display and Ranges


trisyoungs:
**Describe the feature / issue**
In a general sense having the ability to test structure factor (or other) data from an `Accumulate` module against an experimental reference is desirable. For example, we would want the ability to test accumulated total F(Q) data against experimental reference data in a `NeutronSQ` module.

This really feels like a specific use case for the `DataTest` module, which to date has been used solely for system testing in the CI. However, it has the basic components required, but lacks the ability to display the data being compared and the difference between the two datasets (and the difference), and the error range(s) over which the testing is made.

One caveat - the input keywords for the `DataTest` module do not currently have UI widgets, and it is probably beyond the scope of this PR to implement them.

trisyoungs:
After discussion it was decided that one or more new modules will be introduced to achieve this, with the `DataTest` module itself removed as part of #1636.

PR #1645, opened 2023-08-22
Just Tests - Part 2

### LINKED ISSUE #1631: Superfluous information in GeometryListKeyword Serialisation
Currently, the `GeometryListKeyword` serialises the `GeometryType` and requires that information to be present in the deserialisation.  However, during the development of PR #1619, we noticed that the actual type is defined by the context in the program.  As a result, there is only one relevant type at any possible location.  This makes the type information redundant.  Even if we were concerned about a type mismatch, having the correct type label there does nothing to ensure that the correct parameters for that type are present.

Removing the `type` parameter with the `GeometryListKeyword` serialisation and deserialisation will simplify the code and make input files easier to create without having any adverse effect on the program.

trisyoungs:
Following on from #1636, this PR removes unneeded modules, keyword types, and test-specific keywords from user-facing modules. There are quite a few!

Modules removed: `DataTest`, `Checks`, and `CheckSpecies`
Keywords removed: `Data1DStore`, `Data2DStore`, `Data3DStore`, `VectorIntDouble`, `VectorIntString`, `VectorStringPair`, `VectorStringDouble`, `ValueStore`, and `GeometryList`.

The `DataTest` module actually has some utility as a well-formed user-facing module, but this functionality will be reinvented in new module(s) as part of #1620.

Closes #1631.


rprospero (approved):
Not only does everything look good, but every class and keyword that is dropped is one with an ugly TOML serialisation story.

jws-1 (approved):

trisyoungs:
> Not only does everything look good, but every class and keyword that is dropped is one with an ugly TOML serialisation story.

I thought you might like this one! :smile:

PR #1646, opened 2023-08-23
1620 Compare Module and Data Source keyword

### LINKED ISSUE #1620: (DataTest) Data Display and Ranges

rhinoella:
### Compare Module

- Module to replace the old 'DataTest' module with added functionalities
- Currently this PR does not include the GUI implementation, as I think it is better in a different PR

### Data Source Keyword
- A keyword that creates Data objects (Data1D, Data2D, Data3D, etc.) from user-specified internal or external data
- Can accept up to two different data sources depending on the needs of the module

### Notes
- Compare module is currently only able to handle Data1D sources

Closes #1620


src/io/import/dataBase.h
```diff
@@ -0,0 +1,35 @@
...
+     */
+    protected:
+    // Set up keywords for the format
+    virtual void setUpKeywords() = 0;
+
+    /*
+     * Import Functions
+     */
+    public:
+    // Import Data1D using current filename and format
```
trisyoungs:
```suggestion
    // Import data using current filename and format
```

src/io/import/dataBase.h
```diff
@@ -0,0 +1,35 @@
...
+    bool importData(DataBase &data, const ProcessPool *procPool = nullptr) { return false; };
+    // Import Data1D using supplied parser and current format
```
trisyoungs:
```suggestion
    // Import data using supplied parser and current format
```

src/keywords/dataSource.cpp
```diff
@@ -0,0 +1,168 @@
...
+        {
+            break;
+        }
+
+        // Add the data to the vectors
+        dataVector.push_back(data);
+        dataSources.push_back(std::make_pair(source, std::reference_wrapper<T>(data)));
+    }
+
+    // Once finished reading the arguements, add the data to the module
```
trisyoungs:
```suggestion
    // Once finished reading the arguments, add the data to the module
```

src/math/data1D.h
```diff
@@ -26,7 +26,7 @@ class Data1D : public Data1DBase, public Serialisable<>
      */
     private:
     // Tag for data (optional)
-    std::string tag_;
+    std::string tag_{""};
```
trisyoungs: Don't need to brace-initialise this since `std::string` is constructed empty.

src/keywords/dataSource.cpp
trisyoungs, line 1: What's the benefit of having this code in the cpp file, and explicitly instantiating the necessary types, over having a pure header-only class (which is what it feels like it should be)?

trisyoungs (commented):
I may not be fully awake while going through this, but it feels kinda overkill for what we want to do, and I fear is dragging us away from just getting a working 1D comparison up and running correctly, with multiple datasets, which is what we really need.

Perhaps I'm mistaking your intent, but at present this feels like its written to be a catch-all class that can store multiple different dimensionalities of data (which we don't want) rather than something more akin to a templated class taking, for instance, the type and file format (`Data1D` and `Data1DImportFileFormat`) and which then has specialisations to handle the difficult parts that don't translate across dimensionalities (e.g. the fact that `Data1D` derives from a base class, and which complicates the search for internal data of that type).

rhinoella:
> I may not be fully awake while going through this, but it feels kinda overkill for what we want to do, and I fear is dragging us away from just getting a working 1D comparison up and running correctly, with multiple datasets, which is what we really need.
> 
> Perhaps I'm mistaking your intent, but at present this feels like its written to be a catch-all class that can store multiple different dimensionalities of data (which we don't want) rather than something more akin to a templated class taking, for instance, the type and file format (`Data1D` and `Data1DImportFileFormat`) and which then has specialisations to handle the difficult parts that don't translate across dimensionalities (e.g. the fact that `Data1D` derives from a base class, and which complicates the search for internal data of that type).

@trisyoungs The intent is to have a `DataSource` template class that can handle Data1D, 2D and 3D objects depending on what the type is specified as:

https://github.com/disorderedmaterials/dissolve/blob/713c3528a866636c2e900480d2dae601d2843f72/src/modules/compare/compare.cpp#L10C5-L18C5

 The specialisation (Data1D requiring a different search function) has to be handled in Compare.cpp, as it requires the `ModuleContext` object which is provided in the `Compare::process()` function. So the main function of DataSource is to accept a set of internal/external data (1D, 2D or 3D depending on what is specified), construct a vector of the relevant object (`Data1D`, `Data2D` or `Data3D`) and add it to the module's data through the module's `addData()` function that is passed into its constructor.

trisyoungs:
@rhinoella Because of the need to have the processing module data to search? Couldn't you just pass this in as a const reference when creating the keyword?

rhinoella:
@trisyoungs Unfortunately the `moduleContext` object is only known and constructed during runtime, when it is passed into the `Compare::process()` function during `Dissolve::iterate()`, so the keyword is constructed before the `moduleContext` object is created.

trisyoungs:
@rhinoella True true. Alas, module constructors used to get passed the main `Dissolve&` object, but no longer, and so you could have gotten the `GenericList` from there. Curses to this design!  I guess you'd better just work this up, focussing on the 1D case, and see where you get to.

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,84 @@
...
+        // End the block
+        if (!parser.writeLineF("End{}", dataSourceTypes().keyword(External)))
+            return false;
+    }
+
+    return true;
+}
+
+// Express as a serialisable value
+SerialisedValue DataSource::serialise() const { return {}; }
```
trisyoungs: Something like this should work:
```suggestion
SerialisedValue DataSource::serialise() const
{
    Serialisedvalue result;
    if (dataSourceType_ == Internal)
    {
        result["internal"] = internalDataSource_;
    }
    else
    {
        result["external"] = externalDataSource_.get();
    }
    return result;
}
```

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+// Read values from a serialisable value
+void DataSource::deserialise(const SerialisedValue &node, const CoreData &data) {}
```
trisyoungs: And here:
```suggestion
void DataSource::deserialise(const SerialisedValue &node, const CoreData &data)
{
    internalDataSource_ = toml::find_or<std::string>(node, "internal", "");
    if (internalDataSource_.isEmpty())
    {
        dataSourceType_ = External;
        externalDataSource_.get()->deserialise(toml::find(node, "external"), coreData);
    }
    else
    {
        dataSourceType_ = Internal;
    }
}
```

src/classes/dataSource.h
```diff
@@ -0,0 +1,130 @@
...
+    std::optional<std::string> internalDataSource() const;
+    // Return external data source
+    OptionalReferenceWrapper<FileAndFormat> externalDataSource() const;
+
+    /*
+     * I/O
+     */
+    public:
+    // Write through specified LineParser
+    bool serialise(LineParser &parser, std::string_view keywordName, std::string_view prefix) const;
```
trisyoungs, line 91: Where's your normal `deserialise()` function?

src/classes/dataSource.h
```diff
@@ -0,0 +1,130 @@
...
+    }
+    // Returns the data casted to the requested type
+    template <class DataType> DataType data() const
+    {
+        assert(dataExists());
+        auto data = dynamic_cast<DataType &>(*data_);
+
+        return data;
+    }
+
```
trisyoungs, line 85:
```suggestion
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,160 @@
...
+     * Data
+     */
+    private:
+    // Format object for the data
+    DataFormat format_;
+
+    /*
+     * Arguments
+     */
+    // Return minimum number of arguments accepted
```
trisyoungs, line 35:
```suggestion
    public:
    // Return minimum number of arguments accepted
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,160 @@
...
+                DataType data;
+                DataFormat format;
+
+                // Read the supplied arguments
+                if (format.read(parser, 1, fmt::format("End{}", DataSource::dataSourceTypes().keyword(DataSource::External)),
+                                coreData) != FileAndFormat::ReadResult::Success)
+                {
+                    return false;
+                }
+                // Import the data
```
trisyoungs, line 88:
```suggestion

                // Import the data
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,160 @@
...
+                if (!format.importData(data, parser.processPool()))
+                {
+                    return false;
+                }
+
+                // Add data to pair element depending on counter
+                dataCounter == 0 ? dataSourceA.template addData<DataType, DataFormat>(data, format)
+                                 : dataSourceB.template addData<DataType, DataFormat>(data, format);
```
trisyoungs: Why the need for the inline `template` here?  This looks odd!

src/keywords/dataSource.h
```diff
@@ -0,0 +1,160 @@
...
+                return false;
+            }
+        }
+
+        return true;
+    }
+    // Express as a serialisable value
+    SerialisedValue serialise() const override { return {{}}; }
+    // Read values from a serialisable value
+    void deserialise(const SerialisedValue &node, const CoreData &coreData) override {}
```
trisyoungs: Need these!

src/keywords/dataSourceBase.cpp
```diff
@@ -0,0 +1,9 @@
+#include "keywords/dataSource.h"
+
+DataSourceKeywordBase::DataSourceKeywordBase(std::vector<DataSourceKeywordBase::DataPair> &dataSources,
+                                             std::string_view endKeyword)
+    : KeywordBase(typeid(this)), dataSources_(dataSources), endKeyword_(endKeyword)
+{
+}
+// Returns reference to module data
```
trisyoungs, line 9:
```suggestion

// Returns reference to module data
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,160 @@
...
+                // If not, print accepted options
+                return DataSource::dataSourceTypes().errorAndPrintValid(parser.argsv(0));
+            }
+
+            // If data is internal
+            if (DataSource::dataSourceTypes().enumeration(parser.argsv(0)) == DataSource::Internal)
+            {
+                // Add data to pair element depending on counter
+                dataCounter == 0 ? dataSourceA.addData(parser.argsv(1)) : dataSourceB.addData(parser.argsv(1));
+                dataCounter++;
```
trisyoungs: I completely understand the need for the counter as-is, but there's probably a way to get rid of that by storing the data references in wrappers in a `std::vector` and using an iterator.

src/keywords/dataSource.h
```diff
@@ -0,0 +1,160 @@
...
+    bool serialise(LineParser &parser, std::string_view keywordName, std::string_view prefix) const override
+    {
+        for (auto &[dataSourceA, dataSourceB] : dataSources_)
+        {
+            // Write the keyword name
+            if (!parser.writeLineF("{}{}\n", prefix, keywordName))
+            {
+                return false;
+            }
+            // Serialise the first data source
```
trisyoungs, line 134:
```suggestion

            // Serialise the first data source
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,160 @@
...
+            if (!dataSourceA.serialise(parser, keywordName, prefix))
+            {
+                return false;
+            }
+            // Skip to next iteration if dataSourceB is undefined
```
trisyoungs, line 140:
```suggestion

            // Skip to next iteration if dataSourceB is undefined
```

src/math/dataBase.h
```diff
@@ -0,0 +1,38 @@
...
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#pragma once
+
+#include <vector>
+
+// Data Base
+class DataBase
+{
+
```
trisyoungs:
```suggestion
```

src/math/dataBase.h
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#pragma once
+
+#include <vector>
+
+// Data Base
```
trisyoungs: Perhaps a little more detail on what this class' purpose actually is here? Just a full line comment, really!

src/modules/compare/compare.h
```diff
@@ -0,0 +1,54 @@
...
+#include "math/data1DBase.h"
+#include "math/data2D.h"
+#include "math/data3D.h"
+#include "math/error.h"
+#include "math/range.h"
+#include "math/sampledData1D.h"
+#include "module/context.h"
+#include "module/module.h"
+
+using rangeErrorPair = std::pair<std::vector<Range>, std::vector<double>>;
```
trisyoungs: I think by convention we make and `using` pseud-type fully upper camelcase.

src/modules/compare/process.cpp
```diff
@@ -0,0 +1,97 @@
...
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "math/interpolator.h"
+#include "modules/compare/compare.h"
+
+Module::ExecutionResult CompareModule::process(ModuleContext &moduleContext)
+{
+
+    Messenger::print("COMPAREE");
+
```
trisyoungs: Bit of debugee output!
```suggestion
```

src/modules/compare/process.cpp
```diff
@@ -0,0 +1,97 @@
...
+    // Mapping from data pair to ranges and error
+    std::map<DataSourceKeywordBase::DataPair *, rangeErrorPair> dataSourcesErrors;
+
+    int index;
```
trisyoungs:
```suggestion
    auto index = 0;
```

src/modules/compare/process.cpp
```diff
@@ -0,0 +1,97 @@
...
+    for (auto &dataPair : dataSources_)
```
trisyoungs, line 13: Wouldn't it be nicer to do this here?
```suggestion
    for (auto &[dataSourceA, dataSourceB] : dataSources_)
```

src/modules/compare/process.cpp
```diff
@@ -0,0 +1,97 @@
...
+    {
+
```
trisyoungs:
```suggestion
```

src/modules/compare/process.cpp
```diff
@@ -0,0 +1,97 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "math/interpolator.h"
+#include "modules/compare/compare.h"
+
+Module::ExecutionResult CompareModule::process(ModuleContext &moduleContext)
```
trisyoungs, line 7: Obviously this is written purely with a focus on 1D data for now, which is fine, but when we move forward to other dimensionalities it will be nice to do the bulk of high-level processing in this function (i.e. checking for dataset validity etc.) and then using templated functions to do the actual comparisons.

trisyoungs (changes requested):
A large handful of comments and suggestions, but overall I like this a lot (and I can follow it now!). We will need to get the (de)serialisation complete before this gets merged in.

src/keywords/dataSource.h
rhinoella: It's required for the MacOS build to pass unfortunately! Its because addData is an overloaded function so the compiler wants us to be very specific.

src/classes/dataSource.h
rhinoella: It can't be done in the dataSource class because it doesn't know the `FileAndFormat` type of the data. The `bool serialise()` function is just a little helper for the keyword.

src/modules/compare/process.cpp
rhinoella: I need the pointer to the `dataPair` as they are used as the key in the `dataSourcesErrors` map:
https://github.com/disorderedmaterials/dissolve/blob/00bd84c54c2a2223f8c26f563f9c4729b6311769/src/modules/compare/process.cpp#L26
https://github.com/disorderedmaterials/dissolve/blob/00bd84c54c2a2223f8c26f563f9c4729b6311769/src/modules/compare/process.cpp#L51
https://github.com/disorderedmaterials/dissolve/blob/00bd84c54c2a2223f8c26f563f9c4729b6311769/src/modules/compare/process.cpp#L54

rhinoella:
### Changes:
- `DataSource` no longer inherits from `Serialisable` as there is no good way to implement the TOML `deserialise` function, as `DataSource` does not know the `FileAndFormat` type of the data. The only way to do this is to have all the data file and formats inherrit from a new base class that specifies a virtual `importData` function
- TOML serialisation/deserialisation functions properly implemented
- Used a `std::queue` instead of a counter in `DataSourceKeyword`
- `DataSource` now stores a variable called `dataName_` which has either the tag or the filename, in order to have a naming system for the GUI implementation

src/classes/dataSource.h
```diff
@@ -0,0 +1,135 @@
...
+                    return Messenger::error("No data with tag '{}' exists.\n", internalDataSource_);
+                }
+                Messenger::print("Sourced data '{}'", internalDataSource_);
+
+                // Create unique pointer for data
+                data_ = std::make_unique<DataType>(optData->get());
+            }
+        }
+
+        return data_.get() != nullptr ? true : false;
```
trisyoungs:
```suggestion
        return data_ != nullptr;
```

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    result["dataSourceType"] = dataSourceTypes().keyword(dataSourceType_);
+    if (dataSourceType_ == Internal)
+    {
+        result["data"] = internalDataSource_;
+    }
+    else
+    {
+        result["data"] = externalDataSource_.get()->serialise();
+    }
```
trisyoungs, line 165:
```suggestion
    }
    
    return result;
```

rhinoella:
### Changes

- Type definition for a Data classes' associated file and format is stored in the class. This improves type-safety (prevents someone from using a `Data2D` object with a` Data1DImportFileFormat`) and also means that the template class `DataSourceKeyword` only needs one template parameter- the data class
- `DataSource` class stores the Data object and Format object as an `std::variant` instead of a `std::unique_ptr` as this further improves type safety

### Note
Due to the removal of `unique_ptr`, the `DataBase` class (parent class for `Data1DBase`, `Data2DBase` and `Data3DBase`) can also be removed. I left it in anyways because it reduces code-repetition but please let me know if it should be removed.

src/classes/dataSource.h
```diff
@@ -5,17 +5,25 @@
...
+#include "math/data2D.h"
+#include "math/data3D.h"
 #include "math/sampledData1D.h"
+#include <variant>
 
 class DataSource
 {
+    public:
+    using DataType = std::variant<Data1D, SampledData1D, Data2D, Data3D>;
+    using Format = std::variant<Data1DImportFileFormat, Data2DImportFileFormat, Data3DImportFileFormat>;
```
rprospero, line 23: Good call on using the type alias

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        // Data2D, Data3D
+        else
+        {
+            return std::visit(
+                [&]<class T>(T &data)
+                {
+                    // Locate target data from tag
+                    auto optData = processingModuleData.search<const T>(internalDataSource_);
```
rprospero:
```suggestion
                [&](auto &data)
                {
                    // Locate target data from tag
                    auto optData = processingModuleData.search<const std::remove_reference<decltype(T)>>(internalDataSource_);
```

This removes the template on the lambda, which is illegal at the moment and breaking the Windows build.  You'll also need to `#include <type_traits` at the start of the file to use `std::remove:reference`.

rprospero (dismissed):
Everything looks really good.  I've just included one small change to fix the Windows build.

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,167 @@
...
+DataSource::Format &DataSource::externalDataSource() { return externalDataSource_; }
+
+// Function to source data (only required for internal data sources)
+bool DataSource::sourceData(GenericList &processingModuleData)
+{
+    if (!dataExists())
+    {
+        return false;
+    }
+    if (dataSourceType_ == Internal)
```
trisyoungs, line 49:
```suggestion

    // For Internal data sources, try to locate the data now
    if (dataSourceType_ == Internal)
```

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,167 @@
...
+            {
+                return Messenger::error("No data with tag '{}' exists.\n", internalDataSource_);
+            }
+            // Set data
+            data_ = optData->get();
+        }
+
+        // Data2D, Data3D
+        else
+        {
```
trisyoungs: I kind of dislike the way the comment breaks up the "flow" of the `if`/`else`, so perhaps move it inside?
```suggestion
        }
        else
        {
            // For other data types just visit the variant
```

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,167 @@
...
+                        return Messenger::error("No data with tag '{}' exists.\n", internalDataSource_);
+                    }
+                    // Set data
+                    data_ = optData->get();
+                    return true;
+                },
+                data_);
+        }
+    }
+    // Return whether or not variant contains value
```
trisyoungs, line 82:
```suggestion

    // Return whether or not variant contains value
```

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,167 @@
...
+    return !data_.valueless_by_exception();
+}
+
+// Function to add internal data
```
trisyoungs:
```suggestion
// Set internal data source
```

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    // If data is internal
+    if (dataSourceType_ == Internal)
+    {
+        if (!parser.writeLineF("'{}'\n", internalDataSource_))
+            return false;
+    }
+    // If data is external
+    else
```
trisyoungs:
```suggestion
    // For internal data sources just need to write the tag. Externals write the file and format
    if (dataSourceType_ == Internal)
    {
        if (!parser.writeLineF("'{}'\n", internalDataSource_))
            return false;
    }
    else
```

src/classes/dataSource.cpp
```diff
@@ -0,0 +1,167 @@
...
+    {
+        return std::visit(
```
trisyoungs, line 117: This is actually a really nice use of `variant`/`visit`!

src/classes/dataSource.h
```diff
@@ -0,0 +1,96 @@
...
+    public:
+    // Return enum options for DataSourceType
+    static EnumOptions<DataSourceType> dataSourceTypes();
+    // Return data source type enum
+    DataSourceType dataSourceType() const;
+
+    /*
+     * Data
+     */
+
```
trisyoungs:
```suggestion
```

src/classes/dataSource.h
```diff
@@ -0,0 +1,96 @@
...
+    ~DataSource() = default;
+
+    public:
+    enum DataSourceType
+    {
+        Internal,
+        External
+    };
+
+    public:
```
trisyoungs:
```suggestion
    public:
    // Data Source Types
    enum DataSourceType
    {
        Internal,
        External
    };
```

src/classes/dataSource.h
```diff
@@ -0,0 +1,96 @@
...
+    std::string_view dataName() const;
+    // Return if data exists and has been initialised
+    bool dataExists() const;
+    // Return internal data source
+    std::optional<std::string> internalDataSource() const;
+    // Return external data source
+    Format &externalDataSource();
+    // Function to source data (only required for internal data sources)
+    bool sourceData(GenericList &processingModuleData);
+    // Function to add internal data
```
trisyoungs: Don't really need to describe these functions as functions, so...
```suggestion
    // Source specified data (only required for internal data sources)
    bool sourceData(GenericList &processingModuleData);
    // Set internal data source
```

src/classes/dataSource.h
```diff
@@ -0,0 +1,96 @@
...
+    template <class D> void addData(D data, class D::Formatter &fileAndFormat)
+    {
+        dataSourceType_ = External;
+        // Create format object in place in variant
+        externalDataSource_.emplace<class D::Formatter>(fileAndFormat);
+        data_ = data;
+        // Set data name to be base filename
+        dataName_ = fileAndFormat.filename().substr(fileAndFormat.filename().find_last_of("/\\") + 1);
+    }
+    // Returns the data of the requested type
```
trisyoungs:
```suggestion
    // Returns data in the requested type
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,225 @@
...
+#include "io/import/data3D.h"
+#include "keywords/dataSourceBase.h"
+#include "math/data1D.h"
+#include "math/data2D.h"
+#include "math/data3D.h"
+#include "templates/optionalRef.h"
+#include <queue>
+
+// Keyword managing data sources
+// Template arguments: data class (Data1D, Data2D ...), data import file format
```
trisyoungs:
```suggestion
// Keyword managing a DataSource
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,225 @@
...
+                break;
+            }
+            // If data source type supplied is valid
+            if (!DataSource::dataSourceTypes().isValid(parser.argsv(0)))
+            {
+                // If not, print accepted options
+                return DataSource::dataSourceTypes().errorAndPrintValid(parser.argsv(0));
+            }
+
+            // If data is internal
```
trisyoungs:
```suggestion
            // Next parse depends on whether we have internal / external data
```

src/keywords/dataSource.h
```diff
@@ -0,0 +1,225 @@
...
+            if (DataSource::dataSourceTypes().enumeration(parser.argsv(0)) == DataSource::Internal)
+            {
+                // Add data to dataSource
+                sourceQueue.front()->addData(parser.argsv(1));
+                // Remove dataSource from queue
+                sourceQueue.pop();
+            }
+            // If data is external
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
Looking really very nice indeed.   I have a bunch of formatting comments, but nothing more than that.

PR #1647, opened 2023-08-24
1505 Generalise Parameter Access

### LINKED ISSUE #1505: Generalise Parameter Access for Nodes
Adding a usable, externally-visible parameter (`ExpressionVariable`) to a procedure node is quite laborious at present since one needs to implement both the storage for the parameters as well as override the base access functions.  Given that many nodes will eventually present their information, and the fact that there are not going to be vast numbers of nodes in existence at any one point, we should make this general in the base `ProcedureNode` rather than having to reimplement on a case-by-case basis.

trisyoungs:
This PR simplifies parameter (`ExpressionVariable`) creation, storage, and handling, moving it all to the `ProcedureNode` base class rather than requiring override functions in any node type that required to use them.

To handle the automatic prefixing of `ExpressionVariables` in the `parameters_` vector with the node's name (necessitating keeping a record of the original name), a couple of approaches were initially tried and discounted:
1) Modifying the storage of `parameters_` from `std::vector<std::shared_ptr<ExpressionVariable>>` to `std::vector<std::pair<std::shared_ptr<ExpressionVariable>,std::string>>` with the string component storing the original, base name of the variable for reference - this has the effect of "polluting" a lot of other routines using the parameters, and so was abandoned.
2) Determining the original basename of the variable by searching the string for the last `.` (if it existed). However this didn't feel particularly robust and required other routines to determine whether or not a prefix existed.

So, the preferred solution in the end was to extend `ExpressionVariable` to keep track of both it's original name _and_ any applied prefix. This way, impact on the rest of the code is kept to a minimum.

Closes #1505.

### DIFF
--- a/src/expression/ExpressionVisitor.cpp
+++ b/src/expression/ExpressionVisitor.cpp
@@ -154,7 +154,7 @@ antlrcpp::Any ExpressionVisitor::visitVariable(ExpressionParser::VariableContext
     // Check local variables first
     const std::vector<std::shared_ptr<ExpressionVariable>> &localVars = *localVariables_;
     auto it = std::find_if(localVars.begin(), localVars.end(),
-                           [ctx](auto var) { return DissolveSys::sameString(var->name(), ctx->Name()->getText()); });
+                           [ctx](auto var) { return DissolveSys::sameString(var->baseName(), ctx->Name()->getText()); });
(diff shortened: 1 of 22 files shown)


rhinoella (approved):
All looks great, it really cleans up the code nicely.

PR #1658, opened 2023-09-05
Region Potentials

trisyoungs:
This PR implements a new style of global potential based on a "region"-style method. Given some form of `Expression` describing a function in 3D space, energy and force terms are derived on a per-voxel basis. This allows complex or "interesting" formations to be constructed in a fairly natural manner.

The intention is not to necessarily provide a "proper" potential / simulation, but rather to help enforce exotic structural features.


src/kernels/potentials/regional.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+/*
+ * Definition
+ */
+
+// Return functional form of the potential, as a string
+const std::string RegionalPotential::formString() const { return "Expression"; }
+
+// Return parameters of the potential, as a string
+const std::string RegionalPotential::formParametersString() const { return "N/A"; }
```
jws-1, line 100: Probably beyond the scope of this PR, but I wonder if should add an `Expression` column to our Global Potentials table, or something alike?

src/kernels/potentials/regional.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+    // Set fractional voxel sizes
+    Vec3<int> nVoxels;
+    for (auto n = 0; n < 3; ++n)
+        nVoxels.set(n, std::max(int(box->axisLength(n) / voxelSize), 1));
+    voxelSizeFrac_.set(1.0 / nVoxels.x, 1.0 / nVoxels.y, 1.0 / nVoxels.z);
+
+    // Initialise 3D maps
+    energyVoxels_.initialise(nVoxels.x, nVoxels.y, nVoxels.z);
+    forceVoxels_.initialise(nVoxels.x, nVoxels.y, nVoxels.z);
```
jws-1, line 114: Again, not related to this PR - but I feel like it'd be nice for `initialise` to take a `Vec3<int>`, perhaps..

jws-1 (approved):
Code looks good, and the functionality works very nicely! Couple of minor comments that aren't really related to this PR...

Danbr4d (approved):
Looks good to me

src/kernels/potentials/regional.cpp
trisyoungs: Yep. I'll add a general issue to capture this across the whole code.

trisyoungs: #1660.

src/kernels/potentials/regional.cpp
trisyoungs: I wondered about this, but decided to ignore it until we need something more in the GUI.  Since we do not allow potentials to be edited through those tables we could just rename the `formString()` and `formParametersString()` virtuals to something more general to accommodate the present PR. But I would probably do that as a follow-up PR.

jws-1: I agree - happy for this to be left for another future PR.

trisyoungs: #1661

Issue #1652, opened 2023-08-31
[Epic] Workflow 2.0


jws-1:
### Focus
Implement an "overview" that displays all major elements of the simulation, as well as a summary of each, in some diagrammatic way. Initially, this could be a simple hierarchy of "blocks". The idea is that this can replace the current usage of tabs, by being able to view and edit a specific element (e.g. a species) by double-clicking on it. Alongside this, we can move away from using `Layers` and instead define `Tasks`, which each execute a `Procedure` (which could include `RunModule`/`RunModuleList` nodes, as well as running other `Tasks`). With a `Master Task` that is executed in-place of the current simulation loop - this task could expose certain information, such as `Configurations`, to its child tasks through a context, rather than giving every task access to the entirety of `CoreData`. These `Tasks` would also be displayed in the Overview, along with the modules that they run. The GUI side of things should be done using QML.

#### Basic Overview
Implement an Overview tab with the necessary QML "Dissolve" model, and which offers basic navigation capabilities to existing tabbed data.
- [x] #1639
- [ ]  (Lower priority) Beautify QML
- [ ]  (Lower priority) Show relationships between elements of the simulation, i.e. draw a line between a `Species` and each `Configuration` that it is present in.
- [x] #1655
- [x] #1656

#### Dialog-ise Tabs
Convert the `MainTab` derived classes to instead derive from `QDialog`, which is then popped-up when the corresponding item in the overview is clicked. Feasibility of this will be proven by first dialogising the `Species` tab.
- [ ] Convert `SpeciesTab` to enable display in a dialog as well as a tab
- [ ] Convert `ConfigurationTab`
- [ ] Convert `ForcefieldTab`

#### Layer Display
Although we want to move away from `Layers`, it would be a useful exercise to include them, along with their `Modules`, in the Overview - this code would likely be reusable when it comes to `Tasks`. Since users typically want to quickly flick between the information of multiple modules, it would be more suitable to provide a tab-like view. When a `Module` is double-clicked in the Overview, we could show it in a separate tab. Single-clicking / selecting a `Module` could just display its full options in a side-pane to allow setup, but not necessarily show any output data.
 - [x] Display existing `ModuleLayer`s in the QML overview (#1659)
 - [ ] Allow summary information for each module to be displayed on its QML entity
 - [ ] Allow modules to be "popped out" to separate tabs for easier interrogation

#### Task Implementation
Introduce the new `Task`-based workflow.
[...]

PR #1659, opened 2023-09-05
Overview Tab - Layers

jws-1:
This small PR adds `ModuleLayers` to the `OverviewTab`, as well as displaying (but not offering means to edit) their corresponding `Modules`. Future work (see #1652) will make this view more aesthetically pleasing (for example, ensuring that `Nodes` are sized to fit their contents), as well as providing means to edit `ModuleLayers` and control their `Modules`. Furthermore, the `ModuleListView` currently cannot be navigated, but this will be dealt with in a future PR along with the previously described visual improvements.


src/gui/models/dissolveModelImageProvider.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "gui/models/dissolveModel.h"
+#include "gui/models/moduleLayerModel.h"
+
+DissolveModelImageProvider::DissolveModelImageProvider(DissolveModel *dissolveModel)
+    : QQuickImageProvider(QQmlImageProviderBase::Image)
+{
+    dissolveModel_ = dissolveModel;
+}
+
+QImage DissolveModelImageProvider::requestImage(const QString &id, QSize *size, const QSize &requestedSize)
```
trisyoungs, line 14: Hmm. Perhaps a contender for my new most-despised Qt override!

src/gui/models/moduleLayersModel.cpp
```diff
@@ -2,34 +2,61 @@
...
     endResetModel();
 }
 
-ModuleLayer *ModuleLayersModel::rawData(const QModelIndex &index) const { return layers_->get()[index.row()].get(); }
+ModuleLayer *ModuleLayersModel::rawData(const QModelIndex &index) const
+{
+    if (!layers_)
+        return {};
+    if (!index.isValid())
+        return {};
```
trisyoungs:
```suggestion
    if (!layers_ || !index.isValid())
        return {};
```

src/gui/models/moduleLayersModel.cpp
```diff
@@ -2,34 +2,61 @@
...
+
     switch (role)
     {
         case (Qt::DisplayRole):
-            return QString::fromStdString(std::string(rawData(index)->name()));
+            return QString::fromStdString(std::string(data->name()));
         case (Qt::UserRole):
-            return QVariant::fromValue(rawData(index));
+            return QVariant::fromValue(data);
+        case (Qt::UserRole + 1):
```
trisyoungs: I think we definitely need a custom `enum` using `Qt::UserRole` as the base here.

trisyoungs (approved):
Apart from the ugliness this is looking pretty good!

src/gui/models/moduleLayersModel.cpp
jws-1: Ah yep, I meant to do that but forgot to come back!

PR #1662, opened 2023-09-07
1620 Compare Module GUI Implementation

rhinoella:
## GUI implementation for the Compare Module (#1646 )

Fixes serialisation issues with the Compare Module, implements a test and reverts back to using templates instead of variants.

Displays both datasets along with the absolute difference (Delta) between their values:

image


rhinoella:
- [x] Have the buttons separate Data1D, Data2D and Data3D data
- [x] Have all the data in the Data1D category on the same graph but with different offsets

src/modules/compare/compare.cpp
```diff
@@ -12,7 +12,7 @@ CompareModule::CompareModule() : Module(ModuleTypes::Compare)
                                              "EndData1D");
     keywords_.add<EnumOptionsKeyword<Error::ErrorType>>("ErrorType", "Type of error calculation to use", errorType_,
                                                         Error::errorTypes());
-    keywords_.add<RangeVectorKeyword>("RFactorRanges", "Ranges over which to calculate RFactors", ranges_);
+    keywords_.add<RangeVectorKeyword>("ErrorRanges", "Ranges over which to calculate RFactors", ranges_);
```
trisyoungs: I think we should be more general in our description here - "RFactors" is quite specific to EPSR and, while that might be the desired metric for a lot of people, it's just "errors" really.

src/modules/compare/gui/compareWidget.h
```diff
@@ -0,0 +1,50 @@
...
+    /*
+     * UI
+     */
+    private:
+    // Main form declaration
+    Ui::CompareModuleWidget ui_;
+    // DataViewers contained within this widget
+    DataViewer *graph_;
+    // Data1D Button
+    std::array<QPushButton *, 3> buttons_;
```
trisyoungs: Not sure I like stashing the widget pointers in an array like this. See below for a different suggestion...

src/modules/compare/gui/compareWidgetFuncs.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+        // If no buttons are checked, check the first one by default
+        if (index == buttons_.size() && buttons_.back()->isChecked() == false)
+        {
+            buttons_.front()->setChecked(true);
+            index = 0;
+        }
+
+        // Create the renderables for Data1D
+        if (index == 0)
```
trisyoungs: Why can't we just do:
```suggestion
        if (ui_.Data1DButton->isChecked())
```
You can set the state of the 1D button to be checked in Designer and presumably they're in a button group, so everything should work you want.

src/modules/compare/gui/compareWidgetFuncs.cpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+void CompareModuleWidget::on_Data1DButton_clicked(bool checked)
+{
+    if (!checked)
+    {
+        return;
+    }
+    updateControls(ModuleWidget::RecreateRenderablesFlag);
```
trisyoungs: Neater?
```suggestion
    if (checked)
    {
        updateControls(ModuleWidget::RecreateRenderablesFlag);
    }
```

trisyoungs (changes requested):
Looking good, but a strong suggestion on changes to the handling of the display logic.

src/modules/compare/compare.cpp
rhinoella: Ah! I missed that, thanks!

src/classes/dataSource.h
```diff
@@ -9,86 +9,221 @@
...
-        externalDataSource_.emplace<typename D::Formatter>(fileAndFormat);
-        data_ = data;
-        // Set data name to be base filename
-        dataName_ = fileAndFormat.filename().substr(fileAndFormat.filename().find_last_of("/\\") + 1);
+        return (!internalDataSource_.empty() && dataSourceType_ == Internal) ||
+               (externalDataSource_.hasFilename() && dataSourceType_ == External);
     }
-    // Returns data in the requested type
-    template <class D> D data() const
+    // Function to source data
```
trisyoungs:
```suggestion
    // Obtain data from the relevant source
```

src/classes/dataSource.h
```diff
@@ -9,86 +9,221 @@
...
+                return Messenger::error("Error importing data from '{}'", externalDataSource_.filename());
+            }
+
+            return true;
+        }
+
+        return false;
     }
 
+    // Returns data
```
trisyoungs:
```suggestion
    // Return the data
```

src/classes/dataSource.h
```diff
@@ -9,86 +9,221 @@
...
+            dataName_ = toml::find<std::string>(node, "source");
+        }
+        // If data source type is external
+        else if (dataSourceTypes().enumeration(dataSourceType) == External)
+        {
+            dataSourceType_ = External;
+            // Read the file and format
+            externalDataSource_.deserialise(node.at("source"), coreData);
+            // Set the data name as root filename
+            dataName_ = externalDataSource_.filename().substr(externalDataSource_.filename().find_last_of("/\\") + 1);
```
trisyoungs, line 178: Would it be better to just store the full filename path as `dataName_`, in case a user is trying to compare two datasets of the same name from, e.g., different simulations?

src/classes/dataSource.h
```diff
@@ -9,86 +9,221 @@
...
+    }
+
     // Write through specified LineParser
-    bool serialise(LineParser &parser, std::string_view keywordName, std::string_view prefix) const;
+    bool serialise(LineParser &parser, std::string_view keywordName, std::string_view prefix) const
+    {
+        if (!dataExists())
+        {
+            return false;
+        }
```
trisyoungs: Do we need to be so strict here? What if the user adds a new DataSource in the GUI but just leaves it empty for the time being and (for instance) disables the parent `CompareModule` so it doesn't run?  That's a valid simulation and we shouldn't complain about writing the file because the data source isn't quite set up properly.

trisyoungs: Also, you don't make the same check for the TOML version.

src/keywords/dataSource.h
```diff
@@ -13,20 +14,33 @@
...
+    }
     ~DataSourceKeyword() override = default;
 
     /*
      * Data
      */
     private:
-    // Format object for the data
-    typename DataType::Formatter format_;
+    // Reference to module data
```
trisyoungs:
```suggestion
    // Vector of data source pairs
```

src/keywords/dataSource.h
```diff
@@ -13,20 +14,33 @@
...
+    std::vector<DataPair> &dataSources_;
+    // End keyword
+    const std::string endKeyword_;
+
+    public:
+    // Returns reference to module data
```
trisyoungs: Return data source pairs

src/keywords/dataSourceBase.cpp
```diff
@@ -1,10 +1,3 @@
...
-DataSourceKeywordBase::DataSourceKeywordBase(std::vector<DataSourceKeywordBase::DataPair> &dataSources,
-                                             std::string_view endKeyword)
-    : KeywordBase(typeid(this)), dataSources_(dataSources), endKeyword_(endKeyword)
-{
-}
-
-// Returns reference to module data
-std::vector<DataSourceKeywordBase::DataPair> &DataSourceKeywordBase::dataSources() { return dataSources_; }
\ No newline at end of file
+DataSourceKeywordBase::DataSourceKeywordBase() : KeywordBase(typeid(this)) {}
```
trisyoungs, line 3: This feels redundant. Is it not enough to call the KeywordBase constructor from your template class?

src/modules/compare/gui/compareWidget.h
```diff
@@ -0,0 +1,48 @@
...
+    DataViewer *graph_;
+
+    public:
+    // Update controls within widget
+    void updateControls(const Flags<ModuleWidget::UpdateFlags> &updateFlags = {}) override;
+
+    /*
+     * Widgets / Functions
+     */
+
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
The end of a long journey!  Some small suggestions mostly on comments.

src/classes/dataSource.h
rhinoella: Can do, but only concern is that the full path will take a lot of space on the graph legend

src/keywords/dataSource.h
```diff
@@ -13,20 +14,33 @@
...
     private:
-    // Format object for the data
-    typename DataType::Formatter format_;
+    // Vector of data source pairs
+    std::vector<DataPair> &dataSources_;
+    // End keyword
+    const std::string endKeyword_;
+
+    public:
+    // Returns reference to module data
```
rhinoella:
```suggestion
    // Return data source pairs
```

src/keywords/dataSourceBase.cpp
rhinoella: Having the DataSourceKeywordBase class is necessary for registerProducer() (It rejects the template class without the base)

src/classes/dataSource.h
```diff
@@ -75,6 +75,15 @@ template <typename DataType> class DataSource : public Serialisable<const CoreDa
         }
     }
 
+    std::string_view getFilepath()
+    {
+        if (dataSourceType_ == External)
+        {
+            return externalDataSource_.filename();
+        }
+        return "";
```
trisyoungs: May be able to get away with a ternary one-liner here (but GH won't let me make a suggestion in line) provided the types match up:
```
return dataSourceType_ == External ? externalDataSource_.filename() : "";
```

src/keywords/dataSource.h
```diff
@@ -38,7 +38,7 @@ template <class DataType> class DataSourceKeyword : public DataSourceKeywordBase
     // End keyword
     const std::string endKeyword_;
     // Gets path basename
-    std::string getBasename(std::string_view filename) const { return filename.substr(filename.find_last_of("/\\") + 1) }
+    std::string_view getBasename(std::string_view filename) const { return filename.substr(filename.find_last_of("/\\") + 1); }
```
trisyoungs, line 41: :+1:

PR #1664, opened 2023-09-08
EmpiricalFormula - The Next Generation

trisyoungs:
This unexpected PR updates the `EmpiricalFormula` functions to be a little more useful, principally the ability to generate a formula without a `Range` of some objects, which has become a requirement in #1607.

The pre-sized counting `std::vector` is replaced with a `std::map` since it is ordered by default, and a simple reverse iterator over the map generates the formula string in order of decreasing atomic mass. All in all, a very pleasing little bit of code.


rhinoella (approved):
Looks great! Appreciate the efficiency 👍

PR #1666, opened 2023-09-14
Add default image to image provider

rprospero:
If we always return an image, Qt doesn't complain when we don't.


jws-1 (approved):
This looks good to me, I'm happy to merge this straight in to `beautify-overview-01` and deal with the QC there.

PR #1672, opened 2023-09-25
Beautify Overview (Part 1)

jws-1:
This PR implements the first pass of beautification of the Overview Tab, with the main goal of ensuring that nodes are properly layed out and resized.

### DIFF
--- a/src/gui/qml/overview/OverviewTab.qml
+++ b/src/gui/qml/overview/OverviewTab.qml
@@ -13,23 +13,24 @@ ColumnLayout {
     signal masterTermsClicked
     signal moduleLayerClicked(int index)
     signal speciesClicked(int index)
-    anchors.fill: parent
 
+    anchors.fill: parent
 
     RowLayout {
         Layout.alignment: Qt.AlignHCenter
-        Layout.fillWidth: true
         Layout.fillHeight: true
+        Layout.fillWidth: true
         Layout.preferredHeight: 100
+
         AtomTypes {
-            nAtomTypes: dissolveModel.nAtomTypes
+            Layout.fillHeight: true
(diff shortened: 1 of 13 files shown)


src/gui/qml/overview/OverviewTab.qml
```diff
@@ -25,6 +38,12 @@ ColumnLayout {
             }
         }
         Bonds {
+            Layout.fillHeight: true
+            Layout.fillWidth: true
+            Layout.maximumHeight: 200
+            Layout.maximumWidth: 200
+            Layout.preferredHeight: 100
+            Layout.preferredWidth: 200
```
trisyoungs, line 46: Since this is repeated for all intramolecular types, can these properties be captured in a subclass / custom widget type of way?

trisyoungs (approved):
Definitely looking better.  Just a single question on perhaps reusing property definitions in some way, but that's about it from me!

src/gui/qml/overview/OverviewTab.qml
jws-1: That'd be nice, although it seems harder to do than it sounds - I'll look into it a bit more though.

jws-1: It turns out that this is quite difficult to do, since the `Layout...` properties are "attached" by the `RowLayout` that the various intramolecular objects live in. This means we couldn't, as far as I can tell, have a base `Intramolecular` class with these properties that we then inherit from, as the attached properties wouldn't be defined...

trisyoungs: I suspected it might be!  Time to move on I guess...

PR #1678, opened 2023-10-13
Beginning TOML Testing

rprospero:
This PR accomplishes two things:

1. It adds a small framework for tagging tests.  We can mark tests with a flag to enable them to load and run slightly differently.  This enables us to add a *failing* test without breaking the CI (simply flag the test and don't run it).
2. It adds basic TOML testing to the main CI.  This ensures that no future commit will introduce regressions in the TOML parsing and serialisation code.  Three tests are flagged as failing, as they are known bugs.  TOML cannot be merged until these three flags are removed.

What this PR does *not* accomplish is the *full* testing framework where the TOML code.  Specifically, we test that we can round trip the file through TOML, but it does *not* confirm that the TOML code produces identical results to the input file.  That testing will be added in a follow-up PR, but I wanted this PR in so to prevent regressions before adding progress.


tests/testData.h
```diff
@@ -17,6 +17,44 @@
...
+// of two in order for masks to be composable.
+//
+// Eventually, once we move to C++23, this can be replaced by a
+// std::bitset, but bitset isn't constexpr until then.
+enum TestFlags
+{
+    TomlFailure = 1, // tests where the TOML testing is known to fail
+};
+
+void compare_toml(std::string location, SerialisedValue toml, SerialisedValue toml2)
```
trisyoungs: This Pythonically-named function doesn't conform to the style of the rest of the code, so suggest changing to:
```suggestion
void compareToml(std::string location, SerialisedValue toml, SerialisedValue toml2)
```

tests/testData.h
```diff
@@ -69,8 +106,23 @@ class DissolveSystemTest
...
         if (!dissolve_.prepare())
             throw(std::runtime_error("Failed to prepare simulation.\n"));
+
+        if (dissolve_.toml_testing_flag || !(flags & TomlFailure))
+        {
+            auto toml = dissolve_.serialise();
+
+            CoreData other_;
+            Dissolve trial_{other_};
+            // dissolve_.clear();
```
trisyoungs:
```suggestion
```

tests/testData.h
```diff
@@ -69,8 +106,23 @@ class DissolveSystemTest
...
+
+            trial_.deserialise(toml);
+            trial_.setInputFilename(std::string(inputFile));
+            auto repeat = trial_.serialise();
+            compare_toml("", toml, repeat);
+        }
     }
-    void setUp(std::string_view inputFile, const std::function<void(Dissolve &D, CoreData &C)> &additionalSetUp)
+    void setUp(std::string_view inputFile, const std::function<void(Dissolve &D, CoreData &C)> &additionalSetUp,
+               bool known_toml_failure = false)
```
trisyoungs, line 124: Again, a very Pythonic name!

trisyoungs (approved):
Some small comments regarding names, but that's it!

PR #1679, opened 2023-10-22
Reorganise CIF Import GUI, Molecular Supercell Generation & Configuration Outputs

### LINKED ISSUE #1575: Handle molecular supercell generation
Out molecular CIF handling needs to also deal with supercells.

jws-1:
This PR fixes/enhances various parts of the CIF import process. In particular:
- Fixed bug preventing progression through the wizard.
- Removed `Species Partitioning` page, replaced with an `Output` page where the partitioning can be performed.
- The `Output` page gives a preview of the `Configuration` that will be outputted.
- Through the `Output` page, a supercell can be created, the partitioning of the species can be changed and a configuration can be optionally outputted (for the molecular and framework cases).
- The molecular species that will be created are also listed.
- Supercell generation has been extended to support molecular species (closes #1575).

This definitely needs thorough testing and analysis to ensure that the initial intentions of the code have not been lost!


trisyoungs (changes requested):
Nice. Overall I think this puts us exactly where we want to be, but I spotted one significant thing during testing, and that is that when generating a supercell of molecules we don't handle the bonding correctly, or at least we aren't storing and using the "reassembled" molecular fragments in the first place:
image
(This output is from 251754.cif, with a 2,1,1 supercell).

A further small thing is that we default to a framework species on the final output page, but if we know we have molecular species in the system we should default to generating molecular output.

trisyoungs (approved):
:+1: 

This has highlighted some other things we can do, but I have captured these as separate issues as this PR is already big enough.

PR #1680, opened 2023-10-23
Fix import of node name when deserialising

rprospero:
This fixes the last three round trip issues with TOML.  I've marked those tests as working.


trisyoungs (approved):
Ah, obvious when you see it! At least from my perspective as the author of the "breaking changes" in that sense.

Good stuff.  :+1:

PR #1681, opened 2023-10-26
Windows docker setup

RobBuchananCompPhys:
This PR aims to provide documentation for getting Dissolve set up on a Windows system using Docker.
As well as a Dockerfile, it also includes a small change to the `flake.nix` file including a wrapper package, nixGLIntel, to help with running the software in the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+# Nix builder
+FROM nixos/nix:latest AS builder
+
+RUN echo "substituters = https://cache.nixos.org https://dissolve-nix.cachix.org" > /etc/nix/nix.conf
+RUN echo "trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs= dissolve-nix.cachix.org-1:b7kp954+mmuowr7XZsAIBugv+xp7+n3aSF/5obidRDE=" >> /etc/nix/nix.conf
+RUN echo "experimental-features = nix-command flakes" >> /etc/nix/nix.conf
+RUN /root/.nix-profile/bin/nix build github:disorderedmaterials/dissolve#devShells.x86_64-linux.default
+RUN nix-env -iA nixpkgs.xorg.xauth nixpkgs.xorg.xorgserver
+VOLUME /dissolve
+WORKDIR /dissolve
+
+CMD "/root/.nix-profile/bin/nix develop --extra-experimental-features nix-command --extra-experimental-features flakes /dissolve#"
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Nix builder
+FROM nixos/nix:latest AS builder
+
+RUN echo "substituters = https://cache.nixos.org https://dissolve-nix.cachix.org" > /etc/nix/nix.conf
+RUN echo "trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs= dissolve-nix.cachix.org-1:b7kp954+mmuowr7XZsAIBugv+xp7+n3aSF/5obidRDE=" >> /etc/nix/nix.conf
+RUN echo "experimental-features = nix-command flakes" >> /etc/nix/nix.conf
+RUN /root/.nix-profile/bin/nix build github:disorderedmaterials/dissolve#devShells.x86_64-linux.default
+RUN nix-env -iA nixpkgs.xorg.xauth nixpkgs.xorg.xorgserver
```
rprospero: Are we sure that this line is required?  Since XMing acts as our server, I wouldn't think that we should need it on the guest.  Similar for xauth, since we never established authentication anyway.

web/docs/userguide/developers/docker.md
```diff
@@ -0,0 +1,64 @@
...
+For use of the dissolve GUI software, we will need to have installed an X11-forwarding utility on the Windows host system.
+This workflow uses the [Xming X Server](http://www.straightrunning.com/XmingNotes/).
+
+## Setting up the environment
+
+### 1) Build the Docker image
+
+In the dissolve code, open a command prompt of your choice (for instance cmd.exe, powershell, or git bash) and run
+```shell
+Docker build -t dissolve .
```
rprospero:
```suggestion
docker build -t dissolve .
```

I believe that docker has to be lower case to work in `git bash` and I thought that the other two shells were case sensitive, but I may be wrong on this point.

web/docs/userguide/developers/docker.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+This will run a new instance of the built image in an interactive terminal - you should now see the container shell running.
+
+Notice that this command maps the IP address of the host machine to the container (a requirement for X11 forwarding of graphics to take place) as well as mapping the dissolve code into the container. 
+
+### 4) Launch the dissolve GUI
+
+Moving over to the interactive terminal of the container, we should see that we are in the container `/dissolve` directory.
+Run `nixGLIntel ./build/bin/dissolve-gui` to start the software.
```
rprospero: That command should fail, since we haven't actually built dissolve yet.  The user will first need to run

```shell
cmake --preset GUI-nix
cmake --build build
```

rprospero (changes requested):
A couple of bits of the documentation that might need some checking.

web/docs/userguide/developers/docker.md
RobBuchananCompPhys: I can confirm that `Docker` works (for me atleast) with git bash, but happy to change to lowercase `docker` for generality.

Dockerfile
RobBuchananCompPhys: I tried this previously without those packages and the forwarding didn't seem to work but I'm having another look at it now.

RobBuchananCompPhys: Looks like we didn't need that line

scripts/setup.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+dos2unix flake.nix
+"/root/.nix-profile/bin/nix" "develop" "--extra-experimental-features" "nix-command" "--extra-experimental-features" "flakes" "/dissolve#"
```
rprospero, line 3:
```suggestion
"/root/.nix-profile/bin/nix" "develop" "/dissolve#"
```
Since we added the nix-command and flakes features to nix.conf in line six of the docker file, we shouldn't need to specify them in the script.

rprospero (approved):
One optional change to clean the code, but otherwise looks good.

PR #1682, opened 2023-10-26
Correctness testing with TOML

rprospero:
Now that TOML properly performs the round trip, we need to test for correctness.  This PR performs the following changes

1. For each test, we convert the input file to TOML and then run the tests on the imported TOML.  This confirms that the TOML still produces the correct results
  a. We still perform the round trip testing to prevent regressions
  b. Tests tagged with TomlFailure only run the original input file unless the toml testing flag is activated.  This marks the 17 tests currently failing under TOML
2. The TOML serialisation and deserialisation no longer throws an exception when the TOML testing flag is inactive, since we are now using these procedures during the normal testing builds.
3. We now perform proper TOML serialisation for ForceChargeSource and AutoChargeSource in the pair potentials.  This ten line fix saves flagging the fifty tests that would fail without it.

### DIFF
--- a/src/classes/serializablePairPotential.cpp
+++ b/src/classes/serializablePairPotential.cpp
@@ -5,10 +5,10 @@
 #include "classes/serializablePairPotential.h"
 
 SerializablePairPotential::SerializablePairPotential(double &range, double &delta, bool &source,
-						     bool &forceCharge, 
+						     bool &forceCharge, bool &autoCharge,
                                                      std::vector<std::shared_ptr<AtomType>> &types)
     : range_(range), delta_(delta), atomTypeChargeSource_(source), atomTypes_(types),
-      forceCharge_(forceCharge), 
+      forceCharge_(forceCharge), autoCharge_(autoCharge),
       coulombTruncationScheme_(PairPotential::coulombTruncationScheme_),
       shortRangeTruncationScheme_(PairPotential::shortRangeTruncationScheme_){};
 
@@ -47,6 +47,7 @@ SerialisedValue SerializablePairPotential::serialise() const
         {"delta", delta_},
         {"includeCoulomb", atomTypeChargeSource_},
         {"forceChargeSource", forceCharge_},
+        {"autoChargeSource", autoCharge_},
         {"coulombTruncation", PairPotential::coulombTruncationSchemes().serialise(coulombTruncationScheme_)},
         {"shortRangeTruncation", PairPotential::shortRangeTruncationSchemes().serialise(shortRangeTruncationScheme_)}};
     for (auto &atomType : atomTypes_)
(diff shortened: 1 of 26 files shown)


tests/testData.h
```diff
@@ -84,47 +84,83 @@ class DissolveSystemTest
...
-                throw(std::runtime_error(fmt::format("Input file '{}' failed to rewrite correctly.\n", inputFile)));
-
-            dissolve_.clear();
-            if (!dissolve_.loadInput(newInput))
-                throw(std::runtime_error(fmt::format("Input file '{}' failed to reload correctly.\n", newInput)));
-        }
+            SerialisedValue toml;
+            {
+                CoreData other_;
+                Dissolve trial_{other_};
```
trisyoungs: These names are perhaps a bit too generic!  I might suggest:
```suggestion
                CoreData otherCoreData_;
                Dissolve otherDissolve_{otherCoreData_};
```

src/classes/serializablePairPotential.cpp
```diff
@@ -4,10 +4,10 @@
 
 #include "classes/serializablePairPotential.h"
 
-SerializablePairPotential::SerializablePairPotential(double &range, double &delta, bool &source,
-                                                     std::vector<std::shared_ptr<AtomType>> &types)
-    : range_(range), delta_(delta), atomTypeChargeSource_(source), atomTypes_(types),
-      coulombTruncationScheme_(PairPotential::coulombTruncationScheme_),
+SerializablePairPotential::SerializablePairPotential(double &range, double &delta, bool &source, bool &forceCharge,
```
trisyoungs, line 7: Do we still need the `SerializablePairPotential` class?  It feels like its just wrapping functionality that could be added into `Dissolve::serialise()` and `Dissolve::deserialise()`...

trisyoungs (approved):
All looks good.  Those failing tests are somewhat annoying, though!

Couple of comments, nothing major.

src/classes/serializablePairPotential.cpp
```diff
@@ -57,6 +59,8 @@ void SerializablePairPotential::deserialise(const SerialisedValue &node)
     range_ = toml::find_or<double>(node, "range", 15.0);
     delta_ = toml::find_or<double>(node, "delta", 0.005);
     atomTypeChargeSource_ = toml::find_or<bool>(node, "includeCoulomb", false);
+    forceCharge_ = toml::find_or<bool>(node, "forceChargeSource", false);
+    autoCharge_ = toml::find_or<bool>(node, "autoChargeSource", false);
```
trisyoungs: The default value here should be `true` in order to be consistent with what we do in, for example, `Dissolve::clear()`.
```suggestion
    autoCharge_ = toml::find_or<bool>(node, "autoChargeSource", true);
```

tests/testData.h
rprospero: I've gone ahead and renamed them `otherCoreData` and `otherDissolve`.  I dropped the trailing underscore because these aren't members of the class, but just variables in the function.

src/classes/serializablePairPotential.cpp
rprospero: I agree that we need to review `SerializablePairPotential`, as I'm increasingly unhappy with the design.  However, I'm not sure that `Dissolve` is the right place for it, either.  I'll add an issue for it and we can review that before the final TOML addition.

PR #1684, opened 2023-11-02
EPSR Prune AtomTypes mk2

trisyoungs:
This PR supersedes #931, and modifies the EPSR module to only attempt to generate potentials for atom types present in the `Configuration` targeted by the module's reference datasets, rather than doing it for every defined atom type in `CoreData`.

The original emphasis of this work was to circumvent crashes when unused atom types were present, but since we now prune these before starting the simulation this should no longer be relevant.  A much more relevant, current focus is the hope that this will address crashes when legitimately running multi-configuration simulations where atom types are not guaranteed to be present in all `Configuration`s all of the time.


rprospero (approved):

Danbr4d (approved):
Works for the simpler simulations, but still getting a crash on the Multi-comp.

PR #1692, opened 2023-11-07
Fix Add Forcefield Dialog

rprospero:
The Add Forcefield Dialog has a bug where that causes a crash if there are no Master terms.  Since no values were ever added to the node while serialising, the node was never properly registered as a `table`.  I've now intialised it to an empty table so that the deserialiser can correctly load the result.


RobBuchananCompPhys (approved):
Nice!

trisyoungs (approved):
:+1:

PR #1695, opened 2023-11-09
Building & Packaging for OSX Apple Silicon

rhinoella:


rhinoella:
Due to the inefficiency of this method, it has been decided to abandon this PR and focus instead on upgrading ANTLR4 and TBB.

PR #1701, opened 2023-11-14
Fix serialisation of Species Site with TOML

rprospero:
Support for Dynamic and Fragment sites has now been added.  Please note that

* Dynamic Sites support a *list* of `element`.
* Fragment Sites take a `description` string
* *All* site types support an `originMassWeighted` toggle

I'm happy to add other parameters if they are need.  These were just the ones required to pass the tests.

### DIFF
--- a/src/classes/speciesSite.cpp
+++ b/src/classes/speciesSite.cpp
@@ -749,11 +749,17 @@ bool SpeciesSite::write(LineParser &parser, std::string_view prefix)
 SerialisedValue SpeciesSite::serialise() const
 {
     SerialisedValue site;
+    if (originMassWeighted_)
+        site["originMassWeighted"] = originMassWeighted_;
     switch (type_)
     {
         case SiteType::Dynamic:
             site["type"] = "Dynamic";
             site["element"] = dynamicElements_;
+            break;
+        case SiteType::Fragment:
+            site["type"] = "Fragment";
+            site["description"] = fragment_.definitionString();
             break;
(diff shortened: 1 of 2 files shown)


src/classes/speciesSite.cpp
```diff
@@ -749,28 +749,64 @@ bool SpeciesSite::write(LineParser &parser, std::string_view prefix)
...
 
 void SpeciesSite::deserialise(const SerialisedValue &node, CoreData &coreData)
 {
-    if (node.contains("dynamic"))
+    auto typeString = toml::find_or<std::string>(node, "type", "Static");
+    if (typeString == "Static")
+        type_ = SiteType::Static;
+    else if (typeString == "Fragment")
+        type_ = SiteType::Fragment;
+    else if (typeString == "Dynamic")
```
trisyoungs, line 783: Rather than going to the effort of putting this string check in here it would be generally more useful to put in an `EnumOptions` associated with the `SiteType`. That would necessitate reducing it to a simple `enum` rather than an `enum class`.  Small thing though, and not related to TOML, so feel free to leave or capture as a new issue.

trisyoungs (approved):
Just one comment, otherwise all looks nice.

src/classes/speciesSite.cpp
rprospero: I've added issue #1706 to track this and I'm adding it to my master issue for putting TOML before the users.

PR #1711, opened 2023-11-15
Add EnumOption for SpeciesSite::SiteType

### LINKED ISSUE #1706: Create EnumOptions for SpeciesSite::SiteType
As mentioned by @trisyoungs in #1701 , the SiteType should be an EnumOptions to simplify serialisation and also future-proof the code.

rprospero:
This closes #1706

### DIFF
--- a/src/classes/speciesSite.cpp
+++ b/src/classes/speciesSite.cpp
@@ -23,6 +23,13 @@ SpeciesSite::SpeciesSite(const Species *parent, std::string name, SiteType type)
  * Basic Information
  */
 
+// Enum of types
+EnumOptions<SpeciesSite::SiteType> SpeciesSite::types()
+{
+    return EnumOptions<SiteType>(
+        "SiteType", {{SiteType::Static, "static"}, {SiteType::Dynamic, "dynamic"}, {SiteType::Fragment, "fragment"}});
+}
+
 // Set name of site
 void SpeciesSite::setName(std::string_view newName) { name_ = newName; }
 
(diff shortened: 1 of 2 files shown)


src/classes/speciesSite.h
```diff
@@ -25,12 +25,15 @@ class SpeciesSite : public Serialisable<CoreData &>
...
     // Site Type
-    enum class SiteType
+    enum SiteType
     {
         Static,  /* Site is based on fixed atom indices within the species */
         Dynamic, /* Site is atomic and based on elements and atom types */
         Fragment /* Site is based on a NETA description */
     };
+    // EnumOptions for SiteType
+    static EnumOptions<SiteType> types();
```
trisyoungs: To be consistent with the rest of the code:
```suggestion
    static EnumOptions<SiteType> siteTypes();
```

src/classes/speciesSite.cpp
```diff
@@ -23,6 +23,13 @@ SpeciesSite::SpeciesSite(const Species *parent, std::string name, SiteType type)
  * Basic Information
  */
 
+// Enum of types
+EnumOptions<SpeciesSite::SiteType> SpeciesSite::types()
+{
+    return EnumOptions<SiteType>(
+        "SiteType", {{SiteType::Static, "static"}, {SiteType::Dynamic, "dynamic"}, {SiteType::Fragment, "fragment"}});
```
trisyoungs: Now, here's an annoying point for discussion: historically all enum strings had first letter capitalised so that the inputfile (.txt) looked "nice", but of course moving away from that we don't really care, and the lowercase first letter is of course more consistent with the TOML styling we're moving towards.  However, as just discussed in the stand-up, capitalising enums to distinguish them from variables is a sound way forward!

trisyoungs (approved):
Couple of comments, as discussed.

:+1:

PR #1713, opened 2023-11-15
Drop AtomType, Master term count properties

### LINKED ISSUE #1655: Drop Master Term / AtomType count properties in DissolveModel
Currently, the `DissolveModel` stores a property for the atom types and each master term that indicates their counts, which exactly corresponds to the `rowCount()` method on each of their respective models. It'd be better if we dropped these properties, and called `rowCount()` on each model explicitly from within the QML, when we want to display such numbers.

### LINKED ISSUE #1656: Registering QML Types
Currently, there are two places in Dissolve where QML is utilised - the `AddForcefieldTermsDialog` and the `OverviewTab`. Each of these instances necessitate that some C++ types are registered with QML (the `DissolveModel` and `AddForcefieldDialogModel`. Currently, these types are registered in the respective C++ constructors, however as Dissolve begins to lean towards QML, it'd be better if we had a central method that registers the various types upon loading the GUI.

RobBuchananCompPhys:
Closes #1655
Closes #1656


src/gui/models/atomTypeModel.h
```diff
@@ -15,6 +15,9 @@ class AtomTypeModel : public QAbstractListModel
 {
     Q_OBJECT
 
+    // The number of rows
+    Q_PROPERTY(int rowCount READ rowCount NOTIFY modelsUpdated)
```
rprospero, line 20: We notify on the `modelsUpdated` signal, but I don't see where we ever actually emit the signal, so I don't think that the row count will ever be updated.  What we might want to do is bind a slot to all of the [pre-existing](https://doc.qt.io/qt-6/qabstractitemmodel.html#signals) signals that get thrown when the model changes (e.g. `modelReset`, `rowsRemoved`) and then emit a `modelsUpdated` when any of those signals occur.

rprospero (changes requested):

src/gui/qml/overview/OverviewTab.qml
```diff
@@ -8,6 +8,13 @@ import "../widgets" as D
 
 ColumnLayout {
     id: root
+    property AtomTypeModel atomTypeModel
+    property MasterAngleModel masterAngleModel
+    property MasterBondModel masterBondModel
+    property MasterImproperModel masterImproperModel
+    property MasterTorsionModel masterTorsionModel
```
rprospero: I'm also getting an error message that none of these types have been registered with the QML.  This might be a good time to think about issue #1656

src/gui/models/atomTypeModel.h
RobBuchananCompPhys: I've written a class to take care of connected up all these signals since there are quite a few that could change due to the QAbstractItemModel and so it feels like it makes sense to avoid repeating for each model - let me know what you think.

src/gui/models/modelUpdater.h
```diff
@@ -0,0 +1,17 @@
+#include <QAbstractItemModel>
+
+class ModelUpdater
+{
+    Q_Object
```
rprospero:
```suggestion
    Q_OBJECT
```

rprospero: The macro names are case sensitive

src/gui/models/modelUpdater.h
```diff
@@ -0,0 +1,17 @@
+#include <QAbstractItemModel>
```
rprospero, line 6: Make sure to include to the SPDX file header.  Also, the header file will need the once pragma to ensure it doesn't get double included.

src/gui/models/modelUpdater.cpp
```diff
@@ -0,0 +1,26 @@
+#include "gui/models/modelUpdater.h"
```
rprospero, line 4: SPDX file header

src/gui/models/atomTypeModel.h
```diff
@@ -15,6 +16,9 @@ class AtomTypeModel : public QAbstractListModel
...
+
     public:
     AtomTypeModel() = default;
     explicit AtomTypeModel(const CoreData &coreData);
```
rprospero, line 28: So this is a problem because there's some places where we need to assign the AtomTypeModel at Runtime.  I can't give you a simple patch for this, but you'll want to set this up so that AtomTypeModel can still use the default constructor.

I know that this may be considered sacriligeous, but it may be worth doing this with multiple inheritance instead of composition.  Having the classes inherit from ModelUpdater means that we can just call the `connectModelSignals` whenever it's convenient for the class at hand.  Note that this is just a theory - I'm happy to stick with composition if you can make it work.

RobBuchananCompPhys: Ok, that's something I can have a go at. What is exactly the crux of of the problem, though? Is it that we have not assigned an instance of `modelUpdater` when it is required by the `Q_PROPERTY`?

rprospero (approved):

PR #1714, opened 2023-11-15
Upgrade TBB and ANTLR

### LINKED ISSUE #1712: Upgrade to oneTBB 21.10.0
Currently Dissolve is using a depreciated version of TBB (2020.3) which does not natively support Apple Sillicon.

rhinoella:
Change from the depreciated TBB (2020.3) to the new oneAPI version (2021.10.0). as well as ANTLR from version 4.9.3 to 4.13.1. Would allow for Dissolve to work on Apple Sillicon.

Note:
Due to naming conflicts `QT_NO_KEYWORDS` was set meaning we can't use the QT generated MOC keywords anymore such as `slots` or `emit`. These now have to be replaced with `Q_SLOTS` and `Q_EMIT`, respectively.

Closes #1712


.github/workflows/build/osx/action.yml
```diff
@@ -61,15 +56,6 @@ runs:
...
 
-  - name: Retrieve Conan Cache
-    id: cache-conan
-    uses: actions/cache@v3
-    with:
-      key: osx-conan-${{ env.conanHash }}
-      path: |
-        ~/.conan
-        ~/.conancache
-
```
trisyoungs: Presumably this should be reinstated, and the old cache files stored on GitHub deleted and regenerated?

.github/workflows/build/osx/action.yml
```diff
@@ -92,18 +78,19 @@ runs:
       export PATH="$(python3 -m site --user-base)/bin:$PATH"
       Qt6_DIR=${{ runner.temp }}/qt/${{ env.qtVersion }}/macos/lib/cmake/Qt6
       QT_BASE_DIR=${{ runner.temp }}/qt/${{ env.qtVersion }}/macos
-      ANTLR_EXE=`pwd`/antlr.jar
+      ANTLR_EXE=/usr/local/Cellar/antlr/4.13.1/antlr-4.13.1-complete.jar
```
trisyoungs: A magic path!  :)  What we should really do is add our desired ANTLR version to `.github/workflows/setup/action.yml` and grab it direct from GitHub, rather than relying on homebrew to give us the jar.

.github/workflows/build/windows/action.yml
```diff
@@ -123,6 +123,16 @@ runs:
       cmake ../ftgl-latest -G Ninja -DCMAKE_BUILD_TYPE:STRING="Release" -DCMAKE_C_COMPILER=cl -DCMAKE_CXX_COMPILER=cl -DCMAKE_INSTALL_PREFIX:path=../ftgl-install
       cmake --build . --target install --config Release
 
+  - name: Get ANTLR and Java
+    working-directory: ${{ runner.temp }}
+    shell: bash
+    run: |
+      set -ex
+      wget https://www.antlr.org/download/antlr-4.13.1-complete.jar -Oantlr-4.13.1-complete.jar
```
trisyoungs: Exactly, like this!  Still good to add a setup variable with the version in though.

.github/workflows/package/osx/action.yml
```diff
@@ -41,13 +41,17 @@ runs:
       Qt6_ROOT=${{ runner.temp }}/qt/${{ env.qtVersion }}/macos/
       export PATH="$(python3 -m site --user-base)/bin:$PATH"
 
-      # Handle ANTLR
-      cp ./deploy/antlr4-cppruntime/lib/*.dylib /usr/local/lib/
-      cp ./deploy/tbb/lib/*.dylib /usr/local/lib/
+      # Handle ANTLR & TBB
+      cp ./deploy/antlr4-cppruntime/lib/libantlr4-runtime.4.13.1.dylib /usr/local/lib/libantlr4-runtime.4.13.1.dylib
```
trisyoungs:
```suggestion
      cp ./deploy/antlr4-cppruntime/lib/libantlr4-runtime.*.dylib /usr/local/lib
```

.github/workflows/pr.yml
```diff
@@ -33,8 +33,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest]
+        include:
+          - os: macos-latest
+            cmake_target: x86_64
+            conan_target: x86_64
```
trisyoungs: Do you think it would be worth renaming these parameters, and the `env:` entries, here and in the other workflows to show that they are OSX-specific?

ci/windows/dissolve-gui.iss
```diff
@@ -52,6 +52,11 @@ Source: "{#FTGLDir}\ftgl.dll"; DestDir: "{app}\bin"
 Source: "{#QtDir}\bin\Qt6Gui.dll"; DestDir: "{app}\bin"; Flags: ignoreversion
 Source: "{#QtDir}\bin\Qt6Core.dll"; DestDir: "{app}\bin"; Flags: ignoreversion
 Source: "{#QtDir}\bin\Qt6OpenGL.dll"; DestDir: "{app}\bin"; Flags: ignoreversion
+Source: "{#QtDir}\bin\Qt6Qml.dll"; DestDir: "{app}\bin"; Flags: ignoreversion
```
trisyoungs, line 55: Eeek! That's one to get into 1.4 as well!

cmake/Modules/FindANTLR.cmake
```diff
@@ -2,11 +2,11 @@
 # - From https://github.com/antlr/antlr4/ at 6ad9646
 # - Modified by TGAY to find and use JPackage antlr4 script if present
 
-find_package(Java QUIET COMPONENTS Runtime)
+find_package(Java 11 QUIET COMPONENTS Runtime)
 
 if(NOT ANTLR_EXECUTABLE)
   find_program(ANTLR_EXECUTABLE
-               NAMES antlr4 antlr.jar antlr4.jar antlr-4.jar antlr-4.8-complete.jar)
+               NAMES antlr4 antlr.jar antlr4.jar antlr-4.jar antlr-4.13.1-complete.jar)
```
trisyoungs: Hmmm. I want to say that having the version-specific option is unnecessary since, when required, we set `ANTLR_EXECUTABLE` to exactly that anyway, in which case this check isn't performed.

conanfile.txt
```diff
@@ -3,16 +3,15 @@ cli11/1.9.1
 fmt/8.1.1
 gtest/1.10.0
 pugixml/1.11
-tbb/2020.3
-onedpl/20200330
+onetbb/2021.10.0
+onedpl/2021.7.0
 toml11/3.7.0
-antlr4-cppruntime/4.9.3
+antlr4-cppruntime/4.13.1
```
trisyoungs, line 9: Since the version of ANTLR4 we specify here is the same in the Java executable we need to download, in the `setup` action you could scan `conanfile.txt` and extract the version automatically, rather than hardcoding one in...

src/templates/parallelDefs.h
```diff
@@ -7,14 +7,14 @@ struct FakeParallelPolicy
...
+#include <oneapi/dpl/execution>
+#include <oneapi/dpl/numeric>
 namespace ParallelPolicies
 {
-constexpr auto par = pstl::execution::par;
-constexpr auto seq = pstl::execution::seq;
-constexpr auto par_unseq = pstl::execution::par_unseq;
+constexpr auto par = dpl::execution::par;
+constexpr auto seq = dpl::execution::seq;
+constexpr auto par_unseq = dpl::execution::par_unseq;
```
trisyoungs, line 17: Is that the only change required for OneTBB?  Well that was easy! :)

trisyoungs (changes requested):
Looking very good indeed.  I quite like it now that the Qt stuff (e.g. Q_SIGNALS, Q_EMIT) stand out from "standard" C++, rather than insidiously sneak in!

Couple of queries and suggestions for harmonising up the workflows.

.github/workflows/package/osx/action.yml
```diff
@@ -41,13 +41,17 @@ runs:
       Qt6_ROOT=${{ runner.temp }}/qt/${{ env.qtVersion }}/macos/
       export PATH="$(python3 -m site --user-base)/bin:$PATH"
 
-      # Handle ANTLR
-      cp ./deploy/antlr4-cppruntime/lib/*.dylib /usr/local/lib/
-      cp ./deploy/tbb/lib/*.dylib /usr/local/lib/
+      # Handle ANTLR & TBB
+      cp ./deploy/antlr4-cppruntime/lib/libantlr4-runtime.*.dylib /usr/local/lib
+      cp ./deploy/antlr4-cppruntime/lib/libantlr4-runtime.dylib /usr/local/lib/libantlr4-runtime.dylib
```
trisyoungs, line 46: Do we need both of these?

trisyoungs (approved):
:+1: Just one small question.

PR #1715, opened 2023-11-16
1614 explicit toml context

### LINKED ISSUE #1614: Add explicit type for TOML context
Pr #1609 introduced an issue where calling
`Vec3<NodeValue>::deserialise(node)` will throw an except because
the context is missing.  The correct behaviour is to call
`Vec3<NodeValue>::deserialise(node, context)` , but the compiler
cannot enforce this behaviour.  `Vec3` currently has no context
template parameter because it is not possible to determine the
context type for a given type.

The proposal is to create `SerialisableContext`

```
template <typename T> struct SerialisableContext
{
    using type = typename T::Context;
};
```

We would then define the `SerialisableContext` for each type.  This
would enable container types to correctly pass on the arguments of
their children.  Beyond fixing the land mine in `Vec3`, the following
may be possible, depending on the success of the implementation:

- Make the `Serialisable` class not templetated
- Make the keyword types use the correct context.  Currently, every
  `KeywordBase` has a `CoreData` argument, but only very few of the
  subclasses actually use this parameter.

rprospero:
All the real information is contained in the issue below.  The short version is that calling `deserialise(node)` on a `Vec3<NodeValue>` would cause Dissolve to crash.  We never actually called that function, but the inheritance semantics insisted that it had to exist.  I've added a phantom type `SerialisableContext` to enable us to attach a context to primitive types, not just classes.  I've also created a unit type `SerialisableEmpty` to denote data types that have no context.

Closes #1614

### DIFF
--- a/src/base/serialiser.h
+++ b/src/base/serialiser.h
@@ -31,7 +31,8 @@ template <typename... Contexts> class Serialisable
     // Read values from a serialisable value
     virtual void deserialise(const SerialisedValue &node, Contexts... context) { return; }
 
-    template <typename = std::enable_if<sizeof...(Contexts) == 1>> void deserialise(const SerialisedValue &node)
+    template <typename = std::enable_if<sizeof...(Contexts) == 1 && (std::is_empty<Contexts>::value && ...)>>
+    void deserialise(const SerialisedValue &node)
     {
         deserialise(node, {});
     }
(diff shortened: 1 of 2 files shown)


trisyoungs (approved):
Actually rather elegant. :+1:

Shame that the more general possibilities can't be easily done, but lets not do a rewrite when we don't have to!

PR #1722, opened 2023-11-20
1087 toml checking boilerplate

### LINKED ISSUE #1087: Remove optional term checking boilerplate
We have a common patter in the code of

``` c++
if node.contains("quux")
    xyzzy_ = toml::find<TYPE>(node, "quux");
```

This can be wrapped into a utility function of the Serialisable class.

rprospero:
This PR creates an `optionalOn` wrapper function to handle cases where terms in the TOML file can be optional.  The original idea was to simplify out some of the boilerplay (e.g. the parameter name only needs to be given once).  However, I'm concerned about whether this monadic style is clearer to other developers instead of just using an `if` statement.  Hence I'm asking for more reviews than usual.

This closes #1087


rhinoella (commented):
I think it is clear, however I think it might be worth renaming the function to `ifNodeContains` or `ifContains` as it provides a higher level of self documentation.

trisyoungs (commented):
I do like the replacement of bare `if`s with the new function, and I would have to argue that `optionalOn` is a more suitable name than `ifContains`.  While the latter is certainly a more descriptive name I don't think it reflects the behaviour of the function accurately enough (which is not just to check for, but to act upon) so I'd vote for keeping `optionalOn`.

trisyoungs (approved):
Looking good. :+1:

It feels like there's scope for having a template parameter which would determine the conversion of the toml node to a "useful" type for use in the lambda, rather than passing the toml node itself, but on reflection there are no doubt many edge cases (e.g. `to_string()`) which would require lots of specialisations and so this probably isn't worth it.  Certainly not in this PR, anyway!

PR #1725, opened 2023-11-22
Elapsed Timer Fix- zero on each run, update status label

### LINKED ISSUE #1707: Stopped simulation doesn't update status bar / time elapsed label
**Describe the bug**
Stopping the simulation through the main menu `Stop` doesn't update the status bar, so even when it's actually stopped it still shows "Running (ESC to stop)" and the time elapsed string.

**Operating System:**
- IDAaaS

rhinoella:
The status bar is updated so that:
- Timer resets to zero on the start of each run
- The status label bug is fixed: said Dissolve was still running despite stopped

Closes #1707.

### DIFF
--- a/src/gui/guiFuncs.cpp
+++ b/src/gui/guiFuncs.cpp
@@ -316,7 +316,7 @@ void DissolveWindow::updateStatusBar()
     }
     else if (!dissolveIterating_ && elapsedTimer_.secondsElapsed() > 0.0)
     {
-        statusLabel_->setText("Idle");
+        statusLabel_->setText("Stopped");
         statusIndicator_->setPixmap(QPixmap(":/general/icons/true.svg"));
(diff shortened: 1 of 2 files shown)


src/gui/guiFuncs.cpp
```diff
@@ -316,6 +316,8 @@ void DissolveWindow::updateStatusBar()
     }
     else if (!dissolveIterating_ && elapsedTimer_.secondsElapsed() > 0.0)
     {
+        statusLabel_->setText("Idle");
```
trisyoungs: Just a small suggestion, since this reflects when a simulation has been stopped after being actively run in the current GUI session.
```suggestion
        statusLabel_->setText("Stopped");
```

trisyoungs (approved):
:+1:

PR #1726, opened 2023-11-23
Dl poly dev

jounaidr:


CMakeLists.txt
trisyoungs, line 1: If you need to hack the `CMakeLists.txt` to build on your local machine then make sure not to commit those changes since this will break the CI in various places.  I'm not sure exactly why you had to comment these bits out since they are disabled by default, however?

flake.nix
trisyoungs, line 1: I appreciate that you added this so you could compile DL_POLY within nix, but again this is outside of the Dissolve scope and shouldn't be included in this PR.

src/io/export/CMakeLists.txt
trisyoungs, line 1: Alphabetical file order within suffix sections.

src/io/export/dlPolyField.cpp
```diff
@@ -0,0 +1,114 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "io/export/dlPolyField.h"
+#include "base/lineParser.h"
+#include "base/sysFunc.h"
+#include "classes/configuration.h"
+#include "data/atomicMasses.h"
+
```
trisyoungs, line 9:
```suggestion
```

src/io/export/dlPolyField.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+DlPolyFieldExportFileFormat::DlPolyFieldExportFileFormat(std::string_view filename, DlPolyFieldExportFormat format)
+    : FileAndFormat(formats_, filename, (int)format)
+{
+    formats_ = EnumOptions<DlPolyFieldExportFileFormat::DlPolyFieldExportFormat>(
+        "FieldExportFileFormat", {{DlPolyFieldExportFormat::DLPOLY, "dlpoly", "DL_POLY FIELD File"}});
+}
+
```
trisyoungs, line 17:
```suggestion
```

src/io/export/dlPolyField.cpp
```diff
@@ -0,0 +1,114 @@
...
+            const auto params = angle.interactionParameters();
+            if (!parser.writeLineF("{}   {}   {}   {} {} {}\n", AngleFunctions::forms().keyword(angle.interactionForm()).substr(0, 4), angle.indexI() + 1, angle.indexJ() + 1, angle.indexK() + 1, params[0], params[1]))
+                return false;
+        }       
+    }
+    
+    if (!parser.writeLineF("finish\n"))
+        return false;
+        
+    int vdw = 0;
```
trisyoungs, line 61:
```suggestion
    auto vdw = 0;
```

tests/modules/CMakeLists.txt
trisyoungs, line 1: Alphabetical file ordering.

trisyoungs (commented):
This is good - you have the guts of the thing in place.  Seeing it written out like this and the way you have leveraged the `FileAndFormat` for this purpose makes me reconsider how we should be passing data around for these more complex file formats.  Specifically, passing the long list of variables to the export function is functional, but not too pretty.  It's beyond the scope of this PR, but creating intermediate classes to store the data in (e.g. `DLPolyFieldFile`) may be a better approach long term.  This would make the DL_POLY stuff more like the implementation of Crystallographic Information File (CIF) handling, moving it beyond a simple I/O "option".

PR #1729, opened 2023-11-27
Rework TOML backend

### LINKED ISSUE #1705: Base TOML Tables on std::map
There are many places in the file format where we are using vectors, while I table would have been more ergonomic.  Unfortunately, the default table implementation for the toml11 library is `std::unordered_map`, which fails us when we need the elements loaded in a specific order.  Using `std::map` as the backend may solve this issue.

rprospero:
This PR reworks the backend `SerialisedValue` type to use an ordered map.  This enables the ability to process elements of a table in order.  The ordered map isn't as efficient as it could be, but only in a Big O sense.  This can be corrected if we ever need to process tables with hundreds of millions of elements.

Closes #1705


src/base/serialiser.h
```diff
@@ -72,7 +73,7 @@ template <typename... Contexts> class Serialisable
     static SerialisedValue fromVectorToTable(const std::vector<T> &vector, Lambda getName)
     {
         SerialisedValue group;
-        for (auto it = vector.rbegin(); it < vector.rend(); it++)
+        for (auto it = vector.begin(); it < vector.end(); it++)
```
trisyoungs: Minor thing, but is there a good reason for not using a range here?
```suggestion
        for (const auto &value : vector)
```

src/classes/box.cpp
```diff
@@ -428,8 +428,8 @@ Vec3<double> Box::scaleFactors(double requestedVolume, Vec3<bool> scalableAxes)
 SerialisedValue Box::serialise() const
 {
     SerialisedValue box;
-    box["lengths"] = toml::array{a_, b_, c_};
-    box["angles"] = toml::array{alpha_, beta_, gamma_};
-    box["nonPeriodic"] = toml::array{!periodic_.x, !periodic_.y, !periodic_.z};
+    box["lengths"] = {a_, b_, c_};
+    box["angles"] = {alpha_, beta_, gamma_};
+    box["nonPeriodic"] = {!periodic_.x, !periodic_.y, !periodic_.z};
```
trisyoungs, line 433: Worth it for this neatness alone!

src/templates/ordered_map.h
```diff
@@ -0,0 +1,87 @@
...
+// Access is still O(1) and insertion is O(log(N)).  Erasure hasn't
+// been implemented yet and will likely be O(N), unlike the O(log(N))
+// of std::map.  However, seeing as removal hasn't even been required
+// at this point, it seems unlikely that it will ever serve as a true
+// bottleneck.
+template <typename Key, typename Value> class ordered_map
+{
+
+    private:
+    // A maping between keys and indices
```
trisyoungs:
```suggestion
    // A mapping between keys and indices
```

src/templates/ordered_map.h
```diff
@@ -0,0 +1,87 @@
...
+    std::map<Key, int> index_;
+    // An association array of the keys and values
+    std::vector<std::pair<Key, Value>> values_;
+    // If a key does not exist, add it to the array
+    void insistKey_(const Key &key)
```
trisyoungs: Underscore after the name?

src/templates/ordered_map.h
```diff
@@ -0,0 +1,87 @@
...
+            if (values_[i] != other.values_[i])
+                return false;
+        for (auto [k, v] : index_)
+            if (other.index_.at(k) != v)
+                return false;
+        return true;
+    }
+    Value &operator[](const Key &key)
+    {
+        insistKey_(key);
```
trisyoungs: You could return the found/new map entry from `insistKey_()` and save a second lookup in `values_` if we wanted to be optimal here.

tests/algorithms/ordered_map.cpp
```diff
@@ -0,0 +1,30 @@
...
+namespace UnitTest
+{
+
+using SerialisedValue = toml::basic_value<toml::discard_comments, dissolve::ordered_map, std::vector>;
+TEST(OrderedMapTest, BasicOrderedMap)
+{
+    SerialisedValue example;
+    std::vector<std::string> keys = {"foo", "bar", "quux", "xyzzy"};
+    std::vector<int> values = {2, 7, 5, 3};
+    for (int i = 0; i < keys.size(); i++)
```
trisyoungs:
```suggestion
    for (auto i = 0; i < keys.size(); i++)
```

tests/algorithms/ordered_map.cpp
```diff
@@ -0,0 +1,30 @@
...
+        example[keys[i]] = values[i];
+
+    int index = 0;
```
trisyoungs:
```suggestion
    auto index = 0;
```

src/templates/ordered_map.h
```diff
@@ -0,0 +1,87 @@
...
+        if (index_.find(key) == index_.end())
+        {
+            index_[key] = values_.size();
+            values_.emplace_back(key, Value{});
+        }
+    }
+
+    public:
+    // Constructors
+    ordered_map(){};
```
trisyoungs: I have to say it, but I feel particularly pedantic for doing so! This goes against the class naming convention we use everywhere else in Dissolve, and really should be `OrderedMap` (source file then => `orderedMap.h`). However, I appreciate that this might cause readability issues in the context where its used since the other toml-related namings are `xxx_yyy`.

trisyoungs (changes requested):
It's surprising how much of a positive difference this makes to the code where its used, so a big :+1: for that.  I have a couple of comments on the naming of the class and some small suggestions.

If there is a small negative aspect to this refactor it's that the serialisation of the `name_` elements of particular classes (e.g. `Module`) is moved out of their local serialisation routines since it forms the key of the map and so it the parent's responsibility. However, that is by far the best way of expressing the data, so remind me that I said this when I complain in a future PR that somebody's forgotten to write the `name_` member out.

src/base/serialiser.h
rprospero: Range is just fine.  This was just a legacy from when we went through the vector in reverse order.

src/templates/ordered_map.h
rprospero: Force of habit, but I've changed it to match the project conventions.

src/templates/ordered_map.h
rprospero: You're right that that was a much cleaner implementation.

src/templates/ordered_map.h
rprospero: I originally went with `ordered_map` to match the standard library naming convention that had given us `unordered_map`.  However, switching to the project's naming convention hammers home the point that this is a bespoke class and not in the standard, so I will make the change.

src/templates/ordered_map.h
```diff
@@ -0,0 +1,81 @@
...
+        for (int i = 0; i < values_.size(); i++)
+            if (values_[i] != other.values_[i])
+                return false;
+        for (auto [k, v] : index_)
+            if (other.index_.at(k) != v)
+                return false;
+        return true;
+    }
+    Value &operator[](const Key &key) { return insistKey(key).second; }
+    Value &operator[](Key &&key) { return insistKey(key).second; }
```
trisyoungs: Just spotted this - assuming this should just be `&key` rather than `&&key`?

src/templates/ordered_map.h
trisyoungs, line 1: File should be named `templates/orderedMap.h` following our lowerCamelCasing of everything.

src/base/serialiser.h
```diff
@@ -95,7 +96,7 @@ template <typename... Contexts> class Serialisable
     static SerialisedValue fromVectorToMap(const std::vector<T> &vector, Lambda getName, Lambda2 getValue)
     {
         SerialisedValue group;
-        for (auto it = vector.rbegin(); it < vector.rend(); it++)
+        for (auto it = vector.begin(); it < vector.end(); it++)
```
trisyoungs: Another opportunity for a range if you wanted to.

trisyoungs (approved):
Final comments, then happy for this to go in.

 :+1:

src/templates/ordered_map.h
rprospero: The second ampersand is intentional.  There are two overloads for [] on `std::map` and I wanted to make sure that we implemented both of them.  I supposed that I could add a `std::move` in there to take advantage of the move semantics, but I'm hoping that it will never matter.

Issue #1425, opened 2023-05-11
Preamble text in keyword options group?


trisyoungs:
**Describe the bug**
A short preamble "explainer" text at the beginning of each keyword group would be useful in order to explain in light terms what the following group of controls relates to, and offers some "in-place" documentation for the users.

PR #1730, opened 2023-11-28
1425 Preamble text in options groups - Part 1

trisyoungs:
This PR lays the foundation to enable #1425 by refactoring the way we store and organise our keyword data. Previously, the `KeywordStore` class had a simple `std::vector` of `KeywordStoreData`, with each containing a pointer to a `KeywordBase` object and the section / group to use when displaying it in the GUI.  This disconnect between actual organisation and the data being organised meant that the hierarchical map of the intended layout had to be generated before being used, and required a bunch of objects such as `std::map<std::string,std::vector<std::pair<....` making the code a bit messy and hard to understand.

Here we take the approach of storing the keywords in a hierarchical class structure, introducing `KeywordStoreSection` and `KeywordStoreGroup` classes to contain the data and provide `find()` functions etc. This also leaves the door wide open for an easy implementation of #1425.

Works towards #1425.


src/gui/keywordWidgets/widgetFuncs.cpp
```diff
@@ -24,35 +24,32 @@ KeywordsWidget::KeywordsWidget(QWidget *parent) : QScrollArea(parent)
...
+    auto *sectionLayout = new QGridLayout(sectionWidget);
     auto row = 0;
-    for (auto sectionName : keywordIndexInfo.second)
+    for (auto &group : keywordSection.groups())
     {
-        // Get keyword map
-        auto &keywords = groupMap.at(sectionName);
+        // Nothing to do if this group is hidden
+        if (group.name() != "_HIDDEN")
+            continue;
```
rprospero: Am I misunderstanding the code or should that be == instead of != ?

rprospero (changes requested):
There's one line which I'm pretty sure is a bug.  If I'm wrong about that, I'm happy for it to merge.

src/gui/keywordWidgets/widgetFuncs.cpp
trisyoungs: Well, that's embarrassing! This should indeed be `==`.

src/gui/keywordWidgets/dialogFuncs.cpp
```diff
@@ -4,24 +4,23 @@
...
-        Messenger::warn("There are {} keyword groups defined, but only one can be displayed. Tell the developer!\n");
+    if (keywordStore.nVisibleKeywords() == 0)
+        return;
 
-    if (!keywordIndex.empty())
-    {
-        ui_.Keywords->setUp(keywordIndex.front(), keywordMap, coreData);
+    // We can only display a single section of widgets at present
+    if (keywordStore.sections().size() != 1)
+        Messenger::warn("There are {} keyword sections defined, but only one can be displayed. Tell the developer!\n");
```
rhinoella, line 19:
```suggestion
        Messenger::warn("There are {} keyword sections defined, but only one can be displayed. Tell the developer!\n", keywordStore.sections().size());
```

src/gui/moduleControlWidgetFuncs.cpp
```diff
@@ -30,16 +30,11 @@ ModuleControlWidget::ModuleControlWidget(DissolveWindow *dissolveWindow, Module
...
     {
-        if (currentGroupName == groupName)
-            continue;
-
-        // Create a button for the group
-        auto &&[b, alignRight] = KeywordsWidget::buttonForGroup(groupName);
+        // Create a button for the section
+        auto &&[b, alignRight] = KeywordsWidget::buttonForGroup(section.name());
         b->setCheckable(true);
         b->setAutoExclusive(true);
```
rhinoella: Redundant code because these attributes are already set during the construction of the button in `KeywordsWidget::buttonForGroup()`

src/gui/nodeControlWidgetFuncs.cpp
```diff
@@ -21,13 +21,15 @@ NodeControlWidget::NodeControlWidget(DissolveWindow *dissolveWindow, NodeRef nod
...
-        Messenger::warn("There are {} keyword groups defined, but only one can be displayed. Tell the developer!\n");
-
-    if (!keywordIndex.empty())
-        ui_.NodeKeywordsWidget->setUp(keywordIndex.front(), keywordMap, dissolve_.coreData());
+    // Set up any keyword widgets we have
+    if (!node_->keywords().sections().empty())
+    {
+        // We can only display a single section of widgets at present
+        if (node_->keywords().sections().size() > 1)
+            Messenger::warn("There are {} keyword sections defined, but only one can be displayed. Tell the developer!\n");
```
rhinoella, line 29:
```suggestion
            Messenger::warn("There are {} keyword sections defined, but only one can be displayed. Tell the developer!\n", node_->keywords().sections().size());
```

src/keywords/organiser.cpp
```diff
@@ -0,0 +1,96 @@
...
+// Create new group
+KeywordStoreGroup &KeywordStoreSection::createGroup(std::string_view groupName,
+                                                    std::optional<std::string_view> groupDescription)
+{
+    auto oldGroup = getGroup(groupName);
+    if (oldGroup)
+        return *oldGroup;
+
+    return groups_.emplace_back(groupName, groupDescription ? *groupDescription : "");
+}
```
rhinoella:
```suggestion
KeywordStoreGroup &KeywordStoreSection::createGroup(std::string_view groupName,
                                                    std::string_view groupDescription = "")
{
    auto oldGroup = getGroup(groupName);
    return oldGroup ? *oldGroup : groups_.emplace_back(groupName, groupDescription);
}
```
A shorter way of possibly doing this

src/gui/moduleControlWidgetFuncs.cpp
```diff
@@ -50,15 +45,14 @@ ModuleControlWidget::ModuleControlWidget(DissolveWindow *dissolveWindow, Module
...
         keywordWidgets_.push_back(w);
         ui_.ModuleControlStack->addWidget(w);
 
         // Map our button to the new stack page
         controlStackMap_[b] = ui_.ModuleControlStack->count() - 1;
 
-        currentGroupName = groupName;
         if (!firstButton)
             firstButton = b;
     }
```
rhinoella:
```suggestion
        if (!firstButton)
        {
            firstButton = b;
            firstButton->setChecked(true);
        }
    }
```

Better to put these together (remove the if statement after this)

src/keywords/organiser.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    auto groupIt = std::find_if(groups_.begin(), groups_.end(), [groupName](auto &group) { return group.name() == groupName; });
+    if (groupIt != groups_.end())
+        return *groupIt;
+
+    if (createIfRequired)
+        return groups_.emplace_back(groupName);
+
+    return {};
+}
```
rhinoella: Doesn't this function make the `createGroup()` function kind of redundant? They are both essentially doing the same thing except that `createGroup()` allows for a description to be specified. Maybe it's better to remove the group creation responsibilities of this function.

rhinoella (changes requested):
Fixes to messenger warnings and some small suggestions!

src/gui/moduleControlWidgetFuncs.cpp
trisyoungs: Nice!

src/gui/moduleControlWidgetFuncs.cpp
trisyoungs: :+1:

src/keywords/organiser.cpp
trisyoungs: You're absolutely right here.  This whole PR went through a lot of to-and-fro before getting to this point, so there are improvements to be made!

rhinoella (approved):

rprospero (approved):

PR #1732, opened 2023-11-29
1709 Enabled Buttons for Modules

### LINKED ISSUE #1677: Right-clicking module to enable/disable causes program to crash
Bug occurs only when using the right-click menu, the toggle buttons work correctly.

### LINKED ISSUE #1709: Put enabled buttons for modules in list
**Describe the bug**
Currently we have one module "Enabled" button which acts on the currently-selected module. This is confusing because there is also the layer-scope button at the top of the tab.  We would probably be better off having an associated button with each module in the list.

trisyoungs:
This PR moves the `ModuleLayerModel` from `QAbstractListModel` to `QAbstractTableModel` with a view to removing the "floating" controls for the enabled/disabled state and frequency of individual modules in the list.

Below, the old style is on the left, with this PR moving us to the right (see that we now have the means to clearly indicate the module type as well).
image

Closes #1709.
Closes #1677 (or at least obfuscates it).

### DIFF
--- a/src/gui/models/moduleLayerModel.cpp
+++ b/src/gui/models/moduleLayerModel.cpp
@@ -233,13 +233,8 @@ bool ModuleLayerModel::canDropMimeData(const QMimeData *data, Qt::DropAction act
     Q_UNUSED(row);
     Q_UNUSED(parent);
 
-    if (column > 0)
-        return false;
-
-    if (data->hasFormat("application/dissolve.module.move") || data->hasFormat("application/dissolve.module.create"))
-        return true;
-
-    return false;
+    return true;
+    return (data->hasFormat("application/dissolve.module.move") || data->hasFormat("application/dissolve.module.create"));
 }
 
 bool ModuleLayerModel::dropMimeData(const QMimeData *data, Qt::DropAction action, int row, int column,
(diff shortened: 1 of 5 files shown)


src/gui/models/moduleLayerModel.cpp
```diff
@@ -172,13 +233,8 @@ bool ModuleLayerModel::canDropMimeData(const QMimeData *data, Qt::DropAction act
...
 
-    if (column > 0)
-        return false;
-
-    if (data->hasFormat("application/dissolve.module.move") || data->hasFormat("application/dissolve.module.create"))
-        return true;
-
-    return false;
+    return true;
+    return (data->hasFormat("application/dissolve.module.move") || data->hasFormat("application/dissolve.module.create"));
```
rhinoella:
```suggestion
    return (data->hasFormat("application/dissolve.module.move") || data->hasFormat("application/dissolve.module.create"));
```

rhinoella (changes requested):
One issue, everything else looks great!

src/gui/models/moduleLayerModel.cpp
trisyoungs: Ah - found a bit of my debugging I forgot to remove!

rhinoella (approved):

trisyoungs:
@rhinoella - Requested a re-review as I missed a couple of bits before!

rhinoella (approved):

PR #1737, opened 2023-11-30
Fix ScatteringMatrix referencing

trisyoungs:
A quick PR to fix a segfault cause by EPSR tring to reference nonexistent elements of the scattering matix due to bad indexing.


Danbr4d (approved):
Worked well when used

PR #1738, opened 2023-11-30
Size factor configuration and display

### LINKED ISSUE #1599: SizeFactor node does not require a name
**Describe the bug**
The `SizeFactor` node wants to be named (`mustBeNamed() == true`), but it doesn't need to be as we will not reference it anywhere.

### LINKED ISSUE #1605: SizeFactor (1.0) displayed after loading restart file
**Describe the bug**
Our restart file contains information on the applied and requested size factors for a configuration (see `src/classes/configuration_io.cpp` line 22 for the serialise and line 100+ for the deserialise) but since moving the storage type of `Configuration::appliedSizeFactor_` from `double` to `std::optional<double>` we need to check for a value of 1.0 (i.e. no applied size factor) and set the `appliedSizeFactor_` to `std::nullopt` accordingly. If we don't then the GUi displays a warning that we have a sizefactor of 1.0 (which is fine, and not a warning!).

Danbr4d:
Closes #1605 and closes #1599  

- [x] SizeFactor (1.0) displayed after loading restart file
- [x] SizeFactor node does not require a name

### DIFF
--- a/src/classes/configuration_io.cpp
+++ b/src/classes/configuration_io.cpp
@@ -100,7 +100,10 @@ bool Configuration::deserialise(LineParser &parser, const CoreData &coreData, do
      */
     if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
         return false;
-    appliedSizeFactor_ = parser.argd(3);
+    if (parser.argd(3) == 1.0)
+        appliedSizeFactor_ = std::nullopt;
+    else
+        appliedSizeFactor_ = parser.argd(3);
     requestedSizeFactor_ = parser.argd(4);
     auto nonPeriodic = parser.argb(5);
     const auto lengths = parser.arg3d(0) / appliedSizeFactor_.value_or(defaultSizeFactor_);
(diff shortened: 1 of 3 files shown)


src/classes/configuration_io.cpp
```diff
@@ -100,7 +100,10 @@ bool Configuration::deserialise(LineParser &parser, const CoreData &coreData, do
      */
     if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
         return false;
-    appliedSizeFactor_ = parser.argd(3);
+    if (parser.argd(3) == 1.0)
+        appliedSizeFactor_ = std::nullopt;
+    else
+        appliedSizeFactor_ = parser.argd(3);
```
trisyoungs: Comparing real values directly is always a bit dangerous.  Although this should be fine, I might suggest reversing the logic and doing:
```suggestion
    if (parser.argd(3) > 1.0)
        appliedSizeFactor_ = parser.argd(3);
    else
        appliedSizeFactor_ = std::nullopt;
```

trisyoungs (approved):
:+1:
Just one small suggestion.

PR #1740, opened 2023-11-30
#1076 enable backward compatibility

### LINKED ISSUE #1076: Support infinite backwards compatibility in TOML files

As a matter of basic decency to the user community, each Dissolve version should at least maintain backwards compatibility with the previous version.  However, with the proper architecture, it may be possible to support complete backwards compatibility.

There are two ways to handle the issue of backward compatibility.  The first is to parse the two different versions differently.  The second is to take the files from the old format, upgrade them to the new format, and then use a single parser.  This second method has the enormous advantage that the compatibility is handled entirely at the file level and doesn't need to touch any of the actual Dissolve data structures.  The obvious disadvantage is that it requires a way to modify files, but moving to TOML gives us this capability.

As mentioned before, we must always write a function to convert between the previous version's format and the one for the current version.  To maintain infinite backwards compatibility, all that we must do is not delete these functions and then thread the input file between the functions to bring it up to the latest state.  This means that we would only support writing out the latest file format.

rprospero:
This PR adds the framework for backward compatibility.  This includes some code for upgrading files from version 0.

To ease users who are creating files by hand, including a version is optional.  This will cause the program to assume that the file was written for the current version.  This *will* trigger a warning to alert the user.  Files saves by Dissolve will always include the version number.

Closes #1076

### DIFF
--- a/src/main/compatibility.cpp
+++ b/src/main/compatibility.cpp
@@ -13,8 +13,8 @@ namespace dissolve
 SerialisedValue unrealUpgrade(SerialisedValue contents)
 {
     if (contents["pairPotentials"].contains("coulombTruncation"))
-	contents["pairPotentials"]["coulombTruncation"] =
-	    toml::find<std::string>(contents["pairPotentials"], "coulombTruncation") + "ed";
+        contents["pairPotentials"]["coulombTruncation"] =
+            toml::find<std::string>(contents["pairPotentials"], "coulombTruncation") + "ed";
 
     return contents;
 }
@@ -32,12 +32,12 @@ SerialisedValue backwardsUpgrade(SerialisedValue contents)
 
     for (auto &[key, value] : breakingChanges)
     {
-	DissolveVersion next(key);
-	if (current < next)
-	{
-	    contents = value(contents);
-	    current = next;
-	}
+        DissolveVersion next(key);
+        if (current < next)
+        {
+            contents = value(contents);
+            current = next;
+        }
     }
 
     return contents;
--- a/src/main/compatibility.cpp
+++ b/src/main/compatibility.cpp
@@ -10,8 +10,8 @@
 SerialisedValue unreal_upgrade(SerialisedValue contents)
 {
     if (contents["pairPotentials"].contains("coulombTruncation"))
-	contents["pairPotentials"]["coulombTruncation"] =
-	    toml::find<std::string>(contents["pairPotentials"], "coulombTruncation") + "ed";
+        contents["pairPotentials"]["coulombTruncation"] =
+            toml::find<std::string>(contents["pairPotentials"], "coulombTruncation") + "ed";
 
     return contents;
 }
@@ -29,12 +29,12 @@ SerialisedValue backwards_upgrade(SerialisedValue contents)
 
     for (auto &[key, value] : breaking_changes)
     {
-	DissolveVersion next(key);
-	if (current < next)
-	{
-	    contents = value(contents);
-	    current = next;
-	}
+        DissolveVersion next(key);
+        if (current < next)
+        {
+            contents = value(contents);
+            current = next;
+        }
     }
 
     return contents;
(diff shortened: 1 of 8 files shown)


src/main/compatibility.h
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#pragma once
+
+#include "base/serialiser.h"
+#include "main/version.h"
+
```
trisyoungs, line 8: Should we put these functions in the `dissolve` namespace?

src/main/compatibility.cpp
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2023 Team Dissolve and contributors
+
+#include "main/compatibility.h"
+#include <iostream>
+
+// Upgrade from version zero, which never supported TOML in the first
+// place.  This exists purely to allow testing the backward
+// compatibility code before there is any pass worth checking.
+SerialisedValue unreal_upgrade(SerialisedValue contents)
```
trisyoungs:
```suggestion
SerialisedValue unrealUpgrade(SerialisedValue contents)
```

src/main/compatibility.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+    if (contents["pairPotentials"].contains("coulombTruncation"))
+        contents["pairPotentials"]["coulombTruncation"] =
+            toml::find<std::string>(contents["pairPotentials"], "coulombTruncation") + "ed";
+
+    return contents;
+}
+
+// Upgrade a file to the current version.
+SerialisedValue backwards_upgrade(SerialisedValue contents)
```
trisyoungs:
```suggestion
SerialisedValue backwardsUpgrade(SerialisedValue contents)
```

src/main/compatibility.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+    using Version::DissolveVersion;
+
+    // A map of version numbers and the corresponding function to
+    // upgrade the file.
+    std::map<std::string, std::function<SerialisedValue(SerialisedValue)>> breaking_changes = {{"1.2.0", unreal_upgrade}};
```
trisyoungs:
```suggestion
    std::map<std::string, std::function<SerialisedValue(SerialisedValue)>> breakingChanges = {{"1.2.0", unreal_upgrade}};
```

src/main/io.cpp
```diff
@@ -183,8 +190,13 @@ void Dissolve::deserialisePairPotentials(const SerialisedValue &node)
 }
 
 // Read values from a serialisable value
-void Dissolve::deserialise(const SerialisedValue &node)
+void Dissolve::deserialise(const SerialisedValue &node_orig)
 {
+    // Default to current version if no version info is given.
+    bool hasVersion = node_orig.contains("version");
```
trisyoungs:
```suggestion
    auto hasVersion = node_orig.contains("version");
```

src/main/version.h
```diff
@@ -15,4 +15,21 @@ std::string_view info();
 std::string_view repoUrl();
 // Return app type
 std::string_view appType();
+
+// This class handles the version numbers of the app in a structured
+// manor.
```
trisyoungs:
```suggestion
// manner.
```

src/main/io.cpp
```diff
@@ -183,8 +190,13 @@ void Dissolve::deserialisePairPotentials(const SerialisedValue &node)
 }
 
 // Read values from a serialisable value
-void Dissolve::deserialise(const SerialisedValue &node)
+void Dissolve::deserialise(const SerialisedValue &node_orig)
```
trisyoungs:
```suggestion
void Dissolve::deserialise(const SerialisedValue &originalNode)
```

trisyoungs (approved):
Yes, this should do the job nicely. - just have some pedantry regarding naming of things!

Obviously its hard to say how much compatibility fixing we'll have to do as we progress, so we'll see how the storage for backwards upgrade data pans out.

PR #1741, opened 2023-11-30
1299 Add CLI option for converting files to TOML

### LINKED ISSUE #1299: Add command line option for converting input file
Users will need to upgrade their old input files to the new TOML format.  We'll need a command line option for this.  Additionally, we need to make a decision on how to handle files saved from the GUI (save in the same format as the original or always save as TOML).

rprospero:
This adds a command line option (--to-toml) to convert a file into the TOML format.

Closes #1299


src/main.cpp
```diff
@@ -55,22 +55,34 @@ int main(int args, char **argv)
...
+                    dissolve.worldPool().decideFalse();
+            }
             else
-                dissolve.worldPool().decideFalse();
+            {
+                auto toml = dissolve.serialise();
+                std::ofstream outfile(options.toTomlFile().value());
+                outfile << toml;
+                outfile.close();
+                result = 1;
```
trisyoungs:
```suggestion
                result = true;
```

trisyoungs (approved):
:+1:

PR #1742, opened 2023-12-01
1724 Enable loading of TOML files

### LINKED ISSUE #1724: Rework file loader to support TOML
Currently, loading a file assumes that the file is in our bespoke input file format.  What we would like to do is have the editor intelligently handle multiple formats.  The process should look like.

1) If the file name ends in TOML, use the TOML loader and fail if the file does not parse.
2) Check the file for the header of a restart file.  If the header is present, stop and display an error to the user.
2) Try the TOML loader and, if it succeeds, use that interpretation.  We might also want to display a message to the user that they should update the file extension.
3) Use the bespoke file format parser.

rprospero:
The file loading now goes through four stages:

1. If the file name ends in ".toml", force the user of the TOML parser and fail if it fails
2. Check if the file is a restart file and fail with an error telling the user about the problem if it is
3. Try parsing the file as TOML
4. Parse using the original parser

Closes #1724


trisyoungs (approved):
Looking good.  It's a shame that C++ users have to wait until C++20 to get `std::string_view::ends_with()`, though.  Urfff!

PR #1743, opened 2023-12-01
Build for Apple Silicon

rhinoella:
This PR introduces a GitHub workflow for apple silicon. This specific workflow is to be used on intel GH runners, as it is configured to crossbuild. Currently there is no codesigning step, so in order to open the application, you will have to run these commands first:

```
# Run in the folder with the app bundle

xattr -rd com.apple.quarantine dissolve-gui.app
xattr -rd com.apple.quarantine dissolve-gui.app/Contents/MacOS/dissolve-gui
chmod +x dissolve-gui.app/Contents/MacOS/dissolve-gui
```

You may have to right click + 'Open' the first time in order to bypass security.

## Notes

- The bundle ships with the necessary shared dependencies (.dylib) in the bundle
- Currently, FTGL must be downloaded by the user using homebrew: `brew install ftgl`


rhinoella:
- [x] Pass CONAN_FREETYPE_ROOT to find_package()
- [ ] Fix packaging

.github/workflows/build/osx-silicon/action.yml
```diff
@@ -0,0 +1,135 @@
...
+      cd ../
+
+      # Deploy Conan dependencies for ease
+      mkdir deploy && cd deploy
+      conan install .. -g deploy -s arch=armv8
+
+      # Copy over ftgl so it can be shipped with the bundle
+      mkdir ftgl
+      cp $(brew --prefix ftgl)/lib/libftgl.2.dylib ftgl
+      cp $(brew --prefix ftgl)/lib/libftgl.2.1.3.dylib ftgl
```
trisyoungs: Regarding the rename and copy, would it be slightly neater to make our `ftgl` dir (at the end of the stage), copy the libs, and _then_ apply `install_name_tool`?

.github/workflows/continuous.yml
```diff
@@ -54,6 +57,7 @@ jobs:
 
   BuildLinux:
     needs: QC
+    if: false
```
trisyoungs: Best remove this line! :smile:
```suggestion
```

.github/workflows/build/osx-silicon/action.yml
rhinoella: I actually realised its best to remove the install_name_tool call as it will not work without notorisation!

trisyoungs (approved):

PR #1746, opened 2023-12-06
Disable generate button when running

### LINKED ISSUE #1735: Disable Generate button when running
**Describe the bug**
When Dissolve is running the `Generate` button on configurations is not disabled, allowing the user to recreate the box while we're using it.  Not ideal.

Danbr4d:
Closes #1735


trisyoungs (changes requested):
Yup, this would work, but isn't in line with how we control the state of other "editing" widgets.  Look at the two functions following this one (`preventEditing` and `allowEditing`) which are called at the appropriate times when a simulation is stated / stopped.  You can put `setEnabled()` calls in there.

trisyoungs (approved):
:+1:

PR #1747, opened 2023-12-06
Reinventing QSpecies

trisyoungs:
This PR reimplements the `QSpecies` module completely outside of the current `Procedure` framework. A lightweight and, currently, small set of "analyser" classes is introduced to achieve the goal of allowing developers to easily write analysis routines and not get bogged down in the manual construction of complex `Procedure`s.

Your thoughts on this new approach would be greatly appreciated. If comments are positive and we think this is the future direction we should take I will construct an Epic to encapsulate the work, but for me the pros/cons are fairly straightforward:

- [PRO] Enables developers to write analysis code much more easily
- [CON] Assuming the old `Procedure`-based analysis is eventually removed, this will mean that user-created analysis routines will not be possible until we provide a suitable scripting mechanism (which we may not even want to do)
- [PRO] Parallelisation of analysis routines can now be performed with PSTL/TBB.


src/analyser/siteFilter.cpp
```diff
@@ -0,0 +1,45 @@
...
+        neighbours.clear();
+        for (auto &&[nbr, nbrIndex] : otherSites)
+        {
+            if (range.contains(box->minimumDistance(site->origin(), nbr->origin())))
+                neighbours.emplace_back(nbr, nbrIndex);
+
+            // Check current count against maximum value
+            if (neighbours.size() > maxCount)
+                break;
+        }
```
rhinoella, line 34:
```suggestion
            if (range.contains(box->minimumDistance(site->origin(), nbr->origin())))
                neighbours.emplace_back(nbr, nbrIndex);

            // Check current count against maximum value
            if (neighbours.size() >= maxCount)
                break;
        }
```
Shouldn't it be `>=` otherwise the loop will only break at  `maxCount + 1`

src/modules/qSpecies/process.cpp
```diff
@@ -17,19 +18,45 @@ Module::ExecutionResult QSpeciesModule::process(ModuleContext &moduleContext)
...
+    // involved in NF-BO-NF interactions once we have the available NF sites
+    SiteSelector allOxygenSites(targetConfiguration_, bridgingOxygenSites_);
 
-    // Execute the analysis
-    if (!analyser_.execute({moduleContext.dissolve(), targetConfiguration_, name()}))
+    // Select all NF centres
+    SiteSelector NF(targetConfiguration_, networkFormerSites_);
+
+    // Filter the oxygen sites into those surrounded by exactly two NF sites
+    SiteFilter filter(targetConfiguration_, allOxygenSites.sites());
```
rhinoella, line 30: What is the purpose of having the `allOxygenSites` when it is never used again? Could it not just be `SiteFilter filter(targetConfiguration_, bridgingOxygenSites_);`

rhinoella (approved):
Clears up the code and reduces complexity!

src/analyser/siteFilter.cpp
trisyoungs: This is actually deliberate since we are checking for being strictly within a specific range - so, if there happen to be more things than we want, we break out as soon as we exceed the specified limits.

src/modules/qSpecies/process.cpp
trisyoungs: Essentially a two-step process - first, select all possible sites that _might_ be a bridging oxygen ( == `allOxygenSites`).  Second, determine which of those sites in fact _are_ bridging oxygens ( == `BO`).  The `bridgingOxygenSites_` is in fact a vector of which type of `SpeciesSite`s to choose as candidates for `allOxygenSites` rather than a list of those `Site`s.  I appreciate that the naming is unclear here, and I shall change it!

src/modules/qSpecies/process.cpp
```diff
@@ -17,19 +18,45 @@ Module::ExecutionResult QSpeciesModule::process(ModuleContext &moduleContext)
...
+    // The returned 'neighbourMap' maps BO sites to nearby NF sites *only if* there were exactly two NF sites within range.
+    // So, we can use this to determine the Q numbers for each NF by counting the number of times a NF site appears in the map.
+    std::map<const Site *, int> qSpecies;
+    for (const auto &[siteBO, nbrNF] : neighbourMap)
     {
-        Messenger::error("Q Species experienced problems with its analysis.\n");
-        return ExecutionResult::Failed;
+        for (const auto &[nbr, nbrIndex] : nbrNF)
+            ++qSpecies[nbr];
     }
```
Danbr4d, line 47: Could we also bin the O sites that are filtered so we can capture any free oxygens that are not connected to any NFs and display this?

trisyoungs: Yes, absolutely!  I didn't do this in the PR yet since I just wanted to illustrate a proof-of-principle with the proposed new classes.

To achieve everything we need we could extend the allowable range of NF sites in the filter from [2,2] to [0,2]. Then, the iteration over the `neighbourMap` can bin the size of the `nbrNF` array to get the numbers of free, singly-, and doubly-bound oxygens, and bin into `qSpecies` if `size() == 2`.

Since this is your module, @Danbr4d, do you want to take this modification on?

Danbr4d: Yeah I can definitely make a start on it, let me know when you're done with modifications and I'll give it a go

trisyoungs: @Danbr4d All yours! Let me know if you have any questions or see any areas for improvement.

src/modules/qSpecies/process.cpp
```diff
@@ -17,19 +18,50 @@ Module::ExecutionResult QSpeciesModule::process(ModuleContext &moduleContext)
...
+    for (const auto &[siteBO, nbrNF] : neighbourMap)
     {
-        Messenger::error("Q Species experienced problems with its analysis.\n");
-        return ExecutionResult::Failed;
+        for (const auto &[nbr, nbrIndex] : nbrNF)
+        {
+            ++oxygenTypes[nbrNF.size()];
+            if (nbrNF.size() == 2)
+                ++qSpecies[nbr];
+        }
```
trisyoungs: @Danbr4d Careful!  As written this will increase your `oxygenTypes` counting for every NF site defined for each oxygen. So, you want to move the increment and the size check outside of the loop
```suggestion
        ++oxygenTypes[nbrNF.size()];
        if (nbrNF.size() == 2)
        {
            for (const auto &[nbr, nbrIndex] : nbrNF)
            {
                    ++qSpecies[nbr];
            }
        }
```

rprospero (commented):
On the plus side, this is very clean and readable.  On the other hand, it feels like it should be possible to make a *significant* refactoring to the nodes and achieve the same outcome.

rhinoella (approved):

PR #1752, opened 2024-01-04
Move to using just the dissolveModel in the OverviewTab

rprospero:
Since the OverviewTab is just using the `DissolveModel` as a single source of truth, we need to get the values that we want out of the DissolveModel.

Also, `AtomTypes.qml` needs to import the Dissolve library if it wants to use the `AtomTypesModel` type


src/gui/qml/overview/OverviewTab.qml
```diff
@@ -36,7 +30,7 @@ ColumnLayout {
             Layout.maximumWidth: 200
             Layout.preferredHeight: 100
             Layout.preferredWidth: 200
-            nAtomTypes: atomsTypeModel.rowCount
+            atomTypeModel: dissolveModel.atomTypesModel
```
RobBuchananCompPhys, line 32: This may need explaining: why not `nAtomTypes: dissolveModel.atomTypesModel.rowCount` as per the other corrections?

rprospero: If you check out `AtomTypes.qml`, you'll see that we no longer have an `nAtomTypes` property, but we do have an `atomTypeModel` property.  Of course, since it takes an `AtomTypeModel` and not an int, we have to pass the full `dissolveModel.atomTypesModel` and not just the row count.

On a deeper level, the advantage of this setup is that `AtomTypes.qml` now has access to the full AtomTypeModel, so we can have it display more information in the future without needing to update the OverviewTab to pass in more information.

PR #1753, opened 2024-01-05
Update copyright year in source

rprospero:
🎉 Happy new year!


RobBuchananCompPhys (approved):

rprospero:
So conan has dropped onetbb/2023.3, so I've  added an ill-concieved solution of just bumping the version number.  If it perfectly passes, it might be worth discussing, but I'll otherwise roll it back and we'll have a large discussion about the version upgrade that is breaking out build.

In case you were curious about the title

PR #1754, opened 2024-01-10
Actually assign intramolecular terms in add forcefield dialog

### LINKED ISSUE #1750: Medic not working in dissolve (nightly) on idaas.
When using dissolve to set up a simple molecule (in this case ethanol or methanol) after adding the appropriate forcefield (alcohol), using the medic button to clean up the molecule is not working. No errors appear in the message box.

Images shown:

image

image

rprospero:
So the assigning of intramolecular terms was not actually getting called in the Add Forcefield Dialog.  This should fix the issues with the medic.  As basic testing, I created a methanol molecule, added the forcefield, and successfully ran the medic afterwards.

Closes #1750

### DIFF
--- a/src/gui/models/addForcefieldDialogModel.h
+++ b/src/gui/models/addForcefieldDialogModel.h
@@ -116,6 +116,9 @@ class AddForcefieldDialogModel : public QObject
     Radio atomTypeRadio_ = Radio::All;
     // The choice for how to handle intramolecular terms
     Radio intramolecularRadio_ = Radio::All;
+
+    private:
+    // Assign intramolecular terms to species
     void assignIntramolecularTerms(const Forcefield *ff);
 
     public:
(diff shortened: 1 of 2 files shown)


src/gui/models/addForcefieldDialogModel.h
```diff
@@ -116,6 +116,7 @@ class AddForcefieldDialogModel : public QObject
     Radio atomTypeRadio_ = Radio::All;
     // The choice for how to handle intramolecular terms
     Radio intramolecularRadio_ = Radio::All;
+    void assignIntramolecularTerms(const Forcefield *ff);
```
trisyoungs, line 122: Just being picky about separating functions from variables.
```suggestion

    private:
    // Assign intramolecular terms to species
    void assignIntramolecularTerms(const Forcefield *ff);
```

trisyoungs (approved):
Well, not sure how this slipped through the testing net in the first place!  One small addition then good to go.

:+1:

PR #1757, opened 2024-01-12
Disable arm64 build

rhinoella:
The arm64 build was mistakenly included in the continuous workflow, this PR removes that.


trisyoungs (approved):
:+1:

PR #1760, opened 2024-01-15
Link qt-idaaas nixpkgs version to our primary nixpkgs version

rprospero:
This should fix the issue with Qt Designer.  It will also ensure that future upgrades of nixpkgs will also upgrade the nixpkgs of our patched Qt.  Granted, I'm hoping that IDAaaS will have upgraded to a more modern kernel before the next nixpkgs release and we won't still be patching Qt.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -196,22 +196,6 @@
         "type": "github"
       }
     },
-    "nixpkgs_6": {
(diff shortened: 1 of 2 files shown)


trisyoungs (approved):
:+1:

PR #1763, opened 2024-01-16
display qml scale charges dialog

RobBuchananCompPhys:
After selecting an atomic species and clicking "Scale Charges", you should see something like this: 
(The spin box is narrower than before with xml, but maybe that is alright? - the inputs aren't expected to get too large)
scale_charges


src/gui/models/scaleChargesDialogModel.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+ScaleChargesDialogModel::ScaleChargesDialogModel() {}
+
+ScaleChargesDialogModel::~ScaleChargesDialogModel() {}
+
+double ScaleChargesDialogModel::value() const { return currentValue_; }
+
+void ScaleChargesDialogModel::updateValue(double newVal) { currentValue_ = newVal; }
+
+void ScaleChargesDialogModel::setOption(int option)
```
rprospero: `option` would probably be better as an ENUM. It would make the switch statement clearer as to its intent.

src/gui/qml/ScaleChargesDialog.qml
```diff
@@ -0,0 +1,77 @@
...
+        height: root.height
+        width: root.width
+
+        Column {
+            anchors.centerIn: parent
+            spacing: 10
+
+            Text {
+                font.pixelSize: 12
+                text: "Enter the scaling factor to apply to all atoms / the target sum to determine scaling factor from"
```
rprospero, line 37:
```suggestion
                text: "Enter the scaling factor to apply to all atoms / the target sum to determine scaling factor from"
                wrapMode: Text.Wrap
```

We might be able to make the dialog smaller if the text could wrap.

src/gui/qml/ScaleChargesDialog.qml
```diff
@@ -0,0 +1,77 @@
+import Dissolve
```
rprospero:
```suggestion
import Dissolve
import "widgets" as D
```

We should pull in our custom widgets to ensure that we maintain a consistent look across the application.

src/gui/qml/ScaleChargesDialog.qml
```diff
@@ -0,0 +1,77 @@
...
+
+    contentItem: Rectangle {
+        height: root.height
+        width: root.width
+
+        Column {
+            anchors.centerIn: parent
+            spacing: 10
+
+            Text {
```
rprospero: Go ahead and use the `D.Text` widget here to keep our customisations.

src/gui/models/scaleChargesDialogModel.h
```diff
@@ -0,0 +1,32 @@
...
+{
+    Q_OBJECT
+
+    Q_PROPERTY(double value READ value NOTIFY valueSet)
+
+    public:
+    ScaleChargesDialogModel();
+    ~ScaleChargesDialogModel();
+
+    bool scale_, scaleTo_;
```
rprospero: If `option` is a enum, we might be able to replace both these bools with  single enum (e.g. `scale_direction_`).

rprospero (changes requested):
Okay, I've made a bunch of comments, so I want to take a moment here to say that I really like what you've done here.  This is a great use of the model setup and the QML code is pretty clean.

Besides the individual comments that I've made (and which I'm more than willing to listen, if you disagree), there's one larger change that I would suggest.  Your model is very good, but a bunch of the work is still ultimately performed by the `DissolveWindow::on_SpeciesScaleChargesAction_triggered` function (which is the way it's always been).  I think it would be worth adding a function to the model where you pass in a reference/pointer to the species and it does all the scaling within the model.  That gives us a couple of advantages:

1. It's now possible (and recommended) two write a unit test to confirm that the model performs the correct scaling.
2. There's less code in the menu GUI, which should really only be for dispatch
3. When we replace the main GUI with QML, there's no concern that we won't translate the logic correctly

src/gui/models/scaleChargesDialogModel.cpp
RobBuchananCompPhys: Nice, yeah agreed on the clarity

src/gui/qml/ScaleChargesDialog.qml
RobBuchananCompPhys: agreed!

src/gui/menu_species.cpp
```diff
@@ -436,14 +436,14 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
         return;
 
     static ScaleChargesDialog scaleChargesDialog(this);
-    double scaleFactor = 1.0;
+    double scaleFactor = scaleChargesDialog.model->value();
```
trisyoungs:
```suggestion
    auto scaleFactor = scaleChargesDialog.model->value();
```

src/gui/menu_species.cpp
```diff
@@ -436,14 +436,14 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
...
     if (scaleChargesDialog.exec() == QDialog::Accepted)
     {
-        if (scaleChargesDialog.scale_)
-            scaleFactor = scaleChargesDialog.scaleValue();
+        if (scaleChargesDialog.model->scale_)
+            scaleFactor = scaleChargesDialog.model->value();
         else
         {
-            double scaleTarget = scaleChargesDialog.scaleValue();
+            double scaleTarget = scaleChargesDialog.model->value();
```
trisyoungs:
```suggestion
            auto scaleTarget = scaleChargesDialog.model->value();
```

src/gui/qml/ScaleChargesDialog.qml
```diff
@@ -0,0 +1,77 @@
...
+                font.pixelSize: 12
```
trisyoungs: Here we set `pixelSize`, but at the root of the `Page` we use `pointSize`, and neither value corresponds to the "standard" of `11` used in the old UI files!

src/gui/qml/ScaleChargesDialog.qml
trisyoungs: It kinda feels like there is some inherent bad UI design being highlighted by the label text here (and which is obviously not your fault @RobBuchananCompPhys since you're just porting it over!).  Essentially we have two functions "Scale" and "ScaleTo" which are trying to make use of a single spin box, hence the dual description in  the label. A simple pair of radio buttons would clear this up no end, and while requiring a separate spinbox for each option, one could then remove one of the buttons.

trisyoungs (changes requested):
Looking good, and I think @rprospero makes a convincing case for moving the functionality into the model.

Have added some other comments, but notice that in the screengrab the buttons are aligned bottom-left in the dialog, when convention suggests bottom-right is the correct location.

src/gui/menu_species.cpp
RobBuchananCompPhys: Is there a particular reason why we should `auto` here? Not sure this will be necessary after merging this code into the model, since it is getting the value from within the model instance and it's definitely a double.

src/gui/menu_species.cpp
RobBuchananCompPhys: Same as below

src/gui/menu_species.cpp
rprospero: It's not that there's a particular reason to use `auto` as much as there is no particular reason *not* to use it.  In general, using `auto` is preferable when it's available because it makes it easier to change types elsewhere in the program without needing to update the types *everywhere*.  For example, if we later decide to change the type of `value` to some other custom class, then we would also need to update this line, adding noise to the eventual PR.  Making this an `auto` gives us the freedom to change the type of `value` as long as it continues to conform to the same interface.

This is known as Generic Programming and is what C++ has been moving toward (and away from Object Oriented Programming).  It's basically a type-safe approach to duck-typing and using `auto` where possible helps enable this workflow.

RobBuchananCompPhys: Sounds like that's the way to go then - thanks for the explanation @rprospero!
While I was aware that `auto` can have advantages like seamlessly altering APIs, my take stemmed from the fact that the main consumer of this particular method was now going to be the model class itself (via the member function `setScale`), but I can see that the situation *might* still arise where we would want to specify a different return type for `value` without messing things up elsewhere.

RobBuchananCompPhys:
I've added a unit test for the model. This should be good to go.

One outstanding issue is that user manually resizing the dialog's window doesn't seem cause the contained UI components to adapt their size/positioning, or enforce a minimum size. (Perhaps I'm missing something @rprospero?) 

scale_charges2

src/gui/menu_species.cpp
```diff
@@ -435,29 +435,15 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
...
-    double scaleFactor = 1.0;
-    if (scaleChargesDialog.exec() == QDialog::Accepted)
+    static ScaleChargesDialog dialog(this);
+    if (dialog.exec() == QDialog::Accepted)
     {
-        if (scaleChargesDialog.scale_)
-            scaleFactor = scaleChargesDialog.scaleValue();
-        else
+        bool scaled = dialog.model->scaleCharges(species);
+        if (!scaled)
```
trisyoungs:
```suggestion
        if (!dialog.model->scaleCharges(species))
```

src/gui/models/scaleChargesDialogModel.cpp
```diff
@@ -0,0 +1,38 @@
...
+    auto scaleFactor = value();
+    if (scaleType_ != Scale)
+    {
+        auto scaleTarget = scaleFactor;
+        if (scaleTarget == 0.0)
+        {
+            return false;
+        }
+
+        double sum = 0.0;
```
trisyoungs:
```suggestion
        auto sum = 0.0;
```

src/gui/models/scaleChargesDialogModel.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+ScaleChargesDialogModel::Option ScaleChargesDialogModel::getOption() { return scaleType_; }
+
+bool ScaleChargesDialogModel::scaleCharges(Species *species)
+{
+    auto scaleFactor = value();
+    if (scaleType_ != Scale)
+    {
+        auto scaleTarget = scaleFactor;
+        if (scaleTarget == 0.0)
```
trisyoungs: Literal comparisons of doubles always makes me uneasy, even if `0.0` is pseudo-well-defined. Maybe we should do something like `if (fabs(scaleTarget) < 1.0e-6)`??

src/gui/models/scaleChargesDialogModel.cpp
```diff
@@ -0,0 +1,38 @@
...
+        {
+            return false;
+        }
```
trisyoungs: Moving the error message dialog from `menu_species.cpp` into here, rather than returning `false` from `ScaleChargesDialogModel` to handle this specific failure, feels like a more solid and self-contained approach. This function can then just be `void`.

src/gui/models/scaleChargesDialogModel.h
```diff
@@ -0,0 +1,42 @@
...
+    Q_INVOKABLE void setOption(Option);
+    Q_INVOKABLE void updateValue(double);
+
+    Option getOption();
+
+    double value() const;
+    bool scaleCharges(Species *);
+
+    private:
+    double currentValue_ = 1.0;
```
trisyoungs:
```suggestion
    double currentValue_{1.0};
```

src/gui/models/scaleChargesDialogModel.h
trisyoungs, line 1: Even though this is a relatively simple header, ersonally I'd like to see a few more single-line comments etc. separating related parts and data, just so it's brutally obvious what everything is for.

src/gui/models/scaleChargesDialogModel.cpp
trisyoungs: Of course, this will break your unit test, so perhaps still return the `bool` for success but add a second argument `bool showDialogOnError = true` to the `scaleCharges` function so you can prevent the dialog being raised when testing, but still test for success?

trisyoungs:
The dialog looks really nice now - great work on the harmonisation of styles!  One small thing - the `Cancel` button should be leftmost rather than occupying the default "human click" position of bottom-right.

src/gui/models/scaleChargesDialogModel.h
RobBuchananCompPhys: Sure thing!

src/gui/models/scaleChargesDialogModel.cpp
RobBuchananCompPhys: Yeah, I think it's best to go for one or the other, and I get the impression there is not much use returning a bool since the main work of the function is being done on the `*species` instance. In the test I will call the `void` method and then make a test on each atom to check the scaling is correct.

RobBuchananCompPhys: I do like the idea of the parameter to variably display the error message for testing situations

src/gui/scaleChargesDialogFuncs.cpp
```diff
@@ -1,21 +1,19 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // Copyright (c) 2024 Team Dissolve and contributors
 
+#include "gui/models/scaleChargesDialogModel.h"
 #include "gui/scaleChargesDialog.h"
+#include <QQmlContext>
+#include <QQuickItem>
+#include <QQuickWidget>
```
rprospero, line 9:
```suggestion
#include <QQuickWidget>
#include <QHBoxLayout>
```

Includes needed for resizing.

src/gui/scaleChargesDialogFuncs.cpp
```diff
@@ -1,21 +1,19 @@
...
 
-ScaleChargesDialog::ScaleChargesDialog(QWidget *parent) { ui_.setupUi(this); }
-ScaleChargesDialog::~ScaleChargesDialog() {}
+ScaleChargesDialog::ScaleChargesDialog(QWidget *parent) : QDialog(parent)
+{
+    QQuickWidget *view = new QQuickWidget(QUrl("qrc:/dialogs/qml/ScaleChargesDialog.qml"), this);
```
rprospero, line 13:
```suggestion
    QQuickWidget *view = new QQuickWidget(QUrl("qrc:/dialogs/qml/ScaleChargesDialog.qml"), this);
    view-setResizeMode(QQuickWidget::SizeRootObjectToView)
```

Tell QML to perform a resize when `view` is resized.

src/gui/scaleChargesDialogFuncs.cpp
```diff
@@ -1,21 +1,19 @@
...
-    scale_ = true;
-    accept();
-}
-void ScaleChargesDialog::on_ScaleToButton_clicked(bool checked)
-{
-    scaleTo_ = true;
-    accept();
-}
+    QObject::connect(model, SIGNAL(cancelSelection()), this, SLOT(reject()));
+    QObject::connect(model, SIGNAL(acceptSelection()), this, SLOT(accept()));
```
rprospero, line 25:
```suggestion
    QObject::connect(model, SIGNAL(acceptSelection()), this, SLOT(accept()));
    
    QHBoxLayout *topLeftLayout = new QHBoxLayout;
    topLeftLayout->addWidget(view);
    setLayout(topLeftLayout);
```

Tell the old-style Qt dialog to update the size of the view object when the window is resized.

rprospero (approved):

src/gui/scaleChargesDialogFuncs.cpp
```diff
@@ -1,21 +1,26 @@
...
+#include <QHBoxLayout>
+#include <QQmlContext>
+#include <QQuickItem>
+#include <QQuickWidget>
 
-ScaleChargesDialog::ScaleChargesDialog(QWidget *parent) { ui_.setupUi(this); }
-ScaleChargesDialog::~ScaleChargesDialog() {}
+ScaleChargesDialog::ScaleChargesDialog(QWidget *parent) : QDialog(parent)
+{
+    QQuickWidget *view = new QQuickWidget(QUrl("qrc:/dialogs/qml/ScaleChargesDialog.qml"), this);
```
trisyoungs, line 13:
```suggestion
    auto view = new QQuickWidget(QUrl("qrc:/dialogs/qml/ScaleChargesDialog.qml"), this);
```

src/gui/scaleChargesDialogFuncs.cpp
```diff
@@ -1,21 +1,26 @@
...
-    accept();
-}
-void ScaleChargesDialog::on_ScaleToButton_clicked(bool checked)
-{
-    scaleTo_ = true;
-    accept();
-}
+    view->setResizeMode(QQuickWidget::SizeRootObjectToView);
+
+    QHBoxLayout *topLeftLayout = new QHBoxLayout;
```
trisyoungs, line 20:
```suggestion
    auto topLeftLayout = new QHBoxLayout;
```

src/gui/models/scaleChargesDialogModel.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+double ScaleChargesDialogModel::value() const { return currentValue_; }
+
+void ScaleChargesDialogModel::updateValue(double newVal) { currentValue_ = newVal; }
+
+void ScaleChargesDialogModel::setOption(Option option) { scaleType_ = option; }
+
+ScaleChargesDialogModel::Option ScaleChargesDialogModel::getOption() { return scaleType_; }
+
+bool ScaleChargesDialogModel::scaleCharges(Species *species, bool showDialogOnError = true)
```
trisyoungs, line 19:
```suggestion
bool ScaleChargesDialogModel::scaleCharges(Species *species, bool showDialogOnError)
```

src/gui/models/scaleChargesDialogModel.h
```diff
@@ -0,0 +1,49 @@
...
+    // Sets the new selected scale value
+    Q_INVOKABLE void updateValue(double);
+
+    // Returns the current option
+    Option getOption();
+
+    // Returns the current scale value
+    double value() const;
+    // Apply scaling operation to species atoms
+    bool scaleCharges(Species *, bool);
```
trisyoungs, line 39:
```suggestion
    bool scaleCharges(Species *species, bool showDialogOnError = true);
```

trisyoungs (approved):
Excellent stuff.  Some small suggested corrections (more `auto` and a few other bits) but this is good to go.

PR #1766, opened 2024-01-17
1637 dynamic keywords

### LINKED ISSUE #1637: Remove Static Singletons

rprospero:
The primary change here is that every keyword is now contained within a specific KeywordStore.  There is no longer a static singleton vector in either KeywordBase or in KeywordStore.  This leads to a couple of other changes.

1. Since there is no static singleton, removing references to an object in CoreData involves removing the references in every module of every layer.  This is handled in the `objectNoLongerValid` function in `coreData.cpp`.  This is now a function in the module and not a private function in the class because it needs access to the Module and ModuleLayer classes in order to be built.  This wouldn't be an issue, but it's a template function, so leaving it in the `.h` file creates an import loop.
2. The Module and ModuleLayer classes now need mutable access to CoreData, since they might be removing references from it (an obviously mutable action).  This has resulted in the change from `OptionalReferenceWrapper<const CoreData>` to `Core Data*`.  I did use a `const_cast` in `layerFuncs.cpp` rather than update the entire call chain to provide a mutable `CoreData` reference, but I'm willing to change this if other people find it upsetting.

This closes #1637


src/classes/coreData.h
```diff
@@ -254,4 +254,5 @@ class CoreData
     void removeReferencesTo(Configuration *data);
     void removeReferencesTo(Species *data);
     void removeReferencesTo(SpeciesSite *data);
+    void removeReferencesTo(Isotopologue *data);
```
trisyoungs: Pedantic as I am I would move this up a few lines so as to make the "list" alphabetical.

src/keywords/store.h
```diff
@@ -281,11 +281,11 @@ class KeywordStore
      */
     private:
     // Vector of all keywords globally
```
trisyoungs:
```suggestion
    // Vector of all keywords in the store
```

trisyoungs (commented):
I confess I'm a little confused why the `const_cast` and the conversion to a pointer is necessary since, as far as I can see, the `CoreData` reference was passed non-const down the chain anyway. Indeed I just tried to revert that change locally (i.e. go back to `OptionalReferenceWrapper<CoreData>`) for both models and all seems to work as expected...

rhinoella (approved):
👍 
Also agree with what @trisyoungs mentioned regarding the `const_cast`

src/classes/coreData.h
rprospero: Not only did I move Isotopologue up a few lines, I also moved Module down a few lines.  That leaves gets the whole thing alphabetised.

PR #1770, opened 2024-01-22
1767 structure of arrays

rprospero:
This PR converts the array of atoms into a structure of array format.  This should ease the addition of GPU processing support.

Progress:
- [x] Replace atom pointers with array indices
- [x] Replace atom vector with structure of arrays
- [ ] Confirm via benchmark that we have not weakened the performance (an improvement is expected, but not mandatory).


trisyoungs:
@rprospero Is this worth closing now?

rprospero:
This PR never saw the desired speed increases, so it's best to wait until we have a proper GPU branch to do this work.

PR #1771, opened 2024-01-22
Remove need for QC before build workflow

rhinoella:
Have the QC workflow act as a soft-check instead of a hard-check, as currently the other tests won't run if the QC fails.


Danbr4d (approved):
Looks good and prefer this so tests run regardless and QC can be fixed later

trisyoungs (approved):

PR #1772, opened 2024-01-22
1733 Auto Qt

### LINKED ISSUE #1733: Qt AutoMoc
We don't, for historical reasons lost to the sands of time, use Qt's automoc option in our `cmake` chain.  This is fine, but leads to deliberate naming of files like `myWidgetFuncs.cpp` in order to avoid clashes with Qt-generated `myWidget.cpp` from a ui file, but using automoc this isn't required and would tidy up our naming a bit.

trisyoungs:
This PR started off as a quick "30 minutes left on a Friday" job, but has turned into a "waste the whole of Monday" task instead. This is a big PR, but most of it is just renaming and only the first and last commits are particularly interesting.

So to begin we were already using CMake's `AUTOMOC` option, but not `AUTORCC` or `AUTOUIC`, and were kind of half-blending the automated integration with the `qt6_wrap_*` mechanism (not recommended). Switching out to full auto cleans up a lot of our filenames, removing the "custom" `*.hui` and `*Funcs.cpp` namings in favour of plain old `*.h` and `*.cpp` (the latter was a legacy introduced many years ago when I was getting clashes between Qt-generated `*.cpp` files, from processing `ui` files, with my own).

However, moving to full auto introduces a problem whereby now autogenerated files are stored in many separate locations, rather than in a single build dir), meaning `CMakeLists.txt` have had to be adjusted in a few places in order to find the necessary `ui_*` files.  This was a particular issue for the `moduleWidgetProducer` target which needed to find N `ui_*` files for N modules in N different locations. A solution is in place, but perhaps someone with a better relationship with CMake can implement a better one (see final commit).

Closes #1733.


rprospero (approved):

trisyoungs:
@RobBuchananCompPhys are you OK with me merging this in now, since it will cause a conflict with your other charges PR #1787, or would you like me to wait?

PR #1773, opened 2024-01-23
Module Add Crash

trisyoungs:
This PR addresses a critical issue with adding modules to any layer.  The root cause was a bit of bad interplay between the renovated `CoreData` function to return all module instances and the `ModuleLayerModel`.  Basically, the model would create an empty slot in a layer containing a null pointer ready to receive the newly-created module pointer in a different model function, but in-between we attempt to create a unique name which involves iterating over a list of all module instances.  This contains our null pointer placeholder, and so the `DissolveSys::uniqueName()` function would cause a segfault when it tried to get the name of this null module.

Rather than modifying the `CoreData` routine to avoid null pointers I have modified the model to create the new `Module`s at the point of requirement.

### DIFF
--- a/src/classes/coreData.cpp
+++ b/src/classes/coreData.cpp
@@ -38,8 +38,8 @@ std::shared_ptr<AtomType> CoreData::addAtomType(Elements::Element Z)
     atomTypes_.push_back(newAtomType);
 
     // Create a suitable unique name
-    newAtomType->setName(DissolveSys::uniqueName(
-        Elements::symbol(Z), atomTypes_, [&](const auto &at) { return newAtomType == at ? std::string() : at->name(); }));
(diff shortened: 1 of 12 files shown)


Danbr4d (commented):
This fixes the crash with the new layer and modules, but does not fix the crash when implementing a Standard EPSR module for refinement module. It seems that when there isn't a Neutron S(Q) module as well Dissolve crashes when the EPSR module is added.

trisyoungs:
> This fixes the crash with the new layer and modules, but does not fix the crash when implementing a Standard EPSR module for refinement module. It seems that when there isn't a Neutron S(Q) module as well Dissolve crashes when the EPSR module is added.

On it!

Danbr4d (approved):
All seems to be working now with no crashes

rprospero (approved):

PR #1785, opened 2024-01-30
Conan Deps

### LINKED ISSUE #800: Invoke conan when performing CMake configuration
Currently `conan` must be run in the build directory before `CMake`, but it would be nice if this could be automated as part of the `CMake` processing.

I tire of typing `conan ../ -s compiler.version=9`...

trisyoungs:
A ~quick~ massive, week-long struggle of a PR to prevent unnecessary dependencies on conan-provided libs for platforms that don't need them. This ballooned out of its original scope quite quickly, but ultimately clarifies and improves our cross-platform build system, IMHO.

Our GitHub workflows and the root `CMakeLists.txt` get the most attention, the the latter now calling `conan` automatically rather than leaving this to the user.

Closes #800.


trisyoungs:
Apologies @rhinoella and @rprospero that this is a bit of a beast, but most changes beyond the workflow yml and the root CMakeLists.txt is reduction, which is nice!

.github/workflows/build/osx/action.yml
```diff
@@ -18,11 +16,6 @@ runs:
   # Setup / Install Dependencies
   #
 
-  - name: Setup Python
-    uses: actions/setup-python@v4
-    with:
-      python-version: ${{ env.pythonVersion }}
-
   - name: Install Homebrew Dependencies for Intel
```
rhinoella:
```suggestion
  - name: Install Homebrew Dependencies
```

rhinoella (approved):
👍 A much improved way of handling dependencies

rprospero (approved):

.github/workflows/build/osx/action.yml
trisyoungs: Nice catch!

PR #1786, opened 2024-01-31
Q-Species Normalise Frequency

### LINKED ISSUE #1782: Q-Species output of Normalised Frequency
Currently the Q-Species module outputs the exact amount of atoms for each BO, NBO and FO. It also exports the values in the graph as the number of atoms. It would be much better if this module outputted and exported these values as normalised frequency.

Danbr4d:
Closes #1782


src/math/integerHistogram1D.cpp
```diff
@@ -143,7 +143,7 @@ Data1D IntegerHistogram1D::data() const
     return result;
 }
 
-std::map<int, SampledDouble> IntegerHistogram1D::averages() const { return averages_; }
+std::map<int, SampledDouble> IntegerHistogram1D::averages() const & { return averages_; }
```
trisyoungs: Sorry, wasn't clear! I actually meant the return value should be `const &`:
```suggestion
// Return averages map
const std::map<int, SampledDouble> &IntegerHistogram1D::averages() const { return averages_; }
```

src/math/integerHistogram1D.h
```diff
@@ -62,7 +62,7 @@ class IntegerHistogram1D
     // Return current data
     Data1D data() const;
     // Return averages map
-    std::map<int, SampledDouble> averages() const;
+    std::map<int, SampledDouble> averages() const &;
```
trisyoungs:
```suggestion
    const std::map<int, SampledDouble> &averages() const;
```

src/modules/qSpecies/gui/qSpeciesWidgetFuncs.cpp
```diff
@@ -40,9 +41,10 @@ void QSpeciesModuleWidget::updateControls(const Flags<ModuleWidget::UpdateFlags>
 
     // Update Oxygen Sites Labels
     auto oSitesHistogram = dissolve_.processingModuleData().valueOr("OSitesHistogram", module_->name(), IntegerHistogram1D());
-    ui_.FOResultFrame->setText(oSitesHistogram.averages()[0]);
-    ui_.NBOResultFrame->setText(oSitesHistogram.averages()[1]);
-    ui_.BOResultFrame->setText(oSitesHistogram.averages()[2]);
+    auto sum = Integrator::absSum(oSitesHistogram.data());
+    ui_.FOResultFrame->setText(oSitesHistogram.averages()[0] / sum);
+    ui_.NBOResultFrame->setText(oSitesHistogram.averages()[1] / sum);
+    ui_.BOResultFrame->setText(oSitesHistogram.averages()[2] / sum);
```
trisyoungs: Do you want to include another label one for remaining "odd" cases that are not FO, NBO, or BO? You could capture each value individually first for use in their respective labels, then display `(nFO+nNBO+nBO - sum) / sum`.

trisyoungs (approved):

PR #1787, opened 2024-02-01
Modify charge dialogs qml

RobBuchananCompPhys:
This PR aims to implement the three dialogs "Scale Charges", "Smooth Charges" and "Reduce Sig Figs in Charges", in QML.
I have used a generic `ModifyChargesModel` to handle the resources corresponding to each operation, since all dialogs share a common, simple format of components, and similarly updated the dialog class to handle connections between all three QMLs and the model.
Therefore, this supersedes (and incorporates) the work I did recently on the ScaleChargesDialog/Model.


trisyoungs:
Something corrupt / missing in the qml files, or at least their location as I get this error (as does the CI):
`ninja: error: '/home/tris/src/dissolve/src/gui/qml/ModifyChargesDialog.qml', needed by 'src/gui/qrc_main.cpp', missing and no known rule to make it`

src/gui/menu_species.cpp
```diff
@@ -434,10 +435,14 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
     if (!species)
         return;
 
-    static ScaleChargesDialog dialog(this);
+    if (!dissolve_.modifyChargesModel) { dissolve_.modifyChargesModel = new ModifyChargesModel; }
+
+    auto model = dissolve_.modifyChargesModel;
```
trisyoungs: I'm a bit confused by the way this is implemented for two reasons:
1) Why start with an empty pointer to a model and then new it on first use? If the model needs to persist between uses then why not just have it is a full member in the first place? If not, then why not just instantiate it as a temporary variable here in this routine just before the `ModifyChargesDialog` is created?
2) Why is the model pointer a member of the main `Dissolve` class and not the GUI?  This can only introduce GUI-related dependencies / code into the pure CLI version of the code which is likely to cause problems in the long run!

src/gui/modifyChargesDialog.h
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#include "gui/models/modifyChargesModel.h"
+#include <QObject>
+#include <QDialog>
+#include <QQuickWidget>
+#include <QCloseEvent>
+
+class ModifyChargesModel;
```
trisyoungs, line 12:
```suggestion
// Forward Declarations
class ModifyChargesModel;
```

src/gui/models/modifyChargesModel.h
```diff
@@ -0,0 +1,70 @@
...
+    // Reduce significant figures on species atoms charges
+    void reduceSignificantFigures(Species*);
+
+    private:
+    double currentScaleValue_{1.0};
+    double currentSmoothValue_{0.0};
+    int currentSigFigValue_{3};
+
+    ScaleType scaleType_;
+    Modify modify_;
```
trisyoungs: Convention for Dissolve is to define the `private` member variables before the `public` setters/getters. Also a single line comment for each to express what they are.

src/gui/models/modifyChargesModel.h
```diff
@@ -0,0 +1,70 @@
...
+    Q_PROPERTY(double scaleValue READ scaleValue NOTIFY valueSet)
+    Q_PROPERTY(double smoothValue READ smoothValue NOTIFY valueSet)
+    Q_PROPERTY(int sigFigValue READ sigFigValue NOTIFY valueSet)
+
+    public:
+    ModifyChargesModel();
+    ~ModifyChargesModel();
+
+    // Enum type to differentiate between operations for mofiying charges
+    enum Modify { Scaling, Smoothing, ReduceSigFig };
```
trisyoungs: Just a naming suggestion to aid code readability:
```suggestion
    enum ModifyBy { Scaling, Smoothing, ReduceSigFig };
```

src/gui/qml/modifyCharges/Layout.qml
```diff
@@ -0,0 +1,76 @@
...
+                break
+            default:
+                break
+        }
+    }
+
+    D.Text {
+        Layout.fillHeight: true
+        Layout.fillWidth: true
+        font.pixelSize: 11
```
trisyoungs: So interestingly this line causes odd behaviour on my system (Linux / Qt 6.2).  If the line is present it doesn't matter what value I set for `pixelSize` (or `pointSize`) - it is apparently ignored but the font is made tiny. If I remove the line altogether I get what looks like the default 12 px font.  I'm increasingly wondering if this is just a bug in the version of Qt I have, so I'm upgrading to see if that sorts things out.

src/main/dissolve.h
```diff
@@ -102,6 +103,8 @@ class Dissolve : public Serialisable<>
     // Revert potentials to reference state, clearing additional potentials
     void revertPairPotentials();
 
+    ModifyChargesModel *modifyChargesModel{nullptr};
```
trisyoungs: :scream:

src/gui/menu_species.cpp
RobBuchananCompPhys: The full member option sounds preferable, I will try and compile without initializing that as a `nullptr`.
The `new` keyword is used since all three dialogs use that model, but can't guarantee which one will be called first, hence checking if it is a `nullptr` in the start of the function.

RobBuchananCompPhys: I see about the CLI issue, I will move this to `gui.h`

trisyoungs:
@RobBuchananCompPhys An update from my machine.  Upgrading my distribution has now furnished me with Qt 6.4.2 which solves all of the odd errors with text / widget colouring I was seeing. I also now understand the font scaling issue - I still see odd results when setting `font.pixelSize` but `font.pointSize` gives the expected results when set to 11. This must be down to a mis-calculation of screen DPI or something similar at the Qt end since my display is not scaled, but it seems that the recommended approach is to use `pointSize` wherever possible.

trisyoungs:
> @RobBuchananCompPhys An update from my machine. Upgrading my distribution has now furnished me with Qt 6.4.2 which solves all of the odd errors with text / widget colouring I was seeing. I also now understand the font scaling issue - I still see odd results when setting `font.pixelSize` but `font.pointSize` gives the expected results when set to 11. This must be down to a mis-calculation of screen DPI or something similar at the Qt end since my display is not scaled, but it seems that the recommended approach is to use `pointSize` wherever possible.

We can set this in the base D.Text class, presumably, in order to provide a consistent result?

src/gui/menu_species.cpp
RobBuchananCompPhys: Seems to be a circular dependency between `src/gui/models/ModifyChargesModel.h` and `src/gui/gui.h` when I try and add it as a full member to `DissolveWindow`.

src/main/dissolve.h
RobBuchananCompPhys: I think I need more explanation as to why this is terrible!
We currently declare some objects in classes with this pattern (i.e. `QQuickWidget *view_{nullptr};`), so what exactly is the problem, especially if we can sidestep a circular dependency error by using a pointer-to-type rather than a fully initialised member?

RobBuchananCompPhys: For instance, could we use ` ModifyChargesModel *modifyChargesModel_{nullptr};` as a member in the `gui` class?

trisyoungs: Yeah, the emoji here was only related to the placement of a GUI element within the `dissolve` class!

src/gui/menu_species.cpp
trisyoungs: Ah, OK.  So why not just create the necessary model as a local variable just ahead of the dialog in the various functions? If the dialog isn't `static` and we're happy to create it every time, then I don't see why we wouldn't do the same for the associated model as well.

src/main/dissolve.h
RobBuchananCompPhys: Ah got it!

src/gui/menu_species.cpp
```diff
@@ -434,10 +437,15 @@ void DissolveWindow::on_SpeciesScaleChargesAction_triggered(bool checked)
     if (!species)
         return;
 
-    static ScaleChargesDialog dialog(this);
+    if (!modifyChargesModel)
+    {
+        modifyChargesModel = new ModifyChargesModel;
+    }
+
+    ModifyChargesDialog dialog(this, modifyChargesModel, ModifyChargesModel::Scaling);
```
trisyoungs: If the model is not required to persist between uses of `ModifyChargesDialog` I still wonder whether we're better just doing something like this and avoiding the class member variable altogether:
```suggestion
    ModifyChargesDialog dialog(this, new ModifyChargesModel(), ModifyChargesModel::Scaling);
```

src/gui/menu_species.cpp
```diff
@@ -27,6 +28,8 @@
 #include <qpushbutton.h>
 #include <valarray>
 
+#include <QDebug>
+
```
trisyoungs:
```suggestion
```

src/gui/models/modifyChargesModel.cpp
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#include "gui/models/modifyChargesModel.h"
+#include <QMessageBox>
+
+#include <QDebug>
+
```
trisyoungs:
```suggestion
```

src/gui/modifyChargesDialog.cpp
```diff
@@ -0,0 +1,31 @@
...
+    : QDialog(parent)
+{
+    view_ = new QQuickWidget(this);
+    view_->rootContext()->setContextProperty("modify", modify);
+    view_->rootContext()->setContextProperty("dialogModel", dialogModel);
+    view_->setSource(QUrl("qrc:/dialogs/qml/modifyCharges/Loader.qml"));
+
+    view_->setResizeMode(QQuickWidget::SizeRootObjectToView);
+
+    QHBoxLayout *topLeftLayout = new QHBoxLayout;
```
trisyoungs:
```suggestion
    auto *topLeftLayout = new QHBoxLayout;
```

trisyoungs (commented):
Looking good.  Like the unit test!

I still have a question over whether we really need to stash a pointer to a `ModifyChargesModel` anywhere - rather, can't we just `new` one when we need it (see my comment)?

src/gui/menu_species.cpp
RobBuchananCompPhys: The model _should_ persist between uses of the dialog, so as the user switches between different instances of the dialog (scale, smooth, reduce sig fig) the established values for each do not simply go out of scope and are lost. These values must be contained in the `ModifyChargesModel`, so they can be found again next time the dialog is created.

trisyoungs (approved):
Looking very nice now. Regarding the `font.pointSize` issue, do we then just remove that for now until we understand the (cross-platform) effects a bit better?

Also, I have pushed a small commit on to the PR which just moves the declaration of the model in `gui.h` to more of its own section. Felt silly explaining that and requesting a change!

RobBuchananCompPhys:
> Looking very nice now. Regarding the `font.pointSize` issue, do we then just remove that for now until we understand the (cross-platform) effects a bit better?
> 
> Also, I have pushed a small commit on to the PR which just moves the declaration of the model in `gui.h` to more of its own section. Felt silly explaining that and requesting a change!

Thanks @trisyoungs !
I have asked that @rprospero takes a look at how this looks post-`font.pointSize: 11` commit just to get another opinion, currently sounds like it may well be a graphics issue. But I'd certainly consider holding out and reverting that particular commit if we can't pin down exactly why the odd size changes are occurring on diff platforms.

On another note, I noticed that the simulation menu also utilises `Reduce significant figures in charges...` function, so I'll add a new commit which essentially does the same thing in the species menu for sig figs. We can maybe discuss that one more in stand up tomorrow.

rprospero:
@trisyoungs @RobBuchananCompPhys Here's the dialog on my machine.

dialog

RobBuchananCompPhys:
> @trisyoungs @RobBuchananCompPhys Here's the dialog on my machine.
> 
> dialog

Looks promising! Shall we merge?

PR #1789, opened 2024-02-04
Analysis Refactoring - Part 1

jws-1:
This PR continues work towards #1769. Namely it addresses the following:

- [x] Refactor `HistogramCN`
- [x] Refactor `sdf`
- [x] Refactor `OrientedSDF`
- [x] Refactor `IntraDistance`
- [x] Refactor `IntraAngle` and add system test.


src/modules/intraAngle/process.cpp
```diff
@@ -20,19 +19,89 @@ Module::ExecutionResult IntraAngleModule::process(ModuleContext &moduleContext)
...
+            {
+
+                if (siteC->molecule() != siteA->molecule())
+                    continue;
+
+                if (siteC == siteA)
+                    continue;
+
+                if (siteC == siteB)
+                    continue;
```
trisyoungs, line 63: Feels like here there is a way more efficient way of doing things - basically we only pre-select all the A sites, then generate B and C for the specific molecule A exists in.  That said, this may not be a simple refactor and so is probably best left as a follow-up issue.

src/modules/intraAngle/process.cpp
```diff
@@ -20,19 +19,89 @@ Module::ExecutionResult IntraAngleModule::process(ModuleContext &moduleContext)
...
+    hist.accumulate();
+
+    // Angle(ABC)
+    auto &dataNormalisedHisto = processingData.realise<Data1D>("Angle(ABC)", name(), GenericItem::InRestartFileFlag);
+    dataNormalisedHisto = hist.accumulatedData();
+
+    // Normalise
+    DataNormaliser1D normaliser(dataNormalisedHisto);
+    // Normalise by sin(x)
+    normaliser.normaliseByExpression("value/sin(toRad(x))");
```
trisyoungs: Just thinking out loud here but we could have a "developer-specific" lambda-style equivalent here:
```
normaliser.normalise([](const auto x, const auto value) { return value / sin(x / DEGRAD); });
```

trisyoungs: This means we can avoid the slowness of user-defined expression parsing. It's only normalisation, granted, but it's overkill to set an expression in the code in string form, then parse it into machine form!

src/analyser/dataNormaliserBase.h
```diff
@@ -0,0 +1,28 @@
...
+
+    protected:
+    DataND &targetData_;
+
+    public:
+    void normaliseByNumberDensity(double population, Configuration *targetConfiguration)
+    {
+        targetData_ /= (population / targetConfiguration->box()->volume());
+    }
+    void normaliseBySitePopulation(double population) { targetData_ /= population; }
```
trisyoungs: I think that we can simplify things here and replace these with a single function `normaliseDivide` which takes the `double` value to divide the data through by. That way the class doesn't need to use a `Configuration` at all (this is just a very specific divisor).

src/analyser/dataNormaliserBase.h
```diff
@@ -0,0 +1,28 @@
...
+    virtual void normaliseBySphericalShell() = 0;
+    virtual void normaliseByValue(double value = 1.0, bool absolute = true) = 0;
```
trisyoungs: In the interests of both sanity and clarity I suggest we rename this one to avoid confusion:
```suggestion
    virtual void normaliseTo(double value = 1.0, bool absolute = true) = 0;
```

trisyoungs (changes requested):
I think this is all absolutely glorious - seeing the module code shrink down to what you have here is quite satisfying!

Couple of comments to think about, but we should be good to merge once they're done.

src/modules/intraAngle/process.cpp
jws-1: Yep, I totally agree. The same will go for #1795. I think a general "optimisation" pass across the analysis modules would follow nicely on both pieces of work.

src/modules/intraAngle/process.cpp
jws-1: Again, agreed, and this change would be more in-line with the direction we are going with the analysis modules.

trisyoungs:
I'll leave you to capture the optimisation as a new issue. Approving this now.

trisyoungs (approved):

jws-1:
> I'll leave you to capture the optimisation as a new issue. Approving this now.

Just want to double check that you saw https://github.com/disorderedmaterials/dissolve/pull/1789/commits/6c45ec2bf98b83ad763a8589f70134c75c6489e1 before I merge!

trisyoungs:
> > I'll leave you to capture the optimisation as a new issue. Approving this now.
> 
> Just want to double check that you saw [6c45ec2](https://github.com/disorderedmaterials/dissolve/commit/6c45ec2bf98b83ad763a8589f70134c75c6489e1) before I merge!

I did. I like it!

PR #1792, opened 2024-02-06
1690 Molecularize CIF

### LINKED ISSUE #1690: Molecularize CIF handling a bit better
Data management within the `CIFHandler` could be tidied significantly if we introduce a sort of `StaticMolecule` class - one which inherits `Molecule` but which maintains and references its own `Atom` vector.  That way we can harmonise storage of data, make use of functions within the `Molecule` class (principally `unFold()`) and reduce code.

trisyoungs:
This PR focuses on our CIF import functionality, refactoring a few things here and there and generally making life a bit easier.  A bug is fixed in the generation of the supercell whereby individual `Species` were generated for every molecule instance, and the handling of various output options in the wizard was kinda broken.

It ended up being a bit larger than intended!

Closes #1690.

### DIFF
--- a/src/io/import/cif.cpp
+++ b/src/io/import/cif.cpp
@@ -584,9 +584,9 @@ bool CIFHandler::detectMolecules()
     auto idx = 0;
     while (!allAtomIndices.empty())
     {
-        // Choose a fragment
-        auto fragment = cleanedUnitCellSpecies_->fragment(idx);
-        std::sort(fragment.begin(), fragment.end());
+        // Select a fragment
+        auto fragmentIndices = cleanedUnitCellSpecies_->fragment(idx);
+        std::sort(fragmentIndices.begin(), fragmentIndices.end());
 
         // Set up the CIF species
         auto *sp = coreData_.addSpecies();
@@ -596,8 +596,8 @@ bool CIFHandler::detectMolecules()
         std::vector<int> speciesAtomIndices(sp->nAtoms());
         std::iota(speciesAtomIndices.begin(), speciesAtomIndices.end(), 0);
         std::vector<int> indicesToRemove;
-        std::set_difference(speciesAtomIndices.begin(), speciesAtomIndices.end(), fragment.begin(), fragment.end(),
-                            std::back_inserter(indicesToRemove));
+        std::set_difference(speciesAtomIndices.begin(), speciesAtomIndices.end(), fragmentIndices.begin(),
+                            fragmentIndices.end(), std::back_inserter(indicesToRemove));
         sp->removeAtoms(indicesToRemove);
 
         // Give the species a name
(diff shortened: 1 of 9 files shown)


src/io/import/cif.h
```diff
@@ -140,7 +140,7 @@ class CIFHandler
 
     private:
     // Create basic unit cell
-    bool createBasicUnitCel();
+    bool createBasicUnitCell();
```
jws-1, line 143: Oops! Not sure how that survived here for so long!

src/io/import/cif.cpp
```diff
@@ -577,61 +577,86 @@ bool CIFHandler::detectMolecules()
...
+        sp->setName(EmpiricalFormula::formula(sp->atoms(), [&](const auto &at) { return at.Z(); }));
 
-        // Find copies of this fragment
-        auto copies = speciesCopies(cleanedUnitCellSpecies_, neta.value());
+        // Fix geometry of the extracted species
+        fixGeometry(sp, cleanedUnitCellSpecies_->box());
 
-        // Remove the current fragment and all copies
-        for (auto &copy : copies)
+        // Find instances of this fragmentIndices. For large fragments that represent > 50% of the remaining atoms we don't even
```
jws-1:
```suggestion
        // Find instances of this fragment. For large fragments that represent > 50% of the remaining atoms we don't even
```

src/io/import/cif.cpp
```diff
@@ -577,61 +577,86 @@ bool CIFHandler::detectMolecules()
...
+        // attempt to create a NETA definition etc. For cases such as framework species this will speed up detection no end.
+        std::vector<CIFLocalMolecule> instances;
+        if (fragmentIndices.size() * 2 > allAtomIndices.size())
```
jws-1, line 612: :+1:

src/io/import/cif.cpp
```diff
@@ -577,61 +577,86 @@ bool CIFHandler::detectMolecules()
...
-                                         [&](int value) { return std::find(copy.begin(), copy.end(), value) != copy.end(); }),
-                          indices.end());
-        }
-
-        // Determine coordinates
-        auto coords = speciesCopiesCoordinatesFromUnitCell(sp, cleanedUnitCellSpecies_->box(), copies);
-
-        // Fix geometry
-        fixGeometry(sp, cleanedUnitCellSpecies_->box());
+            // Create an instance of the current fragmentIndices
```
jws-1:
```suggestion
            // Create an instance of the current fragment
```

src/io/import/cif.cpp
```diff
@@ -577,61 +577,86 @@ bool CIFHandler::detectMolecules()
...
+            allAtomIndices.erase(std::remove_if(allAtomIndices.begin(), allAtomIndices.end(),
+                                                [&](int value) {
+                                                    return std::find(fragmentIndices.begin(), fragmentIndices.end(), value) !=
+                                                           fragmentIndices.end();
+                                                }),
+                                 allAtomIndices.end());
+        }
+        else
+        {
+            // Determine a unique NETA definition describing the fragmentIndices
```
jws-1:
```suggestion
            // Determine a unique NETA definition describing the fragment
```

src/io/import/cif.cpp
```diff
@@ -577,61 +577,86 @@ bool CIFHandler::detectMolecules()
...
+            auto neta = uniqueNETADefinition(sp);
+            if (!neta.has_value())
+                return Messenger::error("Couldn't generate molecular partitioning for CIF - no unique NETA definition for the "
+                                        "fragmentIndices {} could be determined.\n",
```
jws-1:
```suggestion
                                        "fragment {} could be determined.\n",
```

src/io/import/cif.cpp
```diff
@@ -577,61 +577,86 @@ bool CIFHandler::detectMolecules()
...
+                                        EmpiricalFormula::formula(sp->atoms(), [](const auto &i) { return i.Z(); }));
```
jws-1:
```suggestion
                                        sp->name());
```

src/io/import/cif.cpp
```diff
@@ -577,61 +577,86 @@ bool CIFHandler::detectMolecules()
...
+
+            // Find instances of this fragmentIndices
+            instances = getSpeciesInstances(sp, *neta);
+
+            // Remove the current fragmentIndices and all copies
```
jws-1:
```suggestion
            // Remove the current fragment and all copies
```

jws-1 (approved):
I really like this piece of work, particularly the introduction of the `CIFLocalMolecule` class. I think this cleans things up nicely. Some very minor comments, which may be simply my misunderstanding.

src/io/import/cif.cpp
trisyoungs: Ah, yes. Lazy search/replacing on my part!

src/io/import/cif.h
trisyoungs: :shrug:

PR #1795, opened 2024-02-11
Analysis Refactoring - Part 2

jws-1:
This PR continues work towards #1769, following on from #1789.

 Addresses the following:
- [x] Refactor `angle`
- [x] Add `angle` system test
- [x] Refactor `axisAngle`
- [x] Add `axisAngle` system test
- [x] Refactor `dAngle`


src/analyser/dataNormaliser2D.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    // We expect x values to be centre-bin values, and regularly spaced
+    const auto &xAxis = targetData_.xAxis();
+    const auto &yAxis = targetData_.yAxis();
+    auto &values = targetData_.values();
+    if (xAxis.size() > 2)
+    {
+        // Derive first left-bin boundary from the delta between points 0 and 1
+        double leftBin = xAxis[0] - (xAxis[1] - xAxis[0]) * 0.5, rightBin, divisor;
+        double r1Cubed = pow(leftBin, 3), r2Cubed;
```
trisyoungs:
```suggestion
        auto leftBin = xAxis[0] - (xAxis[1] - xAxis[0]) * 0.5, rightBin, divisor;
        auto r1Cubed = pow(leftBin, 3), r2Cubed;
```

src/analyser/dataNormaliser2D.h
```diff
@@ -0,0 +1,24 @@
...
+
+using NormalisationFunction2D =
+    std::function<double(const double &, const double &, const double &, const double &, const double &)>;
+class DataNormaliser2D : public DataNormaliserBase<Data2D, NormalisationFunction2D>
+{
+    public:
+    DataNormaliser2D(Data2D &targetData);
+
+    /*
+     * Normalisation functions
```
trisyoungs: Picky picky!
```suggestion
     * Normalisation Functions
```

src/modules/axisAngle/process.cpp
```diff
@@ -21,25 +22,109 @@ Module::ExecutionResult AxisAngleModule::process(ModuleContext &moduleContext)
...
+    auto &dAxisAngleNormalised = processingData.realise<Data2D>("DAxisAngle", name(), GenericItem::InRestartFileFlag);
+    dAxisAngleNormalised = dAxisAngle.accumulatedData();
+    DataNormaliser2D dAxisAngleNormaliser(dAxisAngleNormalised);
+    // Normalise by value / sin(y) / sin(yDelta)
+    dAxisAngleNormaliser.normalise([&](const auto &x, const auto &xDelta, const auto &y, const auto &yDelta, const auto &value)
+                                   { return (symmetric_ ? value : value * 2.0) / sin(y / DEGRAD) / sin(yDelta / DEGRAD); });
+    // Normalise by A site population
+    dAxisAngleNormaliser.normaliseDivide(double(a.sites().size()));
+    // Normalise by B site population density
+    dAxisAngleNormaliser.normaliseDivide(double(b.sites().size()) / targetConfiguration_->box()->volume());
```
trisyoungs, line 110: Looking at all these uses of the `DataNormaliserND` classes it feels more and more like we should rename them to reflect what they are actually doing, and that is (for the most part) just "operating" on data. Normalisation to a value or to spherical shell are specific examples of operations, but `normaliseDivide` is actually just `divide` in reality.  As a follow-on issue I'd suggest we do a rename of the classes to `DataOperatorND` which can contain both the specific functions (e.g. `normaliseTo` and `normaliseBySphericalShell`) as well as generics like `divide()` and `multiply()`.

src/modules/dAngle/process.cpp
```diff
@@ -21,25 +22,131 @@ Module::ExecutionResult DAngleModule::process(ModuleContext &moduleContext)
...
-        return ExecutionResult::Failed;
+        for (const auto &[siteB, indexB] : b.sites())
+        {
+
+            if (siteB->molecule() != siteA->molecule())
+                continue;
+
+            nBCumulative++;
+            nCSelections++;
+            nBAvailable++;
```
trisyoungs: It's a tiny, tiny thing, but prefer prefix increment over postfix (not for speed, but for intent in the sense that "we don't want use the value, just increase it").
```suggestion
            ++nBCumulative;
            ++nCSelections;
            ++nBAvailable;
```

src/analyser/dataExporter.h
```diff
@@ -0,0 +1,30 @@
...
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#pragma once
+
+#include "base/processPool.h"
+
+template <typename DataND, typename DataNDExportFileFormat> class DataExporter
+{
+    public:
+    static bool exportData(const DataND &targetData, DataNDExportFileFormat &fileAndFormat, const ProcessPool &procPool)
```
trisyoungs, line 12:
```suggestion
    // Try to export the specified data, if a valid filename has been provided
    static bool exportData(const DataND &targetData, DataNDExportFileFormat &fileAndFormat, const ProcessPool &procPool)
```

trisyoungs (approved):
More fine work - the positive difference in readability between the two still makes me very happy!

PR #1799, opened 2024-02-16
1791 Mugglify CIF Importer

### LINKED ISSUE #1791: Mugglify the CIF Importer
Our CIF import currently revolves around a wizard, which is fine, and provides a nice separation between steps of the CIF processing that we do, but for larger crystals or "bespoke" cases containing a lot of raw atoms and no symmetry elements it is painfully slow.  This is because we repeat the whole processing of the CIF at each stage of the wizard, even if nothing has changed.

The proposal here is to create a single-window import dialog, with adjustment controls on the left (e.g. bonding, moiety removal), a viewer in the centre, and output options / data (e.g. cell size) on the right. This would also preclude the need to maintain configurations for each stage in the `CIFHandler` class.

trisyoungs:
This PR performs a significant refactoring on the `CIFImportDialog`, turning it from a wizard into a fully-featured single-page dialog. Aside from, IMHO, a better UX, a significant secondary reason was the desire to streamline the underlying CIF generation and move away from doing the full, multi-stage generation every time a parameter was changed.

The dialog itself now looks like this, with a toolbox on the left-hand side containing all the tweakable parameters, but the majority of space given over to actually seeing what the generated structure will be:
image

In other news, we also remove the local `CoreData` object and just go for local instances of the necessary (temporary) objects. A few bug fixes also along the way. Lovely.

Closes #1791.

### DIFF
--- a/src/gui/importCIFDialog.cpp
+++ b/src/gui/importCIFDialog.cpp
@@ -62,9 +62,9 @@ void ImportCIFDialog::updateWidgets()
     ui_.InfoPublicationTitleLabel->setText(
         QString::fromStdString(cifHandler_.getTagString("_publ_section_title").value_or("<Unknown>")));
     ui_.InfoPublicationReferenceLabel->setText(QString::fromStdString(fmt::format(
-        "{}, {}, <b>{}</b>, {}", cifHandler_.getTagString("_journal_name_full").value_or("???"),
-        cifHandler_.getTagString("_journal_year").value_or("???"), cifHandler_.getTagString("_journal_volume").value_or("???"),
-        cifHandler_.getTagString("_journal_page_first").value_or("???"))));
+        "{}, {}, <b>{}</b>, {}", cifHandler_.getTagString("_journal_name_full").value_or("N/A"),
+        cifHandler_.getTagString("_journal_year").value_or("N/A"), cifHandler_.getTagString("_journal_volume").value_or("N/A"),
+        cifHandler_.getTagString("_journal_page_first").value_or("N/A"))));
     ui_.InfoAuthorsList->clear();
     auto authors = cifHandler_.getTagStrings("_publ_author_name");
     for (auto &author : authors)
--- a/src/gui/importCIFDialog.cpp
+++ b/src/gui/importCIFDialog.cpp
@@ -69,7 +69,7 @@ void ImportCIFDialog::finalise()
  */
 
 // Update all controls
-void ImportCIFDialog::update()
+void ImportCIFDialog::updateWidgets()
 {
     Locker updateLock(widgetsUpdating_);
 
@@ -149,7 +149,7 @@ void ImportCIFDialog::on_InputFileEdit_editingFinished()
     else
     {
         cifHandler_.generate();
-        update();
+        updateWidgets();
     }
 }
 
@@ -172,7 +172,7 @@ void ImportCIFDialog::on_SpaceGroupsCombo_currentIndexChanged(int index)
 
     cifHandler_.setSpaceGroup((SpaceGroups::SpaceGroupId)(ui_.SpaceGroupsCombo->currentIndex() + 1));
 
-    update();
+    updateWidgets();
 }
 
 void ImportCIFDialog::on_NormalOverlapToleranceRadio_clicked(bool checked)
@@ -180,7 +180,7 @@ void ImportCIFDialog::on_NormalOverlapToleranceRadio_clicked(bool checked)
     if (checked)
     {
         cifHandler_.setOverlapTolerance(0.1);
-        update();
+        updateWidgets();
     }
 }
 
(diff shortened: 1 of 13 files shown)


jws-1:
I see this is not ready for review yet, but just wanted to chime in and say I like the new look of things! A single page dialog is definitely the way to go (I got very fed up of the wizard when I was working on this stuff before!).

trisyoungs:
Yeah, I think my view on wizards has changed to the polar opposite now - can't stand em!  Particularly in the case of the CIF import, it felt like didn't give the users or the developers the best view on things.

src/gui/menu_species.cpp
```diff
@@ -169,7 +169,9 @@ void DissolveWindow::on_SpeciesImportFromCIFAction_triggered(bool checked)
         fullUpdate();
 
         // Select the new species
-        ui_.MainTabs->setCurrentTab(dissolve_.coreData().species().back().get());
+        auto *sp = dissolve_.coreData().species().back().get();
+        if (sp)
+            ui_.MainTabs->setCurrentTab(dissolve_.coreData().species().back().get());
```
rhinoella:
```suggestion
            ui_.MainTabs->setCurrentTab(sp);
```

src/gui/importCIFDialog.cpp
```diff
@@ -326,84 +219,103 @@ bool ImportCIFDialog::createMoietyRemovalNETA(std::string definition)
...
 
-void ImportCIFDialog::on_MoietyRemoveByNETAGroup_clicked(bool checked)
+void ImportCIFDialog::on_MoietyRemoveByNETACheck_clicked(bool checked)
 {
-    if (checked)
-        updateFlags_.setFlag(CIFHandler::UpdateFlags::CleanMoietyRemoveNETA);
-    else
-        updateFlags_.removeFlag(CIFHandler::UpdateFlags::CleanMoietyRemoveNETA);
+    if (ui_.MoietyNETARemovalIndicator->state() != CheckIndicator::OKState)
+        return;
```
rhinoella, line 238: Currently the MoietyNETARemoveFragmentsCheck and the MoietyNETARemovalEdit are disabled, I am assuming here is where they should be enabled.

```suggestion

ui_.MoietyNETARemovalEdit->setEnabled(checked);
ui_.MoietyNETARemoveFragmentsCheck->setEnabled(checked);

```

src/gui/importCIFDialog.cpp
```diff
@@ -2,204 +2,144 @@
...
+    // Chemical Formula
+    ui_.InfoChemicalFormulaLabel->setText(QString::fromStdString(cifHandler_.chemicalFormula()));
+
+    // Publication Data
+    ui_.InfoPublicationTitleLabel->setText(
+        QString::fromStdString(cifHandler_.getTagString("_publ_section_title").value_or("<Unknown>")));
+    ui_.InfoPublicationReferenceLabel->setText(QString::fromStdString(fmt::format(
+        "{}, {}, <b>{}</b>, {}", cifHandler_.getTagString("_journal_name_full").value_or("???"),
+        cifHandler_.getTagString("_journal_year").value_or("???"), cifHandler_.getTagString("_journal_volume").value_or("???"),
+        cifHandler_.getTagString("_journal_page_first").value_or("???"))));
```
rhinoella: Personally, the question marks feel a bit aggressive, I think leaving it blank would look more professional.

src/gui/importCIFDialog.cpp
```diff
@@ -209,111 +149,64 @@ void ImportCIFDialog::on_InputFileSelectButton_clicked(bool checked)
...
-    else
-        updateFlags_.removeFlag(CIFHandler::PreventMetallicBonding);
+    {
+        cifHandler_.setUseCIFBondingDefinitions(true);
+        updateWidgets();
+    }
 
-    if (ui_.MainStack->currentIndex() == ImportCIFDialog::StructurePage)
-        update();
+    updateWidgets();
```
rhinoella:
```suggestion

```

rhinoella: Could also just change to:

```
cifHandler_.setUseCIFBondingDefinitions(checked);
updateWidgets();
```

And remove if statement

rhinoella (changes requested):
Looks good, minor comments.

src/gui/importCIFDialog.cpp
trisyoungs: Actually, the logic to enable/disable those controls based on the state of `ui_.MoietyNETARemoveCheck` is captured in the UI file itself with signals/slots/

src/gui/importCIFDialog.cpp
trisyoungs: Agreed, but I'll switch to `N/A`.

src/gui/importCIFDialog.cpp
trisyoungs: Oh, and which I see now somehow don't work.....

src/gui/importCIFDialog.cpp
trisyoungs: Best not to since these callbacks on the radio buttons should only ever run if the radio has been checked as the  companion radio button represents the other state choice (`false`).

rhinoella (approved):
Looks good!

PR #1801, opened 2024-02-19
1797 Setup After Restart Load

### LINKED ISSUE #1797: Should run setup stages after loading a restart file
**Describe the bug**
We should run the setup stage after loading a restart file, since at this point any Configurations are available, and any Fourier transforms dependent on the configuration density will need to be rerun.

trisyoungs:
This PR ensures that the setup stage of all modules is run (via `Dissolve::prepare()` after loading a restart file in the GUI.

Along the way an old function was removed, and more suitable button texts are now used in the `SelectRestartFileDialog`.

Closes #1797.


Danbr4d (approved):
Looks good to me and loads in the correct G(r) from the start now

PR #1804, opened 2024-02-21
1576 More CIF Unit Tests

### LINKED ISSUE #1552: Molecular CIF - Symmetry
Currently, the functionality implemented for importing CIFs of a molecular nature do not handle symmetry. They should.

### LINKED ISSUE #1576: Testing CIF import

### LINKED ISSUE #1688: Molecule detection in CIF import fails if unique atom match can't be found
**Describe the bug**
Our CIF handler class attempts to create a unique NETA definition in order to determine instances of molecular species, however this will fail if a unique reference atom cannot be found (e.g. as is the case for the Bisphen_n_arenes_1517789.cif example input).

We should be able to match, even if a best-case unique reference can't be found, but some experimentation is needed.

trisyoungs:
This PR refactors a large chunk of the `CIFHandler` class, specifically relating to the detection of molecular species within the generated crystal.

Briefly, this process relies on being able to locate "`Species`-consistent" copies of the same molecule from within, effectively, a set of atoms which we must assume has random order. Previously we relied on a very simple sort of matched `SpeciesAtom`s in the NETA match to enforce this, but that did not get to the root of the issue.

~Here we extend the `getSpeciesInstances()` function to reorder atoms into "`Species`-conformant" sequences by comparing the molecule in 3D to a reference set of coordinates (those first chosen by the molecule detection stage). We also improve the detection of NETA for the target molecule, going for the "best" match possible rather than one which is unique (and may not exist - #1688).~

**Update** We now perform similarity matching by considering "local" connectivity of individual atoms, walking around the molecule from each of the root whole-molecule NETA atom matches we establish.  This seems to work rather well!

Closes #1576.
Closes #1552 (or at least makes it out-of-date).
Closes #1688.


src/io/import/cif.cpp
```diff
@@ -1036,90 +1036,284 @@ void CIFHandler::applyCIFBonding(Species &sp, bool preventMetallicBonding)
...
+        {
+            auto idx = instanceSpecies.addAtom(matchedAtom->Z(), matchedAtom->r(), 0.0, matchedAtom->atomType());
 
-        // Get the (sorted) matched atom indices
-        std::vector<int> indices(matchedAtoms.size());
-        std::transform(matchedAtoms.begin(), matchedAtoms.end(), indices.begin(),
-                       [](const auto matchedAtom) { return matchedAtom->index(); });
+            // Store the index of the root atom in match in our instance species when we find it
+            if (matchedAtom == &atom)
+                rootAtomLocalIndex = idx;
```
rhinoella, line 1136:
```suggestion
            // Store the index of the root atom in match in our instance species when we find it
            if (matchedAtom == &atom)
                rootAtomLocalIndex = instanceSpecies.addAtom(matchedAtom->Z(), matchedAtom->r(), 0.0, matchedAtom->atomType());
```

src/io/import/cif.cpp
```diff
@@ -1036,90 +1036,284 @@ void CIFHandler::applyCIFBonding(Species &sp, bool preventMetallicBonding)
...
+        // Found a fragment that matches the NETA description - we now create a temporary instance Species which will contain
+        // the selected fragment atoms, reassembled into a molecule (i.e. unfolded) and with bonding applied / calculated.
+        // We need to copy the unit cell from the crystal so we detect bonds properly.
+        Species instanceSpecies;
+        instanceSpecies.createBox(unitCellSpecies_.box()->axisLengths(), unitCellSpecies_.box()->axisAngles());
+        auto rootAtomLocalIndex = -1;
+        // -- Create species atoms from those matched in the unit cell by the NETA description.
+        for (auto &matchedAtom : matchedUnitCellAtoms)
+        {
+            auto idx = instanceSpecies.addAtom(matchedAtom->Z(), matchedAtom->r(), 0.0, matchedAtom->atomType());
```
rhinoella, line 1132:
```suggestion
```

rhinoella (approved):
👍

src/io/import/cif.cpp
```diff
@@ -1036,90 +1036,284 @@ void CIFHandler::applyCIFBonding(Species &sp, bool preventMetallicBonding)
...
+
+        // Unfold the molecule and store the unfolded molecule coordinates back into the instance Species.
+        // This represents our full instance coordinates we will be storing (but not their final order)
+        instanceMolecule.unFold(unitCellSpecies_.box());
+        for (auto &&[molAtom, spAtom] : zip(instanceMolecule.localAtoms(), instanceSpecies.atoms()))
+            spAtom.setCoordinates(molAtom.r());
+
+        /*
+         * Now, we have a root match atom on the current instance and a vector of possible matching sites on the reference
+         * species (in referenceRootAtoms). For each of the referenceRootAtoms, try to incremental select along bonds using
```
jws-1:
```suggestion
         * species (in referenceRootAtoms). For each of the referenceRootAtoms, try to incrementally select along bonds using
```

src/io/import/cif.cpp
```diff
@@ -1036,90 +1036,284 @@ void CIFHandler::applyCIFBonding(Species &sp, bool preventMetallicBonding)
...
+        const auto &closestMolSpAtom = molecule.species()->atom(atomIndexMap[spI]);
+        difference += distanceSq;
+        if (spAtom.Z() != closestMolSpAtom.Z())
+            difference += std::max(spAtom.Z(), closestMolSpAtom.Z()) * 10.0;
+    }
 
-    // Update the coordinates of the species atoms
-    for (auto &&[spAtom, atom] : zip(sp->atoms(), molAtoms))
-        spAtom.setCoordinates(atom.r());
+    Messenger::print("NBADATOMS = {} of {}\n", nBadAtoms, species->nAtoms());
```
jws-1:
```suggestion
```
Not sure if this was intentionally left in!

src/io/import/cif.cpp
trisyoungs: Alas we can't do this - need the full loop over atoms.

src/classes/speciesAtom.cpp
```diff
@@ -158,6 +158,38 @@ OptionalReferenceWrapper<SpeciesBond> SpeciesAtom::getBond(const SpeciesAtom *pa
...
+            // Determine angle
+            auto angle = Box::literalAngleInDegrees(i->r(), r_, k->r());
+            if (angle < maxAngle)
+            {
+                pairs.emplace_back(i, k);
+                if (!uniquePairsOnly)
+                    pairs.emplace_back(k, i);
+            }
+        }
+    }
```
rprospero:
```suggestion
    dissolve::for_each_pair(ParallelPolicies:seq, bonds_.begin(), bonds_.end(), [this, uniquePairsOnly, &pairs](int b1, const auto bond1, int b2, const auto bond2) {
      auto *i = bond1.get().partner(this);
      auto *k = bond2.get().partner(this);
      if (i == k)
        return;

      // Determine angle
      auto angle = Box::literalAngleInDegrees(i->r(), r_, k->r());
      if (angle < maxAngle)
      {
          pairs.emplace_back(i, k);
          if (!uniquePairsOnly)
              pairs.emplace_back(k, i);
      }
    });
```
I thought that we might be able to simplify things with our pair algorithm. Please note that I haven't tested this code!

tests/io/cif.cpp
```diff
@@ -143,4 +167,45 @@ TEST_F(ImportCIFTest, CuBTC)
...
+
+TEST_F(ImportCIFTest, MoleculeOrdering)
+{
+    CIFHandler cifHandler;
+    const auto cifFiles = {"cif/molecule-test-simple-ordered", "cif/molecule-test-simple-unordered",
+                           "cif/molecule-test-simple-unordered-rotated"};
+    for (auto cifFile : cifFiles)
+    {
+        // Load the CIF file
+        ASSERT_TRUE(cifHandler.read(fmt::format("{}.cif", cifFile)));
```
rprospero:
```suggestion
    const auto cifFiles = {"cif/molecule-test-simple-ordered.cif", "cif/molecule-test-simple-unordered.cif",
                           "cif/molecule-test-simple-unordered-rotated.cif"};
    for (auto cifFile : cifFiles)
    {
        // Load the CIF file
        ASSERT_TRUE(cifHandler.read(cifFile));
```

I thought that this might be a bit clearer, though there might be a deeper meaning that I missed behind separating the extension.

tests/io/cif.cpp
```diff
@@ -143,4 +167,45 @@ TEST_F(ImportCIFTest, CuBTC)
...
+    }
+}
+
+TEST_F(ImportCIFTest, BigMoleculeOrdering)
+{
+    CIFHandler cifHandler;
+    const auto cifFile = "cif/Bisphen_n_arenes_1517789";
+
+    // Load the CIF file
+    ASSERT_TRUE(cifHandler.read(fmt::format("{}.cif", cifFile)));
```
rprospero:
```suggestion
    const auto cifFile = "cif/Bisphen_n_arenes_1517789.cif";

    // Load the CIF file
    ASSERT_TRUE(cifHandler.read(cifFile));
```

Same as above

rprospero (approved):
A couple of optional changes, but otherwise things look good.  It's going to break my SoA branch, but it'll at least break things in a way that I'm ready for.

src/classes/speciesAtom.cpp
trisyoungs: This looks nice and I should have done it this way before, but you've actually highlighted a bit of now-redundant code! This was previously used to get pairs of atoms to form rotation bases, but since we're not doing things geometrically any more I can just remove the function entirely.

tests/io/cif.cpp
trisyoungs: Another good catch.  This was left-over from when the testing compared against a reference XYZ file on disk, but since now we just use the cleaned unit cell coordinates as the reference the extensions don't need to be swapped.

jws-1 (approved):
:+1:

src/io/import/cif.cpp
trisyoungs: Nope! :+1:

PR #1807, opened 2024-02-23
1806 faster combinations

rprospero:
I've changed our `Combinations` class to use a closed form calculation of the indices that is O(1), instead of the O(N) formula that had been in place.  This does mean that the class can no longer be generic into higher dimensions of pairs, but it never actually supported triplets in the first place, so nothing has been lost.

On the Large CalculateGR benchmark, I've measured an over 300% speed improvement from the improved cell version.


src/math/combinations.cpp
```diff
@@ -2,49 +2,19 @@
...
-    {
-        atPrev = at;
-        at = atPrev + 1;
-        nCtPrev = nCt;
-        nCt = nChooseK(at, k);
-        if (nCt > N)
-            found = true;
-    }
-    return {atPrev, N - nCtPrev};
+    int i = std::floor((-b + std::sqrt(b * b + 4 * a * n)) / (2 * a));
```
RobBuchananCompPhys: Looks like a quadratic formula - could we break this out and modularise into its own function somewhere in the math library? Could be useful to access it down the line.

src/math/combinations.cpp
```diff
@@ -2,49 +2,19 @@
...
+    int j = n - std::floor(a * i * i + b * i) + i + 1;
```
RobBuchananCompPhys: Similar thing here - the readability of this line could benefit a little perhaps if the math were in a function.
If it's a 2nd order polynomial in the `std::floor` arg, it could be
`double polyEval(int order, double i, double &coeffArray)` or something to that effect.

RobBuchananCompPhys (commented):
Speedup sounds very promising! Just raised two points about readability of inlined math, otherwise looks good 👍

src/math/combinations.cpp
```diff
@@ -2,49 +2,19 @@
...
-Combinations::NthCombination Combinations::getCoefficent(int N, int k) const
-{
-
-    auto at = k - 1;
-    auto nCt = 0;
-    if (nCt >= N)
-        return {at, N - nCt};
+  /*
+    If we let x be the 
+   */
```
trisyoungs: Suspenseful!

tests/math/combinations.cpp
```diff
@@ -33,7 +34,7 @@ TEST(CombinationTest, CheckNumberOfCombinations)
     }
     std::vector<std::pair<int, int>> combinations;
 
-    Combinations comb(numTests, 2);
+    Combinations comb(numTests);
     for (int i = 0; i < comb.getNumCombinations(); i++)
```
trisyoungs:
```suggestion
    for (auto i = 0; i < comb.getNumCombinations(); ++i)
```

tests/math/combinations.cpp
```diff
@@ -59,20 +60,20 @@ TEST(CombinationTest, CheckCombinations)
...
```
trisyoungs:
```suggestion
    for (auto i = 0; i < comb.getNumCombinations(); ++i)
```

src/math/combinations.cpp
```diff
@@ -2,49 +2,19 @@
...
+    double a = -0.5;
+    double b = N_ - 0.5;
```
trisyoungs:
```suggestion
    auto a = -0.5;
    auto b = N_ - 0.5;
```

trisyoungs (approved):
:+1:

I'd like to understand the math behind it a bit more, but that's a job for some downtime on another day!

src/math/combinations.cpp
rprospero: @trisyoungs Updated comment should be a little more helpful.

RobBuchananCompPhys (approved):

PR #1809, opened 2024-02-26
Simulation data manager as QML TableView

RobBuchananCompPhys:
Things to keep in mind:
- the QML table is not yet hooked up to the sorting/filtering functionality provided by the simulation proxy model. I think it should be relatively straightforward - essentially set the `simProxy` as the `model` context property, and probably some changes on the QML side, so a bit more work probably needed
- Obviously, a search box is needed for the user to be able to filter by text
- Updating unit tests


src/gui/qml/simulationDataManager/SimulationDataManager.qml
```diff
@@ -0,0 +1,91 @@
...
+    }
+    function getHeaderStringArray(model) {
+        var headerArray = [];
+        for (var i = 0; i < model.columnCount(); ++i) {
+            headerArray.push(qsTr(model.headerData(i, Qt.Horizontal)));
+        }
+        return headerArray;
+    }
+
+    GroupBox {
```
rprospero:
```suggestion
    D.GroupBox {
```

rprospero: Use our custom GroupBox widget.  It doesn't matter much at the moment, but we'll be prepared for the future if we set it up now.

src/gui/qml/simulationDataManager/SimulationDataManager.qml
```diff
@@ -0,0 +1,91 @@
...
+                    implicitWidth: tableText.width
+
+                    D.Text {
+                        id: tableText
+                        padding: 12
+                        text: display
+                    }
+                }
+            }
+            Button {
```
rprospero:
```suggestion
            D.Button {
```

Again

rprospero (approved):
A couple of minor suggestions, but generally approved (pending our discussion from Team about wrapping QSortProxyFilter)

PR #1813, opened 2024-03-08
feat: wrap QSortFilterProxy so it is accessible from QML

### LINKED ISSUE #1812: QSortFilterProxy QML Wrapper
Because QSortFilterProxy isn't directly available from QML, it's not possible to bind the filter regex to a property.  The immediate result of this is that the filter on the Data Manager dialog doesn't update until the user hits enter, which the user may never do.  This problem will only get worse as we add more tables to the project.

We need to create a QML wrapper for QSortFilterProxy that can be bound to properties and updated immediately.

rprospero:
This PR creates a tiny wrapper that make the QSortFilterProxy accessible from QML.  I've then applied this to the Data Manager dialog so that the table is filtered the moment a user types a character, instead of waiting for a carriage return.

closes #1812

### DIFF
--- a/src/gui/qml/simulationDataManager/SimulationDataManager.qml
+++ b/src/gui/qml/simulationDataManager/SimulationDataManager.qml
@@ -6,10 +6,6 @@ import "../widgets" as D
 
 Page {
     id: root
-    height: 500
-    visible: true
-    width: 670
-
     function filterByRegExp(proxy, text) {
         proxy.filterRegularExpression = RegExp(text);
     }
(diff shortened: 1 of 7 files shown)


src/gui/qml/simulationDataManager/SimulationDataManager.qml
```diff
@@ -21,6 +21,11 @@ Page {
         return headerArray;
     }
 
+    SortFilterProxy {
+        id: proxy
+        filterRegularExpression: RegExp(searchBox.text)
```
trisyoungs, line 26: Immediate RE support in our filtering?  Marvellous!

trisyoungs (approved):
:+1:

PR #1819, opened 2024-03-14
feat: Running CNs added to SiteRDF

Danbr4d:
Added into SiteRDF the functionality to calculate the running coordination number for a configuration

### DIFF
--- a/src/modules/siteRDF/process.cpp
+++ b/src/modules/siteRDF/process.cpp
@@ -115,15 +115,18 @@ Module::ExecutionResult SiteRDFModule::process(ModuleContext &moduleContext)
         runningCNHist.bin(countCN);
     }
 
-    // Accumulate runningCNHist
+    // Accumulate histogram averages
     runningCNHist.accumulate();
-    // runningCN
-    auto &dataRunningCN = processingData.realise<Data1D>("RunningCNHistogramNorm", name(), GenericItem::InRestartFileFlag);
-    dataRunningCN = runningCNHist.accumulatedData();
-    // Normalise
-    DataNormaliser1D normaliserRunningCN(dataCN);
+
+    // Averaged values for RunningCN
+    Data1D accumulatedRunningCNData = runningCNHist.accumulatedData();
+    DataNormaliser1D normaliserRunningCN(accumulatedRunningCNData);
+
     // Normalise by A site population
     normaliserRunningCN.normaliseDivide(double(a.sites().size()));
+
+    // Create the display data
+    processingData.realise<Data1D>("RunningCN", name(), GenericItem::InRestartFileFlag) = accumulatedRunningCNData;
 
     // Save RDF data?
     if (!DataExporter<Data1D, Data1DExportFileFormat>::exportData(dataRDF, exportFileAndFormat_, moduleContext.processPool()))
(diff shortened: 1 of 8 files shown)


trisyoungs (commented):
Close! The ultimate result at the maximum of the `distanceRange_` is indeed the integral of the whole histogram, and every bin up to that point is the sum of all those before it. So what you can do here is create your `sumRunNHist` histogram (and maybe rename to `runningCNHist` or similar) using the same `distanceRange_` as `histAB`.

Then, you can simply step over the bins in each with a `zip` and keep track of the running bin total from `histAB` as you go, adding those total values at each bin into your new histogram. Note that we should do this for the raw (i.e. instantaneous, just calculated) bins of `histAB` so that we can form an average.

src/modules/siteRDF/process.cpp
```diff
@@ -98,6 +98,35 @@ Module::ExecutionResult SiteRDFModule::process(ModuleContext &moduleContext)
             }
         }
 
+    auto &dataRunningCN = processingData.realise<SampledData1D>("RunningCNTest", name(), GenericItem::InRestartFileFlag);
+    dataRunningCN.initialise(distanceRange_.x);
```
trisyoungs: Careful - `distanceRange_.x` is just the minimum value of the distance range to use.  In fact, you don't need to initialise this class at all, since that will happen automatically on the first addition (`+=`) to it later on.
```suggestion
```

src/modules/siteRDF/process.cpp
```diff
@@ -98,6 +98,35 @@ Module::ExecutionResult SiteRDFModule::process(ModuleContext &moduleContext)
...
+
+    // Zip over all distanceRange and calculate running CN
+    double countCN{};
+    for (const auto &&[x, currentCN] : zip(histAB.binCentres(), histAB.data().values()))
+    {
+        runningCN[x] += currentCN;
+    }
+
+    // Add data into SampledData1D
+    dataRunningCN += runningCN;
```
trisyoungs: In fact, we can do all of this summation in a `std::transform`...
```suggestion
```

src/modules/siteRDF/process.cpp
```diff
@@ -98,6 +98,35 @@ Module::ExecutionResult SiteRDFModule::process(ModuleContext &moduleContext)
...
+
+    // Accumulate instantaneous binValues
+    auto instBinValues = histAB.data();
+
+    // Normalise Data
+    DataNormaliser1D normaliserInstBinValues(instBinValues);
+
+    // Normalise by A site population
+    normaliserInstBinValues.normaliseDivide(double(a.sites().size()));
```
trisyoungs:
```suggestion
    // Normalise data by A site population
    DataNormaliser1D normaliserInstBinValues(instBinValues);
    normaliserInstBinValues.normaliseDivide(double(a.sites().size()));
```

src/modules/siteRDF/process.cpp
```diff
@@ -98,6 +98,35 @@ Module::ExecutionResult SiteRDFModule::process(ModuleContext &moduleContext)
...
+
```
trisyoungs, line 113: Here we go. So we have the normalised bin data in `instBinValues`, and we can convert that into the running coordination numbers in-place by doing this nice `std::transform`. Essentially we just step over each value and calculate and return a replacement value which is the running coordination number (which we keep track of the in the `sum` variable):
```suggestion
    auto sum = 0.0;
    std::transform(instBinValues.values().begin(), instBinValues.values().end(), instBinValues.values().begin(), [&](const auto &currentBin) {sum += currentBin; return sum; } );
```

trisyoungs (approved):

PR #1826, opened 2024-03-26
feat: Override Pair Potentials (#1805)

### LINKED ISSUE #1805: Allow overriding of interatomic potentials
At present we auto-generate all of our pairwise LJ potentials from the supplied atom types and their parameters. This is convenient for the most part but is inflexible when one wishes to, for example, override a specific pair potential between unlike types with a different set of literature values, or to add on an additional term to an existing pair potential (in a similar spirit to EPSR's minimum distance potentials).

trisyoungs:
Provides new options to define `PairPotential` "overrides" which can add to or completely replace base pair potentials generated autouatically.

Closes #1805.


src/classes/potentialMap.cpp
```diff
@@ -37,28 +37,28 @@ bool PotentialMap::initialise(const std::vector<std::shared_ptr<AtomType>> &mast
 
     // Loop over defined PairPotentials
     int indexI, indexJ;
-    for (auto &pot : pairPotentials)
+    for (auto &&[at1, at2, pp] : pairPotentials)
```
rprospero, line 40: If nothing else, this PR is worth it for this change.

src/main/keywords_pairPotentials.cpp
```diff
@@ -144,7 +174,7 @@ bool PairPotentialsBlock::parse(LineParser &parser, Dissolve *dissolve)
                 }
                 break;
             case (PairPotentialsBlock::ShortRangeTruncationWidthKeyword):
-                PairPotential::setShortRangeTruncationWidth(parser.argd(1));
+                Messenger::warn("The 'ShortRangeTruncationWidth' keyword will be deprecated in a future version.\n");
```
rprospero, line 177: Should we add a backward compatibility adapter to move the ShortRangeTruncationWidth for the TOML branch?

rprospero (approved):
Overall, things look good, but I'm worried that this might be the point where we have to start thinking about the backward compatibility handling code.

PR #1827, opened 2024-03-26
test: Show Testing Throws

trisyoungs:
GTest catches exceptions when calling `ASSERT_NO_THROW` etc. but does not report what the thrown message was.

This PR adds some ugly macros to wrap a couple of GTest functions in a catch/re-throw in order to provide better feedback when testing.

Thoughts welcome!

### DIFF
--- a/tests/ff/testKulmala2010.cpp
+++ b/tests/ff/testKulmala2010.cpp
@@ -12,7 +12,7 @@ class Kulmala2010ForcefieldTest : public ::testing::Test
     protected:
     DissolveSystemTest systemTest;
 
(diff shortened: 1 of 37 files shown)


rprospero:
If you run the unit tests with `--gtest_break_on_failure --gtest_catch_exceptions=0`, then you can both see the exceptions and drop into the debugger at the moment a test fails.

trisyoungs:
@rprospero I see, however trying these flags on my local build still doesn't show the thrown message. I noted that the CI showed the thrown message, but I can't figure out how the tests are being run differently (if indeed they are, and this isn't just a local issue for me...).

rprospero (approved):

PR #1828, opened 2024-04-02
parallelise outer for loop using combinable value

RobBuchananCompPhys:
Some more info about this:
Setting a breakpoint in `Histogram1D::accumulatedData` during the unit test and running `gdb` on both `develop` (single-threaded), and this branch (multithreaded), it looks like all siteRDF calcs are returning the same total of the `bins_`, a value of 71022.
The process generates a lot of data and `gdb` is slow to dump it all to a file so currently pretty challenging to hone in exactly on what is going wrong.


trisyoungs:
Closing politely as implemented elsewhere.

PR #1829, opened 2024-04-02
Fix: Correctly Set Working Directory in GUI

rhinoella:
Currently dissolve does not set the working directory to the input file directory when you start from new, causing the exports to spew out in the GUI binary directory.

This PR fixes that, implementing a new function `setInputFilename` in the GUI class.


src/gui/gui.h
```diff
@@ -126,6 +126,8 @@ class DissolveWindow : public QMainWindow
     const int recentFileLimit_;
 
     private:
+    // Set the input file path
+    void setInputFilename(const QString filename);
```
trisyoungs: Tiny thing, but the Qt people like strings to be passed as const-references:
```suggestion
    void setInputFilename(const QString &filename);
```

src/gui/menu_file.cpp
```diff
@@ -7,6 +7,15 @@
 #include <QMessageBox>
 #include <QSettings>
 
+// Set the input filename and the working directory of dissolve
+void DissolveWindow::setInputFilename(const QString filename)
```
trisyoungs:
```suggestion
void DissolveWindow::setInputFilename(const QString &filename)
```

trisyoungs (approved):
:+1:

PR #1837, opened 2024-04-09
refactor: Separate Forms

trisyoungs:
Following on from #1826 this PR performs more reorganisation on the `PairPotential` class, the significant change being that combination of short-range parameters on atom types is no longer the responsibility of `PairPotential`, and instead is moved to the `ShortRangeFunctions` class.  This is in preparation for a follow-up PR which will further decouple knowledge of functional forms from the `PairPotential` class and move them in to `Function1D`.


rprospero (approved):
I know it's technically 16 more lines, but the better organisation makes it feel more compact.

PR #1838, opened 2024-04-09
fix: removed broken assertion

RobBuchananCompPhys:
`develop` branch was failing due to an undeclared variable `atomTypeI_` that had somehow got through unit tests

### DIFF
--- a/src/classes/pairPotential.cpp
+++ b/src/classes/pairPotential.cpp
@@ -171,7 +171,7 @@ bool PairPotential::setUp(const std::shared_ptr<AtomType> &typeI, const std::sha
     return true;
 }
 
-//  Set Data1D names from source AtomTypes
+// Set Data1D names from source AtomTypes
 void PairPotential::setData1DNames()
 {
     uFull_.setTag(fmt::format("{}-{}", nameI_, nameJ_));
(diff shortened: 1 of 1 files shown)


trisyoungs (approved):
Yep, I caught this in my yet-to-be-reviewed-or-merged PR #1837, but yours can certainly go in ahead of mine. It shouldn't have broken the CI, however, since the `assert` only gets compiled in if we're in DEBUG mode.

src/classes/pairPotential.cpp
```diff
@@ -174,9 +174,6 @@ bool PairPotential::setUp(const std::shared_ptr<AtomType> &typeI, const std::sha
 // Set Data1D names from source AtomTypes
```
rprospero, line 174:
```suggestion
//  Set Data1D names from source AtomTypes
```

src/classes/pairPotential.cpp
```diff
@@ -171,12 +171,9 @@ bool PairPotential::setUp(const std::shared_ptr<AtomType> &typeI, const std::sha
     return true;
 }
 
-// Set Data1D names from source AtomTypes
+//  Set Data1D names from source AtomTypes
```
rprospero:
```suggestion
// Set Data1D names from source AtomTypes
```

PR #1839, opened 2024-04-10
fix: Allow non-pr runs to pass ci

rprospero:
This will enable GitHub runners that do not have a PR title (e.g. the continuous runn) to automatically pass the conventional commit checks.

### DIFF
--- a/.github/workflows/qc/action.yml
+++ b/.github/workflows/qc/action.yml
@@ -19,7 +19,9 @@ runs:
       set -ex
       TITLE="${{ github.event.pull_request.title }}"
       if [ -z "$TITLE" ]
+      then
       TITLE=Continuous
+      fi
       fi
       echo  | awk "/(Continuous)|(^feat:)|(^fix:)|(^doc:)|(^perf:)|(^refactor:)|(^style:)|(^test:)|(^chore:)|(^revert:)|(^build:)/" | grep .
 
--- a/.github/workflows/qc/action.yml
+++ b/.github/workflows/qc/action.yml
@@ -22,7 +22,6 @@ runs:
       then
       TITLE=Continuous
       fi
-      fi
       echo  | awk "/(Continuous)|(^feat:)|(^fix:)|(^doc:)|(^perf:)|(^refactor:)|(^style:)|(^test:)|(^chore:)|(^revert:)|(^build:)/" | grep .
 
   - name: C++ Formatting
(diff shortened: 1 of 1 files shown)


.github/workflows/qc/action.yml
```diff
@@ -17,7 +17,12 @@ runs:
     shell: bash
     run: |
       set -ex
-      echo ${{ github.event.pull_request.title }} | awk "/(^feat:)|(^fix:)|(^doc:)|(^perf:)|(^refactor:)|(^style:)|(^test:)|(^chore:)|(^revert:)|(^build:)/" | grep .
+      TITLE="${{ github.event.pull_request.title }}"
+      if [ -z "$TITLE" ]
+      then
+      TITLE=Continuous
+      fi
+      echo $TITLE | awk "/(Continuous)|(^feat:)|(^fix:)|(^doc:)|(^perf:)|(^refactor:)|(^style:)|(^test:)|(^chore:)|(^revert:)|(^build:)/" | grep .
```
trisyoungs: Just an alternative suggestion, but you could avoid the if/else title set and check by adding `if: github.event_name == 'pull_request'` on to the step perhaps?

rprospero: @trisyoungs That had been my original idea, but I was paranoid that I wouldn't be able to test it.  However, looking at it again, you're right that the simplicity makes it worth the change.

.github/workflows/qc/action.yml
```diff
@@ -14,10 +14,11 @@ runs:
 
   - name: Check Conventional Commits
     # Ensure that the PR title fits the conventional commits framework
+    if: github.event_name == 'pull_request'
     shell: bash
     run: |
       set -ex
-      echo ${{ github.event.pull_request.title }} | awk "/(^feat:)|(^fix:)|(^doc:)|(^perf:)|(^refactor:)|(^style:)|(^test:)|(^chore:)|(^revert:)|(^build:)/" | grep .
+      echo ${{ github.event.pull_request.title }} | awk "/(Continuous)|(^feat:)|(^fix:)|(^doc:)|(^perf:)|(^refactor:)|(^style:)|(^test:)|(^chore:)|(^revert:)|(^build:)/" | grep .
```
trisyoungs: Guess this change is redundant now?

trisyoungs (approved):

.github/workflows/qc/action.yml
trisyoungs: @rprospero Sometimes its hard to have faith in the mess that is GitHub `if` checks, but live dangerously I say!

.github/workflows/qc/action.yml
rprospero: Indeed it is

PR #1844, opened 2024-04-11
fix: Typo in NeutronSQ NormaliseTo Keyword

rhinoella:
Keyword is set to `normaliseTo` instead of `NormaliseTo` which prevents Dissolve-generated input files from being correctly deserialised.


trisyoungs (approved):
:+1:

PR #1848, opened 2024-04-11
fix: 1844 - QML ScaleLayout scale charges dialog to accept floats

### LINKED ISSUE #1847: Scale charges dialog only accepts integers
**Describe the bug**
Species -> Scale Charges only accepts integers, but should allow for floats as well.

rhinoella:
Closes #1847

Creates a custom QML SpinBox that allows for floating-point values with a precision of `0.01`.

### DIFF
--- a/src/gui/qml/modifyCharges/ScaleLayout.qml
+++ b/src/gui/qml/modifyCharges/ScaleLayout.qml
@@ -19,25 +19,24 @@ ColumnLayout {
         width: parent.width - 2 * parent.spacing
         wrapMode: Text.WordWrap
     }
-
     SpinBox {
         id: spinBox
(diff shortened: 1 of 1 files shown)


trisyoungs (commented):
@rhinoella I appreciate the effort that you've gone to in this solution, and I completely get the logic behind it. But _suuuuurely_ QML has a double spin box that we can use?  Having to engineer this as a component for ourselves is absolutely crazy!

rhinoella:
@trisyoungs Unfortunately not! This is the accepted solution. 🤷‍♀️ Qt continues to bring more delights

trisyoungs (approved):
:+1:

PR #1851, opened 2024-04-12
fix: Remove CMakeLists from packaged examples

rhinoella:
Remove unecessary CMake files from Dissolve example data.

### DIFF
--- a/examples/package-examples
+++ b/examples/package-examples
@@ -51,7 +51,10 @@ FILES=( $(find ./ -mindepth 2 -type f -not -regex '.*/_.*') )
 mkdir ${ARCHIVEDIR}
 for file in "${FILES[@]}"
 do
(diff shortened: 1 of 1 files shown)


trisyoungs (changes requested):
Unfortunately it's not that simple!  The examples get registered as tests, hence the `CMakeLists.txt`.  What you'll have to do instead is modify the `examples/package-examples` bash script to exclude the `CMakeLists.txt` from the generated zip...

trisyoungs (changes requested):
I'd suggest that neater way would be to modify the `find` command:
```
FILES=( $(find ./ -mindepth 2 -type f -not -regex '.*/_.*' -not -name CMakeLists.txt) )
```

trisyoungs (approved):

PR #1852, opened 2024-04-12
refactor: Clarify forces and derivatives

trisyoungs:
Depends on #1837.

This PR clarifies forces vs derivatives in interatomic (through-space) force calculations. While the forces calculated throughout were correct at the endpoint, the signs of various terms were mixed throughout.

### Rationale

If the energy of a given interaction between two unconnected atoms at a given distance `r` is `E = U(r)` (for instance via the Lennard-Jones potential) then the force acting on those atoms is equal to the negative of the derivative of the energy, i.e. `F = -dU/dr`. For a pair of atoms `i` and `j` at a distance  `r` the vector between them is `vij = r(j) - r(i)`, and the forces acting on them are given by `F(i) = -F.vij` and `F(j) = F.vij`.

So, as you can see there are plenty of opportunities for absorbing (mistakenly or otherwise) negatives into equations! This PR tidies up the following things and gives us consistency ahead of moving the functions around in the next PR.
1. The `PairPotential` class functions `analyticShortRangeForce()` and `analyticCoulombForce()` actually just returned the derivatives of the functions (i.e. the negation was missing). These functions are used to seed the tabulated potentials.
2. Because of 1) the tabulated values in `dUFull_` were, as the variable name suggested, the derivatives and not the force, but the `PairPotential::force()` function returns the actual values without a negation, and so was returning the derivatives and not the force.
3. All of the `ForceKernel::forcesWith(out)Mim` functions correctly calculated the vectors the forces were acting on as `i -> j` as written above, but added to atom `i` and subtracted from atom `j`, providing the final correcting "reversal" of the signs.

### Outcome

Once all of these aspects were fixed and made consistent all tests passed save for one - `MDModule.BenzeneRestart` - where the forces calculated were slightly different from the reference values in the test data. This turned out to be an age-old bug where the corrective force term in the shifted truncation scheme was correctly used for the energy (upon which tabulated simulation forces are created) but not the analytic forces. This did not affect simulation results.

### DIFF
--- a/src/classes/pairPotential.cpp
+++ b/src/classes/pairPotential.cpp
@@ -411,7 +411,7 @@ double PairPotential::force(double r)
 {
     assert(r >= 0);
 
-    return dUFullInterpolation_.y(r, r * rDelta_);
+    return -dUFullInterpolation_.y(r, r * rDelta_);
 }
 
 // Return analytic force at specified r
--- a/src/classes/pairPotential.cpp
+++ b/src/classes/pairPotential.cpp
@@ -212,11 +212,17 @@ double PairPotential::analyticShortRangeForce(double r, PairPotential::ShortRang
              * Parameter 1 = Sigma
              */
 
-            // f = -48*epsilon*((sigma**12/x**13)-0.5*(sigma**6/x**7))
+            /*
+             * dU/dr = -48*epsilon*((sigma**12/x**13)-0.5*(sigma**6/x**7))
+             *
+             *                        signa**6   (   sigma**6        )
+             *       = 48 * epsilon * -------- * ( - --------- + 0.5 ) * r**-1
+             *                          r**6     (     r**6          )
+             */
 
             auto sigmar = params[1] / r;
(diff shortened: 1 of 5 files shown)


src/classes/pairPotential.cpp
```diff
@@ -375,11 +375,8 @@ double PairPotential::analyticEnergy(double r) const
...
 
-    // Short-range potential
-    auto energy = analyticShortRangeEnergy(r);
-
-    // Coulomb contribution
-    energy += analyticCoulombEnergy(chargeI_ * chargeJ_, r);
+    // Short-range potential and Coulomb contribution
+    auto energy = analyticShortRangeEnergy(r) + analyticCoulombEnergy(chargeI_ * chargeJ_, r);
 
     return energy;
```
RobBuchananCompPhys: We could even one-line the energy directly into the return statement, although we may want to avoid that so it can be easily evaluated during debugging.

rprospero (approved):

PR #1866, opened 2024-04-18
build: Windows QML dlls

trisyoungs:


.github/workflows/pr.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     strategy:
```
rhinoella, line 32: Revert this file

rhinoella (changes requested):

trisyoungs:
Needs more in-depth work.

PR #1869, opened 2024-04-18
build: Probe qml

trisyoungs:


.github/workflows/package/osx/action.yml
```diff
@@ -62,16 +62,17 @@ runs:
       appName=Dissolve-GUI-${{ runner.arch }}
       appDirName=${appName}.app
       packageName=Dissolve-GUI-${{ runner.arch }}-${{ env.dissolveVersion }}
+      dmgName=Dissolve-GUI-${{ env.dissolveVersion }}-${{ runner.arch }}
```
rhinoella, line 65: What is the purpose of dmgName with the arch and version flipped?

.github/workflows/pr.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ macos-latest, macos-14, windows-latest ]
+        os: [ macos-latest, macos-14 ]
```
rhinoella: Revert this file

.github/workflows/package/osx/action.yml
trisyoungs: Soooooo...... Other packages we generate for other platforms all follow the format `Dissolve-GUI-X.Y.Z[-arch].abc` but the OSX ones didn't.  The Hugo shortcode assumes the first format and so was generating links to the dmg files that were not correct.  Rather than being specific for OSX on the Hugo end, I decided a quick mod to the CI was a better approach.  No doubt the scripts and other stuff (e.g. pre-dmg) could be modified too, but that was a bigger job and I was just seeking a quick solution for the workshop.

.github/workflows/pr.yml
trisyoungs: Well of course!

PR #1870, opened 2024-04-18
build: Fix Qt library install for Windows

rprospero:
This uses some Qt magic in cmake to create the deployment script for us.

### DIFF
--- a/.github/workflows/package/windows/action.yml
+++ b/.github/workflows/package/windows/action.yml
@@ -55,7 +55,7 @@ runs:
       export PATH="${Qt6_DIR}\bin;$PATH"
       export FREETYPE_DIR="${RUNNER_TEMP}\freetype-install\bin"
       export FTGL_DIR="${RUNNER_TEMP}\ftgl-install\bin"
-      export DISSOLVE_DIR="${GITHUB_WORKSPACE}\build\bin"
+      export DISSOLVE_DIR="${GITHUB_WORKSPACE}\build\install\bin"
       export DEPLOY_DIR="${GITHUB_WORKSPACE}\build"
 
       # Run Inno Setup Compiler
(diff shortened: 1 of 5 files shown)


rprospero:
I can confirm that this correctly loaded the dialogs under Windows on my laptop.  The one concerning oddity is that it did *not* work when launched from the installer, but running it subsequently from the Start Menu was effective.

My three hypotheses are:
1) It may need to process the QML on its first load, but then works fine afterwards.
2) The installer invokes the program in an odd way that doesn't get access to the QML folders.
3) My Windows laptop is just being weird again.

trisyoungs (commented):
So I cannot try the installer because, you know, Sophos, but on my local Windows machine the zip file is missing the `qml`, `platforms`, `iconengines`, and `imageformats` directories. Looking at the CI the relevant files are there when the installer gets unpacked, but the subsequent directory move and zip misses the subdirs...

rprospero:
@trisyoungs It now requires the full installer to get things in place, so you'll want to use the [installer package](https://github.com/disorderedmaterials/dissolve/actions/runs/8781796049/artifacts/1435043275) and not just the archive.

.github/workflows/package/windows/action.yml
trisyoungs, line 1: @rprospero Just to clarify, the zipped contents derived from the Inno installer should work just the same, but not require installation (and its the only way I can deploy on my own, and other PCs around ISIS! The problem is that the new install methodology puts the deployed libs in a slightly different structure than before. Previously the zip file assumed that the important bits were all in `app/bin` but this is no longer the case. On lines 69/70 you can see:
```
      mv ${exeBase}/app/bin/* $exeBase
      mv ${exeBase}/app ./
```
I think this can just be replaced with:
```
      mv ${exeBase}/app/* $exeBase
      rm -r ${exeBase}/app
```
That should yield a "full" zip.

rprospero: @trisyoungs Good call on the suggestions.  It now runs straight from the zip archive on my machine.  Also cleared up whatever I had going on with the first run through the installer.

trisyoungs (approved):
Zipped distro also now complete and working as expected.
:+1:

PR #1872, opened 2024-04-18
feat: Addition of dataType selection into compare

### LINKED ISSUE #1867: Allow compare module to select target configuration and data
Currently to edit what configuration the compare module runs on you need to edit the input file. We should add this into the GUI. Also, if we could select what data it runs on e.g. Total G(r) instead of S(Q).

Danbr4d:
closes #1867


trisyoungs:
Closing as not implementing.

PR #1883, opened 2024-04-24
fix: enable atom type and master term remove functionality

### LINKED ISSUE #1065: Remove Master Terms from GUI
It is currently not possible to remove defined master terms from the GUI, as the necessary functionality in the derived `MasterTermTreeModel`s does not exist. When implemented, we also must call the relevant detach function in all current `Species`.

### LINKED ISSUE #1586: Cannot remove Atom Types
The Duplicate/Add buttons work as expected, but the remove button is disabled.
image

### LINKED ISSUE #1874: Cannot delete Atom Type and Master Terms
**Describe the bug**
Cannot remove atom type or master terms (can add and duplicate)

**To Reproduce**
1. Import species from LigParGen without "simplify atom types"
2. Try to manually delete duplicating Atom Types

**Expected behaviour**
To be able to remove redundant atom types (and master terms) if the selected atom type is not used in any existing species.

**Operating System:**
 - OS:  Linux
 - OS version / flavour:  IDAaaS

RobBuchananCompPhys:
Closes #1065, Closes #1874, Closes #1586


trisyoungs (commented):
Good start. However there are a couple of "propagation" mechanics to consider.

1) When removing an AtomType we are currently being saved (or perhaps being obscured from problems...) by our use of `std::shared_ptr` both in the main vector of AtomTypes as well as any classes which need it. A bigger discussion which arises from this is "Should we really be using something like `std::weak_ptr` to store references to an `AtomType`? For now, those references must be nullified by extending the `objectNoLongerValid` and `removeReferenceTo` template functions (see `CoreData`).
2) Removing a master term suffers from a similar problem to the AtomType, but when removed we should simply convert any uses into Species-local definitions (i.e. copy the data into a new `SpeciesBond` etc.)

Seems that there's a fair bit more work here than anticipated!

src/classes/coreData.cpp
```diff
@@ -53,9 +53,11 @@ std::shared_ptr<AtomType> CoreData::addAtomType(Elements::Element Z)
 }
 
 // Remove specified AtomType
-void CoreData::removeAtomType(const std::shared_ptr<AtomType> &at)
+void CoreData::removeAtomType(std::shared_ptr<AtomType> &at)
 {
-    atomTypes_.erase(std::remove(atomTypes_.begin(), atomTypes_.end(), at));
+    removeReferencesTo(at);
```
trisyoungs: The `removeReferencesTo(std::shared_ptr<AtomType>)` function in `CoreData` needs to also handle nullification of any atom types pointers on `Species`, something like this:
```
    // Check Species - if an atom references the supplied atom type we must nullify it
    for (auto &sp : species_)
        for (auto &i : sp->atoms())
            if (i.atomType() == data)
                i.setAtomType(nullptr);
```

src/classes/coreData.cpp
```diff
@@ -156,6 +158,18 @@ MasterBond &CoreData::addMasterBond(std::string_view name)
...
 }
 
+// Remove specified master Bond
+void CoreData::removeMasterBond(const std::shared_ptr<MasterBond> &bond)
+{
+    // Copy data into species-local instance
+    for (auto &species : species_)
+    {
+        species->addBond(bond->indexI(), bond->indexJ());
+    }
```
trisyoungs: Your general intention here is correct, but the solution is slightly more detailed. As written here this function just adds on a new _connection_ between two atom indices that I think are both zero.  Bad times!  What you want to do is check every bond on every `Species` to see if the parameters it references correspond to the `MasterBond` we're deleting, and then "detach" from them:
```
    // Check usage in Species
    for (auto &species : species_)
    {
        for (auto &b : species->bonds())
            if (b.masterTerm() == bond.get())
                b.detachFromMasterTerm();
    }
```
Same applies to the other intramolecular types.

src/gui/forcefieldTab.cpp
```diff
@@ -283,10 +292,30 @@ void ForcefieldTab::on_AtomTypeAddButton_clicked(bool checked)
     dissolveWindow_->setModified();
 }
 
-void ForcefieldTab::on_AtomTypeRemoveButton_clicked(bool checked) { Messenger::error("NOT IMPLEMENTED YET.\n"); }
+void ForcefieldTab::on_AtomTypeRemoveButton_clicked(bool checked)
+{
```
trisyoungs: I feel like we should pop up a qmessagebox here to ask the mandatory "Are you sure? THis can't be undone!" question for safety.

PR #1888, opened 2024-05-01
fix: Update runners to GitHub runner change

rhinoella:
GitHub recently updated their runner images so that `macos-latest` is changed from the intel MacOS 12 to the arm64 MacOS 14, which breaks the current workflows.

This PR updates the workflows to accomodate this.


trisyoungs (approved):
:+1:

PR #1889, opened 2024-05-02
fix: Remove atom type and master term from forcefield tab

### LINKED ISSUE #1065: Remove Master Terms from GUI

### LINKED ISSUE #1586: Cannot remove Atom Types

### LINKED ISSUE #1874: Cannot delete Atom Type and Master Terms

RobBuchananCompPhys:
Closes #1065, Closes #1874, Closes #1586

### DIFF
--- a/src/gui/forcefieldTab.cpp
+++ b/src/gui/forcefieldTab.cpp
@@ -73,8 +73,9 @@ ForcefieldTab::ForcefieldTab(DissolveWindow *dissolveWindow, Dissolve &dissolve,
             this, SLOT(masterAnglesSelectionChanged(const QItemSelection &, const QItemSelection &)));
     connect(ui_.MasterTorsionsTable->selectionModel(), SIGNAL(selectionChanged(const QItemSelection &, const QItemSelection &)),
             this, SLOT(masterTorsionsSelectionChanged(const QItemSelection &, const QItemSelection &)));
-    connect(ui_.MasterImpropersTable->selectionModel(), SIGNAL(selectionChanged(const QItemSelection &, const QItemSelection &)),
-            this, SLOT(masterImpropersSelectionChanged(const QItemSelection &, const QItemSelection &)));
+    connect(ui_.MasterImpropersTable->selectionModel(),
+            SIGNAL(selectionChanged(const QItemSelection &, const QItemSelection &)), this,
+            SLOT(masterImpropersSelectionChanged(const QItemSelection &, const QItemSelection &)));
 
     /*
      * Atom Types
@@ -291,8 +292,8 @@ void ForcefieldTab::on_AtomTypeAddButton_clicked(bool checked)
     dissolveWindow_->setModified();
 }
 
-void ForcefieldTab::on_AtomTypeRemoveButton_clicked(bool checked) 
-{ 
+void ForcefieldTab::on_AtomTypeRemoveButton_clicked(bool checked)
+{
     auto index = ui_.AtomTypesTable->currentIndex();
     if (!index.isValid())
         return;
@@ -527,7 +528,7 @@ void ForcefieldTab::on_MasterTermAddBondButton_clicked(bool checked)
     dissolveWindow_->setModified();
 }
 
-void ForcefieldTab::on_MasterTermRemoveBondButton_clicked(bool checked) 
+void ForcefieldTab::on_MasterTermRemoveBondButton_clicked(bool checked)
 {
     auto index = ui_.MasterBondsTable->currentIndex();
     if (!index.isValid())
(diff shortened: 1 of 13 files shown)


src/gui/forcefieldTab.cpp
```diff
@@ -470,6 +510,26 @@ void ForcefieldTab::masterImpropersDataChanged(const QModelIndex &, const QModel
     dissolveWindow_->setModified();
 }
 
+void ForcefieldTab::masterBondsSelectionChanged(const QItemSelection &current, const QItemSelection &previous)
+{
+    ui_.MasterTermRemoveBondButton->setEnabled(!current.empty());
```
trisyoungs, line 515: I use code exactly like this, but there is a small bug somewhere whereby if you delete the last bond or angle the button remains enabled, despite there being no current selection.  Not a merge-killer, but may be worth taking a few minutes to try and figure out.

src/gui/forcefieldTab.cpp
```diff
@@ -479,7 +539,36 @@ void ForcefieldTab::on_MasterTermAddBondButton_clicked(bool checked)
     dissolveWindow_->setModified();
 }
 
-void ForcefieldTab::on_MasterTermRemoveBondButton_clicked(bool checked) { Messenger::error("NOT IMPLEMENTED YET.\n"); }
+void ForcefieldTab::on_MasterTermRemoveBondButton_clicked(bool checked)
+{
+    QMessageBox queryBox;
+    queryBox.setWindowTitle(QString("Remove Master Bond"));
+    queryBox.setText(QString("This operation cannot be undone!"));
```
trisyoungs: Worth elaborating here a bit and elsewhere (and reformatting because this will no doubt break the CI!):
```suggestion
    queryBox.setText(QString("Are you sure? This operation cannot be undone!\nANy uses of this term in species will be converted to a local definition."));
```

trisyoungs (commented):
Very close.  Just a bit of tweaking required!

src/gui/forcefieldTab.cpp
```diff
@@ -479,7 +543,40 @@ void ForcefieldTab::on_MasterTermAddBondButton_clicked(bool checked)
...
+
+    if (queryBox.exec() != QMessageBox::Yes)
+        return;
+
+    if (dissolve_.coreData().nMasterBonds() == 1)
+        ui_.MasterTermRemoveBondButton->setEnabled(false);
+
+    auto index = ui_.MasterBondsTable->currentIndex();
+    if (!index.isValid())
+        return;
```
trisyoungs: Need to do the check for a valid index first so we don't needlessly create & present user choice. Also, not sure what the `if (dissolve_.coreData().nMasterBonds() == 1)` was achieving, so removed it! This change can propagate to the other term types.
```suggestion
    auto index = ui_.MasterBondsTable->currentIndex();
    if (!index.isValid())
        return;

    QMessageBox queryBox;
    queryBox.setWindowTitle(QString("Remove Master Bond"));
    queryBox.setText(QString("Are you sure? This operation cannot be undone!\n") +
                     QString("Any uses of this term in species will be converted to a local definition."));
    queryBox.setInformativeText("Proceed?");
    queryBox.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
    queryBox.setDefaultButton(QMessageBox::No);

    if (queryBox.exec() != QMessageBox::Yes)
        return;
```

trisyoungs (approved):
Looks good.  Just a bit of cleanup / ordering of operations as detailed in the comment.

src/gui/forcefieldTab.cpp
RobBuchananCompPhys: `if (dissolve_.coreData().nMasterBonds() == 1)` was meant to disable the remove button if we are removing the last bond/angle/... etc.
Did this not solve an issue we had previously discussed where the remove button remained enabled after removing all items?

trisyoungs: Oh, I see - sorry.  Not really with it in the logic sense today!

Let's go with this solution for now, but hopefully will be able to revert it in the follow-up PR.

PR #1891, opened 2024-05-07
refactor: qml gui

RobBuchananCompPhys:
Beginning of a refactor to create a fully QML-based GUI that can be built using a feature flag in CMake. Aims to get a QML Dissolve main window as a starting point to experiment with redesign.

We may eventually need to break out the Qt Models directory (currently in `gui/models`) and the QML types register into a destination where it can be accessed by both `GUI` and `QML` builds.


rprospero:
One thing that comes to mind is that it might be better to make the
QML version truly its own target and executable, instead of just being
a switch that gets pulled inside `dissolve-gui.cpp`.

## Advantages

- No extra preprocessor macro logic in the executable
- No CMake flag required (just a new target).
- Easier to make independent changes to each program. This will become important when we start including models to go with the GUI.
- The CI can compile both executables in one pass and ensure that nothing has broken in either GUI.

## Disadvantages

- Still some duplicate code (e.g. CLI argument handling) that would need to be maintained in both executables
- I have no idea what to name the new GUI

trisyoungs:
I definitely think the separate executable, as Adam suggested, is the way to go, and have it configurable at build time with a CMake option perhaps?

Bit of duplicate code I can live with for now, but as for the name.... maybe `dissolve-gui-qml`?

PR #1893, opened 2024-05-10
refactor: Master term models

trisyoungs:
This PR moves the responsibility of adding / removing master terms to the relevant model, rather than leaving this the job of the GUI at the calling point.

A significant change here is the storage of a reference to `CoreData` within the base `MasterTermModel`.  Comments welcome on this, but the allows easy access to the addition / removal of terms, as well as permitting a nice clean up of code throughout all the models.

### DIFF
--- a/src/gui/forcefieldTab.cpp
+++ b/src/gui/forcefieldTab.cpp
@@ -19,7 +19,9 @@
 
 ForcefieldTab::ForcefieldTab(DissolveWindow *dissolveWindow, Dissolve &dissolve, MainTabsWidget *parent, const QString title)
     : MainTab(dissolveWindow, dissolve, parent, title, this), atomTypesModel_(dissolve.coreData()),
-      pairPotentialModel_(dissolve.pairPotentials()), pairPotentialOverrideModel_(dissolve.coreData().pairPotentialOverrides())
+      pairPotentialModel_(dissolve.pairPotentials()), pairPotentialOverrideModel_(dissolve.coreData().pairPotentialOverrides()),
+      masterBondsTableModel_(dissolve.coreData()), masterAnglesTableModel_(dissolve.coreData()),
+      masterTorsionsTableModel_(dissolve.coreData()), masterImpropersTableModel_(dissolve.coreData())
 {
     ui_.setupUi(this);
 
@@ -41,13 +43,9 @@ ForcefieldTab::ForcefieldTab(DissolveWindow *dissolveWindow, Dissolve &dissolve,
         1, new ComboListDelegate(this, new ComboEnumOptionsItems<TorsionFunctions::Form>(TorsionFunctions::forms())));
 
     // Set models for tables
-    masterBondsTableModel_.setSourceData(dissolve_.coreData().masterBonds());
     ui_.MasterBondsTable->setModel(&masterBondsTableModel_);
-    masterAnglesTableModel_.setSourceData(dissolve_.coreData().masterAngles());
     ui_.MasterAnglesTable->setModel(&masterAnglesTableModel_);
-    masterTorsionsTableModel_.setSourceData(dissolve_.coreData().masterTorsions());
     ui_.MasterTorsionsTable->setModel(&masterTorsionsTableModel_);
-    masterImpropersTableModel_.setSourceData(dissolve_.coreData().masterImpropers());
     ui_.MasterImpropersTable->setModel(&masterImpropersTableModel_);
 
     // Ensure fonts for table headers are set correctly and the headers themselves are visible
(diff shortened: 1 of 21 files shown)


rprospero:
Using the `CoreData` as the parameter on the constructor certainly simplifies things.  The one bit that makes me nervous is that requiring a parameter means that the type cannot be created in QML.  Creating the type in the QML itself can be useful (e.g. what we do with `AddForcefieldDialogModel` in `AddForcefieldTermsDialog.qml`).  However, we're not doing that anywhere with the master terms and it doesn't seem likely that we will, so we should be okay.

I'm bringing this up because I've been thinking about the structure of the models for the pure QML GUI.  I'm imagining that we'll ultimately need a bunch of higher level models (e.g. DissolveModel, CoreDataModel) and that these might be able to return inner models (e.g. a CoreDataModel having a function that returns its MasterAngleModel).  That way the models can be passed around as parameters within the QML and can be passed into any other bespoke models that may need them.

src/gui/forcefieldTab.cpp
```diff
@@ -283,10 +291,40 @@ void ForcefieldTab::on_AtomTypeAddButton_clicked(bool checked)
...
+        return;
+
+    auto index = ui_.AtomTypesTable->currentIndex();
+    if (!index.isValid())
+        return;
+
+    // Get selected atomtype
+    auto at = atomTypesModel_.rawData(index);
+    if (!at)
+        return;
```
rprospero: Should we give an error dialog if there was no atom type?

src/gui/forcefieldTab.cpp
```diff
@@ -470,46 +508,176 @@ void ForcefieldTab::masterImpropersDataChanged(const QModelIndex &, const QModel
...
+
+    auto index = ui_.MasterBondsTable->currentIndex();
+    if (!index.isValid())
+        return;
+
+    // Get selected master bond
+    auto bond = masterBondsTableModel_.rawData(index);
+    if (!bond)
+        return;
```
rprospero: Again with error dialog, or at least logging a message to the console

src/gui/forcefieldTab.cpp
```diff
@@ -470,46 +508,176 @@ void ForcefieldTab::masterImpropersDataChanged(const QModelIndex &, const QModel
...
+
+    auto index = ui_.MasterAnglesTable->currentIndex();
+    if (!index.isValid())
+        return;
+
+    // Get selected master angle
+    auto angle = masterAnglesTableModel_.rawData(index);
+    if (!angle)
+        return;
```
rprospero: same

src/gui/forcefieldTab.cpp
```diff
@@ -470,46 +508,176 @@ void ForcefieldTab::masterImpropersDataChanged(const QModelIndex &, const QModel
...
+    queryBox.setInformativeText("Proceed?");
+    queryBox.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
+    queryBox.setDefaultButton(QMessageBox::No);
+
+    if (queryBox.exec() != QMessageBox::Yes)
+        return;
+
+    auto index = ui_.MasterTorsionsTable->currentIndex();
+    if (!index.isValid())
+        return;
```
rprospero, line 626: Same

src/gui/forcefieldTab.cpp
```diff
@@ -470,46 +508,176 @@ void ForcefieldTab::masterImpropersDataChanged(const QModelIndex &, const QModel
...
+    queryBox.setInformativeText("Proceed?");
+    queryBox.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
+    queryBox.setDefaultButton(QMessageBox::No);
+
+    if (queryBox.exec() != QMessageBox::Yes)
+        return;
+
+    auto index = ui_.MasterImpropersTable->currentIndex();
+    if (!index.isValid())
+        return;
```
rprospero, line 664: Same

rprospero (approved):
There's a bunch of places where we might want to add some error logging and/or dialogs, but it looks good otherwise.

trisyoungs:
> Using the `CoreData` as the parameter on the constructor certainly simplifies things. The one bit that makes me nervous is that requiring a parameter means that the type cannot be created in QML. Creating the type in the QML itself can be useful (e.g. what we do with `AddForcefieldDialogModel` in `AddForcefieldTermsDialog.qml`). However, we're not doing that anywhere with the master terms and it doesn't seem likely that we will, so we should be okay.
> 
> I'm bringing this up because I've been thinking about the structure of the models for the pure QML GUI. I'm imagining that we'll ultimately need a bunch of higher level models (e.g. DissolveModel, CoreDataModel) and that these might be able to return inner models (e.g. a CoreDataModel having a function that returns its MasterAngleModel). That way the models can be passed around as parameters within the QML and can be passed into any other bespoke models that may need them.

Yes, this is a very valid concern, and not one which I had thought about when I made that change!  However, your suggestion of a top level model encompassing either or both of `Dissolve` and `CoreData` matches closely with what @jws-1 had started tinkering around with for "Workflow 2.0", so I think we are all of the same mind going forward.

src/gui/forcefieldTab.cpp
trisyoungs: @rprospero If the raw data comes back `nullptr` then we have somehow entered a broken state with null objects in our main arrays, so it feels like `throw` would be the most appropriate action? I know we're not strictly supposed to `throw` in a Qt slot...

rprospero: Very good point.  Considering that it is something that

1. Should never happen
2. Requires the program to stop

Maybe we should just make it an assertion?

```cpp
assert(at);
```

rprospero (approved):

PR #1894, opened 2024-05-10
refactor: Fully QML GUI as separate target executable

RobBuchananCompPhys:
This PR takes on board the comments and discussion over the previous attempt to start a QML GUI for Dissolve.
The idea is now to go with a separate executable, with entrypoint in `dissolve-gui-qml`. 
A couple of lines have been commented out in `main` where the `dissolveWindow` object was used while I figure out how to proceed with the CLI options parsing using a pure QML window, rather than Qt Widgets (basically, trying to make it work for now).
Additionally, the CMake files almost certainly link far more dependencies for the new QML-based executable than is necessary, so over time we should aim to identify what is not needed and remove from the list.


src/dissolve-gui-qml.cpp
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#include "base/messenger.h"
+#include "base/processPool.h"
+#include "gui/gui.h"
```
rprospero:
```suggestion
#include "gui/models/dissolveModel.h"
```

We don't want to include the entire GUI, but only the main model.  I mention this mostly because it enables us to skip linking against some dependencies that we would like to escape.

CMakeLists.txt
```diff
@@ -254,6 +257,24 @@ if(GUI)
...
+    PRIVATE ${PROJECT_SOURCE_DIR}/src
+            ${PROJECT_BINARY_DIR}/src
+            ${CMAKE_BINARY_DIR}/src/gui/gui_autogen/include
+            ${FREETYPE_INCLUDE_DIRS}
+            ${Qt6Core_INCLUDE_DIRS}
+            ${Qt6Gui_INCLUDE_DIRS}
+            ${Qt6OpenGL_INCLUDE_DIRS}
+            ${Qt6Widgets_INCLUDE_DIRS}
+            ${Qt6OpenGL_INCLUDE_DIRS}
+            ${Qt6OpenGLWidgets_INCLUDE_DIRS}
```
rprospero:
```suggestion
            ${Qt6Core_INCLUDE_DIRS}
            ${Qt6Gui_INCLUDE_DIRS}
```

The new GUI hopefully shouldn't need most of these libraries, and currently works without them.

CMakeLists.txt
```diff
@@ -354,6 +375,34 @@ if(GUI)
...
+            # Module gui libs
+            ${MODULEGUI_LINK_LIBS}
+            moduleWidgetProducer
+            # Main gui libs
+            render
+            widgets
+            keywordWidgets
+            delegates
+            models
+            gui
```
rprospero:
```suggestion
            # Main gui libs
            gui-qml
```

Only need to link against out new gui-qml library.  It includes the inner libraries that we need (i.e. `models`).

CMakeLists.txt
```diff
@@ -354,6 +375,34 @@ if(GUI)
...
+    PRIVATE # External libs
+            Qt6::Widgets
+            Qt6::Core
+            Qt6::Qml
+            OpenGL::GL
+            ${GUI_LINK_LIBS}
+            ${FREETYPE_LIBRARIES}
+            ${FTGL_LIBRARIES}
+            ${CORE_LINK_LIBS}
+            ${THREADING_LINK_LIBS}
```
rprospero:
```suggestion
            Qt6::Qml
            ${CORE_LINK_LIBS}
            ${THREADING_LINK_LIBS}
```

Cutting down dependencies again

src/gui-qml/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+# Target 'gui'
+add_library(
+  gui-qml
+  # Qt Resources
+  resources.qrc
```
rprospero:
```suggestion
  # Types for QML access
  ../gui/types.cpp
```

Two bits here.  First, if we link against the resources here, then we get a duplicate linkage error when we link against it again when we define the main executable in the primary CMake file.  Since we *must* link against it in the primary CMake file, we shouldn't link against it here.

The second issue that we run into is that the only file, besides the models, that we need from the main gui is the `types.cpp` file that enables accessing our custom types from QML.  I've done an ugly, direct linkage here, but I'm wondering if `types.cpp` should just be in the `models` folder and linked into that library.

src/gui-qml/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+)
+
+target_include_directories(
+  gui-qml
+  PRIVATE ${PROJECT_SOURCE_DIR}/src
+          ${PROJECT_BINARY_DIR}/src
+          ${Qt6Widgets_INCLUDE_DIRS}
+          ${FREETYPE_INCLUDE_DIRS}
+          Qt6::OpenGL
+          Qt6::OpenGLWidgets
```
rprospero:
```suggestion
```

Dropping more dependencies

src/gui-qml/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+  PRIVATE models
+          base
+          Qt6::OpenGL
+          Qt6::OpenGLWidgets
+          Qt6::Quick
+          Qt6::QuickControls2
+          Qt6::Core
+          Qt6::Gui
+          Qt6::Qml
+          Qt6::QmlModels
```
rprospero:
```suggestion
```

Dropping

rprospero (changes requested):
I haven't had a chance to take a good look at the actual C++ (I'll do that after lunch), but this rewrite is a great opportunity to drop a ton of link time dependencies that we won't really need.  I've gone through and pruned it down a bit, with some suggestions on how to do a little more.

src/dissolve-gui-qml.cpp
RobBuchananCompPhys: Makes sense!

src/dissolve-gui-qml.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    // Flag to track our success in loading the input file and restart file
+    auto loadSuccessful = true;
+
+    // Set restart file frequency
+    dissolve.setRestartFileFrequency(options.noRestartFile() ? 0 : options.restartFileFrequency());
+    
+    // If an input file was specified, load it here
+    if (options.inputFile())
+        // loadSuccessful = dissolveWindow.loadInputFile(options.inputFile().value());
```
rprospero: Since we've commented out the actual action, this if statement will act on the next statement.  Granted, that is also another conditional that tests the same condition, but it might be better to comment out or remove the condition.

src/dissolve-gui-qml.cpp
```diff
@@ -0,0 +1,101 @@
...
+    setlocale(LC_NUMERIC, "C");
+    QLocale::setDefault(QLocale::C);
+
+    // Print GPL license information
+    Messenger::print("Dissolve-GUI-QML {} version {}, Copyright (C) 2024 Team Dissolve and contributors.\n", Version::appType(),
+                     Version::info());
+    Messenger::print("Source repository: {}.\n", Version::repoUrl());
+    Messenger::print("Dissolve comes with ABSOLUTELY NO WARRANTY.\n");
+    Messenger::print("This is free software, and you are welcome to redistribute it under certain conditions.\n");
+    Messenger::print("For more details read the GPL at <http://www.gnu.org/copyleft/gpl.html>.\n");
```
rprospero, line 58: It might be useful to put this boilerplate for in a method on the `Dissolve` class.  You could pass in the program name and it could do this generic setup.  Granted, we don't have to do it on this PR, but it might be worth flagging the issue.

trisyoungs (commented):
Tested and working well on Linux.  :+1:

RobBuchananCompPhys:
> Tested and working well on Linux. 👍

Great, thanks for looking into it. @rprospero do we need to look at restoring some of those dependencies to fix the macos build?

rprospero:
Just a record of where I'm at on this.

1. For some reason, on the Mac build, and only the Mac build, the QML gui depends on our `render` target.
2. The `render` target depends on FTGL.
3. The `render` target also depends on the `StockColours` class, which is in the `gui` target.
4. Having the QML GUI depend on `gui` means brining in *all* of the other dependencies.

There's three possible routes forward from here

1. Try and remove the Apple dependency on the `render` target.
2. Split the classes required by `render` out of `gui` and into a separate target.
3. Abandon all hope and just link all the dependencies back into the QML gui.

I've listed the options in the order that I will attempt them.

trisyoungs:
@rprospero I don't think the dependency issue with `StockColours` can be easily solved, and actually just represents the tip of a rather large and annoying iceberg. Almost every module-related widget, and most of our `MainTab`-derived ones, use a `Viewer` of some kind which brings in the `render` dependency. So, I don't see a clean way of detaching the current GUI from these dependencies for plugging into the new QML one.  Perhaps the beset route is to not link to the current `gui` dependency _at all_ in the new QML frontend, and incrementally build it back up from scratch.  This provides an opportunity to overhaul the workflows that we currently have in place, and should encompass our move to "Workflow 2.0" at the same time, IMO. A lot of work to be sure, but this at least moves the focus of our efforts into a single reimagining of the GUI, rather than worrying about reimplementing the old one and then changing it anyway.

trisyoungs (approved):
Looks like an excellent starting point.

rprospero (approved):

src/dissolve-gui-qml.cpp
trisyoungs: We'll need it eventually probably, so I'd vote for commenting the condition out.

PR #1895, opened 2024-05-13
Qml theme

RobBuchananCompPhys:


rprospero:
There is one large accessibility concern with adding the custom styling.  Certain users will need to be able to pass in their own specific styles to use the software.  The good part about our setup is that they can just plug in the style that they use for other Qt applications, it also means that our *theme* cannot contain custom widgets.  That's not to say that we can't use custom widgets, but merely that they can't be namespaced into the theme.

For this specific moment, it means that we'll need to rework how Button, ButtonPanel, and SquareButton work together.  Button will need to be left in the theme file, but cannot depend on the other two widgets.  Meanwhile, ButtonPanel and SquareButton will need to be moved out of the theme directory and into the main directory (or some other widget subdirectory).

This also means that we'll need an adapter to get the colours from the theme.  I'm suspecting that we'll just be setting the `SystemPalette` in our theme and then accessing the `SystemPalette` in the widgets.

trisyoungs:
@rprospero Agree that accessibility should be a concern of ours.  I had envisaged at the outset two colour schemes (dark and light) accessible through a core style object (`ThemeColours` at present) which also has a couple of configurable parameters in terms of font size etc.  We can merge the `ButtonPanel` styling into `Button` or `SquareButton` at the expense of having that definition in a single source file, but to be honest I would prefer us to have a consistent, easy-to-maintain style / theme, and address specific accessibility concerns as style options within our theme, if you see what I mean. Thoughts?

PR #1899, opened 2024-05-16
feat: (#1890) Support extended trajectory export

### LINKED ISSUE #1890: Trajectory output Atom Types
Accumulated trajectory files can be very useful for post-dissolve analysis and it is a good way to make sure various analysis are performed over the same geometry/energy landscape. Currently, the output trajectory file only contains 4 columns: Element, X-coord, Y-coord, Z-coord.

I think it would be more useful to output: Atom Type, X-coord, Y-coord, Z-coord; so that sites can be differentiated among functional groups (i.e. hydroxyl hydrogen and alkyl hydrogen).

If possible, I think it would be even better if the output could include 6 columns: Element, Atom Type, Atom Index (within the species), X, Y, Z

As an example (2 water molecules per frame):

6 (total number of atoms in 1 frame)
blah blah (box name)
H     Hw     1     X     Y     Z
H     Hw     2     X     Y     Z
O     Ow     3     X     Y     Z
H     Hw     1     X     Y     Z
H     Hw     2     X     Y     Z
O     Ow     3     X     Y     Z
6
blah blah
...

rprospero:
As mentioned in #1890, it would be useful to have additional informtation (i.e. atom index and atom type) along side the trajectory information.  I've added an *extended* trajectory format that includes the extra columns.  I have also added a unit test for checking the trajectory export and updated the documention to include this alternate format.

Closes #1890

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ tests/epsr/H2O-weighted-total.gr.broad
 tests/epsr/H2OX-weighted-total.gr.broad
 tests/epsr/HDO-ReferenceData.q
 tests/epsr/HDO-ReferenceData.r
+tests/data/dissolve/input/TestOutput_*
 
 # Visual Studio
 .vscode/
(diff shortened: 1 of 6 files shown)


trisyoungs (approved):
Lovely stuff.  
:+1:

PR #1901, opened 2024-05-22
refactor: Refactor PairPotentials - Part 1

trisyoungs:
This PR performs some preliminary work on the `PairPotential` class and associated code with a bigger-picture view of implementing #1900.

Herein we tidy the class up a little and improve the update handling of pairpotentials prior to formation of the main `PotentialMap`.

Works towards #1900.

Issue #1900, opened 2024-05-22
Artificial Atoms


trisyoungs:
**Describe the feature / issue**
We require the ability to add artificial or "dummy" atoms to a `Species` for the purposes of being able to define interacting sites such as:
- Shells, permitting the use of core-shell models within Dissolve
- Soft energy "zones", permitting trial implementation of dynamic geometry coordination restraints etc.

Of key importance is that these additional atoms do not constitute parts of the correlation function calculations, for instance in the `GR` and `SQ` modules, and so must be treated / partitioned carefully.

PR #1903, opened 2024-05-24
refactor: Refactor PairPotentials - Part 2

trisyoungs:
Following on from #1901 this PR addresses the user-facing elements of pair potential display and editing from the perspective of the new option to use combination rules or not.

Works towards #1900.

### DIFF
--- a/src/main/io.cpp
+++ b/src/main/io.cpp
@@ -143,6 +143,19 @@ SerialisedValue Dissolve::serialisePairPotentials() const
     if (atomTypeChargeSource_)
         pairPotentials["includeCoulomb"] = true;
     Serialisable::fromVectorToTable(coreData_.atomTypes(), "atomTypes", pairPotentials);
+    if (!useCombinationRules_)
+    {
+        pairPotentials["useCombinationRules"] = false;
+        SerialisedValue pots = SerialisedValue::array_type{};
+        for (const auto &[at1, at2, pot] : pairPotentials_)
+        {
+            SerialisedValue value = pot->serialise();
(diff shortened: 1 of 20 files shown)


src/main/io.cpp
```diff
@@ -143,6 +143,19 @@ SerialisedValue Dissolve::serialisePairPotentials() const
...
+        pairPotentials["useCombinationRules"] = false;
+        SerialisedValue pots = SerialisedValue::array_type{};
+        for (const auto &[at1, at2, pot] : pairPotentials_)
+        {
+            SerialisedValue value = pot->serialise();
+            value["atomTypeI"] = at1->name();
+            value["atomTypeJ"] = at2->name();
+            pots.push_back(value);
+        }
+        pairPotentials["potentials"] = pots;
```
rprospero:
```suggestion
        Serialisable::fromVector(pairPotentials_, "potentials", pairPotentials, [](const auto& term) {
          const auto &[at1, at2, pot] = term;
          auto value = pot->serialise();
          value["atomTypeI"] = at1->name();
          value["atomTypeJ"] = at2->name();
          return value;});
```

rprospero: This should simplify the code a little bit.  However, I have not tested this code, so it might be worth merging your code as in, instead of waiting for my patch to fail on the CI.

rprospero (approved):
One possible change to the TOML parsing, but things otherwise look good.

src/main/io.cpp
trisyoungs: Tested, working, and pushed up.

PR #1905, opened 2024-05-24
fix: Update SpeciesSite keywords after data mutation

### LINKED ISSUE #1896: Newly-added sites don't appear in keyword widgets
**Describe the bug**
When adding a new site to a species, related keyword widgets (e.g. `SiteVector` do not get updated).

RobBuchananCompPhys:
Previously, the species site vector keyword widget wasn't updating in line with newly added species sites. This is now updating when a site is added (see below)

Closes #1896 

WaterSites
WaterSites2


trisyoungs (approved):
Nice and easy!

I might suggest a change to the PR title, bearing in mind how our changelogs are going to be generated from this point - we should perhaps be more specific and call it "fix: Update SpeciesSite keywords correctly on data mutation" or something like that.

PR #1907, opened 2024-05-29
feat: Add Derivative namespace.

trisyoungs:
Represents "Part 3.5" in the `PairPotential` refactoring work.

This PR sits in-between #1906 and #1908, and moves code for calculation of derivatives for `Data1D` from the `PairPotential` class into a separate reusable namespaced function.

Works towards #1900.

### DIFF
--- a/src/math/derivative.h
+++ b/src/math/derivative.h
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#pragma once
+
+#include <vector>
+
+// Forward Declarations
+class Data1D;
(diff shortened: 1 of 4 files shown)


src/math/derivative.h
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#pragma once
+
+#include <vector>
+
```
jws-1: Not sure we need this?
```suggestion
```

jws-1 (approved):
One minor comment to consider. Aside from that, looks good and cleans things up nicely! :+1:

PR #1910, opened 2024-05-30
refactor: Tidy site generation

trisyoungs:
This PR tidies "unique site" generation - these are instances of a `SpeciesSite` description occurring in the same `Species`. These instances are generated from either dynamic element / atom type selection, or via NETA fragment description. Previously, the generated vectors of indices corresponding to the origin, xAxis, yAxis, as well as "all involved atoms" were held in separate `std::vector`s of `std::vector`s which was a bit messy. Here we introduce a `SpeciesSiteInstance` class to knit the information corresponding to individual instances together more tightly.

### DIFF
--- a/src/gui/speciesTab_sites.cpp
+++ b/src/gui/speciesTab_sites.cpp
@@ -16,6 +16,21 @@ SpeciesSite *SpeciesTab::currentSite()
     return sites_.data(ui_.SiteList->selectionModel()->currentIndex(), Qt::UserRole).value<SpeciesSite *>();
 }
 
+// Update instance count group
+void SpeciesTab::updateInstanceCountGroup()
+{
+    auto *site = currentSite();
+    if (!site)
+        ui_.InstanceCountLabel->setText("N/A");
+    else if (site->instances().empty())
+        ui_.InstanceCountLabel->setText("0 (no instances generated)");
+    else
+        ui_.InstanceCountLabel->setText(
+            QString("%1 (%2 atom%3 each)")
+                .arg(QString::number(site->instances().size()), QString::number(site->instances().front().allIndices().size()),
+                     QString::fromStdString(DissolveSys::plural(site->instances().front().allIndices().size()))));
+}
+
 /*
  * Private Slots
  */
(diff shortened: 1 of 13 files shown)


tests/classes/speciesSite.cpp
```diff
@@ -166,19 +166,71 @@ TEST_F(SpeciesSiteTest, FragmentBasic)
...
+    // Re-set with actual origin on the root match
+    EXPECT_TRUE(site.setFragmentDefinitionString("?C,#origin"));
+
+    // Expect six instances, one for each carbon
+    EXPECT_EQ(site.instances().size(), 6);
+    for (auto &instance : site.instances())
+    {
+        // There should be 1 atom in the site overall
+        EXPECT_EQ(instance.allIndices().size(), 1);
+        // There should be a single defined origin atoms
```
jws-1:
```suggestion
        // There should be a single defined origin atom
```

tests/classes/speciesSite.cpp
```diff
@@ -166,19 +166,71 @@ TEST_F(SpeciesSiteTest, FragmentBasic)
...
-    EXPECT_EQ(site.sitesAllAtomsIndices().size(), 1);
-    // There should be 12 atoms in the site
-    EXPECT_EQ(site.sitesAllAtomsIndices().front().size(), 12);
+    // Expect a single instance
+    EXPECT_EQ(site.instances().size(), 1);
+    const auto &instance = site.instances().front();
+
+    // There should be 12 atoms in the site overall
+    EXPECT_EQ(instance.allIndices().size(), 12);
     // There should be a single defined origin atoms
```
jws-1:
```suggestion
    // There should be a single defined origin atom
```

src/gui/speciesTab_sites.cpp
```diff
@@ -16,6 +16,21 @@ SpeciesSite *SpeciesTab::currentSite()
...
+    if (!site)
+        ui_.InstanceCountLabel->setText("N/A");
+    else if (site->instances().empty())
+        ui_.InstanceCountLabel->setText("0 (no instances generated)");
+    else
+        ui_.InstanceCountLabel->setText(
+            QString("%1 (%2 atom%3 each)")
+                .arg(QString::number(site->instances().size()), QString::number(site->instances().front().allIndices().size()),
+                     QString::fromStdString(DissolveSys::plural(site->instances().front().allIndices().size()))));
+}
```
jws-1, line 32: I like it!

jws-1 (changes requested):
This cleans things up nicely. Couple of super minor changes to some comments, and a bug that I've noticed:

If the `SpeciesSite` previously had instances generated, and then an invalid `definitionString` is given,  then either a) the `Instance Count` is not correctly updated in the GUI or b) the prior instances do not get cleared. I have a feeling the problem is the latter, and this must've existed before but didn't get noticed!

jws-1 (approved):
:+1:

PR #1917, opened 2024-06-03
DRAFT Custom model interface

trisyoungs:
@rprospero here is the stuff I came up with when looking into custom model interfaces. Still uses a template class sitting alongside the data (here focussing just on `std::vector` and tables) but the template derives from a concrete base class which the Qt model leverages to do all the work.

Thoughts on how any of this (if any!) can work with your own approach welcome!

### DIFF
--- a/src/keywords/expressionVariableVector.h
+++ b/src/keywords/expressionVariableVector.h
@@ -10,6 +10,91 @@
 class ExpressionVariable;
 class ProcedureNode;
 
+// Data property types
+enum class PropertyType
+{
+    Invalid,
+    Integer,
+    Double,
+    String
+};
+
+// Name / Column Title, Data Type, ReadOnly?
+using DataItemProperty = std::tuple<int, std::string, PropertyType, bool>;
+
+struct DataItemValue
+{
+    PropertyType type;
+    union
+    {
+        int intValue{0};
+        double doubleValue;
+        std::string stringValue;
+    };
+};
+template <class DataItem> class DataModel
+{
+    public:
+    // Data access functions
+    using DataSetFunction = std::function<bool(DataItem&, int, DataItemValue)>;
+    using DataGetFunction = std::function<DataItemValue(DataItem&, int)>;
+    DataModel(std::vector<DataItem> &data, const std::vector<DataItemProperty> &itemProperties) : data_(data), itemProperties_(itemProperties) {}
+
+    private:
+    // Target data for the model
+    std::vector<DataItem> &data_;
+
+    /*
+     * Extent
+     */
+    private:
+    // Return whether the supplied index is valid
+    bool isIndexValid(int childIndex, int columnIndex) const { return (childIndex >= 0 && childIndex < data_.size() && columnIndex >= 0); }
+    // Functions for accessing data extents (table style)
+    std::function<int(int row, int column)> childCountFunction_{
+        [&](const int childIndex, const int columnIndex) { return isIndexValid(childIndex, columnIndex) ? 0 : data_.size(); }};
+    std::function<int(int row, int column)> dataItemCountFunction_{
+        [&](const int childIndex, const int columnIndex) { return !isIndexValid(childIndex, columnIndex) ? 0 : itemProperties_.size(); }};
+
+    public:
+    // Return number of children (rows) for the specified index
+    int nChildren(int childIndex, int columnIndex) { return childCountFunction_(childIndex, columnIndex); }
+    // Return number of data items (columns) for the specified index
+    int nDataItems(int childIndex, int columnIndex) { return dataItemCountFunction_(childIndex, columnIndex); }
+
+    /*
+     * Data Access
+     */
+    private:
+    // Descriptions of relevant item properties within a single object in the container
+    std::vector<DataItemProperty> itemProperties_;
+    // Return whether column index holds / can be set by the given type
+    bool isPropertyType(int columnIndex, PropertyType descriptorType) { return std::get<2>(itemProperties_[columnIndex]) == descriptorType; }
+    // Set / get functions, unique per class
+    DataSetFunction setDataFunction_;
+    DataGetFunction getDataFunction_;
+
+    public:
+    // Set data functions
+    bool setData(int childIndex, int columnIndex, const DataItemValue &newValue) {
+        // Check index validity
+        if (!isIndexValid(childIndex, columnIndex))
+            return false;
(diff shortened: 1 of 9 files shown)


src/keywords/expressionVariableVector.h
```diff
@@ -10,6 +10,170 @@
...
+    bool isPropertyFlagSet(int propertyIndex, PropertyFlag flag)
+    {
+        return std::get<3>(itemProperties_[propertyIndex]).isSet(flag);
+    }
+
+    public:
+    // Return number of children (rows) for the specified index
+    virtual int nChildren(int dataIndex, int propertyIndex) = 0;
+    // Return number of properties (i.e. columns) for the specified index
+    virtual int nProperties(int dataIndex, int propertyIndex) = 0;
```
rprospero, line 84: I'm not sure that these need to depend on the data and property index parameters.  We're not using them for the DataTable and, even when we extend this to trees, I believe that the result depends more upon the raw `parent` index than the specific row and column.

src/keywords/expressionVariableVector.h
```diff
@@ -10,6 +10,170 @@
...
+    // Return number of properties per child (i.e. columns) for the specified index
+    int nProperties(int dataIndex, int propertyIndex) final { return propertyCountFunction_(dataIndex, propertyIndex); }
+
+    /*
+     * Data Access
+     */
+    private:
+    // Set / get functions, unique per class
+    PropertySetFunction setPropertyFunction_;
+    PropertyGetFunction getPropertyFunction_;
```
rprospero, line 134: I might handle this slightly differently.  I would have two global functions `getProp` and `setProp`.  These functions could then be overloaded for each type

```cpp
DataItemValue getProp(const AtomType &at, int index);
bool setProp(AtomType &at, int index, DataItemValue value);
DataItemValue getProp(const MasterBond &at, int index);
bool setProp(MasterBond &at, int index, DataItemValue value);
```

The header for the function can then go in the header for the class and the definition can go in the corresponding cpp file.  The call to `setProperty` in DataModelBase can then just call `setProp(data_[dataIndex], std::get<0>(itemProperties_[propertyIndex]))`

The biggest advantage to this method is that we handle the generics at compile time, instead of runtime.  It gives the compiler more chances to catch mistakes and optimise calls.  It also would allow us to use the `getProp` function in other places in the CLI code.  I believe that it might enable us to automate some parsing and serialisation.

The largest disavantage that I see is that you cannot have two different property accessors for the same type.  This can be alleviated, thought, through the newtype alias pattern.

I'm even tempted to take this a step further and have another overloaded function that handles the item properties

```cpp
template<typename T>
std::vector<DataItemProperty> itemProperties() = 0;

template<>
std::vector<DataItemProperty> itemProperties<AtomType> {...}
template<>
std::vector<DataItemProperty> itemProperties<MasterBondTerms> {...}
```

src/keywords/expressionVariableVector.h
```diff
@@ -10,6 +10,170 @@
...
+    public:
+    // Return value type
+    PropertyType type() const { return type_; }
+    // Return integer value
+    int intValue() const { return intValue_; }
+    // Return double value
+    double doubleValue() const { return doubleValue_; }
+    // Return string value
+    std::string stringValue() const { return stringValue_; }
+};
```
rprospero, line 55: I would have written this as

```cpp
using DataItemValue = std::variant<int, double, std::string_view, std::string>;
```

I prefer the std::variant solution because it enforces type safety and prevents the code below.

```
DataItemValue count(7);
return count.stringValue();
```

Also, with the std::variant, we can eliminate the Invalid enum type because you can only construct valid values.

src/keywords/expressionVariableVector.h
```diff
@@ -10,6 +10,170 @@
...
+
+// Data property flags
+enum PropertyFlag
+{
+    ReadOnly
+};
+
+// Index / Column ID, Name / Column Title, Data Type, ReadOnly?
+using DataItemProperty = std::tuple<int, std::string, PropertyType, Flags<PropertyFlag>>;
```
rprospero, line 29: I think that this might be a bit clearer as a `struct`.  That would allow us to do `diProp.index` instead of `std::get<0>(diProp)`.  Of course, we could still overload std::get if we really needed to.

rprospero (commented):
It looks like a good framework, but I had some thoughts on how we could get more type safety and reusability from the code.

src/keywords/expressionVariableVector.h
trisyoungs: Yes, you're probably right.  I was making the mistake of trying to think of the whole problem rather than deal with a concrete use case (the Table) first.

src/keywords/expressionVariableVector.h
trisyoungs: Now, this is an interesting comment.  I see exactly what you're saying here - indeed I wondered about something semi-templated at the beginning.  However, as-written my approach completely separates the generic model from any knowledge of the underlying class it's accessing, which I find quite satisfying. I concede that supplying lambda functions isn't the cleanest way of doing things, but I like the locality and separation of doing it this way, while with overloads we are creating soft-dependencies on (or at least pseudo-knowledge of) our data classes.

src/keywords/expressionVariableVector.h
trisyoungs: Yes, I agree with you fully on this one.  I had created the horrible `DataItemValue` class as a sort of analogue to `QVariant` and to allow it to be `Invalid` - something that you can't do (easily at least) when using `std::variant`.  That said, I never ended up using `PropertyType::Invalid` anywhere yet, so I think the `std::variant` approach is the one best to follow.  The same behavour could be gained with `std::optional<std::variant<>>` as a return value if we really needed it.

src/keywords/expressionVariableVector.h
trisyoungs: Ah, for sure!  I was using the `std::tuple` just to quickly spin up a working example, although all the `std::get<>` pollution quickly became rather tiresome. `struct` or `class` definitely the way to go.

src/keywords/expressionVariableVector.h
rprospero: Honestly, we should stick with your solution.  Re-reading what I've written, I've just put together half of C++20 Concepts.  If we're not using Concepts or C++20, then there's no reason to use a subpar imitation.

PR #1918, opened 2024-06-03
fix: Remove redundant assertions

trisyoungs:
A quick fix to remove some now-redundant assertions from the code, left in following #1893.


RobBuchananCompPhys (commented):
Looking good! I'll wait for checks to pass before approving

trisyoungs:
> Looking good! I'll wait for checks to pass before approving

Well, I would say that we can't have broken anything since we've just removed some `assert` statements, and the CI is entirely `Release` mode at present, so I'd say just go ahead and merge and fix your local build!

RobBuchananCompPhys (approved):

PR #1928, opened 2024-06-05
refactor: External potentials now use Function1D

trisyoungs:
Following work on the `PairPotential` which focussed on leveraging our `Function1D` class and keeping all such function definitions in a single location, here we apply the same to the various `ExternalPotential`-derived classes.


src/kernels/potentials/cylindrical.cpp
```diff
@@ -0,0 +1,108 @@
...
+    auto xyyx = vector_.x * v.y - vector_.y * v.x;
+    // Cross product between the line and the position of the atom in the xz plane
+    auto xzzx = vector_.x * v.z - vector_.z * v.x;
+    // Cross product between the line and the position of the atom in the yz plane
+    auto yzzy = vector_.y * v.z - vector_.z * v.y;
+
+    // Compute new components
+    v.x = vector_.y * xyyx + vector_.z * xzzx;
+    v.y = vector_.z * yzzy - vector_.x * xyyx;
+    v.z = -vector_.x * xzzx - vector_.y * yzzy;
```
rprospero: I would have found it a bit cleaner to use `zxxz` instead of `xzzx`, since that keeps the signs identical on all the term calculations, but it's not that significant.

rprospero (approved):
A minor comment about the cross product, but otherwise looks good.

jws-1 (approved):
Looks good to me, and from a quick test the potentials still seem to work as expected.

src/kernels/potentials/cylindrical.cpp
trisyoungs: And now you've said that I can't un-see it!  Changed as suggested.

PR #1930, opened 2024-06-05
refactor: Function1D definitions

trisyoungs:
This PR changes the way that `Function1DDefinitions` are defined in the code, moving away from the obscure bunch of strictly-ordered initialisers and introducing specific setters.


rprospero (approved):
This is much cleaner

Danbr4d (approved):
Definitely more clear than before

PR #1931, opened 2024-06-06
refactor: Remove custom analysis nodes

trisyoungs:
Since moving over to new developer-friendly analysis classes in #1789 and #1795 we no longer have need for the set of `ProcedureNode`s that were used in the analysis modules, nor the general-purpose `AnalysisModule` itself. Here we remove all of that.

Given the potentially low number (perhaps zero) of use cases in the user community this is not being flagged as a breaking change.

Closes #1877 (sort of!).
Closes #1653.

PR #1933, opened 2024-06-06
refactor: Remove contexts from procedure

### LINKED ISSUE #1503: Keyword taking a Node / Enum pair
The `CollectND` nodes currently use `NodeAndIntegerKeyword` to take a named node and an optional integer, with the integer indicating the index of the data to take (the named node may offer up a Vec3<double> for instance).

However, the indices are zero-indexed which is not the user-facing convention adopted in Dissolve (all object indices start from 1). For consistency we  should do the +/-1 conversion on I/O.

trisyoungs:
This PR follows #1931 and removes the `NodeContext` property since our `Procedure` is now basically used for one thing (generation of `Configuration`s).  This will probably mean another renaming at some point.

Closes #1503.


rprospero (approved):
Much cleaner without the contexts

PR #1935, opened 2024-06-06
fix: Don't run analysis modules if configuration contents haven't changed

trisyoungs:
After merging of #1789 and #1795, analysis modules no longer used a `Procedure`. This means that we are now bypassing the check on a configuration's version to ensure that its contents have changed since the last run of the module.

This PR reintroduces that check in a more general way across all modules.


src/module/module.cpp
```diff
@@ -156,6 +158,90 @@ bool Module::isDisabled() const { return !enabled_; }
...
+                                { return lastProcessedConfigurations_[currentTarget] == currentTarget->contentsVersion(); }))
+                {
+                    Messenger::warn("One or more target configurations have not changed since module '{}' was last run, so it "
+                                    "will not run in the current iteration.\n",
+                                    name());
+                    return ExecutionResult::NotExecuted;
+                }
+            }
+        }
+    }
```
rprospero:
```suggestion
    if (expectedTargetCount <= 0)
        return ExecutionResult::Success;
    
    // Check basic target count
    if (currentTargets.size() < expectedTargetCount)
    {
        Messenger::error("Not enough configuration targets set for module '{}'.\n", name());
        return ExecutionResult::Failed;
    }

    // Check that the current targets are consistent with the ones we last used
    if (lastProcessedConfigurations_.empty())
        return ExecutionResult::Success;

    // If the vector of previous targets isn't the same size as the current targets then we must clear our data
    if (currentTargets.size() != lastProcessedConfigurations_.size() ||
        !std::all_of(currentTargets.begin(), currentTargets.end(),
                     [&](const auto *currentTarget)
                     {
                         return std::find_if(lastProcessedConfigurations_.begin(), lastProcessedConfigurations_.end(),
                                             [currentTarget](const auto &pair) {
                                                 return pair.first == currentTarget;
                                             }) != lastProcessedConfigurations_.end();
                     }))
    {
        Messenger::warn("Target configuration(s) have changed for module '{}' so processing data for that module will "
                        "be cleared...\n");
        processingModuleData.removeWithPrefix(name());
        lastProcessedConfigurations_.clear();
    }
    else if (!executeIfTargetsUnchanged_)
    {
        // Targets are the same - are _all_ versions different?
        if (std::any_of(currentTargets.begin(), currentTargets.end(),
                        [&](const auto *currentTarget)
                        { return lastProcessedConfigurations_[currentTarget] == currentTarget->contentsVersion(); }))
        {
            Messenger::warn("One or more target configurations have not changed since module '{}' was last run, so it "
                            "will not run in the current iteration.\n",
                            name());
            return ExecutionResult::NotExecuted;
        }
    }
```

We can trade some early returns here to reduce the number of nested conditionals.  Granted, this is entirely stylistic.

src/modules/temperatureSchedule/process.cpp
```diff
@@ -7,17 +7,10 @@
...
-    {
-        Messenger::error("No configuration target set for module '{}'.\n", name());
-        return ExecutionResult::Failed;
-    }
-
     // Get the current configuration temperature
     auto currentT = targetConfiguration_->temperature();
 
-    // Get difference with tergat temperature
+    // Get difference with targat temperature
```
rprospero:
```suggestion
    // Get difference with target temperature
```

rprospero (approved):
Just one typo fix and an optional style change

src/modules/temperatureSchedule/process.cpp
trisyoungs: Whoops!

trisyoungs:
@rprospero Re-requesting a review on this one since it turns out things didn't quite work as expected following further testing!

rprospero (approved):

rprospero (approved):

Issue #1927, opened 2024-06-04
Tidy Benchmarks


trisyoungs:
**Describe the feature / issue**
The organisation and code behind our benchmarks hasn't been touched in a long time and would benefit from being brought up-to-date with the rest of the code.

PR #1937, opened 2024-06-11
perf: Combinable histograms for angle module parallel execution

RobBuchananCompPhys:
I've worked on parallelising the module using a similar approach to the previous multithreading work on #1882 (SiteRDF).
Running into a snag where `Angle` tests are failing *only* when `execution_policy` is `par` (but passing when running sequentially!).
I'm guessing this is due to a part of the histogram binning process being non-thread safe, and I believe I have narrowed it down to the `Histogram3D`, since the output from the 1D and 2d histograms looks correct in `gdb`. Surprising, as the histogram classes seem to all effectively be wrappers to a linear array, so I am still trying to figure out where this is going wrong in parallel 3D binning.

### DIFF
--- a/src/math/histogram2D.cpp
+++ b/src/math/histogram2D.cpp
@@ -195,6 +195,22 @@ void Histogram2D::operator=(const Histogram2D &source)
     averages_ = source.averages_;
 }
 
+Histogram2D Histogram2D::operator+(const Histogram2D &other) const
+{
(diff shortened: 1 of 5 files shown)


trisyoungs:
Regarding this the updated module runs _really_ slowly in parallel, so there's something bad happening somewhere. I have taken a bit of a diversion and decided to handle #1927 and then add one for your module so we can do a before/after comparison too.  This is something we should commit do doing for every module we convert in this way.

Issue #508, opened 2021-01-05
Don't use indexing of AtomTypes for data access


trisyoungs:
Many aspects of the code, in particular the `PartialSet`s generated by the correlation function modules, rely on accessing square 2D arrays of data where the row/column indices are intended to correspond to the `AtomType`s as ordered in an associated `AtomTypeList`.

This is satisfactory for single-configuration simulations which are, at present, the predominant use case, however such an access mechanism is inherently unreliable going forward, as the ordering (and hence indices) of objects in an `AtomTypeList` as constructed across different places in the code cannot be guaranteed to be consistent, or indeed complete.

The dependence on `AtomType` indices should be removed, and a more "dynamic" approach to accessing data in those objects should be considered. Typically, such operations are not used in time-critical portions of code, and so a suitable lookup-style approach should be sufficient.

Related to #289.

trisyoungs:
There _are_, as suggested above, places in the code where performance is not an issue - e.g. when manipulating `PartialSet`s. There, a `std::map` of `std::map`s based around `AtomType`s might be all that is required.

However, the key consideration to make is that calculating any energy and force between `Atom`s requires the indices of the `AtomType` so that they can be looked up as fast as possible in an `Array2D` of `Potential`s. As highlighted in #289 this index information must be local to the `Atom` class in order to remain performant, so we arrive at the only likely possibility for improvement being to ensure that our centralised `PotentialMap` is consistent with the indices of the available `AtomType`s.

trisyoungs:
Closed via work in `develop2`

Issue #289, opened 2020-05-15
Simplify AtomTypeList and Associated Classes


trisyoungs:
PR #287 highlights some issues with the AtomTypeData and AtomTypeList classes, and which should be addressed.

- [x] Remove need for storing listIndex_ in AtomTypeData (consider implications on ordering requirements post-setup)
- [x] Review use of Atom::masterTypeIndex() for location of atom type data in interaction potential
- [x] Remove unused zero() / zeroPopulations() functions
- [x] Remove read()/write() functions from AtomTypeData and IsotopeData, and perform all reading through AtomTypeList for simplicity?

trisyoungs:
It is not possible to replace the use of a `masterTypeIndex()` within the `Atom` class for the purposes of accessing interaction potential data. Any other method of indexing for the 2D array in `PotentialMap` is significantly slower, even just referencing the `SpeciesAtom` to get the `AtomType` and its index.

So this information will have to remain in the `Atom` class, and must be in the form of a POD local to the `Atom` class for performance reasons.

Future workplan is outlined in #508.

PR #1938, opened 2024-06-18
fix: AtomType Indexing

trisyoungs:
This PR starts to touch on the painful world of indices associated to our `AtomType`s, relating to #289 and #508.

Here we begin by removing the local index storage from the `AtomTypeData` class and modify `AtomTypeMix` to only allow `add()`ing atom types, prohibiting `remove()` operations. This indexing is _still_copied over to `Atom`s making up a `Configuration` since there is no faster way of accessing the 2D array when computing forces and energies between atoms, but at least we remove one piece of easily-broken storage in a class.

Works towards #289.

### DIFF
--- a/src/classes/atomTypeMix.cpp
+++ b/src/classes/atomTypeMix.cpp
@@ -47,7 +47,8 @@ std::pair<AtomTypeData &, int> AtomTypeMix::add(std::shared_ptr<AtomType> atomTy
         return {*atd, atd - types_.begin()};
     }
 
-    return {types_.emplace_back(atomType, population), types_.size()};
+    auto &newAtomTypeData = types_.emplace_back(atomType, population);
+    return {newAtomTypeData, types_.size() - 1};
 }
 
 // Add the AtomTypes in the supplied list into this one, increasing populations etc.
--- a/src/classes/atomTypeMix.cpp
+++ b/src/classes/atomTypeMix.cpp
@@ -57,7 +57,7 @@ void AtomTypeMix::add(const AtomTypeMix &source)
     // Loop over AtomTypes in the source list
     for (auto &otherType : source)
     {
-        auto &&[atd, std::ignore] = add(otherType.atomType());
+        auto &atd = std::get<0>(add(otherType.atomType()));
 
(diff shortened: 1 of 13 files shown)


src/classes/atomTypeMix.cpp
```diff
@@ -56,7 +57,7 @@ void AtomTypeMix::add(const AtomTypeMix &source)
     // Loop over AtomTypes in the source list
     for (auto &otherType : source)
     {
-        AtomTypeData &atd = add(otherType.atomType());
+        auto &&[atd, atdIndex] = add(otherType.atomType());
```
rprospero:
```suggestion
        auto &&[atd, std::ignore] = add(otherType.atomType());
```
Since we never use the `atdIndex`, it's best just to ignore it.

src/classes/atomTypeMix.cpp
```diff
@@ -67,17 +68,10 @@ void AtomTypeMix::add(const AtomTypeMix &source)
...
-{
-    types_.erase(
-        std::remove_if(types_.begin(), types_.end(), [&atomType](const auto &atd) { return atd.atomType() == atomType; }));
-}
-
 // Add/increase this AtomType/Isotope pair
 void AtomTypeMix::addIsotope(std::shared_ptr<AtomType> atomType, Sears91::Isotope tope, double popAdd)
 {
-    auto &atd = add(std::move(atomType));
+    auto &&[atd, atdIndex] = add(std::move(atomType));
```
rprospero:
```suggestion
    auto &&[atd, std::ignore] = add(std::move(atomType));
```

rprospero (approved):
A small suggestion to ignore excess variables, but otherwise quite a nice change.

src/classes/atomTypeMix.cpp
trisyoungs: :+1:  Had no idea this was a thing!

trisyoungs: @rprospero ...although annoyingly you cannot use `std::ignore` in a structured binding (yet). As a middle ground I'll refactor these to be `auto &atd = std::get<0>(...)`

PR #1941, opened 2024-06-21
fix: Rename modules bug

### LINKED ISSUE #1594: No graph data when module is renamed
Dissolve lets you rename analysis modules but when they are renamed, the data does not seem to be showing up on the graphs in the module.
I have tried this for HistogramCN and Angle and IntraDistance with the same effect. I run the module first and data is printed on the graph, then I renamed the module, cleared the data and ran the simulation again but this time no data appeared on the graph.
I ran this on IDAaaS.

Danbr4d:
Closes #1594 .

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -10,6 +10,7 @@ dissolve_benchmark(energy)
 dissolve_benchmark(forces)
 dissolve_benchmark(region)
 dissolve_benchmark(math)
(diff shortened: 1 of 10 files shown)


src/math/function1D.cpp
trisyoungs, line 1: Not sure what these changes are doing here!

trisyoungs (changes requested):
Fix looks fine, but there's some extraneous commit data for `Function1D` that has crept in.

PR #1942, opened 2024-06-24
fix: Graph rename bug fix

### LINKED ISSUE #1594: No graph data when module is renamed

Danbr4d:
Closes #1594.

### DIFF
--- a/src/gui/layerTab.cpp
+++ b/src/gui/layerTab.cpp
@@ -237,10 +237,15 @@ void LayerTab::moduleNameChanged(const QModelIndex &index, const QString &oldNam
     // Find the control widget for the module and update it
     auto *mcw = getControlWidget(module);
     if (mcw)
-        mcw->updateControls();
+        mcw->updateControls(ModuleWidget::RecreateRenderablesFlag);
 
     // Rename processing module data
     dissolve_.processingModuleData().renamePrefix(oldName.toStdString(), newName.toStdString());
+
+    if (mcw)
+        mcw->updateControls();
+
+    // another if to recreate controls
 }
 
 // Update the module list
--- a/src/gui/layerTab.cpp
+++ b/src/gui/layerTab.cpp
@@ -244,8 +244,6 @@ void LayerTab::moduleNameChanged(const QModelIndex &index, const QString &oldNam
 
     if (mcw)
         mcw->updateControls();
-
-    // another if to recreate controls
 }
 
 // Update the module list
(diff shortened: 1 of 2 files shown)


Danbr4d:
Not sure how big an issue is but if you rename to something with a number in then it causes a crash too.

src/gui/layerTab.cpp
```diff
@@ -237,10 +237,13 @@ void LayerTab::moduleNameChanged(const QModelIndex &index, const QString &oldNam
...
     auto *mcw = getControlWidget(module);
     if (mcw)
-        mcw->updateControls();
+        mcw->updateControls(ModuleWidget::RecreateRenderablesFlag);
 
     // Rename processing module data
     dissolve_.processingModuleData().renamePrefix(oldName.toStdString(), newName.toStdString());
+
+    if (mcw)
+        mcw->updateControls();
```
trisyoungs: If this doesn't help they we may as well remove it.
```suggestion
```

trisyoungs (approved):
:+1:

trisyoungs:
@Danbr4d the issue with renaming I _think_ boils down to the function which was renaming (badly) the relevant data.  I have updated that particular function in the best way possible.  @rprospero if you wouldn't mind taking a look over the horrible code in question, that'd be great (I'll ask for a review from you). I don't see any other way of achieving the same thing...

rprospero:
@trisyoungs I'm not entirely sure that we need to copy into the other map.  The [cpp reference](https://en.cppreference.com/w/cpp/container/map/extract) seems to indicate that it is a safe action.  The only iterator that is invalidated is the one which is literally being moved.  I think we can do

```
for (auto iter = m.begin(); iter < m.end(); ++m) {
    auto node = m.extract(iter);
    if (test(node.key())) {
        node.key() = new_key(node.key());
        m.insert(std::move(node));
    }
}
```

Even if the iteration invalidation causes an issue, we can still go O(n²) in access.

```
do {
    bool done = true
    for (auto iter = m.begin(); iter < m.end(); ++m) {
        auto node = m.extract(iter);
        if (test(node.key())) {
            node.key() = new_key(node.key());
            m.insert(std::move(node));
        }
        done = false;
        break;
    }
} while(!done)
```

It's still ugly, but at least it's constant memory.

trisyoungs:
@rprospero You're aboslutely right - not sure why I got stuck in the copying methodology. I'm still convinced that there are issues trying to do this in a single loop, so I have reworked in line with your second suggestion, but in a slightly more `std` algorithm way.

rprospero (approved):

PR #1943, opened 2024-06-24
fix: Debug mode assertions error due to const atom not pointer

RobBuchananCompPhys:


src/classes/potentialMap.cpp
```diff
@@ -127,7 +127,7 @@ double PotentialMap::energy(const SpeciesAtom *i, const SpeciesAtom *j, double r
 double PotentialMap::analyticEnergy(const Atom &i, const Atom &j, double r) const
 {
     assert(r >= 0.0);
-    assert(i && j);
+    assert(&i && &j);
```
trisyoungs: Good catch. I think these assertions are best removed, however, since even a malformed reference will have a non-null pointer?  So this is now a bit redundant.
```suggestion
```

src/classes/potentialMap.cpp
```diff
@@ -141,7 +141,7 @@ double PotentialMap::analyticEnergy(const Atom &i, const Atom &j, double r) cons
 double PotentialMap::analyticEnergy(const Atom &i, const Atom &j, double r, double elecScale, double srScale) const
 {
     assert(r >= 0.0);
-    assert(i && j);
+    assert(&i && &j);
```
trisyoungs:
```suggestion
```

src/classes/potentialMap.cpp
```diff
@@ -205,8 +205,8 @@ double PotentialMap::force(const SpeciesAtom *i, const SpeciesAtom *j, double r,
 double PotentialMap::analyticForce(const Atom &i, const Atom &j, double r) const
 {
     assert(r >= 0.0);
-    assert(i && j);
-    assert(i->speciesAtom() && j->speciesAtom());
+    assert(&i && &j);
```
trisyoungs:
```suggestion
```

src/classes/potentialMap.cpp
```diff
@@ -220,8 +220,8 @@ double PotentialMap::analyticForce(const Atom &i, const Atom &j, double r) const
 double PotentialMap::analyticForce(const Atom &i, const Atom &j, double r, double elecScale, double srScale) const
 {
     assert(r >= 0.0);
-    assert(i && j);
-    assert(i->speciesAtom() && j->speciesAtom());
+    assert(&i && &j);
```
trisyoungs:
```suggestion
```

trisyoungs (commented):
:+1:

src/classes/potentialMap.cpp
RobBuchananCompPhys: Ah yes, of course!

trisyoungs (approved):

PR #1949, opened 2024-06-27
feat: voxel density module

### LINKED ISSUE #1915: Voxel Density
A new module to assess the (atomic) number, mass, or scattering length density over voxels in the unit cell would be very useful in determining, for instance, wall density in porous amorphous materials, or general condensed phases.

Restricting to specific `Species` may also be a useful option.

RobBuchananCompPhys:
Closes #1915


src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,67 @@
...
+                    (double)z*numPoints_,
+                    Sears91::boundCoherent(naturalIsotope);
+                ); 
+            }
+            default:
+            {
+                throw(std::runtime_error(
+                fmt::format("'{}' not a valid property.\n", targetProperty_)));
+            }
+        }
```
rprospero:
```suggestion

        double value;
        switch(targetProperty_)
        {
            case TargetPropertyType::Mass: 
                value = AtomicMass::mass(atomicNumber);
                break;
            case TargetPropertyType::AtomicNumber: 
                value = atomicNumber;
                break;
            case TargetPropertyType::ScatteringLengthDensity:
                value = Sears91::boundCoherent(Sears91::naturalIsotope(atomicNumber));
            }
            default:
            {
                throw(std::runtime_error(
                fmt::format("'{}' not a valid property.\n", targetProperty_)));
            }
        }
        density.addPoint(
            (double)x*numPoints_, 
            (double)y*numPoints_, 
            (double)z*numPoints_,
            value);
                ); 
```

I thought it might be a bit easier to read if the switch statement merely selected a value and then the same code was used to add every type of point.

More concerning is that your `switch` doesn't seem to have any `break` statements.  I would have expected that any call to this function would immediately throw and exception as the various cases fall through each other to the final, `default` case.

This brings up the larger point that this module needs a unit test to confirm that it works properly.

rprospero (changes requested):
I think that there's a significant flaw in the switch statement.

src/modules/voxelDensity/process.cpp
RobBuchananCompPhys: Thank you for pointing out the lack of `break`s, honest mistake.

What would be a good test case for this?

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,71 @@
...
+            {
+                auto naturalIsotope = Sears91::naturalIsotope(atomicNumber);
+
+                // Bound coherent natural isotope scattering length density
+                value = Sears91::boundCoherent(naturalIsotope);
+                break;
+            }
+            default:
+            {
+                // throw(std::runtime_error(fmt::format("'{}' not a valid property.\n", targetProperty_)));
```
rprospero: I noticed that you commented out the `throw` statement here.  Honestly, I liked having it there.  We have `throw` statements in the NeutronSQ and XraySQ serving the similar purpose of checking for invalid enums.  Is there a reason that you took it out?

rprospero:
@RobBuchananCompPhys  You'd made a comment asking about what could serve as a good test case.  The first idea that comes to mind is as follows:

Generate a box with a million Helium atoms in it and get the 10x10x10 voxel density.  The first test would be going through the voxel values and ensuring that the individual voxel values add up to a million.  The next part is loops through the individual values and checking for outliers.  If more than fifty of the values are outside 1000 ± 70 or more than four values are outside of 1000 ± 100, then something has gone wrong.

There's probably an even better test to be made by *not* distributing the helium atoms evenly, but the above would be a good start.

src/modules/voxelDensity/process.cpp
RobBuchananCompPhys: It was causing complilation to fail, I'll try again with it uncommented

RobBuchananCompPhys: Turns out the enum value was not formattable, casting it to an `int` works

src/math/data1D.cpp
```diff
@@ -197,6 +197,25 @@ const std::vector<double> &Data1D::values() const { return values_; }
...
+                      if (val > 0)
+                      {
+                          pairs.push_back({x_[idx], val});
+                      }
+                      idx++;
+                  });
+
+    return pairs;
+}
+
```
rprospero:
```suggestion
// Return number of non-zero values present in whole dataset
std::vector<std::tuple<double, double>> Data1D::nNonZeroValues() const
{
    std::vector<std::tuple<double, double>> pairs;
    auto xys = zip(x_, values_);
    
    std::copy_if(std::execution::par_unseq, xys.begin(), xys.end(), std::back_inserter(pairs), [](const auto &val) { return std::get<1>(val) > 0; });
    return pairs;
}
```

Using zip keeps the x_ and values_ terms cleanly together without needing to keep track of a mutable iterator, which suddenly means that we can do it all in parallel.

src/math/data1D.h
```diff
@@ -75,10 +75,14 @@ class Data1D : public Data1DBase, public Serialisable<>
     const std::vector<double> &values() const override;
     // Return number of values present in whole dataset
     int nValues() const override;
+    // Return number of non-zero values present in whole dataset
+    std::pair<double, double> nNonZeroValues() const;
```
rprospero: If you use my suggestion from above, you'll need to update the type here from a pair to a tuple.  It's also probably worth updating the name and the comment, since the function does not return the *number* of non-zero values, but the values themselves.

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,118 @@
...
+                break;
+            }
+            case TargetPropertyType::ScatteringLengthDensity:
+            {
+                auto naturalIsotope = Sears91::naturalIsotope(atomicNumber);
+
+                // Bound coherent natural isotope scattering length density
+                value = Sears91::boundCoherent(naturalIsotope);
+                break;
+            }
```
rprospero:
```suggestion
            case TargetPropertyType::Mass:
                // Atomic mass
                value = AtomicMass::mass(atomicNumber);
                break;
            case TargetPropertyType::AtomicNumber:
                // Atomic number
                value = atomicNumber;
                break;
            case TargetPropertyType::ScatteringLengthDensity:
                auto naturalIsotope = Sears91::naturalIsotope(atomicNumber);

                // Bound coherent natural isotope scattering length density
                value = Sears91::boundCoherent(naturalIsotope);
                break;
```

Cases for switch statements don't take braces.

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,118 @@
...
+    auto &hist = processingData.realise<Histogram1D>(
+        fmt::format("Histogram1D//{}/A^3", targetPropertyTypes().keywordByIndex(static_cast<int>(targetProperty_))), name(),
+        GenericItem::InRestartFileFlag);
+
+    auto max = data3D.maxValue(), min = ((numPoints_ == 1) || (max == data3D.minValue())) ? 0 : data3D.minValue();
+    hist.initialise(min, max, (max - min) / numPoints_);
+    hist.zeroBins();
+
+    dissolve::for_each(std::execution::seq, data3D.values().begin(), data3D.values().end(),
+                       [&hist](auto &value) { hist.bin(value / voxelVolume); });
```
rprospero:
```suggestion
    for(const auto &value: data3D.values())
        hist.bin(value / voxelVolume);
```

I feel like the most contrary and obnoxious reviewer for even making this suggestion, but I think that we might be better off here with just a for loop.  I'm always the guy pushing to move things into the standard algorithms, but there's a couple of reasons that I make an exception here.

1. Unless I'm missing something, you've failed to capture the `voxelVolume` term in the lambda, which should be a compile time error.  We don't have to worry about that with for loops.
1. Unlike most algorithms, `std::for_each` doesn't really inform the compiler.  The nice part about algorthims is that the compiler known more about what we're trying to accomplish and can use that for better optimizations.  `std::for_each` tells the compiler that we're doing this entirely for the side effects, so there's no advantage there over a plain for loop
2. Unlike plain for-loops, `for_each` does have the advantage that it can be run in parallel.  However, we both know what a nightmare it can be to try and bin a histogram in parallel, so this loop will pretty much always be sequential.
3. The other big advantage to algorithms comes from refactorings.  It's easy to turn a `std::copy` into a `std::transform` if your requirements change down the line (all it takes is an extra lambda).  However, since `hist.bin` is such a pain to refactor, that's not likely to happen.  Even if we did find a better solution, it would mot likely no wind up sharing any code with the `for_each` solution.  For example, if we found a good way to do the binning in parallel, we would want it everywhere in the code, not just here.  That would mean that it would likely become part of the Data3D code and the actual result would look something closer to

```cpp
hist.bin_values(data3D.values(),
    [voxelVolume](auto v){return v/voxelVolume;}
);
```
which shares essentially zero code with the `for_each` call.

I'm sorry that this turned into an essay and I'm happy to listen to any arguments you might have for keeping this in a `for_each` call.  I just felt that you deserve some justification when I suddenly pull out "just use a for loop".

tests/modules/voxelDensity.cpp
```diff
@@ -0,0 +1,229 @@
...
+    HeliumBox box(systemTest.coreData());
+
+    box.setProcessing(VoxelDensityModule::TargetPropertyType::Mass, 8);
+    systemTest.dissolve().iterate(20);
+    const auto &data8Bin =
+        systemTest.dissolve().processingModuleData().search<const Data1D>("VoxelDensity//Data1D//Mass/A^3")->get();
+    ASSERT_TRUE(data8Bin.nNonZeroValues().size() == 1);
+    auto max8Bin = data8Bin.maxValueAt();
+    ASSERT_NEAR(max8Bin.first, MASS_HELIUM);
+    ASSERT_NEAR(max8Bin.second, std::pow(8, 3));
```
rprospero: This looks like a nice test setup, but I'll admit that I'm not quite sure what it's testing.  Some extra comments would be helpful in here.

The best understanding that I had was that you're populating each voxel with a single Helium atom and then confirm that it's reading out the correct unit of value from the voxel density.  What I'm not quite understanding is why the maximum value is always supposed to be in the final voxel?

rprospero (commented):
Some suggestions and questions.  I think that we're getting closer.

src/math/data1D.h
RobBuchananCompPhys: Thanks!

src/math/data1D.cpp
RobBuchananCompPhys: Sounds ideal

tests/modules/voxelDensity.cpp
RobBuchananCompPhys: It's testing that different partitioning schemes applied to the unit cell (i.e. diving axes into 8, 4, 2, 1 bins and therefore 8^3, 4^3, 2^3, 1 voxels) result in the same value being binned - in this case, since the system is always 8x8x8 evenly spaced Helium atoms, we expect to always and only be hitting the Mass(He) / Z(He) / SLD(He) per cubic angstrom bin. The maximum value is in the `second` member of the pair returned from `maxValueAt`, and since no other values are being binned (we hope), this will be the maximum value in the series.

src/modules/voxelDensity/process.cpp
RobBuchananCompPhys: No worries at all, I really welcome this level of feedback! Its good to get a long-term take on how the code might evolve, and I'm mindful to take this into consideration (as you alluded to, the histogram parallelisation hasn't been easy). I'm happy to make this change.

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,109 @@
...
+                // Atomic number
+                value = atomicNumber;
+                break;
+            case TargetPropertyType::ScatteringLengthDensity:
+                // Bound coherent natural isotope scattering length density
+                value = scatteringLengthDensity;
+                break;
+            default:
+                throw(std::runtime_error(fmt::format("'{}' not a valid property.\n", static_cast<int>(targetProperty_))));
+        }
```
trisyoungs: My feeling is that having a trio of unary operators tailored to the specific quantities would make this cleaner and more efficient.

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+        data3D.addToPoint(toIndex(x), x, toIndex(y), y, toIndex(z), z, value);
+    };
+
+    dissolve::for_each(std::execution::seq, atoms.begin(), atoms.end(), unaryOp);
+
+    // Calculate voxel density histogram, normalising bin values by voxel volume (property/cubic angstrom)
+    auto &hist = processingData.realise<Histogram1D>(
+        fmt::format("Histogram1D//{}/A^3", targetPropertyTypes().keywordByIndex(static_cast<int>(targetProperty_))), name(),
+        GenericItem::InRestartFileFlag);
```
trisyoungs: When naming the output data we shouldn't worry about appending the name of the property being calculated since the module explicitly deals with one and one only.

src/modules/voxelDensity/voxelDensity.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+    keywords_.addTarget<ConfigurationKeyword>("Configuration", "Set target configuration for the module", targetConfiguration_)
+        ->setEditSignals({KeywordBase::ClearModuleData, KeywordBase::RecreateRenderables});
+
+    keywords_.setOrganisation("Options", "Species");
+    keywords_.add<SpeciesVectorKeyword>("RestrictToSpecies", "Restrict the calculation to the specified Species",
+                                        restrictToSpecies_);
+
+    keywords_.setOrganisation("Options", "Analysis", "Set the voxel density and property for analysis");
+    keywords_.add<IntegerKeyword>("NumberOfPoints", "Number of points over which analysis is performed", numPoints_, 1);
```
trisyoungs: Feels like there's a better name to be had than "NumberOfPoints".  Perhaps "NVoxelsPerSide" or just "NVoxels"?

RobBuchananCompPhys: `NVoxels` sounds succinct and to the point!

src/modules/voxelDensity/process.cpp
RobBuchananCompPhys: Is there somewhere else we can officialise the units being used (i.e. the histogram bins will be in per cubic angstrom), perhaps I can add that to the module keyword description, and I guess the gui part will reflect that too

Issue #1952, opened 2024-06-27
[Epic] Phantom Atoms


trisyoungs:
### Focus
Addresses the full implementation, testing, and usage of phantom atoms (#1900 / #1953).

### Tasks
- [ ] #1954

PR #1953, opened 2024-06-27
feat: Phantom atoms

### LINKED ISSUE #1900: Artificial Atoms

trisyoungs:
This PR completes #1900 although the final product is to introduce "phantom" rather than "artificial" atoms.

Briefy, a phantom atom is a normal `SpeciesAtom` and is part of a `Species`, connected to other `SpeciesAtom`s, and is evolved by MC / MD in the standard way.  However, it makes no contribution to correlation functions or weights as calculated by the various correlation modules.  Thus, they can affect the behaviour of a species / system but are invisible to structural probes.

Closes #1900.
Follow-on issue #1952 captures further work.


src/classes/atomTypeMix.cpp
```diff
@@ -194,19 +183,19 @@ std::pair<int, int> AtomTypeMix::indexOf(const std::shared_ptr<AtomType> &at1, c
             if (index == -1)
                 index = count;
             else
-                return {count, index};
+                return std::pair<int, int>{count, index};
```
rprospero:
```suggestion
                return {{count, index}};
```
Learned this new trick today

src/classes/atomTypeMix.cpp
```diff
@@ -194,19 +183,19 @@ std::pair<int, int> AtomTypeMix::indexOf(const std::shared_ptr<AtomType> &at1, c
...
         }
         if (atd.atomType() == at2)
         {
             if (index == -1)
                 index = count;
             else
-                return {index, count};
+                return std::pair<int, int>{index, count};
```
rprospero:
```suggestion
                return {{index, count}};
```

src/classes/configuration_contents.cpp
```diff
@@ -233,19 +230,29 @@ Atom &Configuration::addAtom(const SpeciesAtom *sourceAtom, const std::shared_pt
...
 
-// Return number of Atoms in Configuration
-int Configuration::nAtoms() const { return atoms_.size(); }
+// Return the number of atoms in the configuration (or only those with the specified presence)
+int Configuration::nAtoms(SpeciesAtom::Presence withPresence) const
+{
+    return withPresence == SpeciesAtom::Presence::Any
+               ? atoms_.size()
+               : std::count_if(atoms_.begin(), atoms_.end(),
+                               [withPresence](const auto &i) { return i.isPresence(withPresence); });
```
rprospero:
```suggestion
    if (withPresence == SpeciesAtom::Presence::Any) {
        return atoms_.size();
    }
    return std::count_if(atoms_.begin(), atoms_.end(),
                               [withPresence](const auto &i) { return i.isPresence(withPresence); });
```

src/classes/neutronWeights.cpp
```diff
@@ -158,71 +157,45 @@ void NeutronWeights::calculateWeightingMatrices()
...
+                                                 : Sears91::boundCoherent(tope->atomTypeIsotope(atd2.atomType()));
+
+                                        // Convert from fm to barns
+                                        bi *= 0.1;
+                                        bj *= 0.1;
+
+                                        intramolecularWeights_[{typeI, typeJ}] += weight * bi * bj;
+                                        intraNorm[{typeI, typeJ}] += weight;
+                                        globalFlag[{typeI, typeJ}] = true;
+                                    });
```
rprospero, line 198: There's a way to do all of this in parallel with the PairIterator and std::transform_reduce, but it's probably not worth the trouble unless this is a serious bottleneck.

src/classes/neutronWeights.cpp
```diff
@@ -259,10 +232,16 @@ void NeutronWeights::createFromIsotopologues(const std::vector<std::shared_ptr<A
...
+            // Loop over Atoms in the Species, searching for the AtomType/Isotope entry in the isotopes list of the
+            // Isotopologue
             for (const auto &i : topes.species()->atoms())
+            {
+                if (!i.isPresence(SpeciesAtom::Presence::Physical))
+                    continue;
+
                 atomTypes_.addIsotope(i.atomType(), top->atomTypeIsotope(i.atomType()),
                                       isoWeight.weight() * topes.speciesPopulation());
+            }
```
rprospero:
```suggestion
            for (const auto &i : topes.species()->atoms())
                if (i.isPresence(SpeciesAtom::Presence::Physical))
                    atomTypes_.addIsotope(i.atomType(), top->atomTypeIsotope(i.atomType()),
                                      isoWeight.weight() * topes.speciesPopulation());
```

Possible rewrite.  Pick whichever you prefer

src/classes/speciesAtom.h
```diff
@@ -38,6 +38,15 @@ class SpeciesAtom : public Serialisable<CoreData &>
     /*
      * Properties
      */
+    public:
+    // Presence
+    enum class Presence
+    {
+        Phantom = -1,
+        Physical,
+        Any
```
rprospero, line 47: A possible suggestion here would be to assign `Any` the value of zero.  The could change a bunch of conditionals from

```
  if (presence == SpeciesAtom::Presence::Any)
```

to merely being

```
   if (presence)
```

src/classes/species_atomic.cpp
```diff
@@ -113,8 +113,14 @@ void Species::removeAtoms(std::vector<int> indices)
...
-// Return the number of Atoms in the Species
-int Species::nAtoms() const { return atoms_.size(); }
+// Return the number of atoms in the species (or only those with the specified presence)
+int Species::nAtoms(SpeciesAtom::Presence withPresence) const
+{
+    return withPresence == SpeciesAtom::Presence::Any
+               ? atoms_.size()
+               : std::count_if(atoms_.begin(), atoms_.end(),
+                               [withPresence](const auto &i) { return i.isPresence(withPresence); });
+}
```
rprospero, line 123: Se above statements about if statements and enum values.

rprospero (approved):
Some stylic suggestions, but all are optional

src/classes/atomTypeMix.cpp
trisyoungs: At last!  This has always bothered me.

src/classes/neutronWeights.cpp
trisyoungs: Shouldn't be as it's called once per module per iteration, so I'm inclined to leave it as-is.

src/classes/neutronWeights.cpp
trisyoungs: Not sure why I had written it that way before!

src/classes/speciesAtom.h
trisyoungs: For clarity (and my own sanity when I look back at this in six months and don't understand what I've written) I think we should keep the explicit comparison and enum as they are.

PR #1958, opened 2024-07-09
build: Naively remove custom Qt

rprospero:
This PR will remove our custom Qt and just use the default one.  This should simplify our builds and development environment.


trisyoungs (approved):
Can't see anything amiss here.
:+1:

PR #1959, opened 2024-07-10
refactor: Common data table model

### LINKED ISSUE #1902: Data Model Interface for Qt
There are many reasons to use models for managing the viewing and editing of our data in the Dissolve GUI, however there is something of a disconnect in the way we manage data between the CLI "core" code and the Qt GUI.

Can we implement a custom "model" which sits alongside, e.g., a `std::vector<CustomData>` and provides "row" addition and deletion, read/write access to data, much as a QtAbstractItemModel does, but without any of the Qt? A companion Qt class in the GUI could then be designed to directly use this non-Qt model to achieve the desired result in the GUI.  Why bother doing this? I'm thinking this would allow:
- Parts of the core "CLI" code to edit data and automatically inform Qt (via, e.g., our GUI registering lambda callbacks to "signals" in our custom model) and hence update the GUI without any manual calling of update functions
- Implementing model/views for stuff in the GUI becomes coherent and simplified, and provides the reverse notification stream to the CLI core.

trisyoungs:
Implements a standard interface between `QAbstractTableModel` and any of our custom classes requiring such a display in the UI.

### TODO
- [x] Implement `insertRows()`
- [x] Implement `removeRows()`
- [x] Flags for items
- [ ] Icons for items
- [x] Test notification of Qt models when change occurs in underlying data programmatically
- [ ] `ExpressionVariableVectorKeyword` needs to update related nodes when a parameter is removed. or added, or renamed.
- [x] ~Filtering~ Standard Qt proxy model when required
- [ ] Need to notify main GUI when data changed

#### Tables with Filter Proxies
- [ ] `AtomTypeModel`
- [ ] `BraggReflectionModel`
- [ ] `SitesModel`
- [ ] `SpeciesSiteModel`

#### Tables
- [ ] `DataManagerSimulationModel`
- [ ] `MasterTermModel` (inherited)
- [ ] `ModuleModel`
- [ ] `RangeVectorModel`
- [ ] `SpeciesAtomModel`
- [ ] `SpeciesBondModel`
- [ ] `SpeciesAngleModel`
- [ ] `SpeciesTorsionModel`
- [ ] `SpeciesImproperModel`
- [ ] `WeightedModuleModel`
- [ ] `XMLBondModel`
- [ ] `XMLAngleModel`
- [ ] `XMLTorsionModel`
- [ ] `XMLImproperModel`

#### `QAbstractListModel`s
- [ ] `ConfigurationModel`
- [ ] `EnumOptionsModel`
- [x] `ExternalPotentialModel`
- [ ] `SpeciesModel`

Closes #1902.


trisyoungs:
@rprospero As mentioned earlier some quite significant evolution has happened here. There's a lot going on, but the summary is captured below. Thoughts, comments, and ideas for alternative approaches very welcome.
- I have namespaced everything within `DataModel` - this makes things quite neat, but is slightly against how the rest of our code is structured.
- A class which is to be modeled now derives from `DataModel::Modelable` - this simple template class defines some useful `using`s for subsequent definition of setter and getter functions, and a single static function `modelableProperties()` which is called by the model itself to determine what properties are available.
- Rather than monolithic setter and getter functions acting on all available properties I elected to move to per-property definition. This removes the need for any enumeration of the type and simplifies things quite a bit. The necessary data setters and getters are defined locally in the class source file by implementing `DataModel::Modelable<T>::modelableProperties()` , next to other related functions.
- I have moved towards a `VectorModelable` class which owns the `std::vector` it is acting as a model for, rather than taking a reference to an object elsewhere. The simple reason for this is to ensure that we can track programmatic mutations of the underlying `std::vector` data, at least to a level that enables the GUI to reflect the changes automatically without impinging too much on us actually writing clear code.
- The template for `VectorModelable` currently takes two classes - the type of the underlying data (e.g. `ExpressionVariable`) and its storage type in the vector (e.g. `std::shared_ptr<ExpressionVariable>`). This is necessary to achieve transparent access via the getters and setters (which just operate on a pointer to an object of the underlying data since the actual mechanism of storage is irrelevant). This isn't terribly neat, and I'm sure there's a clever C++-23 way of achieving the same with an "auto-deduced template mangle", but I'm not clever enough to figure out how.
- Since we will have both plain and pointer types stored in our `std::vector`, the creation function must be manually defined always and cannot be assumed, especially since we may be storing pointers to base class types that are pure virtual on their own.
- Const-access to the underlying vector data is always available, but when programmatic editing of the underlying vector is required we have a function `DataModel::Table::mutableData()` which returns a `VectorMutator` object alongside the non-const vector. The purpose of `VectorMutator` is simply to signal `beginResetModel()` on construction and `endResetModel()` when it goes out of scope. This is the probably going to be the biggest intrusion on the codebase, but feels like a small price to pay for a general model which allows the GUI to react to programmatic changes automatically.

src/templates/dataModelItem.h
```diff
@@ -0,0 +1,93 @@
...
+static std::string propertyAsString(const PropertyValue &value)
+{
+    if (std::holds_alternative<int>(value))
+        return fmt::format("{}", std::get<int>(value));
+    else if (std::holds_alternative<double>(value))
+        return fmt::format("{}", std::get<double>(value));
+    else if (std::holds_alternative<std::string_view>(value))
+        return fmt::format("{}", std::get<std::string_view>(value));
+    else
+        return std::get<std::string>(value);
```
rprospero:
```suggestion
    return std::visit([](auto &arg) {return fmt::format("{}", arg);}, value);
```

Using `std::visit` saves us on some boilerplate and also helps if we ever extend the list of types.

src/gui/dataModelTableInterface.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    // Get the specified data property
+    auto property = dataModel_.getProperty(index.row(), index.column());
+
+    // Construct a QVariant from the contents of our std::variant
+    return std::visit(
+        DataModel::PropertyVisitor{[](int arg) { return QVariant(arg); }, [](double arg) { return QVariant(arg); },
+                                   [](std::string_view arg) { return QVariant(QString::fromStdString(std::string(arg))); },
+                                   [](std::string &arg) { return QVariant(QString::fromStdString(arg)); }},
+        property);
```
rprospero:
```suggestion
    return std::visit(
        [](auto arg) {
            using T = std::decay_t<decltype(arg)>;
            if constexpr (std::is_same_v<T, std::string_view>)
                return QVariant(QString::fromStdString(std::string(arg))); 
            else if constexpr (std::is_same_v<T, std::string>)
                return QVariant(QString::fromStdString(arg)); 
            else
                return QVariant(arg);
        },
        property);
```

I think that we should be able to do this and get rid of the `PropertyVisitor` class, since this is the only place where it is used.

src/templates/dataModelItem.h
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Helper type for PropertyValue visitor
+template <class... Ts> struct PropertyVisitor : Ts...
+{
+    using Ts::operator()...;
+};
+// Explicit deduction guide for PropertyValue visitor
+template <class... Ts> PropertyVisitor(Ts...) -> PropertyVisitor<Ts...>;
```
rprospero: I've put in a comment above trying to work around needing this class.  I'll admit that it scares me a little.

src/templates/dataModelItem.h
```diff
@@ -0,0 +1,93 @@
...
+
+// Modelable - Provides declarations for use by a class defining its properties
+template <class D> class Modelable
+{
+    public:
+    using ModelableGetter = std::function<PropertyValue(const D *)>;
+    using ModelableSetter = std::function<bool(D *, PropertyValue)>;
+    using ModelableProperty = std::tuple<std::string, ItemProperty::PropertyType, Flags<ItemProperty::PropertyFlag>,
+                                         ModelableGetter, ModelableSetter>;
```
rprospero:
```suggestion
    using ModelableProperty = std::tuple<ItemProperty, ModelableGetter, ModelableSetter>;
```

It feels like the ItemProperty class contains the first three parameters of this tuple, so I thought it might be more convenient to condense them.

src/templates/dataModelItem.h
```diff
@@ -0,0 +1,93 @@
...
+};
+// Explicit deduction guide for PropertyValue visitor
+template <class... Ts> PropertyVisitor(Ts...) -> PropertyVisitor<Ts...>;
+
+// Modelable - Provides declarations for use by a class defining its properties
+template <class D> class Modelable
+{
+    public:
+    using ModelableGetter = std::function<PropertyValue(const D *)>;
+    using ModelableSetter = std::function<bool(D *, PropertyValue)>;
```
rprospero, line 66: So there's a trick with profunctor optics where we can combine these combine the getter and setter into a single type.  It would save some boilerplate and all of that.  The catch is it would have the ugliest type signature in the world.  Roughly speaking

```cpp
using ModeableLens = std::function<
  ConstOrIdent<D, PropertyValue>(
    std::function<
      ConstOrIdent<PropertyValue, PropertyValue>(PropertyValue), 
      const D>)>
```

It also has the distinction that it "works" on const values (setters do not operate in place, but create a new object with the appropriate value set).

Honestly, after needing to check that type signature six times to make sure I got it right, I don't think it's worth the offer, but I'd still be remiss not to mention it.

src/templates/dataModelBase.h
```diff
@@ -0,0 +1,103 @@
...
+    bool isPropertyFlagSet(int propertyIndex, ItemProperty::PropertyFlag flag)
+    {
+        return itemProperties_[propertyIndex].flags().isSet(flag);
+    }
+
+    public:
+    // Return number of data items (i.e. rows) for the specified index
+    virtual int nDataItems() const = 0;
+    // Return number of properties (i.e. columns) for the specified index
+    virtual int nProperties() const = 0;
```
rprospero: Wouldn't this always be `iterProperties_.size()`?  I know that we used it for the `DataTableInterface` class, but it seems like it would always need to be that way.

src/templates/dataModelVectorModelable.h
```diff
@@ -0,0 +1,259 @@
...
+     */
+    private:
+    // Target data for the model
+    std::vector<DataItem> data_;
+
+    public:
+    // Return const data
+    const std::vector<DataItem> &data() const { return data_; }
+    // Return mutable data
+    std::pair<VectorMutator<DataItem>, std::vector<DataItem> &> mutableData() { return {{data_, *this}, data_}; };
```
rprospero: While I really like this set-up, there's a minor issue.  Specifically, I can write a function

```cpp
std::vector<DataItem>& invisibleAccess(VectorModelable *source) {
  auto [mutator, data] = source->mutableData();
  // 500 lines of distracting code
  return data;
}
```

This function returns a writeable reference to the vector, but none of the changes that I make will be registered by the mutator (since it's already registered the changes as finished).

On one level, I don't think it's possible to escape this without C++ implementing linear types.  However, if we changed the `mutableData` function to merely return `VectorMutator<DataItem>`, instead of the pair, and we made the `data` member of `VectorMutator` public, then the function becomes

```cpp
std::vector<DataItem>& invisibleAccess(VectorModelable *source) {
  auto mutator = source->mutableData();
  // 500 lines of distracting code
  return mutator.data;
}
```

which feels a bit more likely to get caught in code review.

src/templates/dataModelVectorModelable.h
```diff
@@ -0,0 +1,259 @@
...
+    bool empty() const { return data_.empty(); }
+
+    /*
+     * Setters / Getters
+     */
+    private:
+    // Map of named properties to data getters
+    std::map<std::string, PropertyGetFunction> getters_;
+    // Map of named properties to data setters
+    std::map<std::string, PropertySetFunction> setters_;
```
rprospero, line 78: This is fine as long as we leave the property subsystem to only interact with the GUI.  However, if some future wiseacre decides to use it for meta programming inside of a hot loop, it's going to be a pain with doing a full string comparison each time the property is accessed.  It might be better to simply make this a vector and base the index on a cast from an `enum` that would have contained the name.

On the other hand, it's a lot of work for a performance issue that hopefully should never occur.

src/templates/dataModelVectorModelable.h
```diff
@@ -0,0 +1,259 @@
...
+    // Set property
+    bool setProperty(int dataIndex, int propertyIndex, const PropertyValue &newValue) final
+    {
+        // Check index validity
+        if (!isIndexValid(dataIndex, propertyIndex))
+            return false;
+
+        if (itemProperties_[propertyIndex].flags().isSet(ItemProperty::PropertyFlag::ReadOnly))
+        {
+            fmt::print("Refusing to set data '{}' since it is read-only.\n", itemProperties_[propertyIndex].name());
```
rprospero:
```suggestion
            Messenger::error("Refusing to set data '{}' since it is read-only.\n", itemProperties_[propertyIndex].name());
```

rprospero (commented):
Overall, I like this setup and it feels clean.  I've made some comments about some places where I might have handled the types differently, but there may be corner cases that I haven't considered.

src/templates/dataModelItem.h
trisyoungs: Annoyed that I didn't see this before!

src/gui/dataModelTableInterface.cpp
trisyoungs: And I am very happy to do so since this was voodoo code copy/pasted from cppreference.com.  I cannot pretend that I fully understood the mechanics of it anyway!

src/templates/dataModelItem.h
trisyoungs: It is gone!

src/templates/dataModelItem.h
trisyoungs: :+1:

src/templates/dataModelItem.h
trisyoungs: OK, I believe I get the mechanism here (after doing a bit of reading). Powerful stuff for sure, but I would agree that our code prettiness would suffer somewhat. I did want to try this out though so blindly copied your suggestion above and had a go, but couldn't convince the compiler to go for it...

src/templates/dataModelBase.h
trisyoungs: In the present usages you are absolutely right. This is another leftover from me trying to make `DataModelBase` the base class for _all_ of our future custom models - Tree, List, and Table.  However, the longer this journey goes on the more inclined I am to completely separate the Tree from the List/Table for the sake of maintainability as well as sanity.

src/templates/dataModelVectorModelable.h
trisyoungs: Agreed. Limiting potential misuse is good, plus I really don't like using `std::pair` anyway!

src/templates/dataModelVectorModelable.h
trisyoungs: The implicit intention with all of this is absolutely that only slow, user-based interaction through the GUI will ever call the setters and getters, so performance should not be a concern. The `enum`-based indexing is way faster for sure which is nice in general, but writing the underlying `Modelable::modelableProperties()` function then of course requires one to define the `enum` to be used as well as the name string.  This felt like overkill in the first place so I refactored that and reasoned that string-based access was prettier to define, and the GUI-only performance hit was an acceptable compromise.

PR #1960, opened 2024-07-11
build: Remove last vestiage of our custom Qt for IDAaaS

rprospero:
I've confirmed that stock Qt will work on IDAaaS, so we can remove the framework that I left in place to add it back.


trisyoungs (approved):
:+1:

PR #1961, opened 2024-07-26
build: Fix single-page example generation

trisyoungs:
Starting off as another attempt to get us PDFs of our examples (`pandoc` now just crashes, and an alternative in the form of `md-to-pdf` via npm doesn't work either) this PR just fixes the generation of single-page examples. Users can print those to PDF themselves if required.


.github/workflows/website/action.yml
```diff
@@ -119,7 +119,15 @@ runs:
       do
         echo "Creating pdf for example \"${example}\"..."
         cd ${example}/single/
-        pandoc --pdf-engine=xelatex -V 'mainfont:DejaVuSerif.ttf' -V 'sansfont:DejaVuSans.ttf' -V 'monofont:DejaVuSansMono.ttf' index.html -o ../${example}.pdf
+
+        # "Fix" image links
+        sed "s:/img:../../../static/img:g" index.html > index_pdf.html
```
rprospero, line 124: My mind went on quite a trip trying to parse this sed statement before it finally clicked that you were using the colon as the delimited and not the traditional slashes.  Which was exactly the right call

rprospero (approved):

PR #1963, opened 2024-08-05
build: Add QuickPlot into QML GUI

rprospero:
PR is a minimal proof of concept adding the plotting library from QuickPlot into the Dissolve QML GUI.  Using the Qt Plugin system, we can import both the QML and C++ components of the library.

We load QuickPlot through a git submodule, which I suspect will break a load of the CI.  There's also several changes that need to be made before this could be considered for merge.  However, I wanted to put the code out in front of everyone to ensure that people are happy with how this works.

## Necessary Changes

- [x] Ensure that the new GUI builds on all platforms
- [x] Prevent Quickplot demo executable from being built
- [x] Pick a better qrc folder for QuickPlot folder
- [x] Use `qt_add_qml_module` for dissolve gui qml


src/gui-qml/qml/DissolveMain.qml
```diff
@@ -1,9 +1,16 @@
 import QtQuick
 import QtQuick.Controls
 import QtQuick.Layouts
+import QtQuick3D
+import QtQuick3D.Helpers
+import QtQuick.Controls
+import ProjectDissolve
+import "../../ProjectDissolve"
```
RobBuchananCompPhys, line 8: What does `"../../ProjectDissolve"` reference? The submodule?

RobBuchananCompPhys: Ok, after tracking down the actual library `CMakeLists.txt`, this ref makes more sense

.github/workflows/build/linux/action.yml
```diff
@@ -45,10 +45,26 @@ runs:
...
+      cp -r QuickPlot QuickPlot.bak
+      rm -rf QuickPlot.bak/.git
+      git rm -r QuickPlot
+      git commit -m "Kill submodule"
+
+      mv QuickPlot.bak QuickPlot
+      git add QuickPlot
+      git commit -m "Resurrect submodule"
+
+      nix --version
```
trisyoungs: Is this testing, or a workaround? If the latter, may we a comment as to why it is necessary?

.github/workflows/build/osx/action.yml
```diff
@@ -44,11 +44,11 @@ runs:
       path: ${{ runner.temp }}/qt
 
   - name: Install Qt
-    if: ${{ steps.cache-qt.outputs.cache-hit != 'true' }}
+    # if: ${{ steps.cache-qt.outputs.cache-hit != 'true' }}
```
trisyoungs: Fantastic.  If we no longer need Qt caches for Windows and OSX then we can ditch the "Retrieve Qt Cache"  step prior to this one, although I understand if you want to leave this for a follow-up PR.

trisyoungs (approved):
Two comments to address, but otherwise looks good.

.github/workflows/build/osx/action.yml
rprospero: Sadly, this *was* just a testing issue that leaked through.  I had to temporarily disable the cache because I was changing the Qt install and it kept loading the cached version.  It will build without the cache, but we might get into issues with too many downloads, so it's better to keep it cached.

.github/workflows/build/linux/action.yml
rprospero: This is a work around for something that appears to be a very specific nix bug.  I've added comments to the file explaining the nature of the problem.

PR #1964, opened 2024-08-09
feat: voxel density module NEW

### LINKED ISSUE #1915: Voxel Density

RobBuchananCompPhys:
Closes #1915


src/gui/icons/modules/VoxelDensity.svg
trisyoungs, line 1: I like it!

src/gui/icons/modules/VoxelDensityt.svg
trisyoungs, line 1: Random other icon?

src/math/data1D.cpp
```diff
@@ -215,6 +227,18 @@ double Data1D::maxValue() const
     return *std::max_element(values_.begin(), values_.end());
 }
 
+// Return index and corresponding maximum value over all data points
+std::pair<size_t, double> Data1D::maxValueAt()
+{
+    if (values_.empty())
+        return {0.0, 0.0};
```
trisyoungs:
```suggestion
        return {0, 0.0};
```

src/math/data1D.cpp
```diff
@@ -197,6 +197,18 @@ const std::vector<double> &Data1D::values() const { return values_; }
...
+{
+    std::vector<std::tuple<double, double>> pairs;
+
+    auto xys = zip(x_, values_);
+
+    std::copy_if(xys.begin(), xys.end(), std::back_inserter(pairs), [](const auto &val) { return std::get<1>(val) > 0; });
+
+    return pairs;
+}
+
```
trisyoungs: I would be tempted to follow Dissolve's unwritten rule of keeping the `Data*D` classes as simple as possible in terms of functionality, and adding stuff like this to a functional namespace as we do for things like `Filter`, `Averaging` - `Limits` would be a good choice here.

src/math/data1D.cpp
```diff
@@ -215,6 +227,18 @@ double Data1D::maxValue() const
     return *std::max_element(values_.begin(), values_.end());
 }
 
+// Return index and corresponding maximum value over all data points
+std::pair<size_t, double> Data1D::maxValueAt()
```
trisyoungs: Same comment as above.

src/math/data3D.cpp
```diff
@@ -120,6 +123,32 @@ void Data3D::zero()
...
 
+// Add new data point
+void Data3D::addPoint(double x, double y, double z, double value)
+{
+    x_.push_back(x), y_.push_back(y), z_.push_back(z);
+
+    values_[std::tuple{x, y, z}] = value;
+
+    ++version_;
+}
```
trisyoungs: This function feels extremely dangerous!  The `x_`, `y_`, and `z_` arrays are being extended, but the `values_` array is already the correct size? What if non-contiguous values of x, y, and/or z are passed?
Is this functional actually used anywhere?

src/math/data3D.cpp
```diff
@@ -120,6 +123,32 @@ void Data3D::zero()
...
+
+// Add data to point
+void Data3D::addToPoint(int x, double xDelta, int y, double yDelta, int z, double zDelta, double value)
+{
+    if ((x_.size() < x) || (y_.size() < y) || (z_.size() < z))
+    {
+        addPoint(xDelta, yDelta, zDelta, value);
+        return;
+    }
```
trisyoungs: Again, extending `Data3D` in this way feels very risky!

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,104 @@
...
+    auto t = std::make_tuple((int)std::floor(coords.x * nAxisVoxels_), (int)std::floor(coords.y * nAxisVoxels_),
+                             (int)std::floor(coords.z * nAxisVoxels_));
+    array[t] += value;
+}
+
+Vec3<double> VoxelDensityModule::foldedCoordinates(Atom &atom, const Box *unitCell)
+{
+    atom.set(unitCell->foldFrac(atom.r()));
+    auto x = atom.x(), y = atom.y(), z = atom.z();
+    return Vec3<double>(x, y, z);
```
trisyoungs: This is modifying the actual coordinates of the `Atom` and, while it's only a fold operation, we should be as read-only as possible on a `Configuration`, especially for analysis-type modules which aren't expected to modify contents.

If you change the function signature to just take the non-const `Vec3<double> r` of the atom coordinates, then you can do:
```suggestion
    return unitCell->foldFrac(r);
```
You may even wish to remove the function altogether and just do it in place since it's only one line.

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,104 @@
...
+            if (included)
+            {
+                targetConfiguration_->removeMolecules(sp);
+            }
+        }
+    }
+
+    auto unitCell = targetConfiguration_->box();
+    auto voxelVolume = unitCell->volume() / std::pow(nAxisVoxels_, 3);
+    auto atoms = targetConfiguration_->atoms();
```
trisyoungs: Here you're taking a copy of the entire atoms array which is going to be expensive and might cause issues with other object relationships.
```suggestion
    const auto &atoms = targetConfiguration_->atoms();
```

src/modules/voxelDensity/voxelDensity.h
```diff
@@ -0,0 +1,68 @@
...
+#include "base/enumOptions.h"
+#include "classes/atom.h"
+#include "classes/box.h"
+#include "data/atomicMasses.h"
+#include "data/isotopes.h"
+#include "io/export/data1D.h"
+#include "module/module.h"
+#include "templates/array3D.h"
+
+class Configuration;
```
trisyoungs, line 17:
```suggestion
// Forward Declarations
class Configuration;
```

src/modules/voxelDensity/voxelDensity.cpp
```diff
@@ -0,0 +1,44 @@
...
+    keywords_.setOrganisation("Options", "Species");
+    keywords_.add<SpeciesVectorKeyword>("RestrictToSpecies", "Restrict the calculation to the specified Species",
+                                        restrictToSpecies_);
+
+    keywords_.setOrganisation("Options", "Ranges");
+    keywords_.add<Vec3DoubleKeyword>("BinRange", "Range (min, max, delta) of voxel axis", binRange_,
+                                     Vec3<double>(0.0, 0.0, 1.0e-5), std::nullopt, Vec3Labels::MinMaxDeltaLabels);
+
+    keywords_.setOrganisation("Options", "Analysis", "Set the voxel density and property for analysis");
+    keywords_.add<IntegerKeyword>("NAxisVoxels", "Number of points per axis over which analysis is performed", nAxisVoxels_, 1);
```
trisyoungs: I think here we would make things easier and more intuitive for the user if we took a rough requested voxel length, rather than the number of points along each cell axis.  A default value of `3.0` would give you a requested voxel cube of 27 cubic Angstroms, not too far off the volume of a single water molecule. Of course, the requested value wont always multiply as a whole integer along an axis, so we will need to adjust the requested size based on the actual axis lengths.  This also solves the problem of calculating the voxel density on a non-cubic box shape which may have very different axis lengths - as written, we would get quite non-uniform voxels.

src/templates/array3D.h
```diff
@@ -86,9 +86,19 @@ template <class A> class Array3D
             resize(nX, nY, nZ);
     }
 
+    // Return const reference to values
+    const std::vector<A> &values() const { return array_; }
+
+    // Return modifiable reference to values
+    std::vector<A> &values() { return array_; }
```
trisyoungs: Tens to write these without a separate comment for the overload
```suggestion
    // Return reference to values
    std::vector<A> &values() { return array_; }
    const std::vector<A> &values() const { return array_; }
```

src/templates/array3D.h
```diff
@@ -86,9 +86,19 @@ template <class A> class Array3D
...
+
     // Return specified element as modifiable reference
     A &operator[](std::tuple<int, int, int> index)
     {
+        // Check if array is a single point (sides of length 1)
+        if (nX_ == 1 && nY_ == 1 && nZ_ == 1)
+            return array_[0];
+
```
trisyoungs: Just wondering - would we also have issues here if just one dimension was `1`? A very long stick, or a plane?  Perhaps its worth capturing this in a unit test specifically for the `Array3D`.

tests/modules/voxelDensity.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+/*
+TEST_F(VoxelDensityModuleTest, Water)
+{
+    ASSERT_NO_THROW_VERBOSE(systemTest.setUp("dissolve/input/voxelDensity-water.txt"));
+    auto module = systemTest.getModule<VoxelDensityModule>("AtomicMass");
+    ASSERT_TRUE(module->keywords().getInt("NAxisVoxels") == 1);
+}
+*/
+} // namespace UnitTest
```
trisyoungs: Will we try to implement this?

RobBuchananCompPhys: Yes, we will!

src/math/data1D.cpp
RobBuchananCompPhys: Agreed, I'll address this

src/math/data3D.cpp
RobBuchananCompPhys: This is not actually being used anywhere. I see your point about safety, and I've been uncomfotable about this bit of code since writing it but didn't get round to removing it yet - I think it should go.

src/gui/icons/modules/VoxelDensityt.svg
RobBuchananCompPhys: shouldn't be there, I think it is a copy

src/templates/array3D.h
RobBuchananCompPhys: I'll look at moving this check outside of the `Array3D` class - think all we need to do is check that the number of voxels is 1, and in that case we get the `array[0]`, else we use the `operator[]` as normal.

src/math/mathFunc.cpp
```diff
@@ -89,4 +89,20 @@ double sgn(double a, double signOf) { return signOf >= 0.0 ? fabs(a) : -fabs(a);
...
+// Natural (base-e) logarithm of x
+double log(double x) { return std::log(x); }
+// Base-2 logarithm of n
+int log2(int n) { return int(std::log2(n)); }
+// Base-2 logarithm of x
+double log2(double x) { return std::log2(x); }
+// Base-10 logarithm of n
+int log10(int n) { return int(std::log10(n)); }
+// Base-2 logarithm of x
+double log10(double x) { return std::log10(x); }
```
trisyoungs: Is this something we need to wrap in our own functions? The only benefit is with the pure `int` version, and even then I'm not sure that `DissolveMath::log(n)` is better than `int(std::log(n))`.

src/math/limitsFunc.h
```diff
@@ -0,0 +1,40 @@
...
+// Return index and corresponding maximum value over all data points
+template <typename T> std::pair<std::size_t, T> maxValueAt(std::vector<T> values)
+{
+    if (values.empty())
+        return {0, 0.0};
+
+    std::vector<double>::iterator result = std::max_element(values.begin(), values.end());
+
+    // { index, maxvalue }
+    return {std::distance(values.begin(), result), *result};
```
rprospero:
```suggestion
    return {result - values.begin(), *result};
```

I find the subtraction to be slightly more readable, but I know that sometimes `std::distance` helps clean up a problem.

src/templates/array3D.h
```diff
@@ -86,9 +86,19 @@ template <class A> class Array3D
...
+    std::vector<A> &values() { return array_; }
+
     // Return specified element as modifiable reference
     A &operator[](std::tuple<int, int, int> index)
     {
+        /*
+        // Check if array is a single point (sides of length 1)
+        if (nX_ == 1 && nY_ == 1 && nZ_ == 1)
+            return array_[0];
+        */
```
rprospero: Remove dead code?

src/modules/voxelDensity/voxelDensity.h
```diff
@@ -0,0 +1,77 @@
...
+        Mass,
+        AtomicNumber,
+        ScatteringLengthDensity
+    };
+    // Return enum option info for TargetPropertyType
+    static EnumOptions<VoxelDensityModule::TargetPropertyType> targetPropertyTypes();
+    // Target property for analysis
+    VoxelDensityModule::TargetPropertyType getCurrentProperty() { return targetProperty_; }
+    // Voxel volume (cubic angstroms)
+    const double voxelVolume() { return voxelVolume_; }
```
rprospero:
```suggestion
    double voxelVolume() const { return voxelVolume_; }
```

Did you mean for this to be a const function?  I'm not sure why it would otherwise need to return a `const double`

src/modules/voxelDensity/voxelDensity.h
```diff
@@ -0,0 +1,77 @@
...
+    enum class TargetPropertyType
+    {
+        Mass,
+        AtomicNumber,
+        ScatteringLengthDensity
+    };
+    // Return enum option info for TargetPropertyType
+    static EnumOptions<VoxelDensityModule::TargetPropertyType> targetPropertyTypes();
+    // Target property for analysis
+    VoxelDensityModule::TargetPropertyType getCurrentProperty() { return targetProperty_; }
```
rprospero:
```suggestion
    VoxelDensityModule::TargetPropertyType getCurrentProperty() const { return targetProperty_; }
```

Make this a const function

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+    if (nAxisVoxels_.x == 1 && nAxisVoxels_.y == 1 && nAxisVoxels_.z == 1)
+    {
+        array.values()[0] += value;
+        return;
+    }
+
+    auto t = std::make_tuple((int)std::floor(coords.x * nAxisVoxels_.x), (int)std::floor(coords.y * nAxisVoxels_.y),
+                             (int)std::floor(coords.z * nAxisVoxels_.z));
+    array[t] += value;
```
rprospero: I'm a little concerned why we need the special case for the single element voxel.  If the array accessors doesn't work properly in than instance, are we sure that it works in the other insances.  For instance, if the x and y width are 1, but z is two, does it still give the right answer?

If the tuple access does give the right answer, could we drop the special case?

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,112 @@
...
+            bool included = false;
+
+            for (const auto &spTarget : restrictToSpecies_)
+            {
+                if (sp->name() == spTarget->name())
+                {
+                    included = true;
+                    break;
+                }
+            }
```
rprospero:
```suggestion
            bool included = std::any_oF(restricToToSpecies_.begin(), restrictToSpecies_.end(), [sp](auto &spTarget) {return sp->name() == spTarget->name();});
```

Might be a good place to use `std::any_of`

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+            if (included)
+            {
+                targetConfiguration_->removeMolecules(sp);
+            }
```
rprospero:
```suggestion
            if (included)
                targetConfiguration_->removeMolecules(sp);
```

rprospero (commented):
I have some suggestions and some questions.  Most of the suggestions are fairly trivial, but I'm a bit worried about the questions before we go forward.

src/modules/voxelDensity/gui/voxelDensityWidgetFuncs.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    auto type = module_->getCurrentProperty();
+    switch (type)
+    {
+        case (VoxelDensityModule::TargetPropertyType::Mass):
+            return "Atomic mass unit, u";
+        case (VoxelDensityModule::TargetPropertyType::AtomicNumber):
+            return "Z";
+        case (VoxelDensityModule::TargetPropertyType::ScatteringLengthDensity):
+            return "Scattering length density, cm\\sup{-2}";
```
trisyoungs: Should these x-axis labels by in units of X per voxel or per Angstrom**3?

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,112 @@
...
+    nAxisVoxels_.y = int(round(boxY / voxelSideLength(boxY)));
+    nAxisVoxels_.z = int(round(boxZ / voxelSideLength(boxZ)));
+
+    Messenger::print("Volume of unit cell voxels: {}.\n", voxelVolume_);
+    Messenger::print("Number of voxels along each axis: nX={}, nY={}, nZ={}.\n", nAxisVoxels_.x, nAxisVoxels_.y,
+                     nAxisVoxels_.z);
+
+    // Calculate target property 3d map over unit cell voxels
+    auto array3D = processingData.realise<Array3D<double>>("Array3D", name());
+    array3D.initialise(nAxisVoxels_.x, nAxisVoxels_.y, nAxisVoxels_.z);
```
trisyoungs: Since this `Array3D` is not saved to the restart file there is no need / utility in using the `realise()` function here. Also, we are recreating it every time the module runs, where we could just create it once the first time its used and then zero the values in it on subsequent uses.

src/modules/voxelDensity/voxelDensity.h
trisyoungs: Only a small thing, but we don't define even trivial functions in headers.

tests/data/dissolve/input/voxelDensity-helium-Z.txt
trisyoungs, line 1: These test files could mostly be removed leaving only a single one which we can use in a loop, setting the voxel size and quantity type programmatically in the unit test.

src/modules/voxelDensity/gui/voxelDensityWidgetFuncs.cpp
RobBuchananCompPhys: That's correct I think, thanks for pointing it out!

src/modules/voxelDensity/process.cpp
RobBuchananCompPhys: Agreed

src/math/mathFunc.cpp
RobBuchananCompPhys: In that case, probably not.

src/modules/voxelDensity/process.cpp
RobBuchananCompPhys: This seems to have been a totally unnecessary check - it most likely arose from an oversight on my part where I was using `std::round`, and in the case of a single voxel was rounding up to `0.something * 1 voxel = 1`, and therefore `Array3D` couldn't index it (since there is no `array[1]` to access). Changing to `std::floor` fixed this a while back, and I've removed the check.

src/math/limitsFunc.h
```diff
@@ -0,0 +1,40 @@
...
+    return pairs;
+}
+
+// Return index and corresponding maximum value over all data points
+template <typename T> std::pair<std::size_t, T> maxValueAt(std::vector<T> values)
+{
+    if (values.empty())
+        return {0, 0.0};
+
+    std::vector<double>::iterator result = std::max_element(values.begin(), values.end());
```
trisyoungs:
```suggestion
    auto result = std::max_element(values.begin(), values.end());
```

src/modules/voxelDensity/process.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+        for (const auto &[sp, N] : cfgSpecies)
+        {
+            bool included = std::any_of(restrictToSpecies_.begin(), restrictToSpecies_.end(),
+                                        [sp](auto &spTarget) { return sp->name() == spTarget->name(); });
+
+            if (included)
+                targetConfiguration_->removeMolecules(sp);
+        }
+    }
```
trisyoungs: Just spotted this - this will have the desired effect of restricting the calculation to the listed species, but it is destructive in the sense that it is actually removing molecules from the `Configuration`. If you look at the `MolShake` you'll see that we have to construct a list of `Molecule`s that we work over, based on the `Species` pointers in the `restrictToSpecies_` vector, and then the main loop runs over these molecules.

Either we need to rework the module to work in this way, or we remove the restrict-to-species functionality for the time being and add it in later.

src/modules/voxelDensity/voxelDensity.h
```diff
@@ -0,0 +1,79 @@
...
+#include "classes/box.h"
+#include "data/atomicMasses.h"
+#include "data/isotopes.h"
+#include "io/export/data1D.h"
+#include "module/module.h"
+#include "templates/array3D.h"
+#include <cmath>
+
+// Forward declarations
+
```
trisyoungs:
```suggestion
```

src/templates/array3D.h
```diff
@@ -86,6 +86,11 @@ template <class A> class Array3D
             resize(nX, nY, nZ);
     }
 
+    // Return reference to values
+    const std::vector<A> &values() const { return array_; }
+
```
trisyoungs:
```suggestion
```

trisyoungs (changes requested):
Sorry - just spotted one significant thing that needs changing!
Also I am the Blank Line Police today. :smile:

src/modules/voxelDensity/process.cpp
RobBuchananCompPhys: Removed

trisyoungs (approved):
Good to go. :+1:

Issue #1966, opened 2024-08-14
Optional CMake versioning in Nix shell for CLion compatibility


RobBuchananCompPhys:
**Describe the feature / issue**
CLion IDE on Windows allows us to use a Docker image as a toolchain for building and compiling C++ source code.
However, CMake versions supported by this toolchain interface are apprently limited to 3.16-3.24, while our current nix development shell pulls CMake 3.27.
It is quite possible (based on CLion user forums) that this is the reason that we cannot successfully build Dissolve in CLion in this way.

To test this further, we need to be able to pull a CMake version within the supported bounds, which could be done via a new Nix dev shell (i.e. `nix-clion-shell` or something to that effect). Since the `nix develop` command is run in the Docker file, we need to be able to pass something like an `arg` or `env` var to the Docker build, specifying that we want the older Cmake.

**Additional details**
The CLion toolchain interface discovers the Nix docker image CMake, but presents a warning about compatibility. CLion then fails to find various packages at compile time, despite their paths being recorded in the CMake cache.

PR #1967, opened 2024-08-14
build: Add CLion cmake support

rprospero:
CLion is a bit behind on which versions of CMake it supports.  I've added a new dev shell to the nix flake that uses cmake 3.19 to handle this situation.  The new dev shell is accessible via

```shell
nix develop .#windows_docker
```

I've also added an argument to the docker file to use this dev shell.  It's accessible by adding the following flag to the docker invocation:

```shell
--build-arg=windows_docker
```


RobBuchananCompPhys:
Closes #1966

PR #1969, opened 2024-09-05
feat: ring populations

Danbr4d:
Add in the capability to calculate ring distributions in systems.

Uses https://journals.aps.org/prb/pdf/10.1103/PhysRevB.44.4925

and 

https://github.com/trisyoungs/aten/blob/master/src/plugins/tool_springs/springstool_funcs.cpp


trisyoungs:
Closing as "not doing yet".

PR #1970, opened 2024-09-09
feat: more control of averaging potentials

Danbr4d:
Added in capability to average over a specified number of iterations instead of each time. The hope is that this will fix an issue of spikes in r-factor when running the averaging of potentials currently.

### DIFF
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -40,6 +40,8 @@ class EPSRManagerModule : public Module
     std::optional<int> averagingLength_;
     // Weighting scheme to use when averaging partials
     Averaging::AveragingScheme averagingScheme_{Averaging::LinearAveraging};
+    // Vector of averaged potentials over multiple iterations
+    std::vector<std::map<std::string, EPData>> averagedPotentialsStore;
 
     /*
      * Functions
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -32,7 +32,7 @@ class EPSRManagerModule : public Module
     // Potential scalings
     std::string potentialScalings_;
     // Number of historical partial sets to combine into final partials
-    std::optional<int> averagingLength_;
+    std::optional<int> averagingLength_{1};
     // Weighting scheme to use when averaging partials
     Averaging::AveragingScheme averagingScheme_{Averaging::LinearAveraging};
 
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -32,7 +32,7 @@ class EPSRManagerModule : public Module
     // Potential scalings
     std::string potentialScalings_;
     // Number of historical partial sets to combine into final partials
-    std::optional<int> averagingLength_{1};
+    std::optional<int> averagingLength_{};
     // Weighting scheme to use when averaging partials
     Averaging::AveragingScheme averagingScheme_{Averaging::LinearAveraging};
 
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include "classes/potentialSet.h"
 #include "classes/scatteringMatrix.h"
 #include "generator/generator.h"
 #include "math/averaging.h"
@@ -28,20 +29,12 @@ class EPSRManagerModule : public Module
     std::optional<int> modifyPotential_{1};
     // Vector storing atom pairs and associated potentials
     std::vector<std::tuple<std::shared_ptr<AtomType>, std::shared_ptr<AtomType>, Data1D>> potentials_;
-    struct EPData
-    {
-        Data1D ep;
-        double count{0};
-        std::shared_ptr<AtomType> at1, at2;
-    };
     // Potential scalings
     std::string potentialScalings_;
     // Number of historical partial sets to combine into final partials
     std::optional<int> averagingLength_;
     // Weighting scheme to use when averaging partials
     Averaging::AveragingScheme averagingScheme_{Averaging::LinearAveraging};
-    // Vector of averaged potentials over multiple iterations
-    std::vector<std::map<std::string, EPData>> averagedPotentialsStore;
 
     /*
      * Functions
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -36,12 +36,10 @@ class EPSRManagerModule : public Module
     };
     // Potential scalings
     std::string potentialScalings_;
-    // Number of historical potential sets to combine into final potential
+    // Number of historical partial sets to combine into final partials
     std::optional<int> averagingLength_;
-    // Weighting scheme to use when averaging potential
+    // Weighting scheme to use when averaging partials
     Averaging::AveragingScheme averagingScheme_{Averaging::LinearAveraging};
-    // Vector of averaged potentials over multiple iterations
-    std::vector<std::map<std::string, EPData>> averagedPotentialsStore;
 
     /*
      * Functions
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -31,9 +31,9 @@ class EPSRManagerModule : public Module
     std::vector<std::tuple<std::shared_ptr<AtomType>, std::shared_ptr<AtomType>, Data1D>> potentials_;
     // Potential scalings
     std::string potentialScalings_;
-    // Number of historical partial sets to combine into final partials
+    // Number of historical potentials sets to combine into final potentials
     std::optional<int> averagingLength_{};
-    // Weighting scheme to use when averaging partials
+    // Weighting scheme to use when averaging potentials
     Averaging::AveragingScheme averagingScheme_{Averaging::LinearAveraging};
 
     /*
--- a/src/modules/epsrManager/process.cpp
+++ b/src/modules/epsrManager/process.cpp
@@ -48,6 +48,24 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
             }
         }
     }
+    if (averagingLength_)
+        Messenger::print("EPSRManager: Potentials will be averaged over {} sets (scheme = {}).\n", averagingLength_.value(),
+                         Averaging::averagingSchemes().keyword(averagingScheme_));
+    else
+        Messenger::print("EPSRManager: No averaging of potentials will be performed.\n");
+
+    /*     // Perform averaging of unweighted potentials if requested, and if we're not already up-to-date
+        if (averagingLength_)
+        {
+            // Store the current fingerprint, since we must ensure we retain it in the averaged data.
+            std::string currentFingerprint{unweightedsq.fingerprint()};
+
+            Averaging::average<PartialSet>(moduleContext.dissolve().processingModuleData(), "UnweightedSQ", name_,
+                                           averagingLength_.value(), averagingScheme_);
+
+            // Re-set the object names and fingerprints of the partials
+            unweightedsq.setFingerprint(currentFingerprint);
+        } */
 
     // Form averages
     for (auto &&[key, epData] : potentials)
--- a/src/modules/epsrManager/process.cpp
+++ b/src/modules/epsrManager/process.cpp
@@ -53,27 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
     for (auto &&[key, epData] : potentials)
         epData.ep /= epData.count;
 
-    // Vector of averaged potentials over multiple iterations
-    std::vector<std::map<std::string, EPData>> averagedPotentialsStore;
     std::map<std::string, EPData> averagedPotentials = potentials;
+
+    averagedPotentialsStore.emplace_back(potentials);
     // Check if ran the right amount of iterations before averaging
-    if (averagedPotentialsStore.size() < averagingLength_)
+    if (averagedPotentialsStore.size() > averagingLength_)
     {
-        // If not then add data to vector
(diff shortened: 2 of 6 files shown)


src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,27 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+    // Vector of averaged potentials over multiple iterations
+    std::vector<std::map<std::string, EPData>> averagedPotentialsStore;
+    std::map<std::string, EPData> averagedPotentials = potentials;
+    // Check if ran the right amount of iterations before averaging
+    if (averagedPotentialsStore.size() < averagingLength_)
+    {
+        // If not then add data to vector
+        averagedPotentialsStore.emplace_back(potentials);
+    }
+    else
```
trisyoungs: You should think about this in the sort of reverse way.  You will always `emplace_back` the current `potentials` on to the vector, but if the size of the vector is then greater than the `averagingLength_` you need to `pop_front()` on the vector to take the size back down to `averagingLength_`.

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,27 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
     for (auto &&[key, epData] : potentials)
         epData.ep /= epData.count;
 
+    // Vector of averaged potentials over multiple iterations
+    std::vector<std::map<std::string, EPData>> averagedPotentialsStore;
```
trisyoungs: This will need to be a private member variable of the `EPSRManager` class, otherwise it will get reinitialised every time the module runs and you'll never get more than one set of potentials stored in it.

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,27 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+    std::map<std::string, EPData> averagedPotentials = potentials;
+    // Check if ran the right amount of iterations before averaging
+    if (averagedPotentialsStore.size() < averagingLength_)
+    {
+        // If not then add data to vector
+        averagedPotentialsStore.emplace_back(potentials);
+    }
+    else
+    {
```
trisyoungs: This `else` branch should always run, and average of all the data we currently have in the vector.

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,27 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+        // If yes then average the potentials and replace the map with the new averaged
+        for (auto n : averagedPotentialsStore)
```
trisyoungs: Prefer a const reference here to save copying unnecessarily, and perhaps rename (e.g. to `const auto &pots`)?

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,27 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+        {
+            for (auto &&[key, epData] : n)
+            {
+                averagedPotentials[key].ep += epData.ep;
```
trisyoungs: :+1:

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,27 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+            }
+        }
+    }
+    potentials = averagedPotentials;
```
trisyoungs: Need to do a division here to get the proper average - what you have currently is the sum.

src/modules/epsrManager/epsrManager.h
```diff
@@ -35,6 +36,12 @@ class EPSRManagerModule : public Module
     };
     // Potential scalings
     std::string potentialScalings_;
+    // Number of historical partial sets to combine into final partials
```
trisyoungs:
```suggestion
    // Number of historical partial sets to combine into final potentials
```

src/modules/epsrManager/epsrManager.h
```diff
@@ -35,6 +36,12 @@ class EPSRManagerModule : public Module
...
+    std::optional<int> averagingLength_;
+    // Weighting scheme to use when averaging partials
```
trisyoungs:
```suggestion
    // Weighting scheme to use when averaging potentials
```

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
         epData.ep /= epData.count;
 
+    std::map<std::string, EPData> averagedPotentials = potentials;
+
+    averagedPotentialsStore.emplace_back(potentials);
+    // Check if ran the right amount of iterations before averaging
+    if (averagedPotentialsStore.size() > averagingLength_)
+    {
+        averagedPotentialsStore.pop_back();
+    }
```
trisyoungs: This won't account for cases where the user changes the averaging length and we need to remove more than one. This curation of data is also handled for us if we move to using the proper `Averaging` namespace functions.

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+
+    // Average the potentials and replace the map with the new averaged
+    for (const auto &pots : averagedPotentialsStore)
+    {
+        for (auto &&[key, epData] : pots)
+        {
+            averagedPotentials[key].ep += epData.ep;
+            averagedPotentials[key].ep /= averagingLength_.value();
```
trisyoungs: Careful - the value of `averagingLength_` may very well not be equal to the number of potential sets in the `averagedPotentialStore`.

trisyoungs (commented):
For me, I think this PR is valuable as a proof of concept study that the averaging of potentials makes a positive difference, but we shouldn't merge it in in it's current form, simply because if we did we would immediately have to fix it up in a patch release anyway.

I think the best route forward is to introduce a `PotentialSet` class (distinct from a `PartialSet` and a `PotentialMap`, and living somewhere in the middle - we might consider renaming the latter to make the separation clear, since the new `PotentialSet` class would be based around a `std::map`) into which we can implement the necessary de/serialisation and `+=` / `/=` operators necessary for the `Averaging` functions to work.

src/modules/epsrManager/process.cpp
```diff
@@ -25,33 +25,47 @@ bool EPSRManagerModule::setUp(ModuleContext &moduleContext, Flags<KeywordBase::K
...
+    if (averagingLength_)
+        Messenger::print("Potentials will be averaged over {} sets (scheme = {}).\n", averagingLength_.value(),
+                         Averaging::averagingSchemes().keyword(averagingScheme_));
+    else
+        Messenger::print("Potentials: No averaging of potentials will be performed.\n");
 
+    auto &moduleData = moduleContext.dissolve().processingModuleData();
+    // Potential Set
+    PotentialSet potentials;
     // Loop over target data
```
trisyoungs: Just for neatness I'd shuffle these lines around
```suggestion

    // Loop over target data and form summed / averaged potentials
    PotentialSet potentials;
```

src/modules/epsrManager/process.cpp
```diff
@@ -25,33 +25,47 @@ bool EPSRManagerModule::setUp(ModuleContext &moduleContext, Flags<KeywordBase::K
...
     for (auto *module : target_)
     {
         auto *epsrModule = dynamic_cast<EPSRModule *>(module);
         auto eps = epsrModule->empiricalPotentials();
+        potentials.reset();
```
trisyoungs: Ah, you don't want to do this here, otherwise the sum will only contain the potentials from the last target EPSR module.  In fact, we shouldn't need to do this at all because the `PotentialSet` object we're using is created every time the module runs...

src/modules/epsrManager/process.cpp
```diff
@@ -25,33 +25,47 @@ bool EPSRManagerModule::setUp(ModuleContext &moduleContext, Flags<KeywordBase::K
...
+    // Average the Potentials
+    if (averagingLength_)
+        Averaging::average<PotentialSet>(moduleContext.dissolve().processingModuleData(), "PotentialSet", name(),
+                                         averagingLength_.value(), averagingScheme_);
 
-    // Form averages
-    for (auto &&[key, epData] : potentials)
-        epData.ep /= epData.count;
+    for (auto &&[key, epData] : potentials.potentialMap())
+        epData.ep /= potentials.potentialMap()[key].count;
```
trisyoungs: This needs to go immediately after the loop above, and before we store the "potentials" into the `originalPotentialsObject`.  As written I'm not sure of the effect since this is trying to divide through the potentials _after_ the averaging.

src/modules/epsrManager/process.cpp
```diff
@@ -71,7 +85,7 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
 
         Messenger::print("Apply scaling factor of {} to potential(s) {}-{}...\n", scaleFactor, typeA, typeB);
         auto count = 0;
-        for (auto &&[key, epData] : potentials)
+        for (auto &&[key, epData] : potentials.potentialMap())
```
trisyoungs: Ah! Now this might be a "new" bug - here, and in the "Adjust global potentials" bit below, we're using `potentials.potentialMap()`, but this just contains the summed / divided potentials from the current iteration. It is `originalPotentialsObject.first` that contains the properly averaged data. For clarity I'd consider renaming the `potentials` object to `newPotentials`, and perhaps consider retrieving the `originalPotentialsObject` as `currentPotentials` (just with `realise()`) and use that in these loops.

trisyoungs (commented):
Sorry - few things to consider.  Almost there but not quite!

src/modules/epsrManager/epsrManager.h
trisyoungs: This seems to have gone backwards in the rebase...

src/modules/epsrManager/epsrManager.h
trisyoungs: And here.

tests/classes/potentialSet.cpp
```diff
@@ -63,4 +64,154 @@ TEST(PotentialSetTest, ComplexAddition)
     EXPECT_EQ(2.0, pots.potentialMap()["A-D"].potential.value(0));
 }
 
+TEST(PotentialSetTest, Averaging)
+{
+    GenericList moduleData;
+    auto originalPotentialsObject =
+        moduleData.realiseIf<PotentialSet>("PotentialSet", "module", GenericItem::InRestartFileFlag);
+
+    std::string filename_{"test_restart.txt"};
```
trisyoungs: Being _really_ picky, but there shouldn't be an underscore on this local variable.
```suggestion
    std::string filename{"test_restart.txt"};
```

tests/classes/potentialSet.cpp
```diff
@@ -63,4 +64,154 @@ TEST(PotentialSetTest, ComplexAddition)
...
+    // Open the file
+    LineParser parser;
+    if (!parser.openOutput(filename_))
+    {
+        parser.closeFiles();
+    }
```
trisyoungs: If we can't open the output file we can't do the test, so this should ASSERT:
```suggestion
    ASSERT_TRUE(parser.openOutput(filename_));
```

web/docs/userguide/modules/forcefield/epsrmanager/_index.md
```diff
@@ -0,0 +1,30 @@
+---
+title: EPSR Manager (Module)
+linkTitle: EPSR Manager
+description: Modify and average the empirical potential.
+---
+
+## Overview
+
+The `EPSR Manager` module allows for the modification, by a fixed multiplier, and averaging over a set number of iterations, of the empirial potential that is calculated via the `EPSR` module.
```
trisyoungs:
```suggestion
The `EPSR Manager` module allows for the modification, by a fixed multiplier, and averaging over a set number of iterations, of the empirial potentials calculated via the {{< module "EPSR" >}} module.  It can also be used to combine the potentials from several {{< module "EPSR" >}} modules.
```

tests/classes/potentialSet.cpp
```diff
@@ -63,4 +64,154 @@ TEST(PotentialSetTest, ComplexAddition)
...
+TEST(PotentialSetTest, Averaging2)
+{
+    GenericList moduleData;
+    auto originalPotentialsObject =
+        moduleData.realiseIf<PotentialSet>("PotentialSet", "module", GenericItem::InRestartFileFlag);
+
+    std::string filename_{"test_restart.txt"};
+    // Open the file
+    LineParser parser;
+    if (!parser.openOutput(filename_))
```
trisyoungs: Another ASSERT.

tests/classes/potentialSet.cpp
trisyoungs: Rather than giving this file a generic name you could include the current test name for clarity:
```suggestion
    std::string filename_{std::format("{}.restart.txt", ::testing::UnitTest::GetInstance()->current_test_info()->name()};
```

tests/classes/potentialSet.cpp
```diff
@@ -63,4 +64,154 @@ TEST(PotentialSetTest, ComplexAddition)
...
+    }
+}
+
+TEST(PotentialSetTest, Averaging2)
+{
+    GenericList moduleData;
+    auto originalPotentialsObject =
+        moduleData.realiseIf<PotentialSet>("PotentialSet", "module", GenericItem::InRestartFileFlag);
+
+    std::string filename_{"test_restart.txt"};
```
trisyoungs: Name, underscore as before.

trisyoungs (approved):
Looking good. Couple of little tweaks to the unit test, but that's about it.

PR #1971, opened 2024-09-11
feat: drivenMD setup and testing

Danbr4d:
A new way of processing and driving SANS data has been detailed 
https://biophys.uni-saarland.de/storage/publication/Chatzimagas_MethodsEnzymol_2023.pdf
This is a test module to see how to apply this in Dissolve.


Danbr4d:
Still needs the storing of the error and convert to forces added in

PR #1973, opened 2024-09-13
refactor: Add full pair iterator and for_each_full_pair

rprospero:
This adds a `FullPairIterator` class and a `for_each_full_pair` algorithm that examines each ordered pair (as opposed to `for_each_pair`, which only examines unique unordered pairs).

### DIFF
--- a/src/classes/fullPairIterator.cpp
+++ b/src/classes/fullPairIterator.cpp
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#include "classes/fullPairIterator.h"
+#include <cmath>
+
+FullPairIterator::FullPairIterator(int size, int index) : size_(size)
+{
+    if (index == 0)
+        x_ = y_ = 0;
+    else
+        fromIndex(index);
+}
+
+void FullPairIterator::fromIndex(int index)
+{
+    x_ = index / size_;
+    y_ = index % size_;
+}
+
+int FullPairIterator::toIndex() const { return x_ * size_ + y_; }
+
+bool FullPairIterator::operator<(const FullPairIterator &other) const
+{
+    if (x_ != other.x_)
+        return x_ < other.x_;
+    return y_ < other.y_;
+}
+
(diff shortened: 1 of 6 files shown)


src/classes/fullPairIterator.cpp
```diff
@@ -0,0 +1,66 @@
...
+bool FullPairIterator::operator==(const FullPairIterator &other) const { return x_ == other.x_ && y_ == other.y_; }
+
+bool FullPairIterator::operator!=(const FullPairIterator &other) const { return x_ != other.x_ || y_ != other.y_; }
+
+FullPairIterator::value_type FullPairIterator::operator*() { return {x_, y_}; }
+
+FullPairIterator &FullPairIterator::operator++()
+{
+    y_++;
+    if (y_ >= size_)
```
trisyoungs:
```suggestion
    if (++y_ >= size_)
```

trisyoungs: Just a suggestion!

tests/algorithms/pairIterator.cpp
```diff
@@ -23,4 +24,20 @@ TEST(PairIterTest, Pairs)
...
+    {
+        for (y = 0; y < size; ++y)
+        {
+            auto [i, j] = *it;
+            EXPECT_EQ(i, x);
+            EXPECT_EQ(j, y);
+            ++it;
+        }
+    }
+}
```
trisyoungs: I want to suggest here that we do a less "pathological" test of the new iterator. Specifically something like taking a full (i.e. not halved) `Array2D` of random ints and get the sum through both the dumb double x,y loop as well as via the new iterator.

rprospero: I had a whole argument for why we shouldn't do this, but now I think that you're right.  My argument had been that using the sum test proves that every pair is visited, but doesn't prove that we visited them in the correct order.  However, it then becomes a philospohical question of whether there *is* a correct order or if that should just be considered an implementation detail.  I'm now leaning closer to making it an implementation detail, so none of the tests should depend on it.

trisyoungs (approved):
Think this all looks good, although I'm a bit foggy!

PR #1976, opened 2024-09-19
build: enable ccache by default under nix

rprospero:
This PR enables ccache by default within the nix shell.

To be honest, it re-enables it.  It had always been enabled, but it was being overshadowed by `distcc`, which I don't believe is being used by anyone on the development team.  By entering the `nix develop` in the update flake *and* deleting the build directory and recreating it from the preset, you'll enable ccache.

As a quick benchmark, I managed to recompile the full dissolve GUI build on IDAaaS in 128 seconds.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -157,7 +157,6 @@
               cmake-format
               cmake-language-server
               conan
-              distcc
               direnv
               gdb
               gtk3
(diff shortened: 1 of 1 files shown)


Danbr4d (approved):
Definitely builds faster when used

PR #1977, opened 2024-09-26
refactor: unify for_each_pair and for_each_full_pair

rprospero:
This unifies `for_each_pair` and `for_each_full_pair`  behind a boolean flag that matches the flag on Array2D.


Danbr4d (approved):
Looks good to me

PR #1978, opened 2024-09-27
feat: Peak finding

RobBuchananCompPhys:
Part of #1979 

This class allows us to do the following peak-related operations on 1D data:
- Find peaks, with the option of specifying a vertical threshold or horizontal isolation to constrain the peaks
- Find their prominences
- Rank them by value from highest to lowest
- Struct containers for peaks provides a way of describing each data peak in terms of their location and peak value
- With the use of isolation, the peak finding can be robust to random noise


src/math/peaks.cpp
```diff
@@ -0,0 +1,180 @@
...
+    std::sort(proms.begin(), proms.end(), [](const auto &a, const auto &b) { return a.prominence > b.prominence; });
+}
+
+// Get top n peaks
+std::vector<Peaks::Peak1D> Peaks::top(int n, std::vector<Peaks::Peak1D> peaks)
+{
+    sortPeaks(peaks);
+
+    if (abs(n) == peaks.size())
+        return peaks;
```
rprospero: This means that the `isolation_` parameter is ignored in the specific instance where the length of the array equals `n`.  This can cause number of peaks to change wildly by simply incrementing the value by one.  If this is intended, then we probably need a comment explaining why.

src/math/peaks.cpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Sort prominences in place by prominence value, from highest to lowest
+void Peaks::sortProminences(std::vector<Peaks::Prominence1D> &proms)
+{
+    std::sort(proms.begin(), proms.end(), [](const auto &a, const auto &b) { return a.prominence > b.prominence; });
+}
+
+// Get top n peaks
+std::vector<Peaks::Peak1D> Peaks::top(int n, std::vector<Peaks::Peak1D> peaks)
```
rprospero:
```suggestion
std::vector<Peaks::Peak1D> Peaks::top(std::size_t n, std::vector<Peaks::Peak1D> peaks)
```

This is pedantic, but `n` must logically be between 0 and the largest possible size of `peaks`, which is the exact definition of `std::size_t`.  If also saves us and `abs` call later on.

src/math/peaks.h
```diff
@@ -0,0 +1,80 @@
...
+    void sortProminences(std::vector<Prominence1D> &proms);
+    // Sort peaks in place by index, from first to last
+    void setThreshold(double range);
+    // Set horizontal threshold for peaks
+    void setIsolation(double range);
+    // Get top n peaks
+    std::vector<Peak1D> top(int n, std::vector<Peak1D>);
+    // Find the peaks (local maxima) of data
+    std::vector<Peak1D> find(bool heightOrder = false);
+    // Calculate prominences of peaks
```
rprospero: It might be worth explaining what is meant by prominence here

src/math/peaks.cpp
```diff
@@ -0,0 +1,180 @@
...
+            double pointsRight[3] = {values_[iRight - 1], values_[iRight], values_[iRight + 1]};
+
+            if (isLocalMinimum(pointsRight) || isInflectionPoint(pointsRight))
+            {
+                heightRefRight += values_[iRight];
+                break;
+            }
+            else
+                iRight++;
+        }
```
rprospero: As I understand these lines, you're mostly just looking for the range on each side of the peak where the values are strictly decreasing.  This can be accomplished with the `std::is_sorted_until` algorithm.  That would also be robust against the peak reaching the end of the value, which *looks* like a bug in this setup (e.g. when iLeft=0, the `heightRefLeft` is never set).

On a related note, it feels that `heightRefLeft` should be set with a mere `=` instead of a `+=`, unless there is something that I'm not understand.

src/math/peaks.h
```diff
@@ -0,0 +1,80 @@
...
+    // Characterise only peaks occuring above a given vertical threshold
+    double threshold_;
+    // Characterise only peaks occuring above a given horizontal isolation
+    double isolation_;
+    // Check if neighbouring points correspond to a local maximum
+    bool isLocalMaximum(double *points);
+    // Check if neighbouring points correspond to a local minimum
+    bool isLocalMinimum(double *points);
+    // Check if neighbouring points correspond to an inflection point
+    bool isInflectionPoint(double *points);
```
rprospero: Instead of passing a pointer, it might be better to pass a `std::array<double, 3>`.  This allows the compiler to ensure that the function always has access to three values.

src/math/peaks.cpp
```diff
@@ -0,0 +1,180 @@
...
+        if (values_[i] > threshold_ && isLocalMaximum(neighbours))
+        {
+            peaks.emplace_back(values_[i], domain_[i], i);
+        }
+    }
+
+    if (isolation_ > 0)
+    {
+        auto maxX = *std::max_element(domain_.begin(), domain_.end());
+        auto isolatedPeaks = top(std::round(maxX / isolation_), peaks);
```
rprospero: I'm curious about the choice of `std::round(maxX / isolation_)` for the number of peaks.  As my mental example, I'm imagining taking may data and padding the end with a ton of zeroes.  The function would now start returning more peaks in the exact same data, just because maxX has increased, despite the fact that all the peaks are still in that same unpadded data.

I'm thinking that we might be better off with the `top` function not taking a count at all and simply returning all of the peaks that match the isolation requirements.

rprospero (changes requested):
Your numerical methods background is both more recent and more intensive than mine, so my suggestions might be out of line due to my missing crucial details.  In that case, many of my comments could be corrected with just some explanatory comments.

src/math/peaks.h
RobBuchananCompPhys: Always wanted a reason to use `std::array`!

src/math/peaks.cpp
RobBuchananCompPhys: Good catch. I'm tempted to go with removing the contentious `if` statement and just let the function run without checking if `n == peaks.size()` - I think it's better to always give the `isolation_` chance to take effect.
I say this because in one case, the function is called with zero `isolation_` set, so we will just get back N peaks, ordered by height. In the other case, `isolation_` is non-zero, and the resulting effect is that the isolation radius cuts out a few of the supplied peaks before returning - which is exactly what we'd expect if the object has isolation set on it.

src/math/peaks.cpp
RobBuchananCompPhys: Cheers for pointing this out. The padding out thought experiment is definitely an issue in the current configuration.
I have an idea though - how about instead of removing the explicit target count argument from `top`: what we actually do is replace `std::round(maxX / isolation_)` with `std::round((peaks.back().valueAt - peaks.front().valueAt) / isolation_)`. That way, even if you were to pad out the same data with tonnes of zeros, the ratio of peaks to isolation radius would always be constrained by the actual location of the peaks, which doesn't change when the `Peak1D` vector is passed to `top`.
A user could still call `top` in any other situation with a target integer N and get the N top peaks, but when it's called in `find` as here, it could still handle the logic without the user actually knowing what N was used.

src/math/peaks.cpp
RobBuchananCompPhys: No, you're correct, there was no need to increment after initialising `heightRef` - I have changed it to `=`

RobBuchananCompPhys: The whole method has been refactored in a more simple way that I think better encapsulates the prominence computation.

RobBuchananCompPhys:
@rprospero FYI - I attach [documentation from Mathworks](https://uk.mathworks.com/help/signal/ug/prominence.html) detailing the prominence algorithm we're using here.

src/math/peaks.cpp
RobBuchananCompPhys: Actually I think it would be `std::round((peaks.back().valueAt - peaks.front().valueAt) / (isolation_*2))`, since `isolation_` is a radius, so were looking to see how many full isolation zones fit into the space between the first and last peak. 

Note: If user has set something like isolation = 0.000001, then `top` might get called with a large `n` relative to the number of actual detected peaks, but the function exits when it has reached all the peaks so that shouldn't be an issue.

rprospero (approved):

PR #1980, opened 2024-09-27
fix: Visit each pair in TR module

rprospero:
This should alter the TR process to visit each ordered pair.

### DIFF
--- a/src/classes/partialSet.cpp
+++ b/src/classes/partialSet.cpp
@@ -33,7 +33,7 @@ PartialSet::~PartialSet()
 bool PartialSet::setUp(const AtomTypeMix &atomTypeMix, double rdfRange, double binWidth)
 {
     // Set up partial arrays
-    if (!setUpPartials(atomTypeMix))
+    if (!setUpPartials(atomTypeMix, half_))
         return false;
 
     // Initialise histograms for g(r) calculation
@@ -45,16 +45,17 @@ bool PartialSet::setUp(const AtomTypeMix &atomTypeMix, double rdfRange, double b
 }
 
 // Set up PartialSet without initialising histogram arrays
-bool PartialSet::setUpPartials(const AtomTypeMix &atomTypeMix)
+bool PartialSet::setUpPartials(const AtomTypeMix &atomTypeMix, bool half)
 {
     // Copy type array
     atomTypeMix_ = atomTypeMix;
     auto nTypes = atomTypeMix_.nItems();
+    half_ = half;
 
-    partials_.initialise(nTypes, nTypes, true);
-    boundPartials_.initialise(nTypes, nTypes, true);
-    unboundPartials_.initialise(nTypes, nTypes, true);
-    emptyBoundPartials_.initialise(nTypes, nTypes, true);
+    partials_.initialise(nTypes, nTypes, half_);
+    boundPartials_.initialise(nTypes, nTypes, half_);
+    unboundPartials_.initialise(nTypes, nTypes, half_);
+    emptyBoundPartials_.initialise(nTypes, nTypes, half_);
     emptyBoundPartials_ = false;
 
     // Set up array matrices for partials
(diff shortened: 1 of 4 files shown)


src/modules/tr/process.cpp
```diff
@@ -81,8 +81,8 @@ Module::ExecutionResult TRModule::process(ModuleContext &moduleContext)
             {
                 y *= x * factor;
             }
-        }
-    }
+        },
+        true);
```
Danbr4d: Shouldn't this be false as we do want to go through all the pairs

src/classes/partialSet.cpp
```diff
@@ -268,31 +283,62 @@ bool PartialSet::save(std::string_view prefix, std::string_view tag, std::string
...
-                                parser.writeLineF("{:16.9e}  {:16.9e}  {:16.9e}  {:16.9e}\n", full.xAxis(n), full.value(n),
-                                                  bound.value(n), unbound.value(n));
-                            parser.closeFiles();
-
-                            return EarlyReturn<bool>::Continue;
-                        });
+    if (half_)
+    {
+        // Write partials
+        for_each_pair_early(atomTypeMix_.begin(), atomTypeMix_.end(),
```
trisyoungs: Why are we duplicating the whole code block and not just passing `half_` as the last argument to `for_each_pair_early`?

src/classes/partialSet.cpp
```diff
@@ -421,6 +467,8 @@ void PartialSet::calculateRDF(Data1D &destination, const Histogram1D &histogram,
     }
 }
 
+void PartialSet::setFullMatrix() { half_ = false; }
```
trisyoungs: I think this is quite a dangerous way of doing things, since I can freely set `half_` and it will have no impact on the arrays themselves.  A more robust solution would be to pass an additional argument to the `setUpPartials()` functions.

src/classes/partialSet.cpp
```diff
@@ -534,26 +582,29 @@ int readDataPoint(int argIndex, LineParser &parser, Data1D &data)
 // Read data through specified LineParser
 bool PartialSet::deserialise(LineParser &parser, const CoreData &coreData)
 {
-    if (parser.readNextLine(LineParser::Defaults, fingerprint_) != LineParser::Success)
+    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
         return false;
-
+    fingerprint_ = parser.argsv(0);
+    parser.hasArg(1) ? half_ = parser.argb(1) : half_ = true;
```
trisyoungs:
```suggestion
    half_ = parser.hasArg(1) ? parser.argb(1) : true;
```

src/classes/partialSet.cpp
```diff
@@ -641,43 +692,79 @@ std::string writeDataPoint(int i, Data1D data)
...
 bool PartialSet::serialise(LineParser &parser) const
 {
-    if (!parser.writeLineF("{}\n", fingerprint_))
+    if (!parser.writeLineF("'{}'  {}\n", fingerprint_, half_))
         return false;
 
     // Write out AtomTypes first
     atomTypeMix_.serialise(parser);
     auto nTypes = atomTypeMix_.nItems();
+    if (half_)
```
trisyoungs: Here too you can just pass `half_` as the last argument to `for_each_pair_early()`.

trisyoungs (changes requested):

src/classes/partialSet.h
```diff
@@ -44,12 +44,14 @@ class PartialSet
     Array2D<char> emptyBoundPartials_;
     // Total functions
     Data1D boundTotal_, unboundTotal_, total_;
+    // Check for full or half matrix
+    bool half_ = true;
```
trisyoungs:
```suggestion
    bool half_{true};
```

trisyoungs (commented):
One small stylistic correction, but can't see anything obviously wrong with this PR. I'll run it for a bit and do some testing.

src/classes/partialSet.cpp
```diff
@@ -531,16 +548,18 @@ bool PartialSet::deserialise(LineParser &parser, const CoreData &coreData)
...
+    emptyBoundPartials_.initialise(nTypes, nTypes, half_);
     emptyBoundPartials_ = false;
 
     for (auto typeI = 0; typeI < nTypes; ++typeI)
     {
-        for (auto typeJ = typeI; typeJ < nTypes; ++typeJ)
+        for (auto typeJ = 0; typeJ < nTypes; ++typeJ)
         {
+            if (half_ && typeJ < typeI)
+                continue;
```
trisyoungs: You could turn this into a use of `for_each_pair_early` for consistency:
```
    auto success = for_each_pair_early(
        0, nTypes,
        [&](int typeI, int typeJ) -> EarlyReturn<bool>
        {
            auto &part = partials_[{typeI, typeJ}];
            auto &bound = boundPartials_[{typeI, typeJ}];
            ... etc ...
            return EarlyReturn<bool>::Continue;
        },
        half_};

    if (!success.value_or(true))
        return false;
```

trisyoungs (approved):
Looking good!  One minor change suggestion to use another algorithm, but that's it.

PR #1981, opened 2024-10-01
feat: coulombic as additional potential

Danbr4d:
Added in the ability to add coulombic potentials as an additional potential so that we can manually describe the interactions.

### DIFF
--- a/src/math/function1D.cpp
+++ b/src/math/function1D.cpp
@@ -359,6 +359,28 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
         functions[Functions1D::Form::Coulombic].setDerivativeFunction(
             [](double x, double omega, const std::vector<double> &params)
             { return (-COULCONVERT * params[0] * params[1]) / pow(x, 2); });
+
+        /*
+         * Shifted Coulomb Truncation Potential
+         *
+         * Parameters:
+         * INPUT  0 = q1
+         * INPUT  1 = q2
+         * INPUT  2 = range
+         *
+         *          q1 * q2
+         * F(x)=k * -------
+         *             x
+         */
+        functions[Functions1D::Form::ShiftedCoulombTruncation] = Function1DDefinition(
+            {"q1", "q2", "range"}, [](double x, double omega, const std::vector<double> &params)
+            { return COULCONVERT * params[0] * params[1] * (1.0 / x + x / (params[2] * params[2]) - 2.0 / params[2]); });
+        /*
+         * dYdX(x) = - k * q1 * q2 * r**-2
+         */
+        functions[Functions1D::Form::ShiftedCoulombTruncation].setDerivativeFunction(
+            [](double x, double omega, const std::vector<double> &params)
+            { return COULCONVERT * params[0] * params[1] * (1.0 / (x * x) - 1.0 / (params[2] * params[2])); });
     }
 
     return functions;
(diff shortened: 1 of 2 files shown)


src/math/function1D.cpp
```diff
@@ -337,6 +337,28 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
...
+        /*
+         * Coulombic Potential
+         *
+         * Parameters:
+         * INPUT  0 = k
+         * INPUT  1 = q1
+         * INPUT  2 = q2
+         *
+         *          q1 * q2
+         * F(x)=k - -------
```
trisyoungs:
```suggestion
         * F(x)=-k -------
```

src/math/function1D.cpp
```diff
@@ -337,6 +337,28 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
...
+         *           x * x
```
trisyoungs:
```suggestion
         *             x
```

src/math/function1D.cpp
```diff
@@ -337,6 +337,28 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
...
+         */
+        functions[Functions1D::Form::Coulombic] =
+            Function1DDefinition({"k", "q1", "q2"}, [](double x, double omega, const std::vector<double> &params)
+                                 { return (params[0] * params[1] * params[2]) / pow(x, 2.0); });
```
trisyoungs: The electrostatic energy is dependent on `r`, not `r**2`. You'll also need to convert to kJ/mol (see `src/classes/pairPotential.cpp`, line 300 for the analytic Coulomb potential in Dissolve).

src/math/function1D.cpp
```diff
@@ -337,6 +337,28 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
...
+        /*
+         * dYdX(x) = -2 * k * q1 * q2 * r**-3
```
trisyoungs: _This_ is the dependency on `r**2` (not `r**3`), and also needs converting to suitable energy units.

trisyoungs (changes requested):

src/math/function1D.cpp
```diff
@@ -337,6 +338,49 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
...
+
+        /*
+         * Coulombic Potential
+         *
+         * Parameters:
+         * INPUT  0 = q1
+         * INPUT  1 = q2
+         *
+         *          q1 * q2
+         * F(x)=k * -------
```
trisyoungs: Still have a `k` in here, as with all the other comments, but it isn't in the actual coded function.

src/math/function1D.cpp
```diff
@@ -337,6 +338,49 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
...
+         */
+        functions[Functions1D::Form::Coulombic] =
+            Function1DDefinition({"q1", "q2"}, [](double x, double omega, const std::vector<double> &params)
+                                 { return (COULCONVERT * params[0] * params[1]) / x; });
+        /*
+         * dYdX(x) = - k * q1 * q2 * r**-2
+         */
+        functions[Functions1D::Form::Coulombic].setDerivativeFunction(
+            [](double x, double omega, const std::vector<double> &params)
+            { return (-COULCONVERT * params[0] * params[1]) / pow(x, 2); });
```
trisyoungs: The `pow` function is slow for doing a simple square, so I'd suggest changing this:
```suggestion
            { return (-COULCONVERT * params[0] * params[1]) / (x * x); });
```

src/math/function1D.cpp
```diff
@@ -354,7 +398,9 @@ EnumOptions<Functions1D::Form> Functions1D::forms()
                                            {Functions1D::Form::LennardJones126, "LennardJones126", 2},
                                            {Functions1D::Form::Buckingham, "Buckingham", 3},
                                            {Functions1D::Form::GaussianPotential, "GaussianPotential", 3},
-                                           {Functions1D::Form::Harmonic, "Harmonic", 1}});
+                                           {Functions1D::Form::Harmonic, "Harmonic", 1},
+                                           {Functions1D::Form::Coulombic, "Coulombic", 2},
+                                           {Functions1D::Form::ShiftedCoulombTruncation, "ShiftedCoulombTruncation", 3}});
```
trisyoungs: I'd just call this `ShiftedCoulomb`:
```suggestion
                                           {Functions1D::Form::ShiftedCoulomb, "ShiftedCoulomb", 3}});
```

trisyoungs (approved):
Few small corrections / comments to deal with, then good to go.

PR #1986, opened 2024-10-21
fix: Restart Load Crash

### LINKED ISSUE #1974: Crash when loading a restart file on IDAaaS
**Describe the bug**
When loading a restart file in IDaaS with the current build the error below is given
image

**Screenshots**
image

IDAaaS Build.

trisyoungs:
This PR fixes a bug introduced in #1977 where one of the algorithms used inverted logic when setting the inner loop lower limit.

Closes #1974.

### DIFF
--- a/src/templates/algorithms.h
+++ b/src/templates/algorithms.h
@@ -44,7 +44,8 @@ template <typename T> class EarlyReturn
     std::optional<T> value() const { return value_; }
 };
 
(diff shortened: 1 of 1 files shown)


rprospero:
One annoyance with GitHub Actions is that it caches the PR name, so you have to make a new commit to get it to recognize the name change.

rprospero (approved):

PR #1991, opened 2024-10-22
feat: addition of potentialSet class

Danbr4d:
Addition of potentialSet class for easier averaging of potentials.


src/classes/potentialSet.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    return true;
+}
+
+// Write data through specified LineParser
+bool PotentialSet::serialise(LineParser &parser) const
+{
+    if (!parser.writeLineF("{}\n", fingerprint_))
+        return false;
+
```
rprospero:
```suggestion
    if (!parser.writeLineF("{}\n", potentials_.size())
        return false;
    for (auto &[key, value] : potentials_)
    {
        if (!parser.writeLineF("{} {} {} {}\n", key, value.count, value.at1->name(), value.at2->name()))
            return false;
        if (!value.ep.serialise(parser))
            return false;
    }
```

src/classes/potentialSet.cpp
```diff
@@ -0,0 +1,66 @@
...
+/*
+ * Serialisation
+ */
+
+// Read data through specified LineParser
+bool PotentialSet::deserialise(LineParser &parser, const CoreData &coreData)
+{
+    if (parser.readNextLine(LineParser::Defaults, fingerprint_) != LineParser::Success)
+        return false;
+
```
rprospero, line 72:
```suggestion
    
    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
        return false;
    auto size = parser.argli(0);
    for (auto n = 0; n < size; ++n)
    {
        if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
            return false;
        EPData value;
        auto key = parser.args(0);
        value.count = parser.argi(1);
        value.at1 = coreData.findAtomType(parser.args(2));
        value.at2 = coreData.findAtomType(parser.args(3));
        
        if (!value.ep.deserialise(parser))
            return false;
            
        potentials_[key] = value;
    }

```

src/classes/potentialSet.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+void PotentialSet::operator*=(const double factor)
+{
+    for (auto &[key, potential] : potentials_)
+    {
+        potential.ep *= factor;
+    }
+}
```
rprospero:
```suggestion
PotentialSet& PotentialSet::operator+=(const double delta)
{
    for (auto &[key, potential] : potentials_)
        potential.ep += delta;
    return (*this);
}

PotentialSet& PotentialSet::operator+=(const PotentialSet &source)
{
    for (auto &[key, potential] : source.potentialMap())
        potentials_[key].ep += potential.ep;
    return (*this);
}

PotentialSet& PotentialSet::operator*=(const double factor)
{
    for (auto &[key, potential] : potentials_)
        potential.ep *= factor;
    return (*this);
}
```

A couple of changes here.  First, we're dropping the extra braces on the single line `for` loop.  I'm also having the operator return a reference to the value.  This is generally good practice for the assignment operators because it enables chaining.  For example,

```
  a += b += c
```

Honestly, if someone ever submitted code like that, it's not going to get through code review.  Still, I think that the compiler is a little happier if that is available, so we should leave it in.

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+    std::map<std::string, EPData> &potentialMap();
+    const std::map<std::string, EPData> &potentialMap() const;
+
+    /*
+     * Operators
+     */
+    public:
+    void operator+=(const double delta);
+    void operator+=(const PotentialSet &source);
+    void operator*=(const double factor);
```
rprospero:
```suggestion
    PotentialSet& operator+=(const double delta);
    PotentialSet& operator+=(const PotentialSet &source);
    PotentialSet& operator*=(const double factor);
```

Matchthing the changes I made to the operators.

rprospero (dismissed):
Two bits here.  First, I've made some suggestions on the operators.  Also, we should probably add a unit test for the averaging.  It'll be easier to put it in now while the code is fresh in your mind than six months from now when we're tracking down a bug.

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+    for (const auto &pots : averagedPotentialsStore)
+    {
+        for (auto &&[key, epData] : pots)
+        {
+            averagedPotentials[key].ep += epData.ep;
+            averagedPotentials[key].ep /= averagingLength_.value();
+        }
+    }
+    potentials = averagedPotentials;
+
```
trisyoungs, line 75: Could this section be the issue?  The logic here doesn't quite add up, since you store the current potentials as the beginning of `averagePotentials`, then add `potentials` to the store, then immediately remove it if you have exceeded your averaging length (which past a certain point you always will) and then do the sum over all the store potentials. That will give you N+1 sets of potentials contributing to the average, rather than the N you want, and which will always contain the first N potential sets you ever generate at the beginning of the simulation (since you're doing `pop_back` rather than `pop_front`. The `Averaging` class would sort all this out for you, but I have suggested the necessary modifications to clean this up anyway.

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
     for (auto &&[key, epData] : potentials)
         epData.ep /= epData.count;
 
+    std::map<std::string, EPData> averagedPotentials = potentials;
+
+    averagedPotentialsStore.emplace_back(potentials);
```
trisyoungs, line 58:
```suggestion
```

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+    // Check if ran the right amount of iterations before averaging
+    if (averagedPotentialsStore.size() > averagingLength_)
+    {
+        averagedPotentialsStore.pop_back();
+    }
```
trisyoungs, line 63:
```suggestion
    // Limit data to average to the most recent N-1 sets
    while (averagedPotentialsStore.size() >= averagingLength_)
    {
        averagedPotentialsStore.pop_front();
    }
```

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+    // Average the potentials and replace the map with the new averaged
+    for (const auto &pots : averagedPotentialsStore)
+    {
+        for (auto &&[key, epData] : pots)
+        {
+            averagedPotentials[key].ep += epData.ep;
+            averagedPotentials[key].ep /= averagingLength_.value();
+        }
+    }
+    potentials = averagedPotentials;
```
trisyoungs, line 74:
```suggestion
    // Add the current potentials (before averaging) to the average store before we overwrite them
    averagePotentialsStore.emplace_back(potentials);

    // Set new averaged potentials
    potentials = averagedPotentials;
```

src/modules/epsrManager/process.cpp
```diff
@@ -53,6 +53,26 @@ Module::ExecutionResult EPSRManagerModule::process(ModuleContext &moduleContext)
...
+        averagedPotentialsStore.pop_back();
+    }
+
+    // Average the potentials and replace the map with the new averaged
+    for (const auto &pots : averagedPotentialsStore)
+    {
+        for (auto &&[key, epData] : pots)
+        {
+            averagedPotentials[key].ep += epData.ep;
+            averagedPotentials[key].ep /= averagingLength_.value();
```
trisyoungs, line 71: You definitely don't want to do this division every time you add a set of potentials, just once at the end!

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+#include "math/histogram1D.h"
+#include "modules/epsr/epsr.h"
+#include "modules/epsrManager/epsrManager.h"
+#include "templates/array2D.h"
+
+// Forward Declarations
+class Configuration;
+class Interpolator;
+
+// Set of Partials
```
trisyoungs:
```suggestion
// Set of Potentials
```

src/classes/potentialSet.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "io/export/data1D.h"
+#include "items/deserialisers.h"
+#include "items/serialisers.h"
+#include "templates/algorithms.h"
+
+PotentialSet::PotentialSet() { fingerprint_ = "NO_FINGERPRINT"; }
+
+PotentialSet::~PotentialSet() { potentials_.clear(); }
+
+// Reset partial arrays
```
trisyoungs:
```suggestion
// Reset
```

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+    {
+        Data1D ep;
+        double count{0};
+        std::shared_ptr<AtomType> at1, at2;
+    };
+    // Pair matrix, containing full atom-atom partial
+    std::map<std::string, EPData> potentials_;
+
+    public:
+    // Reset partial arrays
```
trisyoungs:
```suggestion
    // Reset```

File: src/classes/potentialSet.h [hunk 13]
```diff
@@ -0,0 +1,67 @@
...
+    private:
+    // Fingerprint for these partials (e.g. reflecting Configuration indices at which they were calculated)
+    std::string fingerprint_;
+    struct EPData
+    {
+        Data1D ep;
+        double count{0};
+        std::shared_ptr<AtomType> at1, at2;
+    };
+    // Pair matrix, containing full atom-atom partial
```
trisyoungs:
```suggestion
    // Map of named potentials to data
```

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+    PotentialSet();
+    ~PotentialSet();
+
+    /*
+     * Potentials Data
+     */
+    private:
+    // Fingerprint for these partials (e.g. reflecting Configuration indices at which they were calculated)
+    std::string fingerprint_;
+    struct EPData
```
trisyoungs: I want to say that we should rename this because there isn't any reason why it has to have empirical potential (EP) data in it:
```suggestion
    struct PotentialData
```

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+    {
+        Data1D ep;
```
trisyoungs:
```suggestion
        Data1D potential;
```

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+        double count{0};
+        std::shared_ptr<AtomType> at1, at2;
+    };
+    // Pair matrix, containing full atom-atom partial
+    std::map<std::string, EPData> potentials_;
```
trisyoungs:
```suggestion
    std::map<std::string, PotentialData> potentials_;
```

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+
+    public:
+    // Reset partial arrays
+    void reset();
+    // Set new fingerprint
+    void setFingerprint(std::string_view fingerprint);
+    // Return fingerprint of partials
```
trisyoungs:
```suggestion
    // Return fingerprint
```

src/classes/potentialSet.h
```diff
@@ -0,0 +1,67 @@
...
+    std::string_view fingerprint() const;
+    // Return full atom-atom partial specified
```
trisyoungs:
```suggestion
    // Return full map of potentials specified
```

trisyoungs (changes requested):
I've requested changes on this but its just object naming and comment correction.

src/modules/epsrManager/epsrManager.h
```diff
@@ -35,6 +36,12 @@ class EPSRManagerModule : public Module
     };
     // Potential scalings
     std::string potentialScalings_;
+    // Number of historical partial sets to combine into final partials
+    std::optional<int> averagingLength_;
+    // Weighting scheme to use when averaging partials
```
trisyoungs:
```suggestion
    // Number of historical potential sets to combine into final partials
    std::optional<int> averagingLength_;
    // Weighting scheme to use when averaging potential
```

trisyoungs (approved):
Two more small things in comments I just spotted, then good to go.

PR #1992, opened 2024-10-23
feat: adjust potential range against box geometry dialog

### LINKED ISSUE #1206: Auto-adjust pair potential range when adding configurations
After adding a new configuration of "modest" size, the first step afterwards is often adjusting the pair potential range (down) so that it is not longer than the maximum half-cell length (or, more generally the the inscribed sphere radius). The wizard could easily contain a feature to adjust this automatically when the configuration is created.

aboswel:
Auto-adjust pair potential range when adding configurations #1206

Before starting a simulation, current configuration box sizes are retrieved, and their inscribed sphere radius calculated, the smallest of which is compared to the Pair Potential Range. If this radius is lower than the range, a dialog is called asking the user if they want the range to be reduced to the highest acceptable value before running the simulation.


src/gui/CMakeLists.txt
```diff
@@ -110,6 +110,8 @@ add_library(
   intramolecularTermsDialog.ui
   modifyChargesDialog.cpp
   modifyChargesDialog.h
+  pairPotentialCheckDialog.cpp
+  pairPotentialCheckDialog.h
```
trisyoungs: So perhaps the biggest change I would make is to implement this as a single function within the GUI's `DissolveWindow` class, rather than within a separate class. It would absolutely work the way you have designed it, but I feel there is not enough "meat" to warrant a separate class.

src/gui/gui_simulation.cpp
```diff
@@ -51,6 +52,16 @@ void DissolveWindow::setupIteration(int count)
...
 
+    // Check if pair potential range is too large for configuration box, ask to auto adjust if so
+    if (PairPotentialCheck::checkPairPotential(dissolve_.coreData().configurations(), dissolve_.pairPotentialRange())) {
+
+        if (PairPotentialCheck::displayPairPotentialWarning()) {
+
+            dissolve_.setPairPotentialRange(PairPotentialCheck::radius);
+            Messenger::print("Pair potential range updated to {}", PairPotentialCheck::radius);
+        }
+    }
```
trisyoungs: Moving away from the separate class, this would become something like:
```suggestion
    // Check if pair potential range is too large for configuration box, ask to auto adjust if so
    checkPairPotentialRange();
```
Furthermore, I now realise that we can only ever make this check just after a new `Configuration` is generated in the GUI, either by the user pressing "Generate" or though using the AddConfigurationDialog. The `Dissolve::prepare()` function is the one that sets everything up and fails if, for example, the pair potential range is too long, but by this point it is too late for us to intervene!  So, this bit should actually be removed and we should only check after the other two user-triggered events.

src/gui/pairPotentialCheckDialog.cpp
```diff
@@ -0,0 +1,43 @@
...
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#include "pairPotentialCheckDialog.h"
+
+double PairPotentialCheck::radius = 0.0;
+QMessageBox::StandardButton PairPotentialCheck::reply;
+
+// Retrieve the inscribed sphere radius for all configuration boxes find the smallest
+// Compare smallest radius to pair potential range, return radius if smaller
+
```
trisyoungs: A stylistic comment - don't leave blank lines between function comments and the function definition.  The Blank Line Police are always looking for such things.... :smiley:

src/gui/pairPotentialCheckDialog.cpp
```diff
@@ -0,0 +1,43 @@
...
+    reply = QMessageBox::question(parent, "Warning!", QString("Maximum pair potential range exceeds smallest box inradius! Adjust pair potential range to %1? and run?").arg(radius),
+                                    QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
+        
+    if (reply == QMessageBox::Yes)
+        return true;
+    else 
+        return false;
+    
+}
```
trisyoungs: These two functions can be combined into one in the `DissolveWindow` class - determine if the pair potential range is suitable, then raise a dialog asking the user if it should be adjusted.

src/gui/pairPotentialCheckDialog.cpp
```diff
@@ -0,0 +1,43 @@
...
+        if (config->box()->inscribedSphereRadius() < radius) {
+
+            radius = config->box()->inscribedSphereRadius();
+        }
+    }
+
+    if (radius == pairPotentialRange)
+        return 0;
+    else 
+        return radius;
```
trisyoungs: You probably will end up deleting this bit of code, but there are couple of things you could do to make this kind of check neater and more compact in future. If you use the `std::optional` as suggested above then you avoid doing nasty comparisons between floating point numbers which _should_ work, but I always treat them with a bit of distrust.
Second, you could rewrite this into one line using one of my favourite devices, the ternary if:
```
    return radius == pairPotentialRange ? 0 : radius;
```
...or with the `std::optional` `radius`:
```
    return radius.value_or(0.0);
```

src/gui/pairPotentialCheckDialog.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    radius = pairPotentialRange;
+
+    for (const auto& config : currentConfigs) {
+        
+        if (config->box()->inscribedSphereRadius() < radius) {
+
+            radius = config->box()->inscribedSphereRadius();
+        }
+    }
```
trisyoungs: Because you might not find a configuration which necessitates a change in the pair potential range, this would a good use of `std::optional`.
```suggestion
    std::optional<double> radius;

    for (const auto& config : currentConfigs) {
        
        if (config->box()->inscribedSphereRadius() < radius.value_or(pairPotentialRange)) {

            radius = config->box()->inscribedSphereRadius();
        }
    }
```

src/gui/pairPotentialCheckDialog.h
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2024 Team Dissolve and contributors
+
+#pragma once
+
+#include "classes/coreData.h"
+#include "classes/configuration.h"
+#include "classes/box.h"
+#include <QMessageBox>
+
```
trisyoungs: A good IDE will highlight these things, but you don't actually need all these headers. You don't use `CoreData` in any declarations, nor do you use `Box`, and the `Configuration` is named, but not actually accessed of used in any way - i.e. there is a declaration of a function which takes a `std::vector` of `std::unique_ptr`s of `Configuration`s, but there is no actual _usage_ per se.  In such circumstances you can forget the `#include` and use a forward declaration instead. This basically says to the compiler: "Yes, this class _will_ be defined at some point, but you don't need to know the details right now."  This means the compiler loads and parses fewer include files, reduces circular dependencies, and generally speeds compilation up.
```suggestion
#include <QMessageBox>

// Forward Declarations
class Configuration;
```

trisyoungs (changes requested):

src/gui/gui.cpp
```diff
@@ -458,3 +458,33 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
     if (DissolveSys::sameString(link.toStdString(), "Messages"))
         ui_.MainTabs->setCurrentIndex(0);
 }
+
+/*
+ * Check functions
+ */
+
+// Returns maximum pair potential range given all present box geometries
```
trisyoungs: No longer true!
```suggestion
// Checks pair potential range against all present box geometries
```

src/gui/gui.cpp
```diff
@@ -458,3 +458,33 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
...
+void DissolveWindow::checkPairPotential(QWidget *parent) {
+
+    std::optional<double> radius;
+    QMessageBox::StandardButton reply;
```
trisyoungs: Don't actually need this variable...
```suggestion
```

src/gui/gui.cpp
```diff
@@ -458,3 +458,33 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
...
+    }
+
+    // Prompt to auto-adjust
+    if (radius.has_value()) {
+
+        reply = QMessageBox::question(parent, "Warning!", QString("Maximum pair potential range exceeds smallest box inradius! Adjust pair potential range to %1?").arg(radius.value()),
+                                    QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
+        
+        if (reply == QMessageBox::Yes)
+            dissolve_.setPairPotentialRange(radius.value());
```
trisyoungs: Can combine all this and remove the need for the temporary variable:
```suggestion
        if (QMessageBox::question(parent, "Warning!", QString("Maximum pair potential range exceeds smallest allowed by current Configurations! Adjust pair potential range to %1?").arg(radius.value()),
                                    QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes) == QMessageBox::Yes)
            dissolve_.setPairPotentialRange(radius.value());
```

src/gui/gui.h
```diff
@@ -254,4 +254,11 @@ class DissolveWindow : public QMainWindow
     Q_SIGNALS:
     void iterate(int);
     void stopIterating();
+
+    /*
+    * Checks
+    */
+    public:
+    // Returns maximum pair potential range given all present box geometries
```
trisyoungs:
```suggestion
    // Checks pair potential range against all present box geometries
```

src/gui/menu_configuration.cpp
```diff
@@ -81,6 +81,9 @@ void DissolveWindow::on_ConfigurationCreateAction_triggered(bool checked)
 
         ui_.MainTabs->setCurrentTab(newConfig);
     }
+
+    // Check pair potential range against box geometry
+    checkPairPotential();
```
trisyoungs: Should really move this inside the `if` block above since we have no need to run it if the `AddConfigurationDialog` was canceled.

src/gui/gui.cpp
```diff
@@ -458,3 +458,33 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
     if (DissolveSys::sameString(link.toStdString(), "Messages"))
         ui_.MainTabs->setCurrentIndex(0);
 }
+
+/*
+ * Check functions
+ */
+
+// Returns maximum pair potential range given all present box geometries
+void DissolveWindow::checkPairPotential(QWidget *parent) {
```
trisyoungs: I think we can have a more descriptive name for the function - perhaps `checkPairPotentialRange`?

trisyoungs (commented):
Just about there. Need to adjust the formatting!

src/gui/gui.cpp
```diff
@@ -458,3 +458,37 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
...
+        {
+
+            radius = config->box()->inscribedSphereRadius();
+        }
+    }
+
+    // Prompt to auto-adjust
+    if (radius.has_value())
+    {
+
```
trisyoungs:
```suggestion
```

src/gui/gui.cpp
```diff
@@ -458,3 +458,37 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
...
+    std::optional<double> radius;
+
+    // Return smallest inscribed sphere radius if less than current pair potential range
+    for (const auto &config : dissolve_.coreData().configurations())
+    {
+
+        if (config->box()->inscribedSphereRadius() < radius.value_or(dissolve_.pairPotentialRange()))
+        {
+
```
trisyoungs:
```suggestion
```

src/gui/gui.cpp
```diff
@@ -458,3 +458,37 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
...
 }
+
+/*
+ * Check functions
+ */
+
+// Checks pair potential range against all present box geometries
+void DissolveWindow::checkPairPotentialRange(QWidget *parent)
+{
+
```
trisyoungs:
```suggestion
```

src/gui/gui.cpp
```diff
@@ -458,3 +458,37 @@ void DissolveWindow::statusLabelLinkClicked(const QString &link)
...
+    std::optional<double> radius;
+
+    // Return smallest inscribed sphere radius if less than current pair potential range
+    for (const auto &config : dissolve_.coreData().configurations())
+    {
+
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
The blank line police are out in force today, but this looks good to go otherwise!

PR #1995, opened 2024-11-07
feat: Node Graph module system

rprospero:
This contains the intial prototype for the Node Graph viewer.  Full user support isn't available, but it does provide a very basic infrastructure.

20241119_16h01m56s_grim

Above is a basic screenshot of an example of the code.  The nodes of the generator are currently linked linearly to indicate the order of the events, but it's perfectly posible to have them each connect back to the generator individually.

This most interesting files to examine are 
- gui/models/nodeGraph/nodeWrapper.h : Defines all of the functions which node types must ultimately implement
- gui/models/nodeGraph/graphModelBase.h : The lowest foundation of the model
- gui/models/nodeGraph/graphModel.h : The main template model
- gui/models/nodeGraph/graphNodeModel.h : The QAbstractListModel that handles displaying the nodes
- gui/models/nodeGraph/graphEdgesModel.h : The QAbstractListModel that handles displaying the edges
- gui/models/nodeGraph/exampleGraphModel.h : An example of a conctrete graph mode (which isn't actually used anywhere in Dissolve, but shows how the Connection setup works)
- gui/models/nodeGraph/generatorGraphModel : A *very* work in progress example of using a graph model with real data
- gui/models/nodeGraph/instances/configuration.cpp : An example of an slightly more interesting node type implementation
- gui/qml/nodeGraph/GeneratorDelegate.qml : The QML code to display the GeneratorGraphModel models

The following changes should be made eventually, but are out of scope of the current PR:
- `GraphNodeModel` should return an insertion object used to insert new nodes.  This could use RAII for better safety, instead of needing to remember to call `beginInsert` and `endInsert`
- Add function to automatically rearrange nodes (especially useful after file load)


RobBuchananCompPhys:
Is there a reason we aren't using the directory `src/gui-qml` to host the graph view related C++ and qml modules? Just found it surprising since the frontend entrypoint appears to be the new `DissolveMain.qml`.

rprospero:
@RobBuchananCompPhys Ironically, I originally *did* have everything in the `src/gui-qml` folder and spent a whole commit moving all of the models and qml over to `src/gui`.  My thought at the time was that it would be less confusing to keep putting the code in the same place, but I can also see it being the other way around.  There was also a slight benefit of having all the models together for linker simplicity, though this also leads to a slightly larger than necessary executable size.

I'm not opposed to moving the code into the other directory if people find that clearer (as clarity was the entire point of the move in the first place).

rprospero:
GraphNodeJanuary

A screenshot of the current status of the GraphNodes for the generator.  We now just show the graph nodes in order and display the data for the four types I've implemented thus far (with the goal of loading examples/argon/input.txt).

Some immediate improvements that can be made

- Change most of the display values to edits so the nodes can be modified.  My last attempt was bugged, so this might be non-trivial.  Or I might have just done that in a dumb way
- Handling the parameters node needs an obvious improvement
- More node types need to be implemed
- Perhaps we would be better off with a more general structure?  Instead of having a delegate for each node type, we could crate a GeneratorNodeKeywordModel that inherits from AbstractTableModel.  The first column would always just be the name and the second column would be the values.  Since it's an Abstract*Table*Model and not an Abstract*List*Model, we can use extra columns for extra information on te node (e.g. x, y, & z dimensions in the box size can be columns 2, 3, & 4).

rprospero:
This has been superseded by the Node Graph system that is now *in* develop2

PR #1996, opened 2024-11-08
docs: Move to new website server

trisyoungs:
The existing VPS became unsuitable in the past week or so - it had limited memory (0.5 Gb) which proved to be enough in the past but which is now not feasible. This PR makes little change to the workflow, but is designed to test the new secrets by publishing directly to the new server within the PR to ensure things are working correctly.


trisyoungs:
Not required.

PR #2009, opened 2024-11-13
refactor: Single site class

trisyoungs:
This PR merges the `OrientedSite` class into `Site`, giving a single site class with a larger memory footprint (since it _always_ contains an axis matrix, even if the origin site does not specify axes) but which simplifies the rest of the code somewhat and allows future useful functions to be introduced in future PRs.

### DIFF
--- a/src/classes/site.cpp
+++ b/src/classes/site.cpp
@@ -38,7 +38,7 @@ const Vec3<double> &Site::origin() const { return origin_; }
 std::shared_ptr<const Molecule> Site::molecule() const { return molecule_; }
 
 // Return whether local axes are present
-bool Site::hasAxes() const { return false; }
+bool Site::hasAxes() const { return hasAxes_; }
 
 // Return local axes
 const Matrix3 &Site::axes() const { return axes_; }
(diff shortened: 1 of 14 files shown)


src/classes/site.cpp
```diff
@@ -6,17 +6,23 @@
...
+Site::Site(const SpeciesSite *parent, std::optional<int> uniqueSiteIndex, std::shared_ptr<const Molecule> molecule,
+           const Vec3<double> &origin)
+    : parent_(parent), uniqueSiteIndex_(uniqueSiteIndex), molecule_(std::move(molecule)), origin_(origin)
+{
+}
 
 Site::Site(const SpeciesSite *parent, std::optional<int> uniqueSiteIndex, std::shared_ptr<const Molecule> molecule,
-           Vec3<double> origin)
+           const Matrix3 &axes, const Vec3<double> &origin)
+    : parent_(parent), uniqueSiteIndex_(uniqueSiteIndex), molecule_(std::move(molecule)), origin_(origin), axes_(axes),
```
rprospero, line 17: This PR is already worth it for the initialisers.

rprospero (approved):

PR #2011, opened 2024-11-14
feat: Add on sphere

trisyoungs:
This PR implements a new Add-style node which places a number of molecules on the surface of a sphere through a reference site on the Species, with the optional ability to orient one site axis along the surface tangent. Two positioning methods are implemented - fully random and one based on a Fibonacci lattice, with the latter giving eye-pleasing results like this:

image

And for a molecule with a large tail:

image

### TODO
- [x] Node icon
- [x] Documentation

### DIFF
--- a/src/math/matrix3.cpp
+++ b/src/math/matrix3.cpp
@@ -127,6 +127,30 @@ void Matrix3::print() const
 // Set zero matrix
 void Matrix3::zero() { std::fill(matrix_.begin(), matrix_.end(), 0.0); }
 
+// Create orthogonal matrix around supplied single column vector
+void Matrix3::createFromVector(const Vec3<double> &v, int columnIndex)
+{
+    setColumn(columnIndex, v);
+    const auto v2 = v.orthogonal();
+    switch (columnIndex)
+    {
+        case 0:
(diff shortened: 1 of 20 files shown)


src/math/matrix3.cpp
```diff
@@ -127,6 +127,30 @@ void Matrix3::print() const
...
+            break;
+        case 2:
+            setColumn(0, v2);
+            setColumn(1, v * v2);
+            break;
+        default:
+            throw(std::runtime_error("Invalid column index.\n"));
+    }
+}
+
```
rprospero, line 153: I'll admit that I had to read this three times to get how it worked, but, once, I did, I couldn't think of a clearer way to write it.  This is perfectly succinct.

rprospero (approved):

src/math/matrix3.cpp
trisyoungs: I also thought about this for quite a while before settling on the "long-winded" way here. A `std::set` of {0,1,2} with the supplied `columnIndex` subsequently removed allowed you to remove the `switch` completely, but felt a little clunky in practice.

PR #2013, opened 2024-11-14
fix: Size factor persisted after node was removed

### LINKED ISSUE #2010: SizeFactor persists on Box
**Describe the bug**
If I have a generator procedure containing a SizeFactor node and I generate a configuration the size factor is correctly applied. However, if I then remove the SizeFactor node and regenerate the box, the old size factor still persists.

This should be reset on clearing the box contents / generation.

trisyoungs:
This PR fixes an issue where the requested size factor on a configuration was not reset when regenerating the content, leading to unwanted "sizefactor history".

Closes #2010.


Danbr4d (approved):
All looks good

PR #2014, opened 2024-11-18
feat: Use PCG for random number generation

### LINKED ISSUE #227: Better random number generator
Utilise "better than rand()" random number generator.  C++11 provides several built-in RNG - [https://isocpp.org/files/papers/n3551.pdf](https://isocpp.org/files/papers/n3551.pdf)

trisyoungs:
This PR moves us away from the stdlib random number generators and on to an arguably better framework in the form of PCG (https://www.pcg-random.org/).

Please check changes to `DissolveMath` carefully! In particular the generation of inclusive real numbers between 0.0 and 1.0, using (MAX - 1) - looking into the PCG code I believe this is correct since there is a modulo(MAX) operation in the `bounded_rand` function in `pcg_extras.h`.

Closes #227.

### DIFF
--- a/src/dissolve-gui-qml.cpp
+++ b/src/dissolve-gui-qml.cpp
@@ -25,7 +25,8 @@ int main(int args, char **argv)
         return 1;
 
     // Initialise random seed
-    srand(options.randomSeed().value_or((unsigned)time(nullptr)));
+    if (options.randomSeed())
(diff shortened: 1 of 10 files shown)


rprospero (approved):
I've always liked PCGs, but I didn't realise that they had better randomness than mersenne twister!

PR #2015, opened 2024-11-19
build: CMake tweaks

trisyoungs:
This PR addresses a couple of minor, non-critical issues in some `CMakeLists.txt` here and there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ if(GUI)
                3DExtras
                QuickWidgets
                QuickControls2
-    REQUIRED
+               REQUIRED
   )
 
   # Make sure we enable automated handling of Qt files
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -127,7 +124,6 @@ if(GUI)
                QuickControls2
                REQUIRED
   )
-  add_definitions(-DQT_NO_KEYWORDS)
```
rprospero, line 130: Does this mean that we can go back to `slots:` instead of `Q_SLOTS:"

rprospero (approved):

CMakeLists.txt
trisyoungs: Alas no - just removing a duplicate.

PR #2018, opened 2024-11-22
build: Update release publishing workflow

trisyoungs:
This PR modifies the `publish.yml` workflow, simplifying it and making it easier to publish our various flavours of release version. 

Some important things to note, or that I have learned during this escapade:
1) Moving to a workflow `input` variable of (new) type `choice`, rather than having three or four `bool`s define what happens for the various release types was a glorious discovery.
2) A less-than-glorious discovery was that there is precisely zero error checking on the GitHub Actions side for whether you have supplied a valid value to a `choice`-type input.  Which more-or-less makes the whole thing pointless and you might as well use a plain old `string`.
3) The `web.yml` top-level workflow previously was responsible for the actual publishing of the website to our server. I have changed this so that it just builds and sanity-checks the website, but does not push anything.  The `release.yml` and `continuous.yml` workflows now include their call to the website-building workflow with the necessary release type. While this does mean that the website does get built twice for every release and push to `develop`, taking this approach did mean that the web workflow itself could be cleaned up pretty nicely.

Please, see what you think!  One question remains on how we actually flag that a build is a legacy build, or at least how we integrate this into the way we work.

### DIFF
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -92,8 +92,7 @@ jobs:
       - name: Publish
         uses: "./.github/workflows/publish"
         with:
-          isRelease: false
-          publish: true
+          releaseType: 'continuous'
         env:
           HARBOR_USER: ${{ secrets.HARBOR_USER }}
           HARBOR_SECRET: ${{ secrets.HARBOR_SECRET }}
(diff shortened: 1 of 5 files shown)


rprospero (approved):
Everything looks good and the new `choice` setup is much cleaner.

PR #2019, opened 2024-11-26
build:  start building with C++20 Standard

rprospero:
This PR starts compiling with the C++20 standard.  This causes some knock-on effects

- We no longer depend on the `fmt` formatting library
- We need at least GCC 13
- To retrieve GCC 13, we needed to upgrade to a more recent nixpkgs release, which brought on a patch upgrade of TBB
- We've also upgraded LLVM to version 17, which also upgrades `clang-format` from version 13 to version 17
- To get the appropriate clang for OS X, I've had to drop support for Intel Macs. To be honest, it *should* work on Macs running OS 14 or higher, but GitHub Actions don't provide Intel hardward for those OSes, so we wouldn't be able to test them.
- The QC checks now use the commands from the nix shell, to ensure that upgrades are kept in sync.  This has resulted in some format changed with updated version.

The most interesting files to look at are:

- .github/workflows/qc/action.yml
- CMakeLists.txt 
- src/gui/models/sortFilterProxy.h
- src/items/serialisers.cpp
- src/keywords/weightedModuleVector.cpp
- src/templates/vector3.h
- src/templates/vector4.h

A couple of notes to take while looking through the pr

- There is not standard equivalent of `fmt::print`, so I replaced
 those calls with `std::cout`
- `std::format` *requires* that the format string is known at
 compile time.  This means that formats passed as a function or
 switched via a ternary expression cannot be used.  The
 `std::vformat` function handles dynamic formats, though it
 requires an ugly wrapping of the parameters.  For most of the
 ternary cases, I simply moved the conditional outside of the
 format call to regain the benefit of static format checking.
  Allegedly, C++26 will merge `format` and `vformat`.

I'll pull this out of draft once it compiles on all platforms and passes all tests.


src/keywords/weightedModuleVector.cpp
```diff
@@ -99,15 +100,16 @@ void WeightedModuleVectorKeyword::deserialise(const SerialisedValue &node, const
...
                  // Check the module's type if we can
                  if (!moduleTypes_.empty() &&
                      std::find_if(moduleTypes_.cbegin(), moduleTypes_.cend(),
                                   [module](const auto &s) { return s == module->type(); }) == moduleTypes_.cend())
                      throw toml::type_error(
-                         fmt::format("Module '{}' is of type '{}', and is not relevant to keyword '{}' (allowed types = {}).\n",
-                                     moduleName, module->type(), name(), joinStrings(moduleTypes_)),
+                         std::format("Module '{}' is of type '{}', and is not relevant to keyword '{}' (allowed types = {}).\n",
+                                     moduleName, ModuleTypes::moduleType(module->type()), name(),
+                                     joinStrings(moduleTypes_, ", ", [](auto m) { return ModuleTypes::moduleType(m); })),
```
RobBuchananCompPhys, line 112: I'm guessing this change with the lambda expression relates to your last note about `std::format` formatting behaviour at compile time. Just trying to get my head around why else we would need this instead of the previous member `moduleTypes_` !

rprospero: The problem is that the `joinStrings` function used to use `fmt::format` to convert the elements of `moduleTypes_` from a `ModuleTypes::moduleType` to a string.  However, `std::format` doesn't have a default formatter for `ModuleTypes::moduleType`, so I need the lambda to turn them into a string before the strings get joined.

src/base/enumOptions.h
```diff
@@ -64,7 +64,10 @@ template <class E> class EnumOptions : public EnumOptionsBase
     {
         std::string validValueString;
         for (auto n = 0; n < options_.size(); ++n)
-            validValueString += fmt::format(n == 0 ? "{}" : ", {}", options_[n].keyword());
+            if (n == 0)
+                validValueString += std::format("{}", options_[n].keyword());
+            else
+                validValueString += std::format(", {}", options_[n].keyword());
```
trisyoungs: I understand the reason for splitting this into an `if`/`else`, but can this loop be replaced with a call to our `joinStrings` algo?

src/classes/atomTypeData.cpp
```diff
@@ -45,9 +45,9 @@ void AtomTypeData::setIsotope(Sears91::Isotope tope, double pop, double fraction
 {
     if (std::find_if(isotopes_.begin(), isotopes_.end(), [tope](const auto &topeData) { return topeData.isotope() == tope; }) !=
         isotopes_.end())
-        throw(std::runtime_error(fmt::format(
+        throw(std::runtime_error(std::format(
             "Tried to set IsotopeData for isotope A = {} in AtomTypeData for AtomType '{}', but existing data is present.\n",
-            tope, Sears91::A(tope), atomTypeName())));
+            Sears91::A(tope), atomTypeName())));
```
trisyoungs: Ahh - nice catch by `std::format`!

src/classes/box.cpp
```diff
@@ -185,7 +185,7 @@ void Box::scale(Vec3<double> scaleFactors)
         auto abSame = (fabs(scaleFactors.x - scaleFactors.y) < 1.0e-5);
         auto acSame = (fabs(scaleFactors.x - scaleFactors.z) < 1.0e-5);
         if (!abSame || !acSame)
-            throw(std::runtime_error(fmt::format("Irregular scaling of cubic box requested (scale factors = {}, {}, {}\n",
+            throw(std::runtime_error(std::format("Irregular scaling of cubic box requested (scale factors = {}, {}, {}\n",
```
trisyoungs: The more I see `throw(std::runtime_error(std::format("...` crop up in our code I wonder if we shouldn't introduce a `Messenger::throw()` function to tidy things up.

src/generator/sequence.cpp
```diff
@@ -364,7 +364,7 @@ bool GeneratorNodeSequence::check() const
         // Check ownership
         if (&node->scope()->get() != this)
             return Messenger::error("Node '{}' failed parent check ({} is not this {})\n", node->name(),
-                                    fmt::ptr(node->parent()), fmt::ptr(this));
+                                    static_cast<const void *>(node->parent()), static_cast<const void *>(this));
```
trisyoungs: This feels fairly unrecoverable, so perhaps we should make this a `throw` instead.

src/keywords/weightedModuleVector.cpp
trisyoungs: Just to add to this, as-written before the change this was a bug. The `moduleTypes_` variable used to be a vector of names, so just throwing it at `joinStrings()` worked just fine.  Now it's a vector of enums, so conversion required.

src/main.cpp
```diff
@@ -7,6 +7,7 @@
 #include "main/dissolve.h"
 #include "main/version.h"
 #include "math/mathFunc.h"
+#include <format>
```
trisyoungs, line 10: A pernickety blank line request!
```suggestion
#include <format>

```

src/modules/energy/functions.cpp
```diff
@@ -17,16 +17,17 @@ namespace
...
     double improperEnergy;
 
     Energies operator+(const Energies &other) const
     {
-        return {this->bondEnergy + other.bondEnergy, this->angleEnergy + other.angleEnergy,
-                this->torsionEnergy + other.torsionEnergy, this->improperEnergy + other.improperEnergy};
+        return {.bondEnergy = this->bondEnergy + other.bondEnergy,
+                .angleEnergy = this->angleEnergy + other.angleEnergy,
+                .torsionEnergy = this->torsionEnergy + other.torsionEnergy,
+                .improperEnergy = this->improperEnergy + other.improperEnergy};
```
trisyoungs, line 30: Still not a fan of bare `struct`s, but at least this enforced style change makes it much clearer what's going on!

src/modules/geomOpt/functions.cpp
```diff
@@ -39,7 +39,7 @@ template <> void GeometryOptimisationModule::revertToReferenceCoordinates(Specie
 double GeometryOptimisationModule::rmsForce() const
 {
     auto msf = std::accumulate(f_.begin(), f_.end(), 0.0,
-                               [](auto &acc, const auto &f) { return acc + f.x * f.x + f.y * f.y + f.z * f.z; }) /
+                               [](auto acc, const auto &f) { return acc + f.x * f.x + f.y * f.y + f.z * f.z; }) /
```
trisyoungs, line 42: Kinda curious about this one - so we have to pass in the accumulated value by copy rather than reference now, or is the reference implicit somehow?

tests/math/derivatives.cpp
```diff
@@ -6,7 +6,7 @@
 #include "classes/speciesBond.h"
 #include "classes/speciesImproper.h"
 #include "classes/speciesTorsion.h"
-#include <fmt/format.h>
+#include <format>
```
trisyoungs: Pointless `#include`? No other changes in the file.

tests/math/expression.cpp
```diff
@@ -8,7 +8,7 @@
 #include "expression/reference.h"
 #include "expression/variable.h"
 #include "math/mathFunc.h"
-#include <fmt/format.h>
+#include <format>
```
trisyoungs: Same here.

trisyoungs (approved):

trisyoungs:
Like it. Worth doing for the formatting changes in the PCG headers alone!

src/base/enumOptions.h
rprospero: Good call

src/classes/box.cpp
rprospero: Agreed! I'll add an issue for myself to make that change after this PR is merged in.

src/modules/geomOpt/functions.cpp
rprospero: Under c++17, the compiler was letting us play fast and loose, but they're calling us out on our lies in c++20.  We've claimed that our function takes a reference to the accumulator and a reference to the structure and returns the new accumulator.  On the first call to the function, everything is fine, as it gets a reference to the zero value we declared before.  However, on the *second* call, it gets a reference to the result of the first call to the function, which was a temporary and thus cannot be referenced.

It never really mattered before because `acc` is just a double and the compiler elided the references under the hood (passing a 64 bit double versus a 64 bit pointer to a double allows the compiler to save a dereference).  However, something became more strict in C++20 mode and it's tryign to do what we literally said and it can't.

If we *really* wanted to keep the reference, we would write the lambda function as 

```cpp
   [](auto &acc, const auto &f) { acc += f.x * f.x + f.y * f.y + f.z * f.z; return acc;}
```

This threads the original `acc` reference through each call while still updating the value.  However, this creates a race condition if std::accumulate it run in parallel, since the same reference is being updated from multiple threads.  Using the pass by value version keeps things safe to run in parallel while still satisfying the reference rules.

tests/math/derivatives.cpp
rprospero: Good catch.

src/modules/geomOpt/functions.cpp
trisyoungs: Excellent explanation. I'm happy to stop lying and pass by value.

rprospero:
@trisyoungs I've moved in the copyright updates into this PR.  I'm just looking to fix up the windows build.  Or should we make that a separate PR, now that we at least have the QC working again.

trisyoungs:
@rprospero I think, if it turns out to be a quick / easy fix then wrap it in to this PR so we can get the whole CI rolling again. If it turns out to be more of an effort, split it.

PR #2020, opened 2024-11-27
build: Update website and main / legacy workflows

trisyoungs:
This PR does a few things:
- Updates the `web.yml` workflow to run and publish as and when necessary independent of the other workflows (recently-introduced duplicate web builds in the other workflows have been removed.
- Moves the core build, packaging, and publishing jobs into separate callable workflows in `_build_and_package.yml` and `_publishing.yml`, the underscore denoting that these are dependent workflows to be called from others. This significantly reduces repetition in our top-level workflows and is A Good Thing.
- A `legacy.yml` workflow has been added to respond to only to changes on an (as-yet-unused) `legacy` tag, but which will hopefully make publishing of legacy versions a trivial matter.


.github/workflows/_publishing.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup
+        uses: "./.github/workflows/setup"
+      - name: Set Short Hash
+        uses: "./.github/workflows/set-short-hash"
+      - name: Publish
+        uses: "./.github/workflows/publish"
```
trisyoungs, line 22: I realised just now that the `publish.yml` workflow, modified in #2018, will break our IDAaaS integration since the image which gets pulled in is currently tagged `latest`, but this will tag it `release` instead.  I think that the latter is a more sensible naming anyway, so I will ask for the scripts to be adjusted once we have moved to 1.6 and the new workflows have been run.

rprospero (approved):
I like how much cleaner this is.  It's going to break some of my C++20 PR, but the bit it breaks was ugly and the replacement cleaner.

PR #2021, opened 2024-11-27
build: Powershell script to install dependencies for MSVC

RobBuchananCompPhys:
I think this is getting close to completion. Looks good for developing in Dissolve CLI (able to set breakpoints throughout), but still need to successfully build Dissolve GUI.

- [x] Powershell script to build dissolve for development in Visual Studio 2022 (MSVC), **CLI** version
- [x] Successfully link against Qt6 for debugging the GUI version of dissolve in Visual Studio MSVC
- [x] Modularise Cmake workflow

TODO:
- [x] Tidy up and remove unused cmake `cacheVariables`
- [x] Improve Powershell script comments with slightly increased verbosity and proper Powershell comment style
- [x] Developer docs page to guide through steps
- [ ] CI job


.github/workflows/visual-studio/action.yml
```diff
@@ -0,0 +1,36 @@
...
+
+inputs:
+    publishType:
+        type: choice
+        default: 'none'
+        options:
+            - none
+            - continuous
+            - release
+            - legacy
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
Honestly I don't think I have a lot to add to this - it's looking pretty good.  One thing that did make me stop and think was the hardcoded version numbers for things like Qt and Antlr4. In my PR these versions are set at the beginning of the workflow and passed on where needed, so I wonder if, for the sake of consistency, your main `develop.ps1` script shouldn't have optional arguments to take in versions? At least then the CI is self-consistent.

.github/workflows/visual-studio/action.yml
RobBuchananCompPhys: I've just been taking a look at your changelog PR, and I can see there is a bit of harmonisation to do on my part here - agree that making these parameters into script arguments is a good idea

rprospero: My temptation would be to make make using the system qt the default option and the having a flag to specify installing a *specific* Qt version with that version as a parameter.  Then the correct Qt version can easily be passed in on the CI.  It also feels like a better default in that, if I don't have Qt installed and forget the flag, I could run it again with the flag to install Qt, but the current setup makes it easy to accidentally install two versions of Qt and doesn't provide any method for rectifying that problem.

PR #2028, opened 2024-12-12
build: Update version management scripts and control

trisyoungs:
This PR is finally ready for review! I have written it up as something which can be included in the documentation in slightly modified form if we think it is useful for onboarding / reference?

The `changeversion` script is one large change worth mentioning, and is now git-cliff-centric and bash-functionalised.

Aside from external testing in a separate repo, the `pr.yml` workflow itself is the only in-place testing done to date for obvious reasons. Once (if) approved, then I will monitor closely the `continuous` workflow and squash any bugs as a priority. That's probably a merge to take on first thing of a morning! Once that's all working I can handle the `legacy` workflow, and on next release everything might just work.

## 1. Preamble
This PR addresses a few annoying, old, stale, or niggling issues with version updating in the project. The idea is to harmonise and simplify how we manage versions for releases in the repo and keep them consistent with the website etc. The work here assumes that we want to continue with the current release branching strategy (or at least a lighter, more easily-achieved version of it) - i.e. separate branches for releases, with patches appended in those branches and back-merged into `develop`. 

## 2. External-Facing Changes
As well as the repo, the content of the external website changes in this PR:
1) The projectdissolve.com website no longer refers to a specific development version of the code (since only `git-cliff` knows exactly what it will be) but this simplifies its deployment.
2) The projectdissolve.com website no longer lists all legacy releases, because "who cares really", and it simplifies our CI. These are of course still all available via the GitHub repo.
3) Previously the main website was updated from `continuous` pushes, but this meant we had different versions in different files in order to distinguish the development code from the website release. This is now moved to the `release` pipeline, making it the authority on the release version presented on the website.
4) The `README.md` no longer has the current version release number shown at the top of it, because it isn't really necessary and was just a bit awkward to update consistently.

## 3. Daily Work Life (a.k.a. Continuous Builds)
A simple repo may look like this:
```mermaid
---
title: Basic Repo Structure
---
gitGraph
%%{init: { 'gitGraph': {'mainBranchName': 'develop'}} }%%
   commit id: "Initial Commit"
   commit id: "[Release] 0.1.0" tag: "0.1.0"
   commit id: "feat: Feature 1"
   commit id: "feat: Feature 2"
   commit id: "fix: Feature 1"
   commit id: "feat: Feature 3"
```
On merging of a (non-"[Release]") PR the standard `continuous.yml` workflow is in charge of building, packaging, and publishing the continuous version, but what version should this development release have? The main `develop` branch will still be labelled as the latest release "0.1.0" in the source code and configuration files, but the various feature additions and fixes will push this to "0.2.0". The `continuous.yml` workflow will dynamically change the versions accordingly when it runs (but not, of course, commit them back to the `develop` branch) and so every continuous release that gets published will consistently reflect the next release version number. We also get a ChangeLog reflecting all changes since the last release for free, and which is set as the body text of the release.

Worth noting here is a change to the general way the source code is checked out in the workflows. Previously there were multiple checkouts of the code at each stage, but with the version number changing dynamically based on commit history alone it seemed sensible to not do that and have to check / change the version every time. So, the first step in every workflow is a new `_checkout.yml` action which retrieves the entire repository and bumps the version number with git-cliff if it is requested / necessary. The updated source code is then stored in a build artifact which is downloaded by other workflows when required (a copy of the original, unmodified source is also stored for the purposes of code formatting etc.). `_checkout.yml` also handles any re-tagging related to releases, and setting of various version parameters and hashes: this last part was previously the job of the now-removed `setup` action, but explicit passing of values to child actions is now the preferred mechanism.

## 4. Making a Major/Minor Release
Previously, creating a new major/minor release involved branching off from `develop` to create a `release/X.Y.Z` branch which would then build under the `release.yml` workflow and be published automatically. Changelog generation was a manual procedure.

Here we introduce the `prepare_release.yml` workflow which only runs manually from the GitHub Actions page of the repo (i.e. it has a `workflow_dispatch` trigger and no others). On running the workflow on the `develop` branch the new version of the code is determined (as the `continuous.yml` workflow does) but here the modified source code and configuration files committed back to `develop` through an auto-generated PR prefixed by "[Release]".  The idea here is to automate the determination of the next major/minor version, but using the PR as a barrier / choice as to whether to proceed or not, or whether a new release version is even valid / possible- if no changes to the version number with git-cliff are found it will simply fail and not create a release PR.

If successfully  merged, the `detect_release.yml` workflow runs since it is triggered on closed PRs which were successfully merged _and_ whose title begins "[Release]". It's job is to set things up consistently in terms of tagging, and to subsequently generate and publish a full release. This action replaces the existing `release.yml` workflow, and performs the following tasks focussed on the PR merge commit in the `develop` branch:
1. Creates a new tag with the version number determined by `git-cliff` (via `./changeversion -q`).
2. Creates a new branch coming named `release/MAJOR.MINOR.X` if it doesn't already exist. This serves as the entry point for any patch releases to be made should they be required.

So, after all that the repository then looks like this:

```mermaid
---
title: After merge of make_release.yml PR
---
gitGraph
%%{init: { 'gitGraph': {'mainBranchName': 'develop'}} }%%
   commit id: "Initial Commit"
   commit id: "[Release] 0.1.0" tag: "0.1.0"
   commit id: "feat: Feature 1"
   commit id: "feat: Feature 2"
   commit id: "fix: Feature 1"
   commit id: "feat: Feature 3"
   commit id: "[Release] 0.2.0" tag: "0.2.0"
   branch release/0.2.X
   commit id: "(empty)"
   checkout develop
```
(Note that there is no actual commit in the "release/0.2.X" - I just had to add one here to get Mermaid to correctly display the relationship with the PR merge commit on `develop`.)

One final point - following the successful merge of the "[Release]" PR back into `develop`, the `continuous.yml` will run as per normal, bringing the continuous release version also up-to-date with the latest production release.

## 5. Patching an Existing Release

Following the release of 0.2.X work on adding new features to `develop` carries on but we also discover and fix a bug and patch the `release/0.2.X` branch in a couple of commits:

```mermaid
---
title: Patched v0.2.X
---
gitGraph
%%{init: { 'gitGraph': {'mainBranchName': 'develop'}} }%%
   commit id: "Initial Commit"
   commit id: "[Release] 0.1.0" tag: "0.1.0"
   commit id: "feat: Feature 1"
   commit id: "feat: Feature 2"
   commit id: "fix: Feature 1"
   commit id: "feat: Feature 3"
   commit id: "[Release] 0.2.0" tag: "0.2.0"
   branch release/0.2.X
   commit id: "fix: Bug 1"
   commit id: "fix: Bug 2"
   checkout develop
   commit id: "feat: Feature 4"
   commit id: "feat: Feature 5"
```
 In order to push our new patch release we simply run `prepare_release.yml` on the `release/0.2.X` branch rather than `develop`. This will automatically re-version the source code and configuration files, create the tag, but not create the `release/0.2.X` branch because of course it already exists. The fixes can then be back-merged into `develop`, if they haven't been already, and the repo looks like this:
 
```mermaid
---
title: Patched v0.2.X
---
gitGraph
%%{init: { 'gitGraph': {'mainBranchName': 'develop'}} }%%
   commit id: "Initial Commit"
   commit id: "[Release] 0.1.0" tag: "0.1.0"
   commit id: "feat: Feature 1"
   commit id: "feat: Feature 2"
   commit id: "fix: Feature 1"
   commit id: "feat: Feature 3"
   commit id: "[Release] 0.2.0" tag: "0.2.0"
   branch release/0.2.X
   commit id: "fix: Bug 1"
   commit id: "fix: Bug 2"
   commit id: "[Release] 0.2.1" tag: "0.2.1"
   checkout develop
   commit id: "feat: Feature 4"
   commit id: "feat: Feature 5"
   cherry-pick id: "fix: Bug 1"
   cherry-pick id: "fix: Bug 2"
```

## 6. A Note on Version Checking

As noted earlier the `prepare_release.yml` script will bail out if it detects that the version has not changed between the current authoritative version in the source and that determined by git-cliff. In a similar manner, it will also fail if one tries to prepare a release on `develop` which is only a patch version, or conversely if one tries to prepare a release on a `release/?.?.X` branch that is _not_ a patch version.

## 7. A Note on Documentation and the Web Workflow

The `web.yml` workflow is removed in the current PR in favour of having calls to the `website` action made in the relevant top-level workflows. This is enforced by the fact that git-cliff determines (quite rightly) any chance to documentation as a patch on the current release.  This means that updates to documentation must be made as a full CI job (i.e. including build of all targets) while before it could be done separately. However, the present implementation is much more in the spirit of controlled, auto-versioned releases.

## 8. A Note on Creating Releases

Previously we downloaded the `update-release` script from one of our organisational repos and used this to leverage the GitHub API to publish / update releases. Here I have converted the majority of that script to an action `create-release` within the Dissolve workflows, but of course longer-term this could become an action that all organisational projects use.


.github/workflows/get-cliff/action.yml
rprospero, line 1: Love the homophone!

.github/workflows/build/windows/action.yml
```diff
@@ -36,14 +45,14 @@ runs:
...
-      key: windows-qt-${{ env.qtVersion }}
+      key: windows-qt-${{ inputs.qtVersion }}
       path: ${{ runner.temp }}\qt
 
   - name: Install Qt
     # if: ${{ steps.cache-qt.outputs.cache-hit != 'true' }}
     shell: bash
     run: |
-      aqt install-qt --outputdir ${RUNNER_TEMP}/qt windows desktop ${{ env.qtVersion }} win64_msvc2019_64 -m all
+      aqt install-qt --outputdir ${RUNNER_TEMP}/qt windows desktop ${{ inputs.qtVersion }} win64_msvc2019_64 -m all
```
RobBuchananCompPhys, line 55: I'll keep in mind having to make this change in my VS2022 action!

changeversion
```diff
@@ -12,201 +12,259 @@ DaySuffix() {
...
+  if ! GetCodeVersion ${SRC_VERSION}; then RETURN=1; fi
+  exit $(( RETURN ))
+}
+
+PrintUsage() {
+  echo "Usage:  $0 [-v] [-k] [-t] [-q] [-r X.Y.Z] [-s X.Y.Z]"
+  echo "        -r  Explicitly set release version and date in README.md"
+  echo "        -s  Set version explicitly to X.Y.Z"
+  echo "        -v  Get authoritative code version (from src/main/version.cpp)"
+  echo "        -x  Get authoritative code version (from src/main/version.cpp) with patch version as X"
```
rprospero: This flag isn't listed in the usage example.  I wouldn't normally make a big deal out of this (-b,-l, and -L are also not included), but it seems that -x takes an argument and it's a little clear what format the argument should be.  My first thought was -x X.Y.Z, but that makes using X as the patch version a little unclear.  My next assumption is just -x X, but it would be nice to see it in the usage.  Reading the code below, it looks like the call is just -x and that the X isn't a variable to be substitutted but a literal X.

rprospero (approved):
Good job on a lot of hard work.  The new setup is a lot cleaner than what we had before.  I had one minor comment about the docstring for `changeversion`, but everything else looked good to me.

changeversion
trisyoungs: So the `-x` option is there to simplify the formatting of a new branch name for a release. It does not take an argument, but just returns whatever the current code version is with an "X" in place of any patch release.  All patch releases will be branched off from a release branch called, for example, `release/1.2.X` - so tags for 1.2.1, 1.2.2 etc will all exist on the `release/1.2.X` branch.  The `-x` option is just there to make it ease to write simple CI logic in creating / finding this parent branch.

PR #2029, opened 2024-12-16
fix: Correct precedence issue in NeutronSQ Averaging

rprospero:
I got a warning about this line in NeutronSQ module.  Since divide binds faster than comparison, we're taking the inverse of an enum and comparison *that* to another enum, which I'm pretty sure is not the intended action.

If the original code is correct, it could use some documentation to explain the logic.  As I read it now, it looks like we *always* get the average of squares.


trisyoungs (approved):
Yep, that was indeed a problem.  We don't tend to generate data with normalisation applied, so this would have largely gone unnoticed.

trisyoungs:
@rprospero Ah - this is where this fix was previously made!  It has already been merged into `develop` by #2035 so this PR can be closed.

PR #2032, opened 2025-01-06
chore: Update copyright dates

rprospero:
This PR updates all of the copyright text to the new year


trisyoungs (approved):
Always forget that we have to do this!  Along with the Happy New Year also comes a Sad Windows Build, apparently, with some FTGL-related build fails. Second, the Linux QC is broken because `clang-format-13` is no longer in the latest Ubuntu. I did try to put together a quick PR to solve this before we all went on hols, but failed, so perhaps the move to C++20 can be prioritised as I think it already solves this?

rprospero:
@trisyoungs I'll rebase the C++20 build onto this and then try and tackle the windows build issue.

PR #2033, opened 2025-01-08
chore: benchmark fixes

rprospero:
What seems to be the most important part of this PR is that the benchmarks now provide status updates on stderr.  These seem to prevent the process from hanging (I've had two consecutive successful benchmarks).  In the worst case, this will still provide diagnostic information if the problem recurs.


trisyoungs (approved):
Not sure I understand the reasons, but I believe in the solution!

PR #2037, opened 2025-01-16
refactor: Establish Graphable Concept

rprospero:
# This PR merges into GraphViewModules and *not* develop

I'm mostly posting this PR to ensure that the rest of the team will be comfortable with the use of concepts in the module graphs.  One of the biggest changes that comes out of all of this is that several templated functions have been turned into regular, overloaded functions.  It also ensures at compile time that a type is capable of being displayed as a graph.  This check was previously performed at link time with complaints about missing implementations of template functions.

The most interesting files to look at are:

- [src/gui/models/nodeGraph/nodeWrapper.h](https://github.com/disorderedmaterials/dissolve/compare/GraphViewModules...NodeConcept?expand=1#diff-02460a0ca5be38eec6fe68eaf1d8168edc11c2f8c8fc614233be0660ceaa82cd) This is where the concept is defined
- [src/gui/models/nodeGraph/phantom.h](https://github.com/disorderedmaterials/dissolve/compare/GraphViewModules...NodeConcept?expand=1#diff-6a40a72aa2c1cf89d81196b969db7dba99f3d8cfaaa407462502e1a0cd577b11) An implementation of phantom types to allow `nodeRoleNames` to work without being a template function, despite not being passed an instance of the type

### DIFF
--- a/src/gui/models/nodeGraph/generatorGraphModel.h
+++ b/src/gui/models/nodeGraph/generatorGraphModel.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
-// Copyright (c) 2024 Team Dissolve and contributors
+// Copyright (c) 2025 Team Dissolve and contributors
 
 #pragma once
 
--- a/src/gui/models/nodeGraph/generatorGraphModel.h
+++ b/src/gui/models/nodeGraph/generatorGraphModel.h
@@ -10,24 +10,6 @@
 #include "gui/models/nodeGraph/instances/all.h"
 #include "nodeWrapper.h"
 
-// The variant of all of the types that we will examine
-using GeneratorGraphInnerType = std::variant<Configuration *, Generator *, GeneratorNode *>;
-
-// A class to contain the inner type, since we need a constructor that
-// take a QVariant
-class GeneratorGraphNode
-{
-    public:
-    GeneratorGraphNode(QVariant var = {});
-    GeneratorGraphInnerType value;
-};
-
-// All of these types may require access to CoreData
-template <> struct GraphNodeContext<GeneratorGraphNode>
-{
-    using type = CoreData *;
-};
-
 // A graph node model for looking at the generators on a configuration
 class GeneratorGraphModel : public GraphModel<GeneratorGraphNode>
 {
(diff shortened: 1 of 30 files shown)


src/gui/models/nodeGraph/generatorGraphModel.h
```diff
@@ -10,26 +10,8 @@
...
-
-// All of these types may require access to CoreData
-template <> struct GraphNodeContext<GeneratorGraphNode>
-{
-    using type = CoreData *;
-};
-
 // A graph node model for looking at the generators on a configuration
-class GeneratorGraphModel : public GraphModel<GeneratorGraphNode>
+class GeneratorGraphModel : public GraphModel<GeneratorGraphNode, CoreData *>
```
trisyoungs, line 14: It's worth pointing out that this additional `Context` will not be required in the end product since all data will be local to a given graph. The current dependency on `CoreData` as a "data resource" for the code is part of what stopped me in my tracks when doing the generalised custom model PR #1959.

src/gui/models/nodeGraph/instances/all.h
```diff
@@ -1,10 +1,35 @@
...
+void setNodeName(GeneratorGraphNode &value, std::string);
+QVariant nodeGetValue(const GeneratorGraphNode &value);
+bool nodeConnect(GeneratorGraphNode &source, int sourceIndex, GeneratorGraphNode &destionation, int destinationIndex);
+bool nodeConnectable(const GeneratorGraphNode &source, int sourceIndex, const GeneratorGraphNode &destination,
+                     int destinationIndex);
+bool nodeDisconnect(GeneratorGraphNode &source, int sourceIndex, GeneratorGraphNode &destination, int destinationIndex);
+QHash<int, QByteArray> &nodeRoleNames(Phantom<GeneratorGraphNode> proxy, QHash<int, QByteArray> &roles);
+bool nodeDelete(GeneratorGraphNode &item, CoreData *coreData);
+QVariant nodeData(const GeneratorGraphNode &item, int role);
+bool nodeSetData(GeneratorGraphNode &item, const QVariant &value, int role);
```
trisyoungs, line 35: OK, so these sets of "floating" functions taking the specific types to be included in the graph model are what initially confused me, but I get it now. Should / can they be `namespace`d or must they be defined at the most global level?

trisyoungs (approved):
While I fully admit that I need to read-up more on concepts, I get the thrust of the idea and this makes for some nice, clean, template-free code. I think that, in time and once we're used to using them, there are several other places where they can be immediately useful.

PR #2038, opened 2025-01-17
fix: hugoVersion not passed from top-level workflows

trisyoungs:
Fix round 1 following #2028. This is destined for a quick merge without waiting for CI to complete.


Danbr4d (approved):
All looks fine to me

PR #2040, opened 2025-01-17
fix: Missed copyright years in new files.

trisyoungs:
Just a couple of years missed in the new classes added in a recent merge.


Danbr4d (approved):
Beat me to it!

PR #2042, opened 2025-01-21
build: Fix passing of nixHash to QC, run benchmarks as part of main build

trisyoungs:
A few final adjustments to the new CI, in particular removing the separate `benchmarks.yml` workflow and running this as an optional addendum to the main Linux CLI build. 

Still failing on the QC check, but looking good otherwise.  Except, @rprospero, as I have just noticed - the benchmarks is taking an age and basically cripples the Linux CLI build.  It may be stuck, or has hung, or there is another issue at play...


.github/workflows/pr.yml
```diff
@@ -27,11 +24,14 @@ jobs:
...
     needs: [Checkout]
     uses: "./.github/workflows/_qc.yml"
+    with:
+      nixHash: ${{ needs.Checkout.outputs.nixHash }}
 
   BuildAndPackage:
     needs: [Checkout]
     uses: "./.github/workflows/_build_and_package.yml"
     with:
+      benchmark: true
```
rprospero:
```suggestion
      benchmark: false
```
Until I get that part fixed up

rprospero (approved):
I've added a possible temp change to turn of bnechmarks until we get the fix in.

PR #2048, opened 2025-01-31
build: Enable cppCheck tests on ci

rprospero:
With a recent discusson on Mantid trying to close up the thousands of issues detected via cppcheck, I thought it would be worth running it on Dissolve (again?  it feels like we've done this before, but I can't quite remember).  The good news is that we only found fifteen errors, most of which are either from the Space Group Info code or issues with Qt moc.  I've cleaned up the other two errors.  I've also created a QC check to ensure that we're not adding new issues in future PRs.

Running cppcheck adds about two minutes to the QC check, so I've added it at the end.

### DIFF
--- a/.github/workflows/qc/action.yml
+++ b/.github/workflows/qc/action.yml
@@ -81,4 +81,6 @@ runs:
     shell: bash
     run: |
       set -ex
-      nix develop --command scripts/cppCheckTest.sh 0 1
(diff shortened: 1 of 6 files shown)


trisyoungs (approved):
Excellent. And how nice to see that our curation and care has been so good to this point!

PR #2059, opened 2025-02-18
build: Icons QML module

RobBuchananCompPhys:
This compact module should allow us to use Dissolve icons from Dissolve2 QML frontend like `../../IconsModule/clock.svg` or `../../IconsModule/Nodes/box.svg`, rather than going back to `src/gui/icons...`, or having to gather the files together in the top-level CMakeLists.txt.


trisyoungs (approved):
:+1:  Makes sense!  Our icon locations will almost certainly change as we move forward, but this work will still allow us to keep things nicely organised.

PR #2063, opened 2025-02-24
feat: clustering analysis routine through clustering module

aboswel:
Aims to add clustering analysis to dissolve through a new module. Clusters are defined (currently) by two sites and a distance cut-off. Angle constraints will be added for hydrogen bonding. Various metrics are calculated from these clusters. Plans to add 3D GUI visualisation and graph network analysis.


tests/data/Averaging.restart.txt
aboswel, line 1: Not sure what this is?

trisyoungs: Output from unit test for Dan's potential averaging stuff.  It's generated when you run the test so should be removed from this PR.

tests/data/Averaging2.restart.txt
trisyoungs, line 1: This one too!

src/modules/clustering/gui/clusteringWidget.h
```diff
@@ -0,0 +1,42 @@
...
+class ClusteringModuleWidget : public ModuleWidget
+{
+    Q_OBJECT
+
+    private:
+    // Associated Module
+    ClusteringModule *module_;
+    // Pointer to cluster config for visualisation (maybe just use the module config?)
+    Configuration *clusterConfiguration_{nullptr};
+    // I don't understand what this does but the other modules have it
```
trisyoungs: It's just a flag used by some module GUIs to prevent unnecessary or unwanted updates to the GUI when programmatically changing controls (e.g. button states).
```suggestion
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,104 @@
...
+    // Simulation data...
+    // Struct
+    struct BondInfo
+    {
+        const SpeciesSite *a_, *b_;
+        double cutOff;
+
+        BondInfo(const SpeciesSite *site1, const SpeciesSite *site2, double cut) : a_(site1), b_(site2), cutOff(cut) {}
+    };
+
```
trisyoungs: If it was me I would break stuff out of this `BondInfo` structure and just have them as member variables in the module. I get why you added it, but I don't think it's necessary now or in future evolutions of the module TBH.
```suggestion
    // Sites involved in the cluster
    const SpeciesSite *a_, *b_;
    // Maximum allowed distance for two sites to be considered in the same cluster
    double cutoff_{3.0};
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,104 @@
...
+    // GUI site and cutoff selection
+    BondInfo userSites_{nullptr, nullptr, -1};
```
trisyoungs:
```suggestion
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,104 @@
...
+    int maxClusterForConfig_{10000};
+    // Tracks if the clusters have been generated yet
+    bool clustersConfigGenerated_{false};
+    // Pointer to the cluster display configuration
+    Configuration *clusterConfig_{nullptr};
+
+    /*
+     * Processing
+     */
+
```
trisyoungs:
```suggestion
```

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,264 @@
...
+    // Output file for diagnostics
+    if (!parser.openOutput(filename))
+    {
+        Messenger::error("Failed to open output file '{}'.\n", filename);
+        return ExecutionResult::Failed;
+    }
+
+    // Write header with site information
+    parser.writeLineF("# Analysis for sites: {} - {}\n", selectedBonds_[0].a_->parent()->name(),
+                      selectedBonds_[0].b_->parent()->name());
```
trisyoungs: Strictly speaking I would say this is debug output and should be removed, but if you want to keep this in put it behind an `if (false) { ... }` so it doesn't run for the user.

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,264 @@
...
+        const Analyser::SiteVector &siteVectorA = selectionA.sites();
+
+        SiteSelector selectionB(targetConfiguration_, std::vector<const SpeciesSite *>{bond.b_});
+        const Analyser::SiteVector &siteVectorB = selectionB.sites();
+
+        SiteFilter filterA(targetConfiguration_, siteVectorA);
+        std::tie(std::ignore, neighbourMapA) =
+            filterA.filterBySiteProximity(siteVectorB, Range(0.001, bond.cutOff), 1, 100); // min of 0.01 to avoid self-bonding
+
+        // If we're dealing with a bond between the same site, it's already symmetric. Running twice will add dupes
```
trisyoungs: I will say that I'm not keen on the use of the word "bond" here! I think "close contact" or just "contact" is more descriptive of what we're doing.

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,264 @@
...
+            if (!processed.contains(clusterStart))
+            {
+                std::unordered_set<const Site *> visited{clusterStart};
+                buildCluster(clusterStart, visited);
+                processed.insert(visited.begin(), visited.end());
+                clusterMap_.insert({clusterTrack, std::vector<const Site *>(visited.begin(), visited.end())});
+                clusterTrack++;
+            }
+        }
+    }
```
trisyoungs: Any reason why this, and other chunks of code are in their own set of braces?

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,264 @@
...
+            Vec3<double> massWeightedTotalVec{0, 0, 0};
+            const Site *refSite{clusterVec[0]}; // Define a reference site (the first member of the cluster)
+            for (const auto &clusterMem : clusterVec)
+                // Accumlate mass weighted total vector from reference site
+                massWeightedTotalVec +=
+                    (box->minimumVector(refSite->origin(), clusterMem->origin())) * clusterMem->parent()->parent()->mass();
+
+            massWeightedTotalVec /= clusterMasses_[clusterID];
+            clusterCoM_[clusterID] = massWeightedTotalVec;
+            double massWeightedDistanceSqrd{0};
```
trisyoungs:
```suggestion
            auto massWeightedDistanceSqrd = 0.0;
```

trisyoungs (commented):
First batch of comments.  If you want to have a chat about any of this just give me a shout.

src/modules/clustering/clustering.h
aboswel: A better example of why this might be useful is for moonshine: looking at clustering due to hydrophobicity would require Methanol-Methanol, Methanol-Ethanol, and Ethanol-Ethanol to be considered all at once for a complete picture. Will certainly rename it though if you're convinced!

src/modules/clustering/process.cpp
aboswel: I thought it might be better to give these ex-functions their own scope so the temp variables are destroyed immediately when finished? Will get rid of them if you think them not necessary.

src/modules/clustering/process.cpp
aboswel: Fair

src/modules/clustering/process.cpp
aboswel: Out of interest, why?

trisyoungs: There is a trend (which I resisted strongly at first!) to not needlessly use type names wherever possible and use `auto` where the type can be automatically deduced instead. That way, if the class / type you're assigning to the variable ever changes, you don't need to change the type of the assigned variable because it's `auto`. I prefer assignment initialisation rather than braced initialisation because I think the intention is way clearer.

src/modules/clustering/process.cpp
trisyoungs: That's a fair reason, but I think the extra braces introduces a bit of ugliness to the code (and code should be pretty, after all!).  I'd prefer removing them and living with the temporary var lasting until the end of function scope.

src/modules/clustering/clustering.h
trisyoungs: Coulda sworn I replied to this one this morning at around 10:20, but apparently not! I get why you would want to use a struct, but if you do that you will also need a custom keyword to handle it, a custom keyword widget to display/edit it etc.  In the simple use case (which we should be focusing on) and in the one you've cited if we're talking about hydrogen bond networks then the cutoff is more-or-less going to be the same between all the different OH groups, so one can define general A and B sites containing sites from the different molecules.  The analysis, however, can freely partition its results and output by site type if it wanted to.

If there are more complicated cases where there are two distinctly different types of contacts related to the cluster, this sounds like a module extension or entirely new module.

src/modules/clustering/clustering.cpp
```diff
@@ -0,0 +1,24 @@
...
+    // Perhaps add an option to disable evolution from here for convenience?
+    executeIfTargetsUnchanged_ = true;
+
+    keywords_.addTarget<ConfigurationKeyword>("Configuration", "Set target configuration for the module", targetConfiguration_)
+        ->setEditSignals({KeywordBase::ClearModuleData, KeywordBase::RecreateRenderables});
+
+    keywords_.setOrganisation("Options", "Sites");
+    keywords_.add<SpeciesSiteKeyword>("SiteA", "Choose the first site for cluster definition", a_);
+    keywords_.add<SpeciesSiteKeyword>("SiteB", "Choose the second site for cluster definition", b_);
+    keywords_.add<DoubleKeyword>("Cut-off", "The maximum distance between sites for them to be considered clustered",
```
trisyoungs:
```suggestion
    keywords_.add<DoubleKeyword>("Cut-off", "The maximum distance between sites for them to be considered part of the same cluster",
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,86 @@
...
+    ~ClusteringModule() override = default;
+
+    /*
+     * Definition
+     */
+    private:
+    // Target configurations
+    Configuration *targetConfiguration_{nullptr};
+
+    // Simulation data...
```
trisyoungs:
```suggestion
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,86 @@
...
+    // Maximum allowed distance for the two sites to be considered in the same cluster
+    double cutoff_{3.0}; 
+    // Symmetric map of all sites to all neighbour sites
+    Analyser::SiteMap neighbourMap_{};
+    // Map of every member in every cluster
+    std::map<int, std::vector<const Site *>> clusterMap_{};
+    // Map of size to no. of clusters of that size
+    std::map<int, std::vector<int>> sizeDistribution_{};
+    // Map of cluster ID to mass of cluster
+    std::map<int, float> clusterMasses_{};
```
trisyoungs: Here and throughout, I think we can afford a `double` rather than a `float`!

src/modules/clustering/gui/clusteringWidgetFuncs.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    // Check eveything is ready
+    if (!module_->viewingReady)
+    {
+        refreshing_ = false;
+        return;
+    }
+    // Set the new configuration and redraw the visualisation. Is there a way to set the style to spheres view from here (lines
+    // arent showing)?
+    clusterConfiguration_ = module_->generateClustersConfig(&dissolve_, module_->getSourceConfig(), displaySize_, displayID_);
```
trisyoungs: Instead of defining a function explicitly to return the target `Configuration*` you should use the `KeywordList` object to get what you need:
```
module_->keywords().getConfiguration("Configuration");
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,86 @@
...
+    // Set up module for processing
+    bool setUp(ModuleContext &moduleContext, Flags<KeywordBase::KeywordSignal> actionSignals) override;
+    // Generation of the cluster visualisation configuration
+    Configuration *generateClustersConfig(Dissolve *dissolve, Configuration *source, int displaySize, int displayID);
+    // Getter for the target configuration of the module
+    Configuration *getSourceConfig();
+    // Get the size distribution
+    std::map<int, std::vector<int>> &getSizeDistribution() { return sizeDistribution_; }
+    // Ensures we don't visualise the config before it's ready (still necessary?)
+    bool viewingReady{false};
```
trisyoungs: If this is still necessary it should be a private variable with a getter if the UI needs it, unless there's a better way of checking for what you need to check?

src/modules/clustering/gui/clusteringWidget.h
```diff
@@ -0,0 +1,48 @@
...
+     * UI
+     */
+    private:
+    // Main form declaration
+    Ui::ClusteringModuleWidget ui_;
+    DataViewer *sizeDist_, *massDist_;
+
+    private Q_SLOTS:
+    void on_listWidget_itemClicked(QListWidgetItem *item);
+    void on_listWidget2_itemClicked(QListWidgetItem *item);
```
trisyoungs: Better names required here I think!

src/modules/clustering/gui/clusteringWidgetFuncs.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+void ClusteringModuleWidget::buildSizeList()
+{
+    // Produce a list of all the different cluster sizes
+    auto sizeDistribution = module_->getSizeDistribution();
+    ui_.listWidget->clear();
+    ui_.listWidget->addItem("View All");
+    for (const auto &[size, _] : sizeDistribution)
+    {
+        QListWidgetItem *item = new QListWidgetItem(QString::number(size));
```
trisyoungs:
```suggestion
        auto *item = new QListWidgetItem(QString::number(size));
```

src/modules/clustering/gui/clusteringWidgetFuncs.cpp
```diff
@@ -0,0 +1,135 @@
...
+        displaySize_ = 0;
+        updateControls();
+        return;
+    }
+
+    ui_.listWidget2->addItem("View All");
+    displaySize_ = item->text().toInt();
+    for (const auto &ID : sizeDistribution[displaySize_])
+    {
+        QListWidgetItem *cell = new QListWidgetItem(QString::number(ID));
```
trisyoungs:
```suggestion
        auto *cell = new QListWidgetItem(QString::number(ID));
```

src/modules/clustering/gui/clusteringWidgetFuncs.cpp
```diff
@@ -0,0 +1,135 @@
...
+{
+    if (item->text() == "View All")
+    {
+        displayID_ = 0;
+        updateControls();
+        return;
+    }
+
+    displayID_ = item->text().toInt();
+    updateControls();
```
trisyoungs:
```suggestion
    displayID_ = item->text() == "View All" ? 0 : item->text().toInt();
    updateControls();
```

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
+{
+    auto &moduleData = moduleContext.dissolve().processingModuleData();
+
+    if (!(a_ && b_ && (cutoff_ > 0)))
+    {
+        Messenger::error("Cluster definition invalid!");
+        return ExecutionResult::Failed;
+    }
```
trisyoungs: Better to perform this check in `setUp()` rather than in `process()` - that way it is done once before any module runs rather than each time it runs.

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+    // Process code
+    // Produce NeighbourMap
+    neighbourMap_.clear();
+    Analyser::SiteMap neighbourMapA, neighbourMapB;
+
+    // Transform SiteObjects to instances ready for the filter function
+    SiteSelector selectionA(targetConfiguration_, std::vector<const SpeciesSite *>{a_});
+    const Analyser::SiteVector &siteVectorA = selectionA.sites();
```
trisyoungs:
```suggestion
    const auto &siteVectorA = selectionA.sites();
```

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+    SiteSelector selectionB(targetConfiguration_, std::vector<const SpeciesSite *>{b_});
+    const Analyser::SiteVector &siteVectorB = selectionB.sites();
```
trisyoungs:
```suggestion
    const auto &siteVectorB = selectionB.sites();
```

src/modules/registry.cpp
```diff
@@ -53,6 +54,7 @@ ModuleRegistry::ModuleRegistry()
     registerProducer<BenchmarkModule>(ModuleTypes::Benchmark, "Perform benchmarking on a variety of functions",
                                       "Checks & Tests");
     registerProducer<BraggModule>(ModuleTypes::Bragg, "Calculate Bragg intensities", "Correlation Functions");
+    registerProducer<ClusteringModule>(ModuleTypes::Clustering, "Analyse molecular clustering", "Analysis");
```
trisyoungs: Better to say "Analyse clustering between sites"?

trisyoungs (commented):
Looking really good now, and from what you showed earlier, working well too!

A sprinkling of comments, mostly stylistic.

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,97 @@
...
+#include "base/processPool.h"
+#include "classes/configuration.h"
+#include "classes/coreData.h"
+#include "generator/context.h"
+#include "main/dissolve.h"
+#include "module/context.h"
+#include "module/module.h"
+#include <unordered_set>
+
+class Atom;
```
trisyoungs, line 18:
```suggestion
// Forward Declarations
class Atom;
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,97 @@
...
+{
+    public:
+    ClusteringModule();
+    ~ClusteringModule() override = default;
+
+    /*
+     * Definition
+     */
+    private:
+    // WIP
```
trisyoungs: Still WIP, or can we put a proper comment here?

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,97 @@
...
+    std::unordered_map<const SpeciesSite *, std::unordered_set<int>> hydroxylIndexes_;
+    bool strict_{false};
+    double angleDev_{20};
+    // Check angles
+    std::map<const Site *, std::pair<const Site *, double>> angleMap;
```
trisyoungs: If this is a `private` variable it should be suffixed with `_`.

src/modules/clustering/gui/clusteringWidgetFuncs.cpp
```diff
@@ -0,0 +1,159 @@
...
+    sizeDist_->postRedisplay();
+    massDist_->postRedisplay();
+
+    fromBuilder_ = false;
+    refreshing_ = false;
+}
+
+void ClusteringModuleWidget::buildSizeList()
+{
+    // Produce a list of all the different cluster sizes
```
trisyoungs:
```suggestion
// Produce a list of all the different cluster sizes
void ClusteringModuleWidget::buildSizeList()
{
```

src/modules/clustering/gui/clusteringWidgetFuncs.cpp
```diff
@@ -0,0 +1,159 @@
...
+    {
+        auto *item = new QListWidgetItem(QString::number(size));
+        item->setFlags(item->flags() | Qt::ItemIsEnabled | Qt::ItemIsSelectable);
+        ui_.clusterSizeList->addItem(item);
+    }
+}
+
+void ClusteringModuleWidget::buildIDList(QListWidgetItem *item)
+{
+    // Produce a list of all the cluster IDs of the selected cluster size
```
trisyoungs:
```suggestion
// Produce a list of all the cluster IDs of the selected cluster size
void ClusteringModuleWidget::buildIDList(QListWidgetItem *item)
{
```

src/modules/clustering/gui/clusteringWidgetFuncs.cpp
```diff
@@ -0,0 +1,159 @@
...
+    displaySize_ = item->text().toInt();
+    for (const auto &ID : sizeDistribution[displaySize_])
+    {
+        auto *cell = new QListWidgetItem(QString::number(ID));
+        cell->setFlags(cell->flags() | Qt::ItemIsEnabled | Qt::ItemIsSelectable);
+        ui_.clusterIDList->addItem(cell);
+    }
+}
+
+void ClusteringModuleWidget::buildCNList()
```
trisyoungs, line 106:
```suggestion
// Build list of available coordination numbers
void ClusteringModuleWidget::buildCNList()
```

src/modules/clustering/gui/clusteringWidget.h
```diff
@@ -0,0 +1,56 @@
...
+    bool getNewConfig_{false};
+
+    public:
+    ClusteringModuleWidget(QWidget *parent, ClusteringModule *module, Dissolve &dissolve);
+    void updateControls(const Flags<ModuleWidget::UpdateFlags> &updateFlags = {}) override;
+
+    private:
+    void buildSizeList();
+    void buildIDList(QListWidgetItem *item);
+    void buildCNList();
```
trisyoungs:
```suggestion
    // Produce a list of all the different cluster sizes
    void buildSizeList();
    // Produce a list of all the cluster IDs of the selected cluster size
    void buildIDList(QListWidgetItem *item);
    // Build list of available coordination numbers
    void buildCNList();
```

src/modules/clustering/clustering.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+    keywords_.setOrganisation("Options", "Hydrogen Bonding");
+    keywords_.add<BoolKeyword>("Directional Hydrogen Bonding?",
+                               "Static sites based on (at least one) single hydroxyl oxygen atoms must be used", strict_);
+    keywords_.add<DoubleKeyword>("Maximum bond angle deviation", "+- from 180 degrees", angleDev_);
+
+    keywords_.setOrganisation("Export", "Options");
+    keywords_.add<BoolKeyword>("Export Size Distribution", "SizeDist.txt", saveSizeDist_);
+    keywords_.add<BoolKeyword>("Export Mass Distribution", "MassDist.txt", saveMassDist_);
+    keywords_.add<BoolKeyword>("Export Fractal Dimension and Radius of Gyration - Cluster Mass", "RgMass.txt", saveRgMass_);
```
trisyoungs: These keyword names are kinda huge and need shortening, and no spaces is the norm. Suggestions below. Also the descriptive text needs to be more... descriptive!
```suggestion
    keywords_.add<BoolKeyword>("ExportSize", "Export size distribution to SizeDist.txt", saveSizeDist_);
    keywords_.add<BoolKeyword>("ExportMass", "Export mass distribution to MassDist.txt", saveMassDist_);
    keywords_.add<BoolKeyword>("ExportAnalysis", "Export fractal dimension and radius of gyration to RgMass.txt", saveRgMass_);
```

src/modules/clustering/clustering.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+    keywords_.setOrganisation("Options", "Sites");
+    keywords_.add<SpeciesSiteKeyword>("SiteA", "Choose the first site for cluster definition", a_);
+    keywords_.add<SpeciesSiteKeyword>("SiteB", "Choose the second site for cluster definition", b_);
+    keywords_.add<DoubleKeyword>(
+        "Cut-off", "The maximum distance between sites for them to be considered part of the same cluster", cutoff_);
+
+    keywords_.setOrganisation("Options", "Hydrogen Bonding");
+    keywords_.add<BoolKeyword>("Directional Hydrogen Bonding?",
+                               "Static sites based on (at least one) single hydroxyl oxygen atoms must be used", strict_);
```
trisyoungs: Keyword name (as indicated below for others):
```suggestion
    keywords_.add<BoolKeyword>("Directional Hydrogen Bonding?",
                               "Consider directional hydrogen bonding (static sites based on at least one hydroxyl oxygen atom must be used)", strict_);
```

src/modules/clustering/clustering.h
```diff
@@ -0,0 +1,97 @@
...
+    ClusteringModule();
+    ~ClusteringModule() override = default;
+
+    /*
+     * Definition
+     */
+    private:
+    // WIP
+    std::unordered_map<const SpeciesSite *, std::unordered_set<int>> hydroxylIndexes_;
+    bool strict_{false};
```
trisyoungs: If  the `strict_` variable determines whether directional HB is considered, it should be renamed to reflect that more clearly.

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,392 @@
...
+            const auto &hIdx = hydroxylIndexes_[site->parent()];
+            // Check the site is hydroxyl
+            if (hIdx.empty())
+                continue;
+
+            // Iterate neighbours
+            for (auto it = neighbours.begin(); it != neighbours.end();)
+            {
+                auto oOVec = box->minimumVector(site->origin(), std::get<0>(*it)->origin());
+                bool keep = false;
```
trisyoungs:
```suggestion
                auto keep = false;
```

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,392 @@
...
+                    // Get the relevant vectors
+                    auto oHVec = box->minimumVector(site->origin(), site->molecule()->atom(h)->r());
+                    auto angle = box->angleInDegrees(oOVec / oOVec.magnitude(), oHVec / oHVec.magnitude());
+
+                    // Make sure we have the smallest angle possible
+                    if (360.0 - angle < angle)
+                        angle = 360.0 - angle;
+
+                    if (angle <= angleDev_)
+                        keep = true;
```
trisyoungs: If you only care whether one of the neighbour sites is within the angle range, you could `break` here for efficiency..

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,392 @@
...
+                    tempMap[std::get<0>(*it)].emplace_back(Analyser::SiteData(site, 0));
+                }
+                it++;
+            }
+        }
+        neighbourMap_ = tempMap;
+    }
+
+    if (neighbourMap_.empty())
+        Messenger::error("No neighbours found!");
```
trisyoungs: Under these circumstances should the rest of the routine continue? Should this be a warning followed by a `return` instead?

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,392 @@
...
+            if (!o)
+            {
+                Messenger::error("The origin atom for site is inaccessible.");
+                return false;
+            }
+            if (o->Z() != Elements::O)
+            {
+                Messenger::error("For directional hydrogen bonding, the static origin atom for site must be an Oxygen.");
+                return false;
+            }
```
trisyoungs: Or a nitrogen. Or, in some cases, even a carbon.  This is one reason why I was keen on using the NETA-defined sites since this puts the burden of what is a suitable centre and H site fully on the user (as it should be!).

trisyoungs (commented):
Yet another handful of little comments, but this is looking really good!

src/modules/clustering/process.cpp
aboswel: Looking back on this now with fresh eyes I see a much better way to do this which would inherently allow all site types and is user centric...

src/gui/icons/modules/Clustering.svg
trisyoungs, line 1: Nice!

src/modules/clustering/clustering.cpp
```diff
@@ -0,0 +1,43 @@
...
+                               selfClustering_);
+
+    keywords_.setOrganisation("Options", "Directional Hydrogen Bonding");
+    keywords_.add<BoolKeyword>(
+        "DirectionalHBonding",
+        "Static and Dynamics sites should be based on a single atom/element/atom type, with hydrogens bonded to it. Fragment "
+        "sites can be defined for more flexibility, requiring a #origin and #h tag for the origin atom and hydrogen (or other) "
+        "group",
+        directional_);
+    keywords_.add<DoubleKeyword>("AngleDeviation", "Maximum angle deviation from 180 to be considered a valid interaction",
```
trisyoungs: Do you think "AngleTolerance" is a better name for this?

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,485 @@
...
+    if (!(a_ && b_ && (cutoff_ > 0)))
+        Messenger::error("Cluster definition invalid! Set both sites and a positive cutoff.");
+
+    // If we have strict bonding, we need to check and determine index map for hydroxyl group
+    if (directional_)
+    {
+        directionIndexes_.clear();
+        for (const auto &s : {a_, b_})
+        {
+            if (s->type() != SpeciesSite::SiteType::Fragment)
```
trisyoungs: For me, I always find that affirming the positive case in the `if` aids readability - in other words check for `== SpeciesSite::SiteType::Fragment` first.

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,485 @@
...
+                            {
+                                Messenger::error("Inaccessible bond partner found, skipping...");
+                                continue;
+                            }
+                            if (atom->Z() == Elements::H)
+                                directionIndexes_[s].emplace(atom->index());
+                        }
+                }
+            }
+            else if (s->type() == SpeciesSite::SiteType::Fragment)
```
trisyoungs: Don't need to check for the opposite truth - can just have `else` on its own.

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,485 @@
...
+            {
+                // For a fragment site, we're relying on "#origin, -H(#h)" or similar, with hydrogens (or whatever desired in
+                // theory should work) tagged as h If the number of hydrogens exceeds one, the definition should reflect that
+                // else only one will be calculated i.e. -H(n=2,#h)
```
trisyoungs: Now this comment made me think - if you have a water molecule and your fragment site definition origin is the oxygen you will only ever generate one site definition. To get distinct ones involving each H you'll have to flip the definition around, and have #origin as the H and #h (or perhaps #other) as the O.

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,485 @@
...
+                auto instances = s->instances();
+                for (const auto &instance : instances)
+                {
+                    if (instance.originIndices().size() != 1)
+                    {
+                        Messenger::error("NETA defined sites for directional clustering must have a single origin atom!");
+                        return false;
+                    }
```
trisyoungs: `Messenger::error` returns `false` so you can do this in one line:
```suggestion
                    {
                        return Messenger::error("NETA defined sites for directional clustering must have a single origin atom!");
                    }
```

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,485 @@
...
+                    // For each instance find the tagged group
+                    auto &origin = s->parent()->atom(instance.originIndices()[0]);
+                    auto identifiers = s->fragment().matchedPath(&origin).identifiers();
+                    auto it = identifiers.find("h");
+                    if (it == identifiers.end())
+                    {
+                        Messenger::error("NETA defined sites for directional clustering must include a specified group tagged "
+                                         "with #h e.g. '#origin, -H(#h)' - see NETA documentation for more detail");
+                        return false;
```
trisyoungs:
```suggestion
                        return Messenger::error("NETA defined sites for directional clustering must include a specified group tagged "
                                         "with #h e.g. '#origin, -H(#h)' - see NETA documentation for more detail");
```

src/modules/clustering/process.cpp
```diff
@@ -0,0 +1,485 @@
...
+                        directionIndexes_[s].emplace(atom->index());
+                    }
+                }
+            }
+        }
+        // Complain if we don't find any valid hydrogens
+        if (directionIndexes_.empty())
+        {
+            Messenger::error("Failed to find hydroxyl hydrogens - check site set-up!");
+            return false;
```
trisyoungs:
```suggestion
            return Messenger::error("Failed to find hydroxyl hydrogens - check site set-up!");
```

trisyoungs (commented):
There's an important comment regarding how the fragment sites work - the others are just style.

src/modules/clustering/process.cpp
aboswel: With the way I've done this we only need one site, so long as all the hydrogens (or anything else) that the user wants angle checking is included in this #h group (I agree #other is probably a better name!). The code then creates a map of the site to all the indexes of atoms in the #h group for easy retrieval later. It seems to work fine in testing so long as all the H are included in the tag (hence me leaving that comment).

trisyoungs: OK, I see now. Being honest I think I prefer the reverse way of defining things (i.e. H first) as this then makes defining what the H-bonding sites are more explicit. That said, this will only ever be required for water, ammonia / ammonium, and other "exotics" so probably not that big of an issue!

aboswel: I agree. I didn't consider defining it the other way around. In theory there's nothing stopping this but trying `#other, -O(#origin)` doesn't seem to tag the hydrogens as #other for water?

trisyoungs: So the full fragment site description would be `?H,#other,-O(#origin)`.  Not sure if that's what you have, but if I understand the fragment site system (and I definitely knew 90% of it at some point :) ) then this should work and give you two hydrogens for water.

web/docs/userguide/modules/analysis/clustering/_index.md
```diff
@@ -0,0 +1,60 @@
...
+
+Fragment sites require a NETA defintion, for which there is a separate docs page. Beyond that, there are some specific requirements for using Fragment sites with `Clustering`. A typical clustering compatible NETA defintion for a -X-H group (for some origin element X) is `-X(#origin, -H(n=2, #other))` meaning: we're looking for a bond to an X (which is tagged as the origin) where this X has bonded to it two (n=2) hydrogens which will both be tagged as #other. There are a few things to note here:
```
trisyoungs:
```suggestion
Fragment sites require a [NETA connectivity definition]({{< ref neta >}}). Beyond that, there are some specific requirements for using Fragment sites with `Clustering`. A typical clustering compatible NETA defintion for a -X-H group (for some origin element X) is `-X(#origin, -H(n=2, #other))` meaning: we're looking for a bond to an X (which is tagged as the origin) where this X has bonded to it two (n=2) hydrogens which will both be tagged as #other. There are a few things to note here: 
```

trisyoungs (approved):
Top effort on the documentation.
I think this is now a good point to merge in.  Any issues you find moving forward can be handled in small bugfix PRs.

PR #2064, opened 2025-02-25
refactor: Nodes and Parameters

trisyoungs:
A draft PR showing my journey into nodes and parameters.  There are of course lots of questions to ask here about how I've done this vs how we should be doing it, and predictably the way I've gone about it is close to what we had before.


RobBuchananCompPhys:
Good to see the parameters being fleshed out as they emerge from Dissolve-1 `keywords`! 

I'm focussing in more on what you've done for the `node` base class, which is looking very good. I am wondering if there is a way for us to meet in the middle on this, first of all using the node `registry` I drafted the other day. Things like the `name` and `description` can be added to the `Attributes` struct. The actual parameters of the node can also be intialised at compile time as lists of stack-allocated `Parameter` sub-class instances inside node `registry`. Is there an advantage to this over making them `new` as done in the `addInput` method I wonder?

Edit: Just realised why my approach to initialising parameters may be inadequate - I believe a list of `Parameter` objects would mean not utilitising the polymorphism of `Dervied P = new Parameter()`, which is done in the `addInputs` method

rprospero:
As I'm looking at the parameter objects, I'm going to try building off of this with a template parameter class to see if we can avoid needing to create individual classes for all of the parameter types.

trisyoungs:
@RobBuchananCompPhys I would just be concerned that moving this kind of data to a `NodeRegistry` makes sense from the perspective of the registry itself, but then I would be defining the name and description etc. outside of the individual `Node`s themselves, separating what feels like "core data" from the core definition, unless I'm misunderstanding?

I guess my response is the same for the `Parameter`s - again defining them in a `NodeRegistry` entry for a given `Node` makes sense in terms of centralised access, but creates a separation between the `Node` and how it is used. If we did this, then `Node`s would have no knowledge of what are its inputs and what are its outputs, which might make implementing the actual graph traversal harder than it needs to be.  Again, unless I'm misunderstanding!

trisyoungs:
> As I'm looking at the parameter objects, I'm going to try building off of this with a template parameter class to see if we can avoid needing to create individual classes for all of the parameter types.

This is something that occurred while I was writing the simplified `Parameter` classes, but I sidelined that thought because it's not just input types that change between `Parameter`s but also, for instance, how they are clamped or range-restricted.  That said, if there's a good way of generalising / templating then I'm all for it.

src/nodes/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_library(nodes atomShake.cpp atomShake.h node.cpp node.h parameter.cpp parameter.h)
+
+target_include_directories(nodes PRIVATE ${PROJECT_SOURCE_DIR}/src)
+target_link_libraries(nodes PUBLIC base ${THREADING_LINK_LIBS})
+
```
rprospero, line 5:
```suggestion
```

Fix formatting

rprospero (approved):

PR #2065, opened 2025-02-25
refactor: Use templates for parameters

rprospero:
This creates a templated class for the parameters instead of using a new class for each parameter type.  As a bit of proof of concept, there's also a subclass for bounded parameters to handle numeric types that might need extra context.  We can always add more of these as needed.

Edit: Switched to `shared_ptr` from `unique_ptr` to enable the upcasting capability.  Some of my experiences with the QML nodes indicate that this might be helpful in the future).


trisyoungs (commented):
OK, this is looking pretty nice!  I have some questions and queries as this evolves:

1) What about "even more complex" data types - I'm initially thinking of the old `OptionalDoubleKeyword` and `OptionalIntegerKeyword`.  There the default value is based around `std::optional` which might cause some subtle issues.
2) Now that this is fully templated, does this affect how we expose / iterate the vector of inputs or outputs on the QML side?

trisyoungs:
Ah, just remembered my third comment!

3) It feels like we shouldn't pass the default values for parameters as part of the `addInput*` methods, since this means we can / will define default values in two separate places (i.e. in the header file or in the `Node` constructor as well as when we register keywords).  If we can transparently just copy the default data over from the provided source variable this would make things clearer.

rprospero:
@trisyoungs 

Going out of order, point #3 is excellent and I've already made the corresponding changes to the repo.  For #1, I don't believe that we should experience any issues with complex data types, but I'll prototype one and make sure that it doesn't bring any issues.

As for point #2, I see two ways of handling the parameters.  In the trivial way, the QML model wouldn't need much beyond that `ParameterBase`.  All the node would show is a list of names with different coloured connection points that indicate the desired data type.  All of that information is available in `ParameterBase` and the trivial solution wouldn't even touch the template classes.

The caveat to the trivial solution is that it's displaying the names and connection points, but not the *values*.  It we want that, then it does become more complicated, though not terribly so.  The model still gets the datatype, which is passed to the node to select the correct delegate, which can then request the rest of the data from the model.  It's a bit of a hassle, but far less than my original plan of creating a full delegate for each node.

PR #2066, opened 2025-02-25
chore: added concurrency group to the pr workflow to automatically cancel previous workflow

RabiyaF:
This is supposed to be another update for sustainability and optimising resources. I noticed when I pushed two subsequent changes, the previous workflow keep running. As can be seen in the actions list below.

image

I have added [concurrency](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs) to `pr.yml` to stop the previous workflow.

I have tested it as well and it work fine. This can be verified from the the actions list.
image

The workflow associated with the latest commit will run and jobs from the previous commit will automatically be cancelled.
image

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -14,6 +14,10 @@ on:
     - '.github/ISSUE_TEMPLATE/**'
     - 'README.md'
 
+concurrency:
+  group: pull-request-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
 
   Checkout:
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ ui_*
 moc_*
 *.kdev4
 build/
+!/.github/workflows/build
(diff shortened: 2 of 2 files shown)


RabiyaF:
I tried updating the Windows workflow by using `conan 2.*`. I could not fix it. I see there is a PR to disable those tests. The other issue here is the failing linux test. I think those are failing because of the steps between the `unholy hack` comments. It is permission error. Maybe Forks do not have the right permissions.

```
remote: Permission to disorderedmaterials/dissolve.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/disorderedmaterials/dissolve.git/': The requested URL returned error: 403
```

But what I added in this PR is the concurrency group and that is working fine.

RabiyaF:
Closing this PR because it was created through a fork. Similar changes will be added through a branch on the main repo.

PR #2067, opened 2025-02-26
refactor: node registry

RobBuchananCompPhys:


trisyoungs (commented):
I think this is about the only way we can do this at present - although I dislike the repetition of the node name (duplicating what we define in the `Node` itself), I don't see a way around it.

I have one comment which is that writing out the whole `[]() -> std::unique_ptr<Node> { return std::make_unique<AtomShakeNode>(); }` is going to look pretty heavy once we have a page full of `Node` types. Perhaps we could define a template function in the main `Registry` class to handle this and reduce the boilerplate in the code?

PR #2068, opened 2025-02-27
feat: Connect nodes and propagate changes

rprospero:
This PR adds the ability to create connects between parameters.  It also adds a process to recursively update node information and a test to confirm that this functionality works.


rprospero:
@trisyoungs As per conversation from this morning, I've updated the graph propagation to ensure that calculations aren't repeated unless necessary.  Part of this is through respecting the presence (or, most importantly, absence) of the Invalidation flag.

Another part of this comes through two different routes for node invalidation.  When dealing with a parameter and using the setter, the node is already marked for invalidation if necessary.  However, within module code, the variables can still be accessed directly (for better developer ergonomics).  In that case, that's simply an `invalidate` function that can be manually called if your function would require the node to be recalculated.

src/nodes/node.cpp
```diff
@@ -11,17 +11,46 @@
...
+bool Node::preprocess()
+{
+    for (auto &[key, link] : links_)
+    {
+        // Ignore parameters that don't invalidate
+        if (!link.sink().flags().isSet(ParameterBase::Invalidates))
+            continue;
+        // Update unsatisfied sources
+        if (!(link.source().parent()->isSatisfied() || link.updateSource()))
+            return false;
```
trisyoungs: I think we should elaborate on what potential states a `Node` could find itself in here, just for clarity.  If a node is missing one or more "required" data inputs (e.g. pointer types such as target `Configuration`s or `Species` fall neatly into this category since their default value will be `nullptr`) then the `Node` cannot proceed. This is separate from "data consistency" which the `ParameterBase::Invalidates` flag was focussed on - that is, if an input with this flag is modified (set / unset / re-set) then any calculated data the node might be storing is now invalid and should be deleted.  For example, a `Node` which calculates a radial distribution function between two `SpeciesSite`s cannot switch those sites half way through. I suppose what I'm trying to say is that there is a distinction between having all inputs satisfied, having to invalidate data, and being able to run, which I don't think we're clearly capturing here.  We may need to add a `ParameterBase::Required` flag or similar to clean things up.

src/nodes/node.cpp
```diff
@@ -11,17 +11,46 @@
...
+    if (satisfied_)
+        return true;
+    for (auto &[name, link] : links_)
+        if (!link.source().parent()->isSatisfied())
+            satisfied_ = false;
+    return satisfied_;
+}
+
+// Tell node to recalculate results
+void Node::invalidate() { satisfied_ = false; }
```
trisyoungs, line 70: Related to my point above, this isn't really about recalculating results which could be long term averages over many iterations, but is more about being ready to run.  I wonder if it's worth not storing `satisfied_` and just calculating it every time it is required.  Since this would only be done once just prior to when a Graph runs, performance shouldn't be an issue.

src/nodes/node.h
```diff
@@ -23,22 +25,51 @@ class Node
...
 
     public:
+    // Link an input
+    bool link(std::string_view name, ParameterBase &source)
+    {
+
+        // Confirm that this node hasn't already been linked
+        if (std::find_if(links_.begin(), links_.end(), [name](const auto &it) { return name == it.second.sink().name(); }) !=
+            links_.end())
+            return false;
```
trisyoungs: I'm not sure what this is meant to prevent?  We can have multiple inputs to a `Node` from another single `Node` in principle.

src/nodes/parameterLink.h
```diff
@@ -0,0 +1,38 @@
...
+#include "nodes/parameter.h"
+
+// A link between two parameters
+class ParameterLink
+{
+
+    // This class uses references instead of pointers for two reasons:
+    // 1) Neither end of the link should EVER be null
+    // 2) The link itself is immutable.  You can create links and
+    // destroy links, but *never mutate links*.
```
trisyoungs, line 16: I like the interplay between the `assign()` function and `ParameterLink` itself - that's nice and neat.  I also agree that references are a sensible choice here, and keeping links immutable is a good design choice. There is a slight edge case where a link could appear to be mutable to the user, but we can hide this in the code - for an existing link I may click on either end of the connection and drag that point to a new input / output.  That's a valid action on the part of the user, and appears to mutate the link, but we can of course just lie and have a temporary "one-ended" floating link for visual purposes.

trisyoungs (approved):
Good work in here.  How this evolves will become clearer once we start putting together our first test graph in earnest.

src/nodes/node.h
rprospero: I need to make the comment clearer.  While we will have multiple links into the same `Node`, we should not have multiple links into the same `Parameter` of a node.  For example, the `Configuration` should never have two different *inbound* links, since it would be unclear which configuration to use.

PR #2070, opened 2025-02-28
fix: potential range check dialog appearing unnecessarily

aboswel:
Range check operating unnecessarily on not yet generated configurations.

Fix:
Range check no longer checks empty configurations 
Range check occurs after a prepare() failure, to auto-adjust if necessary.


trisyoungs (commented):
Looks good. Just picking up some CI errors for some reason...

trisyoungs (approved):

PR #2074, opened 2025-03-04
feat: Node and edge containers

RobBuchananCompPhys:
Provides a pair of containers (mapping of string->node, vector of edges) at the core level.
Also introduces a type-alias in `Node` to refer to a map of `ParameterLink`.

`Graph` derives from `Node` - the former contains a parent `std::unique_ptr<Node>`, while the latter contains a raw pointer to its parent `Graph`. `Dissolve` owns a singleton top-level graph which contains all nodes (apart from the `DissolveNode`).

graphNode

### DIFF
--- a/src/nodes/graph.h
+++ b/src/nodes/graph.h
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#pragma once
+
+#include "base/messenger.h"
+#include "nodes/node.h"
+#include "nodes/registry.h"
+#include <map>
+#include <string>
+#include <vector>
+
+// Graph
+class Graph
+{
+    public:
+    Graph() {};
+    ~Graph() = default;
+
+    /*
+     * Nodes and edges
+     */
+    using Nodes = std::map<std::string_view, std::unique_ptr<Node>>;
+    using Edges = std::vector<Node::Edge>;
+
+    private:
+    // Container of nodes
+    Nodes nodes_{};
+    // Container of parameter links between nodes
+    Edges edges_{};
+
+    public:
+    // Produce a registered node by type
+    static std::unique_ptr<Node> produceNode(const std::string_view& nodeType) { return registry.find(nodeType)(); }
+    // Add node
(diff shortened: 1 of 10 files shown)


src/nodes/graph.h
```diff
@@ -0,0 +1,46 @@
...
+
+    /*
+     * Nodes and edges
+     */
+    using Nodes = std::map<std::string_view, std::unique_ptr<Node>>;
+    using Edges = std::vector<Node::Edge>;
+
+    private:
+    // Parent node
+    std::unique_ptr<Node> parent_;
```
trisyoungs, line 29: I don't see the `parent_` being set or used anywhere at present, but extending the notion it might be useful for any `Node` to know its parent `Graph`, rather than specifically a `GraphNode` knowing its parent `Node`. The parent could then be set in the `Graph::addNode()` function for convenience.

src/nodes/node.h
```diff
@@ -17,6 +18,8 @@ class Node
     Node() {}
     virtual ~Node() = default;
 
+    using Edge = std::map<std::string_view, ParameterLink>;
```
trisyoungs: "Edge" as a typename for this map of links feels a little odd.  I'd prefer it to be something like "LinkMap".

src/nodes/node.h
```diff
@@ -39,7 +42,7 @@ class Node
     // Input parameters
     std::map<std::string_view, std::shared_ptr<ParameterBase>> inputs_;
     // Inbound Links
-    std::map<std::string_view, ParameterLink> links_;
+    Edge links_;
```
trisyoungs: Perhaps just for clarity, and assuming the typename change above:
```suggestion
    LinkMap inputLinks_;
```

trisyoungs (commented):
Just a few general comments - this feels like the way forward, so run with it!

src/nodes/graph.h
RobBuchananCompPhys: Good point

RobBuchananCompPhys: This `parent_` is set when the constructor for `Graph` is called

trisyoungs (approved):
I don't think I have much else to add here at the present time.  Let's go and get it merged in.

PR #2078, opened 2025-03-10
fix: potential range check not handling empty configs or size factors

aboswel:
Issue:
No check performed when using "Generate" on an empty config.
Not checking box dimensions when using a size factor correctly.

### DIFF
--- a/src/gui/gui.cpp
+++ b/src/gui/gui.cpp
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // Copyright (c) 2025 Team Dissolve and contributors
 
-#include "gui/gui.h"
 #include "base/lineParser.h"
 #include "base/messenger.h"
 #include "gui/configurationTab.h"
+#include "gui/gui.h"
 #include "gui/layerTab.h"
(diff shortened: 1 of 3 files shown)


src/gui/gui.cpp
```diff
@@ -471,10 +471,10 @@ void DissolveWindow::checkPairPotentialRange(QWidget *parent)
     // Return smallest inscribed sphere radius if less than current pair potential range
     for (const auto &config : dissolve_.coreData().configurations())
     {
-        if (config->box()->inscribedSphereRadius() < radius.value_or(dissolve_.pairPotentialRange()) && config->nAtoms() > 0)
-        {
-            radius = config->box()->inscribedSphereRadius();
-        }
+        // Accounting for size factors
+        double reducedRadius = config->box()->inscribedSphereRadius() / config->requestedSizeFactor().value_or(1.0);
```
trisyoungs, line 475:
```suggestion
        auto reducedRadius = config->box()->inscribedSphereRadius() / config->requestedSizeFactor().value_or(1.0);
```

trisyoungs (approved):
Being pick about using an `auto`, but otherwise good to go.

PR #2079, opened 2025-03-11
feat: Add MoleculeTorsion module to calculate a specific torsion distribution

trisyoungs:
This PR adds a new `MoleculeTorsion` module to allow calculation of the distribution of a specific torsion angle in a given `Species`/`Molecule` type in the box, but from atom indices rather than (more cumbersome) `SpeciesSite` definitions.

`MoleculeAngle` and `MoleculeBond` variants could easily follow, but the focus here was on calculating a specific torsion angle distribution.


src/modules/moleculeTorsion/moleculeTorsion.cpp
```diff
@@ -0,0 +1,98 @@
...
+        return Messenger::error("[MoleculeTorsion] Torsion atom index i={} is out of range.\n", i_);
+    if (j_ > species_->nAtoms())
+        return Messenger::error("[MoleculeTorsion] Torsion atom index j={} is out of range.\n", j_);
+    if (k_ > species_->nAtoms())
+        return Messenger::error("[MoleculeTorsion] Torsion atom index k={} is out of range.\n", k_);
+    if (l_ > species_->nAtoms())
+        return Messenger::error("[MoleculeTorsion] Torsion atom index l={} is out of range.\n", l_);
+
+    return true;
+}
```
rprospero: Should we also add a check that i, j, k, and l are unique?  I'm thinking of that partially because they all default to 1 and I thought that the torsion was ill defined if we use the same atom twice.

rprospero (approved):
A small suggestion about an extra integrity check, but otherwise it looks good.

src/modules/moleculeTorsion/moleculeTorsion.cpp
trisyoungs: Yes, agreed.  I've added this check in via a `std::set` - if there's a more elegant way I am all ears!

Danbr4d (approved):
All works for me

PR #2080, opened 2025-03-11
fix: Intramolecular PairPotential energy was double-counted in Molecule-focussed energy calculation

trisyoungs:
This PR fixes a horrible little bug in the `EnergyKernel::pairPotential(Molecule &mol...)` function where the molecule's self-energy was double-counted. This led to odd, biased behaviour of the `IntraShake` module (the only user of the routine in this manner).


rprospero (approved):
Unsurprisingly, the correct code is cleaner and easier to read than the buggy code.  Full approval pending the passing of the test suite.

PR #2083, opened 2025-03-12
fix: Potential scaling in EPSRManager was broken after potential averaging was introduced

Danbr4d:
Reference the correct potential so potScale works again with averaging


trisyoungs (approved):
Yep, love a one-liner!  I changed the PR title with the auto-generated changelog in mind. All good to go once the tests pass.

PR #2084, opened 2025-03-13
fix: Use long int for some accumulators to avoid overflow

### LINKED ISSUE #2022: Large number of molecules overflows int in analysis modules
**Describe the bug**
When one has a large number of molecules (approx 50k) and calculates any site-site correlation between them, if the normalisation depends on the square of the number of sites this overflows a standard 32-bit int (~2.4 billion) and wraps back round to negative numbers. This causes RDFs, for example, to become inverted about y.

trisyoungs:
The PR fixes two modules that were prone to integer overflow when a decent number of molecules / sites were being considered (as detailed in #2022). The solution is to just use `long` but there might be a more general discussion to have regarding how we handle large integer accumulators / counters in future.

Closes #2022.

### DIFF
--- a/src/modules/angle/process.cpp
+++ b/src/modules/angle/process.cpp
@@ -69,8 +69,8 @@ Module::ExecutionResult AngleModule::process(ModuleContext &moduleContext)
 
     auto nAAvailable = a.sites().size(), nACumulative = a.sites().size();
     auto nASelections = 1;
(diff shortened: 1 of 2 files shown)


rprospero (approved):
While it's still possible to trigger that bug with 64bit longs, current tech will need a couple of centuries before users can file the bug report.  Then we can switch to floating point with Kahan summation.

PR #2086, opened 2025-03-19
feat: Argon nodes part I (atomShake, md)

RobBuchananCompPhys:


src/main/dissolve.cpp
```diff
@@ -71,5 +71,10 @@ GenericList &Dissolve::processingModuleData() { return processingModuleData_; }
...
  */
 
-// Set the Dissolve graph node
-void Dissolve::setGraph() { graphNode_ = std::move(Graph::produceNode("Dissolve")); }
+// Set the top-level Dissolve graph node with context
+void Dissolve::setGraph()
+{
+    auto dissolveNode = Graph::produceNode("Dissolve");
+    dissolveNode->setContext(*this);
+    graphNode_ = std::move(dissolveNode);
```
trisyoungs: Can `graphNode_` not be assigned to directly?

src/nodes/node.h
```diff
@@ -25,9 +26,36 @@ class Node : public Serialisable<>
...
+    /*
+     * Context
+     */
+    struct NodeContext
+    {
+        private:
+        // World process pool
+        ProcessPool &processPool_;
+        // Dissolve optional reference wrapper
+        OptionalReferenceWrapper<Dissolve> &dissolve_;
```
trisyoungs: This is a reference to an `OptionalReferenceWrapper` - not sure that's what you want!

src/nodes/node.h
```diff
@@ -25,9 +26,36 @@ class Node : public Serialisable<>
 
     using EdgeMap = std::map<std::string_view, Edge *>;
 
+    /*
+     * Context
+     */
+    struct NodeContext
+    {
+        private:
```
trisyoungs: I'm a touch confused why a new `NodeContext` is being introduced to mirror the old `ModuleContext` etc.  If we consider what data needs to be retrieved from the toplevel Dissolve instance there is `ProcessPool` and there is our forcefield data (which will likely just be a single object) and not a lot else.  Having functions on `Node` which specifically return these data feels like the cleaner way to go:
```
virtual void &Node::processPool() { return parentGraph_->processPool(); }
```
...and then in `DissolveNode` we override that to give:
```
void &Node::processPool() override { return dissolve_.processPool(); }
```

Quite a lot of nodes will require neither of these two objects, so retrieving the data only when its needed, rather than having a reference stored locally, feels cleaner.

src/nodes/atomShake.cpp
```diff
@@ -1,33 +1,30 @@
...
+#include "classes/changeStore.h"
+#include "classes/configuration.h"
+#include "classes/regionalDistributor.h"
+#include "kernels/producer.h"
+#include "main/dissolve.h"
 #include "nodes/atomShake.h"
 
 AtomShakeNode::AtomShakeNode()
 {
     addInput<Configuration *>("Configuration", "Set target configuration for the module", targetConfiguration_)
```
trisyoungs, line 16: This also needs to have the `Required` flag set.

src/nodes/atomShake.cpp
```diff
@@ -1,33 +1,30 @@
...
         ->setFlags(ParameterBase::Invalidates);
 
-    //    keywords_.setOrganisation("Options", "Control", "Number of move attempts per atom and the target acceptance rate.");
-    addInput<int>("ShakesPerAtom", "Number of shakes to attempt per atom", nShakesPerAtom_);
-    addBoundedInput<double>("TargetAcceptanceRate", "Target acceptance rate for Monte Carlo moves", targetAcceptanceRate_, 0,
+    addInput<Number>("ShakesPerAtom", "Number of shakes to attempt per atom", nShakesPerAtom_);
+    addBoundedInput<Number>("TargetAcceptanceRate", "Target acceptance rate for Monte Carlo moves", targetAcceptanceRate_, 0,
                             1);
+    addInput<Number>("StepSizeMax", "Maximum allowed value for step size, in Angstroms", stepSizeMax_);
+    addInput<Number>("StepSizeMin", "Minimum allowed value for step size, in Angstroms", stepSizeMin_);
```
trisyoungs: I think these all qualify as "options" rather than "inputs", since they will rarely (if ever) be set via `Graph` connectivity.

src/nodes/atomShake.cpp
```diff
@@ -1,33 +1,30 @@
...
-    //                              as " "the calculation proceeds in order to approximate the specified acceptance rate.");
-    //    keywords_.addRestartable<DoubleKeyword>("StepSize", "Step size in Angstroms to use in Monte Carlo moves", stepSize_,
-    //    0.001);
-    addInput<double>("StepSizeMax", "Maximum allowed value for step size, in Angstroms", stepSizeMax_);
-    addInput<double>("StepSizeMin", "Minimum allowed value for step size, in Angstroms", stepSizeMin_);
-
-    // THIS KEYWORD HAS NEVER BEEN USED AFAIK, SO I THINK IT (AND OTHERS) CAN BE REMOVED AS WE MOVE FORWARD
-    //    keywords_.setOrganisation("Advanced");
-    addBoundedOptionalInput<std::optional<double>>(
+    addBoundedOptionalInput<std::optional<Number>>(
```
trisyoungs: As I think I wrote in a comment in commit way back, we should remove these `cutoffDistance` parameters from `Module`s as we transition them to `Node`s as I don't believe they hold a useful purpose any more. I certainly have never used them!

src/nodes/atomShake.cpp
```diff
@@ -1,33 +1,30 @@
...
         "CuttOffDistance", "Interatomic cutoff distance to use for energy calculation (0.0 to use pair potential range)",
         cutoffDistance_, 0.0, "Use PairPotential Range", {}, 0.1);
-
-    // executeIfTargetsUnchanged_ = true;
 }
 
 std::string_view AtomShakeNode::name() const { return "Atom Shake"; }
```
trisyoungs, line 25: Once #2099 is merged in this will need to be `type()`.

src/nodes/atomShake.cpp
```diff
@@ -36,3 +33,191 @@ std::string_view AtomShakeNode::summary() const
...
+    // Retrieve control parameters from Configuration
+    double rCut;
+    if (cutoffDistance_.has_value())
+    {
+        rCut = cutoffDistance_.value().asDouble();
+    }
+    else
+    {
+        rCut = dissolve().pairPotentialRange();
+    }
```
trisyoungs: As suggested, removing the `cutoffDistance_` variable means this can be deleted, and wherever `rCut` was used you can simply pass `dissolve().pairPotentialRange()`.

src/nodes/atomShake.cpp
```diff
@@ -36,3 +33,191 @@ std::string_view AtomShakeNode::summary() const
...
+
+    const auto termScale = 1.0;
+    const auto rRT = 1.0 / (.008314472 * targetConfiguration_->temperature());
+
+    // Print argument/parameter summary
+    Messenger::print("AtomShake: Cutoff distance is {}\n", rCut);
```
trisyoungs:
```suggestion
```

src/nodes/atomShake.cpp
```diff
@@ -36,3 +33,191 @@ std::string_view AtomShakeNode::summary() const
...
+    Timer commsTimer(false);
+
+    // Create a Molecule distributor
+    RegionalDistributor distributor(targetConfiguration_->nMolecules(), targetConfiguration_->cells(),
+                                    processPool(), strategy);
+
+    // Create a local ChangeStore and EnergyKernel
+    ChangeStore changeStore(processPool(), commsTimer);
+    auto kernel = KernelProducer::energyKernel(targetConfiguration_, processPool(),
+                                               dissolve().potentialMap(), rCut);
```
trisyoungs:
```suggestion
                                               dissolve().potentialMap(), dissolve().pairPotentialRange());
```

src/nodes/atomShake.h
```diff
@@ -24,22 +25,22 @@ class AtomShakeNode : public Node
     // Target configurations
     Configuration *targetConfiguration_{nullptr};
     // Interatomic cutoff distance to use for energy calculation
-    std::optional<double> cutoffDistance_;
+    std::optional<Number> cutoffDistance_;
```
trisyoungs: Remove.

src/nodes/atomShake.h
```diff
@@ -24,22 +25,22 @@ class AtomShakeNode : public Node
...
-    double targetAcceptanceRate_{0.33};
+    Number targetAcceptanceRate_{0.33};
 
     /*
      * Processing
      */
     private:
     // Run main processing
-    Module::ExecutionResult process(ModuleContext &moduleContext);
+    NodeConstants::ProcessResult process();
```
trisyoungs:
```suggestion
    NodeConstants::ProcessResult process() override;
```

src/nodes/dissolve.h
```diff
@@ -10,17 +10,25 @@
...
-    /*
-     * Definition
-     */
     private:
+    // Dissolve reference
+    Dissolve &dissolve_;
+
+    public:
+    // Set dissolve
+    void setDissolve(Dissolve &dissolve);
```
trisyoungs, line 26: I might be tempted to pass this reference in via the constructor, rather than as a separate function, since without it `DissolveNode` doesn't really offer a lot!

src/nodes/graph.h
```diff
@@ -16,7 +16,7 @@
 class Graph : public Node
 {
     public:
-    explicit Graph(Node *parent) : Node(this), parent_(parent) {}
+    explicit Graph(Node *parent) : Node(this), parent_(parent), nodeContext_(parent->context()) {}
```
trisyoungs: Old ref to `nodeContext_`.

src/nodes/md.cpp
```diff
@@ -0,0 +1,645 @@
...
+#include "keywords/integer.h"
+#include "keywords/optionalDouble.h"
+#include "keywords/optionalInt.h"
+#include "keywords/speciesVector.h"
+#include "main/dissolve.h"
+#include "module/context.h"
+#include "modules/energy/energy.h"
+#include "modules/forces/forces.h"
+#include "nodes/md.h"
+
```
trisyoungs, line 22:
```suggestion
```

src/nodes/md.cpp
```diff
@@ -0,0 +1,645 @@
...
+
+MDNode::MDNode()
+{
+    addInput<Configuration*>("Configuration", "Set target configuration for the module", targetConfiguration_)
```
trisyoungs, line 25: Required parameter.

src/nodes/md.cpp
```diff
@@ -0,0 +1,645 @@
...
+        ->setFlags(ParameterBase::Invalidates);
+
+    addInput<Number>("NSteps", "Number of MD steps to perform", nSteps, 1);
```
trisyoungs, line 28: I think I would keep NSteps as an input, and move the rest to options.

src/nodes/md.cpp
```diff
@@ -0,0 +1,645 @@
...
+    addInput<std::optional<Number>>("EnergyFrequency", "Frequency at which to calculate total system energy",
+                                          energyFrequency_, 0, std::nullopt, 5, "Off");
+    addInput<std::optional<Number>>("OutputFrequency", "Frequency at which to output step information", outputFrequency_,
+                                          0, std::nullopt, 5, "Off");
+    addInput<std::optional<Number>>("TrajectoryFrequency", "Write frequency for trajectory file", trajectoryFrequency_, 0,
+                                          std::nullopt, 5, "Off");
+
+    addInput<bool>("CapForces", "Control whether atomic forces are capped every step", capForces_);
+    addInput<Number>("CapForcesAt", "Set cap on allowable force (kJ/mol) per atom", capForcesAt_, 0.0);
+    addInput<std::optional<Number>>(
```
trisyoungs: And again, `cutoffDIstance_` can be removed entirely.

trisyoungs (commented):
Yep, like where this is heading!  A handful of comments, but most are related to how we transition `Module`s to `Node`s, what should be inputs vs options etc. 

Both nodes should also have an output which is the `Configuration*`.

The monolithic `MD` cpp file makes me think how we should address nodes that have a decent amount of code associated to them. Rather than splitting up into directories we could opt for `md_functions.cpp` for example, to contain auxiliary routines outside of the main processing function.

src/main/dissolve.cpp
RobBuchananCompPhys: I guess we could, but I find it more compelling in this case to have a `setDissolve` method exclusively on `DissolveNode` since we don't want any old graph to hold a potentially conflicting dissolve instance (though it is of course unlikely).

RobBuchananCompPhys:
> The monolithic `MD` cpp file makes me think how we should address nodes that have a decent amount of code associated to them. Rather than splitting up into directories we could opt for `md_functions.cpp` for example, to contain auxiliary routines outside of the main processing function.

@trisyoungs We may still end up with rather a lot of code in the implementation file when combined with the constructor and other definitions, particularly for large processes. An alternative take is that we make a pair of sub-directories `helper`, and `process` - the former being the various functions required by a process (i.e. `helper/md.cpp`), and the latter is just the process (`process/md.cpp`) ?

src/nodes/dissolve.h
RobBuchananCompPhys: Hmm, this might be a little tricky to do. As we are currently producing nodes from lambdas buried in the factory function, it sounds like there could be a bit of `if std::is_same<T, DissolveNode>` 'do this, otherwise do something else', which I worry would spoil what is an otherwise rather elegant solution!

PR #2087, opened 2025-03-19
Numbers, edges, and nodes

trisyoungs:
So a lot has happened here. I still have a TODO list but will park this one for review now (as well as its root #2068). To summarise:

- The `Number` class has been introduced, providing a union-style `int`/`double` container for general use as a `Parameter`, and which retains `int` as the type when it should.
- A basic `NodeRegistry` has been introduced to produce `Node`s
- Math nodes are no longer templated.
- `Node` outputs are stored separately from inputs, but both are flagged appropriately.
- The `Edge` and `ParameterLink` classes have been combined into one since together they form a coherent definition and storage entity.  A `std::vector` of `std::unique_ptr<Edge>` is now stored in `Graph` and `Node`s store raw pointers to `Edge`s they are connected by.
- `Node` versioning is here, as well as a new process result in the form of "Unchanged" (where a node "succeeds` but didn't actually need to do anything).
- A unit test `nodes/flow.cpp` expresses the basic linkage and flow of data over a test graph thanks to the new `Node::run()` and `Edge::pull()` functions.  This basically amounts to a reworking of the various functions that existed before but which are now based on our internal versioning system.


src/nodes/edge.h
```diff
@@ -0,0 +1,65 @@
...
+    EdgeDefinition definition_;
+
+    protected:
+    // The constructor is private because it can only be constructed by the factory method
+    Edge(Node *sourceNode, ParameterBase &sourceOutput, Node *targetNode, ParameterBase &targetInput);
+
+    private:
+    // Pointers to source and target nodes
+    Node *sourceNode_{nullptr};
+    Node *targetNode_{nullptr};
```
rprospero:
```suggestion
    Node& sourceNode_;
    Node& targetNode_;
```

I rather liked having the nodes as references along with the parameters, for the same reasons given in the comment below.  While making this change will have some minor cascading effects with replacing `->` with `.`, I don't think it should break anything major.

src/nodes/edge.h
```diff
@@ -0,0 +1,65 @@
...
+    //
+    // The immutability is important because the factory ensures that
+    // all links are valid at creation time.  By keeping the links
+    // immutable, that means that all links are valid at all times.
+    // This enables us to skip a great deal of error checking when
+    // processing a chain of links.
+    ParameterBase &sourceOutput_;
+    ParameterBase &targetInput_;
+    // Version of the source node when this edge was last pulled by the target node.
+    int sourceNodeVersionIndex_{NodeConstants::InvalidVersion};
```
rprospero, line 47: I had assumed that the last seen version would be tracked by the nodes, but it's a lot smarter to keep it in the edges like this.

src/nodes/number.cpp
```diff
@@ -0,0 +1,154 @@
...
+{
+    value_ = other.value_;
+    return *this;
+}
+
+struct AddNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a + b}; }
+};
+Number Number::operator+(const Number &other) const { return std::visit(AddNumberVisitor{}, value_, other.value_); }
```
rprospero: I'm a little curious whywe need a full class here, instead of just a lambda?  I would have thought that we could get away with:

```suggestion
Number Number::operator+(const Number &other) const { return std::visit([](auto a, auto b) -> Number {return {a+b};}, value_, other.value_); }
```

Is there an advanage to making the visitor class or some other context I've missed here that prevents the lambda from working?  These changes would need to apply to the other operator classes as well, if it's possible.

src/nodes/number.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+struct AddAssignmentNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a += b}; }
```
rprospero:
```suggestion
    template <class T1, class T2> auto operator()(T1 &a, T2 b) const -> Number { return {a += b}; }
```

I believe that `a` will need to be a reference to allow for mutation.

src/nodes/number.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+struct SubtractNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a - b}; }
+};
+Number Number::operator-(const Number &other) const { return std::visit(SubtractNumberVisitor{}, value_, other.value_); }
+
+struct SubtractAssignmentNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a -= b}; }
```
rprospero: See above

src/nodes/number.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+struct MultiplyNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a * b}; }
+};
+Number Number::operator*(const Number &other) const { return std::visit(MultiplyNumberVisitor{}, value_, other.value_); }
+
+struct MultiplyAssignmentNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a *= b}; }
```
rprospero: See above

src/nodes/number.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+struct DivisionNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a / b}; }
+};
+Number Number::operator/(const Number &other) const { return std::visit(DivisionNumberVisitor{}, value_, other.value_); }
+
+struct DivisionAssignmentNumberVisitor
+{
+    template <class T1, class T2> auto operator()(T1 a, T2 b) const -> Number { return {a /= b}; }
```
rprospero: See above

src/nodes/number.cpp
```diff
@@ -0,0 +1,154 @@
...
+        {
+            using T = std::decay_t<decltype(arg)>;
+            if constexpr (std::is_same_v<T, toml::integer>)
+            {
+                set((int)arg);
+            }
+            else if constexpr (std::is_same_v<T, toml::floating>)
+            {
+                set((double)arg);
+            }
```
rprospero, line 127: Love the use of `constexpr` here

src/nodes/node.cpp
```diff
@@ -5,74 +5,161 @@
...
-std::shared_ptr<ParameterBase> Node::findOption(std::string_view name) const
+// Remove edge
+void Node::removeEdge(Edge *edge)
 {
-    if (!options_.contains(name))
-        return {};
-    return options_.at(name);
+    // TODO
+    // inputEdges_.erase(std::remove_if(inputEdges_.begin(), inputEdges_.end(), [edge](const auto &it) { return edge ==
+    // it.second; } ), inputEdges_.end());
```
rprospero, line 41: Can we uncomment this TODO?

src/nodes/node.cpp
```diff
@@ -5,74 +5,161 @@
...
 
-// Return Options
-std::map<std::string_view, std::shared_ptr<ParameterBase>> &Node::options() { return options_; };
+// Invalidate the current node, resetting the version index
+void Node::invalidate() { versionIndex_ = NodeConstants::InvalidVersion; }
 
-// Prepare for processing
-Node::Readiness Node::preprocess()
+// Check that all required inputs are present, and that all inputs are valid
+bool Node::inputsAreValid() const
```
rprospero, line 55: Much better name

src/nodes/node.cpp
```diff
@@ -5,74 +5,161 @@
...
+NodeConstants::ProcessResult Node::run()
 {
-    if (satisfied_)
-        return true;
-    for (auto &[name, link] : inputLinks_)
-        if (!link.source().parent()->isSatisfied())
-            satisfied_ = false;
-    return satisfied_;
+    // Check our input links - if any are out-of-date we must retrieve new values
+    auto nInputLinksChanged = 0;
```
rprospero, line 76: perhaps make this a boolean as to whether or not any links changed?  I'm not sure we ever use the exact number

rprospero (commented):
I'm quite happy with all of this, but I did have some questions.

src/nodes/number.cpp
rprospero: On further reflection, it does not *need* to be a reference, since the template can be resolved by references, but I still feel that it may be clearer with making the reference explicit.

Also, this is all moot if we can just use lambdas instead of needing the visitor structs

src/nodes/node.cpp
rprospero: Nevermind - I saw the other PR with the proper implementation

src/nodes/edge.h
trisyoungs: I moved to pointers because (I think) while I was starting to combine the `Edge` and `ParameterLink` classes I thought I needed to, but you are absolutely right that use of references still stands as the best option (and there's no technical reason why this won't work) so I will revert this.

src/nodes/edge.h
trisyoungs: Agreed. Thought about having it on the `Node`, but this made more sense in a lot of ways.

src/nodes/number.cpp
trisyoungs: Nope. Apparently I overcomplicated things here!  The explicit reference on the paraneter `a` _is_ required for assignment operators though.

src/nodes/node.cpp
trisyoungs: Removed in a subsequent PR.

PR #2099, opened 2025-03-20
Node parameter setting

### LINKED ISSUE #2089: Clarify node name vs node type

### LINKED ISSUE #2090: Handle removal of Edges

### LINKED ISSUE #2094: Setting input parameters should forward invalidated status to connected nodes

trisyoungs:
This PR builds on #2087 and helps to clarify how our graph nodes interact and affect each other.

A node has two principal "state" variables:
1) `versionIndex_` - used to track effectively the number of times the node has been run since the last modification to its edge connectivity.
2) `upToDate_` - whether the node's current data is valid (and it doesn't need to `run()` again at present)

When a node A is `run()` it first checks all of its input `Edge`s and attempts to `pull()` the data from each - if the connected source node B is out of date or the last version index stored in the `Edge` is different from that of B, the node B is `run()` and the relevant data copied to the input parameter on A, and the new version index of B stored in the `Edge`. This automatically sets the `upToDate_` flag on A as `false`, and will force A to call `process()` in order to update itself (once all `Edge`s are checked) and set `upToDate_` back to `true` (as well as incrementing `versionIndex_`.

As alluded to in 1) above, adding or removing an `Edge` to/from a node's input parameters invalidates the node, setting `versionIndex_` back to `InvalidVersion` and clearing any local data.

One other significant change here is moving from storing nodes in a `std::map<std::string_view, std::unique_ptr<Node>>` to a straight `std::vector<std::unique_ptr<Node>>` and setting the name on the `Node` itself. This allows a `Node` to know what its own name is, and allows `Node`s to write data to (say) a file named after itself,  and `Edge`s to return a text description of the connection it represents for serialisation.

Closes #2094.
Closes #2089.
Closes #2090.

### DIFF
--- a/src/nodes/edge.cpp
+++ b/src/nodes/edge.cpp
@@ -108,8 +108,7 @@ NodeConstants::ProcessResult Edge::pull()
      * current versionIndex ready for next time.
      */
     auto result = NodeConstants::ProcessResult::Failed;
-    if (sourceNodeVersionIndex_ == NodeConstants::InvalidVersion ||
-        sourceNode_.versionIndex() == NodeConstants::InvalidVersion || (sourceNodeVersionIndex_ != sourceNode_.versionIndex()))
+    if (!sourceNode_.isUpToDate() || (sourceNodeVersionIndex_ != sourceNode_.versionIndex()))
     {
         result = sourceNode_.run();
         if (result != NodeConstants::ProcessResult::Success && result != NodeConstants::ProcessResult::Unchanged)
--- a/src/nodes/edge.cpp
+++ b/src/nodes/edge.cpp
@@ -12,10 +12,8 @@ Edge::Edge(Node &sourceNode, ParameterBase &sourceOutput, Node &targetNode, Para
 Edge::~Edge()
 {
     // Detach from the source and target nodes
-    if (sourceNode_)
-        sourceNode_->removeEdge(this);
-    if (targetNode_)
-        targetNode_->removeEdge(this);
+    sourceNode_.unlinkEdge(this);
+    targetNode_.unlinkEdge(this);
 }
 
 // Local EdgeConstructor class to allow creation of memory-managed Edge instances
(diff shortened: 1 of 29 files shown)


src/nodes/edge.cpp
```diff
@@ -9,6 +9,15 @@ Edge::Edge(Node *sourceNode, ParameterBase &sourceOutput, Node *targetNode, Para
 {
 }
 
+Edge::~Edge()
+{
+    // Detach from the source and target nodes
+    if (sourceNode_)
+        sourceNode_->removeEdge(this);
+    if (targetNode_)
+        targetNode_->removeEdge(this);
```
rprospero: This is one of the reasons I like using references for the nodes - you never need to do nullptr checking.

src/nodes/graph.h
```diff
@@ -22,7 +22,7 @@ class Graph : public Node
     /*
      * Nodes and edges
      */
-    using Nodes = std::map<std::string_view, std::unique_ptr<Node>>;
+    using Nodes = std::vector<std::unique_ptr<Node>>;
```
rprospero: I agree that the name should be in the Node class to make life easier for serialisation and similar effects, but I believe that we can still store them in the `map`.  I make this argument mostly because accessing nodes by name is the primary operation on this structure, so it saves us calling `find_if` every time we need to find one.

I see two *disadvantages* to moving back to a `map`.  First, having the name stored *only* in the node allows for two Nodes to have the same name (though this may count as an advantage to maps, instead of a disadvantage).  The second is that we have to worry about the name in the node becoming out of sync from the key of the map.  I could see this becoming an issue in the GUI, but it's a pretty simple operation to get the keys back in sync with the node names.

rprospero (approved):
Besides some qualms about the switch of the node storage type, I'm happy with this change.

src/nodes/graph.h
trisyoungs: I had firmly believed in the single source of truth being on the `Node`, but I see your point here. I guess this brings us back to one of @RobBuchananCompPhys's conversations with you about bimaps! If we agree that the `Graph` is the ultimate arbiter and manager of `Node` naming, then having `Node::name_` as well as a `std::map` gives us the both of best worlds - we will just have to live with the fact that `Node::setName()` can be called from anywhere, although we could make a check in that function, querying the parent `Graph` if the proposed name change is consistent with the current `std::map` (assuming we rename the key as the first operation). This gives us protection at least from the "`Node`-abuse" side of things.

PR #2103, opened 2025-03-21
fix: Crash when closing DataManager dialog

### LINKED ISSUE #2050: Closing DataManager on simulation tab crash
**Describe the bug**
When you open and then close the DataManager window on the Simulation Tab it causes Dissolve to crash with an error of:

TypeError: Cannot call method `rowCount` of null
TypeError: Cannot call method `columnCount` of null

trisyoungs:
This quick PR fixes a crash in the `DataManagerDialog` when closing the window.

Closes #2050.


rprospero (approved):
Good catch.  That was my bug

PR #2110, opened 2025-03-25
Graph node naming

trisyoungs:
This PR proposes a method to ensure unique `Node` naming within a `Graph`, avoiding (for the most part) having multiple sources of truth.

Opinions requested before I push on with a unit test.

### DIFF
--- a/src/nodes/node.cpp
+++ b/src/nodes/node.cpp
@@ -3,16 +3,21 @@
 
 #include "nodes/node.h"
 #include "base/sysFunc.h"
+#include "nodes/graph.h"
 
 /*
  * Definition
  */
 
 // Set node name
-void Node::setName(std::string_view newName) { name_ = newName; }
+void Node::setName(std::string_view newName)
+{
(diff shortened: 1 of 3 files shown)


RobBuchananCompPhys (approved):

src/nodes/node.cpp
```diff
@@ -3,16 +3,21 @@
...
 /*
  * Definition
  */
 
 // Set node name
-void Node::setName(std::string_view newName) { name_ = newName; }
+void Node::setName(std::string_view newName)
+{
+    if (parentGraph_)
+        parentGraph_->setNodeName(this, newName);
```
rprospero, line 16:
```suggestion
    if (parentGraph_)
        parentGraph_->setNodeName(this, newName);
    else
        name_ = newName;
```

rprospero (approved):
That came together rather cleanly.  I'm happy that we have a setName function that's part of Node but still does the right thing.  I've added one suggestion, but it's not mandatory

PR #2116, opened 2025-03-27
feat: Dissolve graph, constructor parent graph assignment

RobBuchananCompPhys:
This PR demonstrates two changes to how we manage nodes and the top-level `Dissolve` graph
- Nodes no longer rely on a call to `setParentGraph` post-initialisation: the `parentGraph_` is now a constructor argument, which means that the parent `Graph*` gets passed into the factory lambda via the `produceNode` method
- The top-level `Dissolve` graph is no longer just another registered `Node`, but a new type `DissolveGraph`, and is initialised along with `Dissolve`

### DIFF
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -26,7 +26,7 @@ class Molecule;
 class Dissolve : public Serialisable<>
 {
     public:
-    Dissolve(CoreData &coreData) : graphNode_(std::make_unique<DissolveGraph>(*this)) {}
+    Dissolve(CoreData &coreData);
     ~Dissolve();
 
     /*
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -26,7 +26,7 @@ class Molecule;
 class Dissolve : public Serialisable<>
 {
     public:
-    Dissolve(CoreData &coreData) : graphNode_(*this);
+    Dissolve(CoreData &coreData) : graphNode_(std::make_unique<DissolveGraph>(*this));
     ~Dissolve();
 
     /*
@@ -131,7 +131,7 @@ class Dissolve : public Serialisable<>
 
     private:
     // Dissolve graph node
-    DissolveGraph graphNode_;
+    std::unique_ptr<DissolveGraph> graphNode_;
 
     /*
      * Simulation
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -12,7 +12,6 @@
 #include "data/elements.h"
 #include "module/layer.h"
 #include "module/module.h"
-#include "nodes/node.h"
 
 // Forward Declarations
 class Atom;
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -26,7 +26,7 @@ class Molecule;
 class Dissolve : public Serialisable<>
 {
     public:
-    Dissolve(CoreData &coreData) : graphNode_(this);
+    Dissolve(CoreData &coreData) : graphNode_(*this);
     ~Dissolve();
 
     /*
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -26,7 +26,7 @@ class Molecule;
 class Dissolve : public Serialisable<>
 {
     public:
-    Dissolve(CoreData &coreData);
+    Dissolve(CoreData &coreData) : graphNode_(this);
     ~Dissolve();
 
     /*
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -18,6 +18,7 @@
 class Atom;
 class Box;
 class Cell;
+class DissolveGraph;
 class Isotopologue;
 class Molecule;
 
@@ -130,7 +131,7 @@ class Dissolve : public Serialisable<>
 
     private:
     // Dissolve graph node
-    std::unique_ptr<Node> graphNode_;
+    DissolveGraph graphNode_;
 
     /*
      * Simulation
--- a/src/main/dissolve.h
+++ b/src/main/dissolve.h
@@ -26,7 +26,7 @@ class Molecule;
 class Dissolve : public Serialisable<>
 {
     public:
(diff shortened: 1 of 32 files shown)


src/nodes/graph.h
```diff
@@ -39,7 +39,7 @@ class Graph : public Node
     // Return short summary of the node's purpose
     std::string_view summary() const override;
     // Produce a registered node by type
-    static std::unique_ptr<Node> produceNode(const std::string_view &nodeType) { return registry.find(nodeType)->second(); }
+    static std::unique_ptr<Node> produceNode(const std::string_view &nodeType) { return registry.find(nodeType)->second()(this); }
```
RobBuchananCompPhys: This is sort of a note-to-self, but I've just realised that 
a) the argument `this` may be ineffective in this case since it's a static method and there may not in fact be an instance to point to
b) this method appears to have been superseded by `NodeRegistry::Produce`, so we can actually remove it!

src/main/dissolve.h
```diff
@@ -18,14 +18,15 @@
...
+class DissolveGraph;
 class Isotopologue;
 class Molecule;
 
 // Dissolve Main Class
 class Dissolve : public Serialisable<>
 {
     public:
-    Dissolve(CoreData &coreData);
+    Dissolve(CoreData &coreData) : graphNode_(*this);
```
trisyoungs:
```suggestion
    Dissolve(CoreData &coreData) : graphNode_(*this) {};
```

src/main/dissolve.h
```diff
@@ -130,7 +131,7 @@ class Dissolve : public Serialisable<>
 
     private:
     // Dissolve graph node
-    std::unique_ptr<Node> graphNode_;
+    DissolveGraph graphNode_;
```
trisyoungs: I still think a `std::unique_ptr` is the way to go here, rather than an instance as a member variable, the reason being that cleanup of the whole simulation (e.g. loading an old simulation, starting a new simulation etc.) could be as simple as overwriting or nullifying the `std::unique_ptr`. As currently written we would need at least a `clear()` function on the `DissolveGraph` node.

src/nodes/dissolve.cpp
```diff
@@ -0,0 +1,13 @@
...
+Dissolve& DissolveGraph::dissolve() const
+{ 
+	return dissolve_;
+}
+
+// Return world pool
+ProcessPool& DissolveGraph::processPool() const
+{ 
+	return dissolve_.worldPool();
+}
```
```suggestion
// Return dissolve
Dissolve& DissolveGraph::dissolve() const { return dissolve_; }

// Return world pool
ProcessPool& DissolveGraph::processPool() const { return dissolve_.worldPool(); }
```

src/nodes/dissolve.h
```diff
@@ -4,13 +4,14 @@
 #pragma once
 
 #include "classes/configuration.h"
-#include "nodes/node.h"
+#include "main/dissolve.h"
+#include "nodes/graph.h"
 
 // AtomShake Node
-class DissolveNode : public Node
+class DissolveGraph : public Graph
```
trisyoungs, line 11: Should be `DissolveGraph`.

src/nodes/dissolve.h
```diff
@@ -4,13 +4,14 @@
 #pragma once
 
 #include "classes/configuration.h"
-#include "nodes/node.h"
+#include "main/dissolve.h"
+#include "nodes/graph.h"
 
 // AtomShake Node
```
trisyoungs:
```suggestion
// Main Dissolve node
```

src/nodes/graph.h
trisyoungs: Yep, this can go now we have `NodeRegistry`.

src/main/dissolve.h
RobBuchananCompPhys: Good shout! I did wonder about that

src/main/dissolve.h
```diff
@@ -18,14 +18,15 @@
...
+class DissolveGraph;
 class Isotopologue;
 class Molecule;
 
 // Dissolve Main Class
 class Dissolve : public Serialisable<>
 {
     public:
-    Dissolve(CoreData &coreData);
+    Dissolve(CoreData &coreData) : graphNode_(std::make_unique<DissolveGraph>(*this)) {}
```
trisyoungs: Missed this first time around, but you're redefining the constructor in the header file whereas the actual / existing implementation is already in `src/main/dissolve.cpp`.  Further, I don't think we should default to creating a new `DissolveGraph` on the construction of `Dissolve`, since this implies we are ready with an empty simulation. I think we should handle construction of the unique object only when required - e.g. just prior to loading an input file or (via the GUI) choosing File->New.  I think you can put this in `Dissolve::clear()` for now.

trisyoungs (approved):
Looking good.  One further comment to address, and you've picked up a merge conflict thanks to my merging in the graph naming branch (sorry about that!).

src/nodes/registry.h
```diff
@@ -3,54 +3,31 @@
...
-using NodeProducer = std::function<std::unique_ptr<Node>(void)>;
+using NodeProducer = std::function<std::unique_ptr<Node>(Graph *parent)>;
 
 // Makes unique pointer to derived node instance
 template <typename T> NodeProducer makeDerivedNode()
 {
-    auto nodeLambda = []() -> std::unique_ptr<Node> { return std::make_unique<T>(); };
-    return nodeLambda;
+    return [=](Graph *parent) -> std::unique_ptr<Node> { return std::make_unique<T>(parent); };
 }
```
rprospero: I think that we can move this function into `registry.cpp`, since that's the only place where it's called.  If this header gets imported in a lot of different places, then we might save on a bit of compile time.

rprospero (approved):
I like the changes.  I've added a small suggestion that might give a small improvement to compile times, but it's entirely optional.

src/nodes/registry.h
trisyoungs: Hang on though - it's a template function...

PR #2117, opened 2025-03-28
feat: Serialise nodes

rprospero:
This PR adds serialisation support to the graphs.  I've added a test file to compare with in `tests/data/dissolve/graph.toml`.


src/base/serialiser.h
```diff
@@ -145,6 +150,22 @@ template <typename... Contexts> class Serialisable
         return result;
     }
 
+    // A helper function to add the lements of a map to a node under a name
```
trisyoungs:
```suggestion
    // A helper function to add the elements of a map to a node under a name
```

src/nodes/edge.cpp
```diff
@@ -133,3 +164,16 @@ NodeConstants::ProcessResult Edge::pull()
...
+SerialisedValue Edge::serialise() const { return definition().serialise(); }
+
+// Read values from a serialisable value This is required for the
+// SerialableValue type implementation, but we actually deserialise
+// Edges through an EdgeConnection.  I've added this error to
+// immediately alert us in case this function is ever called.
+void Edge::deserialise(const SerialisedValue &node)
+{
+    throw std::runtime_error("Cannot directly deserialise edges.  Please contact the Dissolve development team if you are "
+                             "seeing this error - this is a bug and NOT your fault.");
```
trisyoungs: Just to ensure the poor recipient of the error sees the message!
```suggestion
                             "seeing this error - this is a bug and NOT your fault.\n");
```

tests/data/dissolve/graph.toml
trisyoungs, line 1: Strictly speaking this is an input file, so I would put it in `data/dissolve/input`.  Perhaps also rename it to be more specific - maybe "simple_addition_graph.toml"?

src/nodes/parameter.h
```diff
@@ -138,15 +139,20 @@ template <typename T> class Parameter : public ParameterBase, public std::enable
         SerialisedValue result = {};
```
trisyoungs, line 139: Here we shouldn't serialise anything if the parameter `isDefault()`, but I'm not sure where you put this check since `toMap` handles the actual serialisation...

trisyoungs (approved):
Looks good to me. One comment which might be complicated (i.e. the `isDefault()` issue) could be left for a follow-up.

tests/data/dissolve/graph.toml
rprospero: Agreed

PR #2119, opened 2025-03-31
feat: Support CBOR file format

rprospero:
This PR adds support for serialising and deserialising TOML structures in the [CBOR](https://www.rfc-editor.org/rfc/rfc8949) format.  This format is both smaller and faster to read or write than true TOML.  Additionally, it's possible to extract individual values from the file without loading the entire structure into memory.

While this PR does fully serialise and deserialise our TOML strctures, certains parts are against the *spirit* of the spec.  Specifically, all segment lengths (i.e. string length, array length, and table element count) use an unsigned 64bit integer to supply the count.  The spec supports using smaller sizes, which could theoretically improve string parsing efficiency by up to 400% for single character strings.  Furthermore, editing our CBOR files in other tooling might use these extra features, which currently our parser ignores.  These issues should be handled in the future, but I believe are outside the scope of this PR.

Additionally, I have not written the parser for extracting smaller elements from a file.  In pseudocode:

```Python
def access(ptr: File, path: Vector<Union<int, string>>):
    if path.empty():
        return fromCBOR(ptr)
    head = path.pop()
    if head.is_int() and ptr.is_array():
        for i in 0..head:
            ignore_element(ptr)
            return access(ptr, path)
    if head.is_string() and ptr.is_table():
        for key, value_ptr in ptr.elements:
            if key == head:
                return access(value_ptr, path)
    error "Bad file"
```


rprospero:
Pointed to wrong branch

PR #2120, opened 2025-03-31
feat: Add cbor support

rprospero:
This PR adds support for serialising and deserialising TOML structures in the [CBOR](https://www.rfc-editor.org/rfc/rfc8949) format.  This format is both smaller and faster to read or write than true TOML.  Additionally, it's possible to extract individual values from the file without loading the entire structure into memory.

While this PR does fully serialise and deserialise our TOML strctures, certains parts are against the *spirit* of the spec.  Specifically, all segment lengths (i.e. string length, array length, and table element count) use an unsigned 64bit integer to supply the count.  The spec supports using smaller sizes, which could theoretically improve string parsing efficiency by up to 400% for single character strings.  Furthermore, editing our CBOR files in other tooling might use these extra features, which currently our parser ignores.  These issues should be handled in the future, but I believe are outside the scope of this PR.

Additionally, I have not written the parser for extracting smaller elements from a file.  In pseudocode:

```Python
def access(ptr: File, path: Vector<Union<int, string>>):
    if path.empty():
        return fromCBOR(ptr)
    head = path.pop()
    if head.is_int() and ptr.is_array():
        for i in 0..head:
            ignore_element(ptr)
            return access(ptr, path)
    if head.is_string() and ptr.is_table():
        for key, value_ptr in ptr.elements:
            if key == head:
                return access(value_ptr, path)
    error "Bad file"
```


src/base/cbor.cpp
```diff
@@ -0,0 +1,192 @@
...
+    std::vector<uint8_t> result;
+
+    switch (node.type())
+    {
+        case toml::value_t::integer:
+        {
+            int64_t number = node.as_integer();
+            if (number >= 0)
+            {
+                result.push_back(0x1b);
```
trisyoungs, line 42: It feels like we can and should `enum`erate the various bit values corresponding to definite types.

tests/io/CMakeLists.txt
```diff
@@ -2,3 +2,4 @@ dissolve_add_test(SRC cif.cpp)
 dissolve_add_test(SRC exportTrajectory.cpp)
 dissolve_add_test(SRC intraParameterParse.cpp)
 dissolve_add_test(SRC version.cpp)
+dissolve_add_test(SRC cbor.cpp)
```
trisyoungs: Picky, I know, but can we add this at the top so it's in alphabetical order?

tests/io/cbor.cpp
```diff
@@ -0,0 +1,86 @@
...
+    std::ranges::subrange sub{cbor.begin(), cbor.end()};
+    auto [copy, remainder] = fromCBOR(sub);
+
+    EXPECT_EQ(remainder.begin(), remainder.end());
+    UnitTest::compare_toml("", node, copy);
+}
+
+} // namespace UnitTest
+
+TEST(CBORTest, BasicTypes)
```
trisyoungs: Might consider reimplementing these using a test fixture so they are more in line with other unit tests requiring their own functions.  As we move forward into streaming more complex datatypes this will probably come in handy.

trisyoungs (commented):
This is actually very elegant, and I like it a lot. Thinking forward a little bit to what our requirements for Dissolve actually are, would our basic process be to serialise the necessary per-node data to a single TOML map and then just `toCBOR` it?

Regarding the structure of the current restart file every piece of data starts with a string tag which has the module name in it, and we will need to do something very similar in our TOML/CBOR-based output too.  My concerns about random access are, however, probably mostly unfounded as the deserialisation can/would still proceed as a single pass read of the file, first determining the target `Node` and then calling it's `deserialise()` on the extracted `fromCBOR()`d TOML? Or am I mangling up the idea here?

rprospero:
@trisyoungs You're exactly right about how we would extract individual pieces of data out of the restart file.  In fact, it would even be possible to take it a step further and go any arbitrary depth into the file before calling `fromCBOR` and `deserialise`.

src/base/cbor.cpp
rprospero: That's a good call, but there's a catch.  During lunch, I figured out the smart way to handle the smart way to handle the variable bit depths.  The key code is really stored in the first three bits, the primary size in the next five, and the extra information in the remaining 0-64 bits.  It would be possible to write a single function that would extract the information from the code and use the primary size to return the rest as an integer.  This would enable use to support all of the bit depths without needing to write case functions for them, as we only need to switch on the 3-bit key code (with some weird subkeys under code 7 (i.e. True, False, and Float)).

The reason that I mention all of this is that the switch values currently in place will *not* be the ones that we use long term, so I'm going to argue that we push the enumeration onto the bit size patch that I'll write next week.

tests/io/CMakeLists.txt
rprospero: Agreed

tests/io/cbor.cpp
trisyoungs: I take this back - apparently we don't use fixtures as much as I thought we did!

PR #2121, opened 2025-04-01
feat: Argon nodes part 1 (atomShake, md)

RobBuchananCompPhys:
This PR implements two nodes relevant to Argon simulation - AtomShake, and MD.
There is also a transition to sub-directories for nodes that can be broken into several files. This changes the structure of the node CMakeLists, for the time being, to a recursive file search in all subdirectories (rather than multiple CMakeLists.txt files).


src/nodes/atomShake/atomShake.cpp
```diff
@@ -0,0 +1,31 @@
...
+    addInput<Configuration *>("Configuration", "Set target configuration for the module", targetConfiguration_)
+        ->setFlags({ParameterBase::Required, ParameterBase::Invalidates});
+
+    addOption<Number>("ShakesPerAtom", "Number of shakes to attempt per atom", nShakesPerAtom_);
+    addOption<Number>("TargetAcceptanceRate", "Target acceptance rate for Monte Carlo moves", targetAcceptanceRate_);
+    addOption<Number>("StepSizeMax", "Maximum allowed value for step size, in Angstroms", stepSizeMax_);
+    addOption<Number>("StepSizeMin", "Minimum allowed value for step size, in Angstroms", stepSizeMin_);
+}
+
+std::string_view AtomShakeNode::type() const { return "Atom Shake"; }
```
trisyoungs: I think all `Node` types should be simple strings - i.e. no spaces etc.

src/nodes/atomShake/process.cpp
```diff
@@ -0,0 +1,176 @@
+#include "atomShake.h"
```
trisyoungs: I would worry that this would break unless `CMAKE_INCLUDE_CURRENT_DIR` is set - elsewhere in the code we always tend to put full paths of header includes relative to the main `src/` directory.

src/nodes/md/md.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "classes/configuration.h"
+#include "classes/species.h"
+#include "data/atomicMasses.h"
+#include "keywords/bool.h"
+#include "keywords/configuration.h"
+#include "keywords/double.h"
+#include "keywords/integer.h"
+#include "keywords/optionalDouble.h"
+#include "keywords/optionalInt.h"
+#include "keywords/speciesVector.h"
```
trisyoungs: Most/all of these are now redundant and can be removed.

trisyoungs (commented):
Subdirectories definitely the way to go!

PR #2123, opened 2025-04-02
build: Add missing GSL libs to OSX package

trisyoungs:
This PR adds some missing GSL libs into the OSX package - while the Dissolve GUI links to two of the available dylibs there are clearly dependencies of those dylibs on two others in the Homebrew-installed location.


RobBuchananCompPhys (approved):
Looks good!

PR #2125, opened 2025-04-03
perf: More efficient CBOR serialisation

rprospero:
This PR ensures that each CBOR element is encoded in an appropriately sized container, instead of using 64 bits for all values.  As a quick check, the example TOML file `tests/data/dissolve/input/simple_addition_graph.toml` is 35% smaller in CBOR format.  Since this is already a fairly minimal and compact TOML file, I suspect that this is a lower bound and we will see greater improvements as the files become larger, especially for files with deeply nested arrays and tables.


src/base/cbor.cpp
```diff
@@ -38,15 +134,9 @@ std::vector<uint8_t> toCBOR(const SerialisedValue &node)
...
-                result.push_back(0x1b);
-            }
+                writeHeader(MajorKey::POS_INT, number, result);
             else
-            {
-                result.push_back(0x3b);
-                number *= -1;
-            }
-            ontoBuffer(number, result);
+                writeHeader(MajorKey::NEG_INT, -1 * number, result);
```
trisyoungs: Just curious - why not `abs(number)` here?  The intent is clearer there, IMO.

src/base/cbor.cpp
```diff
@@ -117,37 +204,33 @@ fromCBOR(std::ranges::subrange<std::vector<uint8_t>::iterator> bytes)
...
             break;
         }
-        case 0x9b: // Array
+        case MajorKey::ARRAY: // Array
         {
             std::vector<SerialisedValue> buf;
-            bytes.advance(1);
-            auto len = fromBuffer<uint64_t>(bytes);
-            for (int i = 0; i < len; ++i)
+            for (int i = 0; i < size; ++i)
```
trisyoungs:
```suggestion
            for (auto i = 0; i < size; ++i)
```

src/base/cbor.cpp
```diff
@@ -156,12 +239,10 @@ fromCBOR(std::ranges::subrange<std::vector<uint8_t>::iterator> bytes)
...
             break;
         }
-        case 0xbb: // Map
+        case MajorKey::TABLE: // Map
         {
             SerialisedValue map;
-            bytes.advance(1);
-            auto len = fromBuffer<uint64_t>(bytes);
-            for (int i = 0; i < len; ++i)
+            for (int i = 0; i < size; ++i)
```
trisyoungs:
```suggestion
            for (auto i = 0; i < size; ++i)
```

trisyoungs: Unless there's a subtlety I'm missing forcing it to `int`?

trisyoungs (approved):
I won't pretend to be able to "natively read" all the bitshifting and stuff myself, but looks good! Just a couple of minor queries.

src/base/cbor.cpp
rprospero: Good call!

PR #2126, opened 2025-04-04
feat: EPSRManager graph updates

Danbr4d:
Added the r-factors and additional potentials to the EPSRManager graphing.

### DIFF
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -50,4 +50,6 @@ class EPSRManagerModule : public Module
     public:
     // Run set-up stage
     bool setUp(ModuleContext &moduleContext, Flags<KeywordBase::KeywordSignal> actionSignals) override;
+    // Return list of target Modules / data for refinement
+    const std::vector<Module *> &targets() const;
 };
--- a/src/modules/epsrManager/epsrManager.h
+++ b/src/modules/epsrManager/epsrManager.h
@@ -50,6 +50,4 @@ class EPSRManagerModule : public Module
     public:
     // Run set-up stage
     bool setUp(ModuleContext &moduleContext, Flags<KeywordBase::KeywordSignal> actionSignals) override;
-    // Return list of target Modules / data for refinement
-    const std::vector<Module *> &targets() const;
 };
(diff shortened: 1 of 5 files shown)


src/modules/epsrManager/gui/epsrManagerWidget.h
```diff
@@ -31,4 +31,18 @@ class EPSRManagerModuleWidget : public ModuleWidget
     private:
     // Main form declaration
     Ui::EPSRManagerModuleWidget ui_;
+    // DataViewer contained within this widget
+    DataViewer *graph_;
+
+    /*
+     * Widgets / Functions
+     */
+
```
trisyoungs:
```suggestion
```

src/modules/epsrManager/epsrManager.h
```diff
@@ -50,4 +50,6 @@ class EPSRManagerModule : public Module
     public:
     // Run set-up stage
     bool setUp(ModuleContext &moduleContext, Flags<KeywordBase::KeywordSignal> actionSignals) override;
+    // Return list of target Modules / data for refinement
+    const std::vector<Module *> &targets() const;
```
trisyoungs: Strictly speaking you shouldn't need to implement this function as you can use `keywords_.get<const std::vector<Module *> >("Target");` to retrieve the info from the Module. See TR module line 26 in `process.cpp` for where you've done it for a (non-vector) before.

trisyoungs (approved):
Looks good. Might be worth trying the change I've suggested, but then good to go.

src/modules/epsrManager/gui/epsrManagerFuncs.cpp
```diff
@@ -11,5 +15,107 @@ EPSRManagerModuleWidget::EPSRManagerModuleWidget(QWidget *parent, EPSRManagerMod
...
+                                        auto pp = dissolve_.pairPotential(at1, at2);
+                                        if (pp)
+                                            graph_->createRenderable<RenderableData1D>(
+                                                std::format("Dissolve//Potential_{}_Additional", id), id, "Phi");
+                                    });
+        }
+        if (ui_.RFactorButton->isChecked())
+        {
+            graph_->groupManager().removeVerticalShifts();
+            auto optEPSRModule = module_->keywords().getVectorModule("Target");
```
trisyoungs: Just that name again!

trisyoungs (approved):
Same comment as the other PR, then good to go.

PR #2127, opened 2025-04-08
fix: Change epsr targets function to get

Danbr4d:
EPSR uses a targets() function that is not necessary as we have keywords.get().


src/modules/epsr/gui/epsrWidgetFuncs.cpp
```diff
@@ -69,12 +69,13 @@ void EPSRModuleWidget::updateControls(const Flags<ModuleWidget::UpdateFlags> &up
 
         const auto &atomTypes = module_->scatteringMatrix().atomTypes();
         const auto nAtomTypes = atomTypes.size();
+        auto optEPSRModule = module_->keywords().getVectorModule("Target");
```
trisyoungs: Surely `epsrModuleTargets` would be a better variable name than `optEPSRModule`?

trisyoungs (approved):
One little comment / request, then good to go.

PR #2128, opened 2025-04-10
feat: Argon nodes (Atomshake, MD)

RobBuchananCompPhys:
- [x] Clang format
- [x] Fix build error (no overloaded function found) in `toml11::find<T>(node, "data")`
^ this was addressed in #2135


src/nodes/atomShake/process.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    Messenger::print("Overall acceptance rate was {:4.2f}% ({} of {} attempted moves)\n", 100.0 * rate, nAccepted, nAttempts);
+
+    // Update and set translation step size
+    stepSize *= (nAccepted == 0) ? 0.8 : rate / targetAcceptanceRate;
+    if (stepSize < stepSizeMin)
+        stepSize = stepSizeMin;
+    else if (stepSize > stepSizeMax)
+        stepSize = stepSizeMax;
+    //keywords_.set("StepSize", stepSize);
```
trisyoungs:
```suggestion
```

src/nodes/atomShake/atomShake.cpp
```diff
@@ -0,0 +1,22 @@
...
+    addOption<Number>("TargetAcceptanceRate", "Target acceptance rate for Monte Carlo moves", targetAcceptanceRate_);
+    addOption<Number>("StepSizeMax", "Maximum allowed value for step size, in Angstroms", stepSizeMax_);
+    addOption<Number>("StepSizeMin", "Minimum allowed value for step size, in Angstroms", stepSizeMin_);
+}
+
+std::string_view AtomShakeNode::type() const { return "AtomShake"; }
+
+std::string_view AtomShakeNode::summary() const
+{
+    return "Perform an atomic Monte Carlo \"shake\" of every atom in each target configuration.";
```
trisyoungs:
```suggestion
    return "Perform a Monte Carlo trial move on every atom in the target configuration.";
```
I actually think I would like to rename this and the other "shake" nodes to reflect their actual intention more clearly - i.e. explicitly "AtomicMonteCarlo".

trisyoungs (commented):
All looking good here so far.

Since one of these nodes (AtomShake) is the first one we've converted which needs to save its data to the restart file (the current step size) we really should consider the mechanism of how this will work. I might do that as my next bit of work once this PR is merged in.

trisyoungs (approved):

PR #2130, opened 2025-04-10
perf: Read arbitrarily small parts of CBOR files

rprospero:
This PR adds the ability to extract an arbitrary element from a CBOR file.  This also necessitated a refctoring of the CBOR code to support both memory and file based backends.

Most of this refactoring is due to the utter stupidity of how file iterators work.  Specifically, the iterators read the stream when advancing, not when dereferencing.  THus, if you try to do:

```
   std::copy(stream.begin(), stream.begin()+10, destination.begin()
```

The stream immediately advances ten bytes before the copy even begins.  It also means that any two iterators on a stream are identical, so every algorithm is a no-op unless the end of the range is the Enod-Of-Stream special value.

As another pointless rage at the C++ committee and the IO library, the fact that

```
stream.get(destination, 10)
```
copies ***9*** characters into the destination array, but advances the stream by ***10*** characters, really deserves to be in giant bold letters in the documentation.


src/base/cbor.cpp
```diff
@@ -5,6 +5,9 @@
 #include "base/messenger.h"
 #include <bit>
 
+namespace CBOR
```
trisyoungs, line 8: :+1:

src/base/cbor.cpp
```diff
@@ -124,7 +195,7 @@ void writeHeader(MajorKey key, uint64_t size, std::vector<uint8_t> &buf)
 }
 
 // Convert a serialed Value to its CBOR representation
```
trisyoungs:
```suggestion
// Convert a serialised value to its CBOR representation
```

src/base/cbor.h
```diff
@@ -4,13 +4,27 @@
...
+#include <iostream>
 #include <ranges>
 #include <tuple>
+#include <variant>
 #include <vector>
 
+namespace CBOR
+{
+
 // Convert a serialed Value to its CBOR representation
```
trisyoungs:
```suggestion
// Convert a serialised value to its CBOR representation
```

trisyoungs (approved):
Looks good, although I hope I never have to to debug it! Particularly if the mantra of the C++ is "off-by-one-by-design" in places. :)

trisyoungs (approved):

PR #2132, opened 2025-04-11
fix: Minimum distance for larger, non-orthogonal cells

trisyoungs:
This PR addresses an issue where the minimum image calculation was not employed in all cases for larger (>= 70 Angstrom) monoclinic and triclinic `Box` types. We introduce a 3D lookup table of minimum corner distances between `Cell`s in the `CellArray` and probe this to determine the absolute distances, and hence mim-requirements, for any pair of cells.

A follow-up PR is now possible which can optimise and simplify other aspects of `Cell`-querying, but in this PR we focus on adding unit tests and fixing the original issue.


rprospero (approved):
Always good to document release procedures

PR #2135, opened 2025-04-11
refactor: Add framework for serialising and deserialising enum

rprospero:
This adds a concept for enums with an associated EnumOptions instance.  If you implement the `hasEnumOptions` function for the type, then serialisation of the parameter can be handled automatically.

One caveat is that this will fail for any enums that are not default constructible.  However, I'm not sure it's possible to make an enum that isn't default constructible without conjuring demons.

Note that this does *not* handle the case of deserialising vectors of pointers.

### DIFF
--- a/src/base/enumOptions.h
+++ b/src/base/enumOptions.h
@@ -219,3 +219,9 @@ template <class E> class EnumOptions : public EnumOptionsBase
 
     E deserialise(const SerialisedValue &node) { return enumeration(std::string(node.as_string())); }
 };
+
+template <typename T>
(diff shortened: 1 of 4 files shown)


rprospero:
Update:  This PR now *does* handle vectors of pointers.  It handles them by ignoring them, just as we ignore pointers, but I believe that that is the right call.  Users aren't entering raw lists of pointers, so every vector like this will ultimately be populated by some upstream module.

RobBuchananCompPhys (approved):

PR #2138, opened 2025-04-15
feat: Configuration node

RobBuchananCompPhys:


src/nodes/configuration.cpp
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#include "configuration.h"
```
trisyoungs:
```suggestion
#include "nodes/configuration.h"
```

src/nodes/configuration.h
```diff
@@ -0,0 +1,35 @@
...
+
+    public:
+    std::string_view type() const override;
+    std::string_view summary() const override;
+
+    /*
+     * Definition
+     */
+    private:
+    // Empty atomic configuration
```
trisyoungs: Small thing, but I would comment this differently as it won't be empty for long!
```suggestion
    // Configuration object
```

src/nodes/configuration.h
```diff
@@ -0,0 +1,35 @@
...
+    ConfigurationNode(Graph *parentGraph);
+    ~ConfigurationNode() override = default;
+
+    public:
+    std::string_view type() const override;
+    std::string_view summary() const override;
+
+    /*
+     * Definition
+     */
```
trisyoungs:
```suggestion

    /*
      * Definition
      */
    public:
    std::string_view type() const override;
    std::string_view summary() const override;

    /*
     * Data
     */
```

trisyoungs (approved):
Small comments in the below.  Not sure quite the best way for addressing the output type issue, but definitely one for another day.

PR #2139, opened 2025-04-17
No more templated vectors

trisyoungs:
This PR moves us away from the `Vec3<T>` and `Vec4<T>` templates in favour of specific typed versions. This greatly improves the clarity and intent in the code, especially given that many functions in the original were not relevant for anything but `T == double`. There should be some improvement on build times now too since we are not generating definitions for a rather large templated include ad nauseum, but I have not benchmarked it.

To summarise:
- `Vec3<double>` is now `Vector3`
- `Vec3<int>` is now `Vector3i` with a reduced, relevant function set
- `Vec3<bool>` is replaced with `std::array<bool, 3>` where appropraite
- `Vec3<Nodevalue>` is replaced with `VectorNodeValue`, a simple wrapper class basically
- `Vec4<double>` and `Vec4<GLfloat>` are replaces with `Vector4` based on a `double` - this largely affects the GUI plotting code, but since we are destined to remove it in future I didn't spend too much time worrying about the implications right now.


RobBuchananCompPhys (approved):
Seems like a sensible way to divide up the different vector types.

PR #2143, opened 2025-04-23
refactor: Tidy math constants, use radians throughout

### LINKED ISSUE #482: Update constants
The unit testing found that there was a slight discrependy in the value of DEGRAD.  We need to go through and make sure that all of the other constants in src/math/constants.h are correct.

trisyoungs:
This PR follows on from #2139 with a principal focus on tidying numerical constants. The result was the removal of `math/constants.h` entirely as they were either unused or were better located elsewhere.  A significant change coming from this was the introduction of `toRadians()` and `toDegrees()` in `DissolveMath` rather than the bare `DEGRAD` constant. This I believe improves readability and intent of the code.  A knock-on effect was then a slight refactoring of some energy and force routines which previously took angles in degrees to instead take them in radians, avoiding pointless conversions along the way.

Closes #482.


RobBuchananCompPhys (approved):
Certainly agree that this represents a big improvement to readability! Very much a fan of constants as namespace declarations rather than big caps, too.

RobBuchananCompPhys (approved):
Certainly agree that this represents a big improvement to readability! Very much a fan of constants as namespace declarations rather than big caps, too.

PR #2145, opened 2025-04-30
build: Fix generation of continuous ChangeLog.md

trisyoungs:
ChangeLog.md was printed out during the continuous CI, but wasn't saved anywhere, breaking publishing.


rprospero (approved):
This is the most coherent use of `tee` that I've ever seen.  I'm so used to seeing it in some monster script

PR #2146, opened 2025-04-30
feat: Atomic species node

RobBuchananCompPhys:


src/nodes/atomicSpecies.h
```diff
@@ -0,0 +1,44 @@
...
+    std::string_view summary() const override;
+
+    /*
+     * Data
+     */
+    private:
+    // Species object
+    std::unique_ptr<Species> species_{std::make_unique<Species>()};
+    // Local raw pointer to species
+    Species *rawSpecies_{species_.get()};
```
RobBuchananCompPhys: should be `const` ?

src/nodes/atomicSpecies.h
```diff
@@ -0,0 +1,44 @@
...
+    public:
+    std::string_view type() const override;
+    std::string_view summary() const override;
+
+    /*
+     * Data
+     */
+    private:
+    // Species object
+    std::unique_ptr<Species> species_{std::make_unique<Species>()};
```
RobBuchananCompPhys: No smart pointer necessary

RobBuchananCompPhys: nor for configuration

trisyoungs (approved):
I think this all looks good for now. Once you implement the `addPointerOutput()` function some things here will of course change, but may as well merge this in and refactor as part of that work IMO.

The next medium-to-large thing that this brings up is the forcefield stuff and how we deal with assigning the single atom type to the atomic species.  Easy enough to think about for the atomic case, but for molecules we will need something way more thought out.

PR #2147, opened 2025-05-01
fix: Visual Studio dev workflow

RobBuchananCompPhys:
A couple of changes to ensure our `msvc` workflow stays in working order
- Downgrade MSVC (for Antlr)
- Python 3.11
- Wrestled with some path finding in the powershell script to get it to work in GitHub (still seems to work on local machine)
- `msvc` CI now tracks changes to the main Windows CI action, to try and catch changes to the build process more readily

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -45,6 +45,7 @@ jobs:
         arch: x64
 
     - name: 'Set up Python 3.11'
+      id: setup-python
       uses: actions/setup-python@v5
       with:
         python-version: '3.11'
@@ -55,6 +56,7 @@ jobs:
       env:
         qtVersion: ${{ needs.Checkout.outputs.qtVersion }}
         antlrVersion: ${{ needs.Checkout.outputs.antlrVersion }}
+        PythonPath: ${{ steps.setup-python.outputs.python-path }}
       run: | 
         ./develop.ps1 -release -qtVersion $env:qtVersion -antlrVersion $env:antlrVersion -noPython
         
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -56,7 +56,7 @@ jobs:
         qtVersion: ${{ needs.Checkout.outputs.qtVersion }}
         antlrVersion: ${{ needs.Checkout.outputs.antlrVersion }}
       run: | 
-        ./develop.ps1 -release -qtVersion $env:qtVersion -antlrVersion $env:antlrVersion
+        ./develop.ps1 -release -qtVersion $env:qtVersion -antlrVersion $env:antlrVersion -noPython
         
     - name: 'Configure and build with CMake'
       shell: bash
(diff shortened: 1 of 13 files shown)


.github/workflows/msvc.yml
```diff
@@ -21,19 +22,45 @@ jobs:
...
+    - name: 'Set up Python 3.11'
+      id: setup-python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.11'
+
+    - name: 'Install Downgraded MSVC'
+      shell: powershell
+      env:
+        msvcVersion: '14.41.17.11'
```
trisyoungs: Since we now use this version ID in two places we should consider adding it as a variable defined at the top-level in `_checkout.yml` and passed down to the workflows as an input (as we do for `qtVersion` and a bunch of other stuff).

develop.ps1
```diff
@@ -3,15 +3,23 @@
         Script to install dependencies for Dissolve development environment in Visual Studio.
     .DESCRIPTION
         Installs the following dependencies for Dissolve (separate and prior to Conan-managed packages):
+            - Python 3.12 (unless stated otherwise)
```
trisyoungs, line 6: We used 3.11 in the script?

develop.ps1
```diff
@@ -3,15 +3,23 @@
...
+            - CMake 3.3
```
trisyoungs: Probably clearer to state "CMake 3.X".

develop.ps1
```diff
@@ -68,28 +79,46 @@ try {
...
+            Write-Output "Found system Python..." @info_colors
+            $pythonVersion = $(python -c "import sys; v = sys.version_info; print(v.major == 3, v.minor == 12)")
+            $versionParts = $pythonVersion -split " "
+            if (-not ($versionParts[0] -eq "True" -and $versionParts[1] -eq "True")) {
+                Write-Output "System Python is version $(python --version) and it is recommended to be version == 3.12 - installing with Chocolatey..." @info_colors
+                choco install -y python --version=3.12.0
+            }
+        } catch {
+            Write-Output "Could not find system Python - installing with Chocolatey..." @info_colors
+            choco install -y python --version=3.12.0
```
trisyoungs, line 106: Should the Python version here be an environment variable?

trisyoungs (approved):
Looks good. Some little suggestions which will no doubt turn into headaches (sorry!).
Will test when I have Windows machine to hand.

develop.ps1
RobBuchananCompPhys: 3.12 is the latest and default version used by this Powershell script, but 3.11 can also be used if the path or version is supplied as an argument.

develop.ps1
RobBuchananCompPhys: I believe it already is supplied as an environment variable, via the new `pythonPath` parameter, and is handled that way in the CI job.

trisyoungs (approved):
All looks good to me.  Let's see if it works on the `continuous` build!

PR #2148, opened 2025-05-02
chore: Fix cells tests

trisyoungs:
So as it turns out I had already fixed the errors resulting from the rebase of `develop2` that we now see in the CI, but had not actually committed them from my home machine.

Here they are!


RobBuchananCompPhys (approved):
Perfect!

PR #2151, opened 2025-05-07
feat: InsertRandom

trisyoungs:
This PR begins the journey of the new set of `Insert` nodes that will replace functionality currently in the various `Add` generator nodes. Given point 1) below, the new `InsertRandom` node is probably better called  just `Insert`.

Things to note:
1) `InsertRandom` is very specific and does not even consider `Region`s.  The intent is to have a specific `InsertRegion` node for that.
2) Because a `MoleculeSet` is now the source of multiple molecules rather than a `CoordinateSets`, there is no longer a guarantee that all molecules derive from the same `Species`. This must now be accounted for in the initial size adjustment of the `Box`.


src/nodes/insertBase.cpp
```diff
@@ -0,0 +1,135 @@
...
+        // Scale existing contents
+        configuration_->scaleContents(scaleFactors);
+
+        // Scale the current Box so there is enough space for our new species
+        configuration_->scaleBox(scaleFactors);
+
+        Messenger::print(" ... Current box scaled by ({},{},{}) - new volume is {:e} cubic Angstroms.\n", scaleFactors.x,
+                         scaleFactors.y, scaleFactors.z, configuration_->box()->volume());
+    }
+}
```
RobBuchananCompPhys: I like the idea of breaking this `adjustBoxVolume` method up a little, just to keep it to a single level of conditionals, since there are nested if/else statements going on. Could we have two private methods `scaleVolume` and `addVolume` to encompass what happens when either enum value is hit:
```
void InsertNodeBase::adjustBoxVolume(Configuration *cfg, int nAtomsToAdd, double massToAdd) const
{
    // Determine volume required to contain the population of the specified Species at the requested density
    auto requiredVolume = 0.0;
    if (densityUnits_ == Units::AtomsPerAngstromUnits)
        requiredVolume = nAtomsToAdd / density_.asDouble();
    else
        requiredVolume = (nAtomsToAdd / DissolveMath::Avogadro) / (density_.asDouble() / 1.0E24);

    // If a density was not given, just add new molecules to the current box without adjusting its size
    switch(boxAction_)
    {
        case InsertNodeBase::BoxActionStyle::AddVolume:
        {
            addVolume();
            break;
        }
        case InsertNodeBase::BoxActionStyle::ScaleVolume:
        {
            scaleVolume();
            break;
        }
        case InsertNodeBase::BoxActionStyle::None: { }
        default:
           Messenger::print(" ... Current box geometry / volume will remain as-is.\n");
           break;
    }
}
```

src/nodes/insertBase.h
```diff
@@ -0,0 +1,60 @@
...
+    protected:
+    // Target configuration to insert into
+    Configuration *configuration_;
+    // The default box action if none is specified
+    static constexpr BoxActionStyle defaultBoxAction_ = BoxActionStyle::AddVolume;
+    // Action to take on the Box geometry / volume on addition of the species
+    BoxActionStyle boxAction_{defaultBoxAction_};
+    // Target density when adding molecules (if adjusting box size)
+    Number density_{1.0};
+    Units::DensityUnits densityUnits_{Units::GramsPerCentimetreCubedUnits};
```
RobBuchananCompPhys: Any chance we could get a brief descriptive comment for this `densityUnits_` member?

src/classes/moleculeSet.cpp
```diff
@@ -3,9 +3,12 @@
...
 
-// Return nth local molecule
-LocalMolecule &MoleculeSet::localMolecule(int n) { return localMolecules_[n]; }
-const LocalMolecule &MoleculeSet::localMolecule(int n) const { return localMolecules_[n]; }
+// Add a new molecule from the specified Species
+LocalMolecule &MoleculeSet::addMolecule(const Species *species) { return localMolecules_.emplace_back(species); }
+
+// Return nth local molecule, wrapping index if necessary
+LocalMolecule &MoleculeSet::localMolecule(int n) { return localMolecules_[n % localMolecules_.size()]; }
+const LocalMolecule &MoleculeSet::localMolecule(int n) const { return localMolecules_[n % localMolecules_.size()]; }
```
RobBuchananCompPhys, line 11: This is interestingly the first time I've come across indexing with the modulo operator, it seems to make sense mathematically and fits nice on one line - however, I'd worry that evaluating the `n % size` inside the square brackets might complicate matters in a debugging context, where you might want to know the value before the indexing happens.

RobBuchananCompPhys, line 11:
```suggestion
const LocalMolecule &MoleculeSet::localMolecule(int n) const
{
   auto index = n % localMolecules_.size();
   return localMolecules_[index];
}
```

src/nodes/insertBase.cpp
trisyoungs: Yes. I was about to make the case that this should be done in the `InsertBaseNode` class since there will be other nodes that require the same check, but on reflection I'm not sure that there are! As such, following your suggestion above.

trisyoungs: In fact, going further I'm not even sure an `InsertBaseNode` is warranted at this point, so I'm going to remove it and absorb everything into `Insert` (yep, renaming this one).

src/classes/moleculeSet.cpp
trisyoungs: I'm not convinced that we need to go this far - if debugging is required then compiling in DEBUG mode will give us out-of-bounds errors on the `[]` operator. Given the array access is using it's own size for the modulo I think I'm not too worried!

RobBuchananCompPhys (approved):

PR #2153, opened 2025-05-07
feat: Node messenger

RobBuchananCompPhys:
This PR aims to capture the requirement for individual `Node` instances to hold a local record of messages, represented by the `MessageStore` type alias. I currently envisage each Node-local message to have a `MessageStatus` enum attributed to it (one of `Info`, `Warn`, `Error` or `Exception`) to allow us to differentiate between message levels.

### DIFF
--- a/src/nodes/node.cpp
+++ b/src/nodes/node.cpp
@@ -27,7 +27,7 @@ std::string_view Node::name() const { return parentGraph_ ? parentGraph_->nodeNa
  */
 
 // Add message to store
-void Node::postMessage(std::string_view message, MessageStatus status) { messages_[status].emplace_back(message); }
+void Node::postMessage(std::string message, MessageStatus status) { messages_[status].emplace_back(message); }
 
 // Message store vector
 const Node::MessageStore &Node::messages() const { return messages_; }
--- a/src/nodes/node.cpp
+++ b/src/nodes/node.cpp
@@ -5,6 +5,7 @@
 #include "base/sysFunc.h"
 #include "nodes/edge.h"
 #include "nodes/graph.h"
+#include <algorithm>
 
 /*
  * Definition
@@ -27,7 +28,18 @@ std::string_view Node::name() const { return parentGraph_ ? parentGraph_->nodeNa
  */
 
 // Message store vector
-const Node::MessageStore &Node::messages() const { return messages_; }
+const Node::MessageStore &Node::messages(MessageStatus status) const
+{
+    if (status != MessageStatus::Any)
+    {
+        MessageStore filtered;
+        std::copy_if(messages_.cbegin(), messages_.cend(), std::back_inserter(filtered),
+                     [status](const auto &entry) { return entry.first == status; });
+        return filtered;
+    }
+
+    return messages_;
+}
 
 /*
  * Inputs, Outputs & Options
--- a/src/nodes/node.cpp
+++ b/src/nodes/node.cpp
@@ -25,11 +25,9 @@ std::string_view Node::name() const { return parentGraph_ ? parentGraph_->nodeNa
 /*
  * Node message
  */
+
 // Add message to store
-void Node::postMessage(std::string_view message, MessageStatus status)
-{
-    messages_.emplace_back(std::make_pair(status, message));
-}
+void Node::postMessage(std::string_view message, MessageStatus status) { messages_[status].emplace_back(message); }
 
 // Message store vector
 const Node::MessageStore &Node::messages() const { return messages_; }
--- a/src/nodes/node.cpp
+++ b/src/nodes/node.cpp
@@ -23,6 +23,18 @@ void Node::setName(std::string_view newName)
 std::string_view Node::name() const { return parentGraph_ ? parentGraph_->nodeName(this) : "UnparentedNode"; }
 
 /*
+ * Node message
+ */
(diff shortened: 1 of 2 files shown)


src/nodes/node.cpp
```diff
@@ -22,6 +22,18 @@ void Node::setName(std::string_view newName)
 // Return node name
 std::string_view Node::name() const { return parentGraph_ ? parentGraph_->nodeName(this) : "UnparentedNode"; }
 
+/*
+ * Node message
+ */
+// Add message to store
```
```suggestion

// Add message to store
```

src/nodes/node.h
```diff
@@ -45,6 +45,28 @@ class Node : public Serialisable<>
...
+     * Node message
+     */
+    enum class MessageStatus
+    {
+        Info,
+        Warn,
+        Error,
+        Exception
+    };
+    using MessageStore = std::vector<std::pair<Node::MessageStatus, std::string_view>>;
```
trisyoungs: :+1: Allows us to maintain the order of messages and decorate them in the GUI with icons according to status if we wanted to (and I think we wanted to). Determining whether a `Node` has encountered _any_ warnings or errors, however, needs a search of the entire vector, but the alternative option - something like `std::map<Node::MessageStatus, std::vector<std::string_view>>` - kills the ordering of messages.

trisyoungs: This has to use `std::string` rather than `std::string_view` because our string messages will be mostly dynamically created at runtime using `std::format`, and there will be nothing guaranteed for the `string_view` to point to.

src/nodes/node.h
```diff
@@ -45,6 +45,28 @@ class Node : public Serialisable<>
...
+
+    private:
+    // Message store vector
+    MessageStore messages_;
+    // Post message to store
+    void postMessage(std::string_view message, MessageStatus status = MessageStatus::Info);
```
trisyoungs: `Messenger` provides us a few convenience functions to take a formatting string and variable argument list and forward them neatly on to `std::format`, as well as provide any necessary additional formatting (e.g. the "*** ERROR" prefix), splitting lines as necessary, and sending the message to the correct place (console or GUI). Now, with this focus on `Node`s we don't need _any_ of this, I think, so the `Messenger` class will basically be obsoleted, but I think we _do_ still need a similar format string plus var args capability in the message functions in `Node`. I would suggest three: `message()`, `warn()`, and `error()`. Each would take a format string and var args and set the relevant `MessageStatus` flag in the vector when storing the message. If we do want to enable CLI printing, this could be as simple as an `echo_` static variable in the `Node` class which we turn on when needed, and which immediately prints the message after storing it in the local vector.

trisyoungs (commented):
Yes, this is what we need. Comments below!

src/nodes/node.h
```diff
@@ -45,6 +45,58 @@ class Node : public Serialisable<>
...
+        Any,
+        Info,
+        Warn,
+        Error,
+        Exception
+    };
+
+    using MessageStore = std::vector<std::pair<MessageStatus, std::string>>;
+    // Print latest message
+    static bool echo_;
```
trisyoungs, line 62:
```suggestion
    static bool echo_{false};
```

src/nodes/node.cpp
```diff
@@ -22,6 +23,30 @@ void Node::setName(std::string_view newName)
 // Return node name
 std::string_view Node::name() const { return parentGraph_ ? parentGraph_->nodeName(this) : "UnparentedNode"; }
 
+/*
+ * Node message
+ */
+
+// Print latest message
+bool Node::echo_ = false;
```
trisyoungs, line 31:
```suggestion
```

src/nodes/node.cpp
```diff
@@ -22,6 +23,30 @@ void Node::setName(std::string_view newName)
...
+const Node::MessageStore &Node::messages(MessageStatus status) const
+{
+    if (status != MessageStatus::Any)
+    {
+        MessageStore filtered;
+        std::copy_if(messages_.cbegin(), messages_.cend(), std::back_inserter(filtered),
+                     [status](const auto &entry) { return entry.first == status; });
+        return filtered;
+    }
+
```
trisyoungs: Surprised this compiles as you're returning a reference to a temporary? I wonder if we really need to retrieve a specific class of message from the `Node`, and would be able to make to with a simpler `bool hasMessages(MessageStatus status) const` function which returns if there are any messages of the specified type. That would be enough to create a warning icon in the GUI on the relevant nodes.

src/nodes/node.h
```diff
@@ -45,6 +45,58 @@ class Node : public Serialisable<>
...
+     */
+
+    // Enumeration for message status
+    enum class MessageStatus
+    {
+        Any,
+        Info,
+        Warn,
+        Error,
+        Exception
```
trisyoungs: We talked about not needing to store exceptions, so:
```suggestion
        Error
```

trisyoungs (commented):
Looking good, but a couple more comments to address. One other thing we need to think about is when we clear messages every time - should be done in process(), but we can think about that later.

src/nodes/node.cpp
RobBuchananCompPhys: Agreed, seems like a sensible solution

trisyoungs (approved):
:+1:

PR #2154, opened 2025-05-09
feat: Argon graph test

### LINKED ISSUE #2092: Unit test for unique Node naming within a Graph

trisyoungs:
Adds a unit test for creating a working Argon example simulation structure. A bunch of stuff had be changed along the way, but mostly small stuff or things that are now in a better place moving forward.

Closes #2092.

### DIFF
--- a/src/nodes/edge.cpp
+++ b/src/nodes/edge.cpp
@@ -117,6 +117,12 @@ bool EdgeDefinition::operator==(const EdgeDefinition &other) const
            targetInput == other.targetInput;
 }
 
+// Return as a string
+std::string EdgeDefinition::asString() const
+{
+    return std::format("Edge( [O] {}@{} -> {}@{} [I] )", sourceOutput, sourceNode, targetInput, targetNode);
+}
+
 // Express as a serialisable value
 SerialisedValue EdgeDefinition::serialise() const
 {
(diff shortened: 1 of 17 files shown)


src/nodes/edge.cpp
```diff
@@ -117,6 +117,12 @@ bool EdgeDefinition::operator==(const EdgeDefinition &other) const
            targetInput == other.targetInput;
 }
 
+// Return as a string
+std::string EdgeDefinition::asString() const
+{
+    return std::format("Edge( [O] {}@{} -> {}@{} [I] )", sourceOutput, sourceNode, targetInput, targetNode);
+}
+
```
RobBuchananCompPhys, line 125: This is a nice string representation of the edge - makes me wonder if there might be a need at some point to cement these symbols in a kind of string 'vocabulary' for graphs/nodes/edges in general (for instance if we wanted to represent as a string the "interior" view of a node, or how a graph relates to a subgraph, we could have virtual methods in those classes that do similar and specific formatting for each).

src/nodes/parameter.h
```diff
@@ -76,8 +77,11 @@ class ParameterBase : public Serialisable<>
     {
         if (std::type_index(typeid(T)) != type_)
             return nullptr;
-        auto casted = static_cast<Parameter<T> *>(this);
-        return casted->shared_from_this();
+        auto cast1 = dynamic_cast<PointerParameter<T> *>(this);
+        if (cast1)
+            return cast1->shared_from_this();
```
rprospero, line 82: I always get a little nervous about using dynamic_cast for this type of runtime type checking, but every solution I can think of is just as ugly.

rprospero (approved):
Beyond my urge to bikeshed around the use of `dynamic_cast`, everything looks good

PR #2155, opened 2025-05-09
feat: Optional bounded Number

RobBuchananCompPhys:
Storing numeric upper and lower bounds in `Number` should alleviate the need for methods like `addBoundedInput<Number>`

- [x] Still a question of how to handle operators for bounded `Number` (i.e. do we blindly apply the operators to the bounds as well as `value_`, or do we need to consider whether `other` and `this` actually both have bounds)
- [ ] Refactor nodes to utilise bounded `Number` instead of `addBoundedInput`

### DIFF
--- a/src/nodes/number.cpp
+++ b/src/nodes/number.cpp
@@ -80,12 +80,12 @@ bool Number::operator<=(const Number &other) const
 
 bool Number::operator>(const Number &other) const
 {
-    return std::visit([](auto a, auto b) -> bool { return b < a; }, value_, other.value_);
+    return std::visit([](auto a, auto b) -> bool { return a > b; }, value_, other.value_);
 }
 
 bool Number::operator>=(const Number &other) const
 {
-    return std::visit([](auto a, auto b) -> bool { return b <= a; }, value_, other.value_);
+    return std::visit([](auto a, auto b) -> bool { return a >= b; }, value_, other.value_);
 }
 
 /*
--- a/src/nodes/number.cpp
+++ b/src/nodes/number.cpp
@@ -138,8 +138,7 @@ std::optional<double> Number::doubleMin()
     if (!hasBounds())
         return {};
 
-    auto min = asDouble(min_.value());
-    return min;
+    return asDouble(min_.value());
 }
 
 // Return optional integer lower bound
--- a/src/nodes/number.cpp
+++ b/src/nodes/number.cpp
@@ -117,44 +117,50 @@ double Number::asDouble(NumberVariant n) const
 // Set from other Number
 void Number::set(const Number &other) { value_ = other.value_; }
 
-// Return whether the number has bounds
-bool Number::hasBounds() const { return min_.has_value() && max_.has_value(); }
+// Return whether the number has lower bound
+bool Number::hasLowerBound() const { return min_.has_value(); }
+
+// Return whether the number has upper bound
+bool Number::hasUpperBound() const { return min_.has_value(); }
 
 // Return optional double upper bound
 std::optional<double> Number::doubleMax()
 {
-    if (!hasBounds())
-        return {};
+    if (hasUpperBound())
+        return asDouble(max_.value());
 
-    return asDouble(max_.value());
+    return {};
 }
 
 // Return optional integer upper bound
 std::optional<int> Number::integerMax()
 {
-    if (!hasBounds())
-        return {};
+    if (hasUpperBound())
+        return asInteger(max_.value());
 
-    return asInteger(max_.value());
+    return {};
 }
 
 // Return optional double lower bound
 std::optional<double> Number::doubleMin()
 {
-    if (!hasBounds())
-        return {};
+    if (hasLowerBound())
+        return asDouble(min_.value());
 
-    return asDouble(min_.value());
+    return {};
 }
 
 // Return optional integer lower bound
 std::optional<int> Number::integerMin()
 {
-    if (!hasBounds())
-        return {};
+    if (hasLowerBound())
+        return asInteger(min_.value());
 
-    return asInteger(min_.value());
+    return {};
 }
+
+// Return whether the number has bounds
+bool Number::isBounded() const { return hasLowerBound() && hasUpperBound(); }
 
 // Return whether the contained value is an integer
 bool Number::isInteger() const { return isInteger(value_); }
--- a/src/nodes/number.cpp
+++ b/src/nodes/number.cpp
@@ -142,7 +142,7 @@ std::optional<double> Number::doubleMin()
     return min;
 }
 
-// Return optional integer lowe bound
+// Return optional integer lower bound
 std::optional<int> Number::integerMin()
 {
     if (!hasBounds())
--- a/src/nodes/number.cpp
+++ b/src/nodes/number.cpp
@@ -34,7 +34,7 @@ Number::Number(double value, std::optional<double> min, std::optional<double> ma
 
 Number &Number::operator=(const Number &other)
 {
-    value_ = other.value_;
+    set(other);
     return *this;
 }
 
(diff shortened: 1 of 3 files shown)


src/nodes/number.cpp
```diff
@@ -10,6 +10,16 @@ Number::Number(int value) : value_(value) {}
 
 Number::Number(double value) : value_(value) {}
 
+Number::Number(int value, std::optional<int> min, std::optional<int> max)
+    : value_(value), min_(std::optional<NumberVariant>(min.value())), max_(std::optional<NumberVariant>(max.value()))
```
trisyoungs: Won't this throw if you pass `std::nullopt` as min and/or max since you're always calling `min.value()`? You may have to resort to a couple of `if` statements checking whether those parameters were given, and setting `min_` and/or `max_` accordingly. Same for the `double` variation.

src/nodes/number.cpp
```diff
@@ -68,33 +78,91 @@ bool Number::operator!=(const Number &other) const { return value_ != other.valu
...
 }
 
+// Set from other Number
+void Number::set(const Number &other)
+{
+    if (other.hasBounds())
+        min_ = other.min_, max_ = other.max_;
+
+    value_ = other.value_;
+}
```
trisyoungs: Now, here is a point for discussion. I don't think we need to pass on the bounds like this, reason being that the limits should only be relevant within the `Number` which was constructed with them.  If the value is passed on (e.g. by copying `Number` and passing it through an `Edge`) the limits don't really apply any more, but I don't think this will be a common thing anyway as the bounds are most relevant for input / option parameters that probably only have relevance within their own `Node`.

trisyoungs: Second point - the function of the `set()` function should not be to copy bounds from the source `Number`, but only to enforce its own limits on the incoming value. So here we should check for each bound being specified and clamping the incoming number to those bounds if necessary.

src/nodes/number.cpp
```diff
@@ -68,33 +78,91 @@ bool Number::operator!=(const Number &other) const { return value_ != other.valu
...
+
+// Return optional integer lowe bound
+std::optional<int> Number::integerMin()
+{
+    if (!hasBounds())
+        return {};
+
+    return asInteger(min_.value());
+}
```
trisyoungs: Here's yet another interesting thought! Obviously you added these functions for the GUI to query for spinbox controls when we're ready, but how does the GUI (or any other bit of the code for that matter) know which value type to use? There's a fair bit to unpack here:
1) There are lots of occasions where we only want a `Number` to contain an `int`, but we don't currently enforce that at all.
2) Similarly, a `Number` can freely change between `int` and `double`, so querying the current type wouldn't help the GUI know what the true limits are ( double -> int loses precision etc.)
3) Any bounds created on construction never change over the lifetime of the `Number`, and so represent the true type of the limiting values.

We may need to add the ability to enforce integral values within `Number`...

src/nodes/number.cpp
```diff
@@ -68,33 +78,91 @@ bool Number::operator!=(const Number &other) const { return value_ != other.valu
...
+std::optional<double> Number::doubleMin()
+{
+    if (!hasBounds())
+        return {};
+
+    auto min = asDouble(min_.value());
+    return min;
+}
+
+// Return optional integer lowe bound
```
trisyoungs:
```suggestion
// Return optional integer lower bound
```

trisyoungs (commented):
Good so far.  Made some comments which may change our direction on this, however!

src/nodes/number.cpp
RobBuchananCompPhys: Absolutely!

src/nodes/number.cpp
```diff
@@ -6,13 +6,35 @@
 
 Number::Number(const std::variant<int, double> &value) : value_(value) {}
 
-Number::Number(int value) : value_(value) {}
+Number::Number(int value, std::optional<int> min, std::optional<int> max) : value_(value)
+{
+    if (min && min.has_value())
+    {
+        min_ = min;
+    }
```
trisyoungs: So assigning an `std::optional<int>` to our `std::optional<NumberVariant>` works with implicit conversion? If so then there's not even a need for the `if` - the incoming `std::optional` can just be assigned directly.

src/nodes/number.cpp
```diff
@@ -57,44 +82,134 @@ Number Number::operator/(const Number &other) const
...
     return *this;
 }
 
 bool Number::operator==(const Number &other) const { return value_ == other.value_; }
 
 bool Number::operator!=(const Number &other) const { return value_ != other.value_; }
 
+bool Number::operator<(const Number &other) const
+{
+    return std::visit([](auto a, auto b) -> bool { return {a < b}; }, value_, other.value_);
```
trisyoungs: Shouldn't need the braces here, i.e. can just have `a < b` etc.

src/nodes/number.cpp
```diff
@@ -57,44 +82,134 @@ Number Number::operator/(const Number &other) const
...
+void Number::set(const Number &other) { value_ = limit(other.value_); }
+
+// Return whether the number has lower bound
+bool Number::hasLowerBound() const { return min_.has_value(); }
+
+// Return whether the number has upper bound
+bool Number::hasUpperBound() const { return max_.has_value(); }
+
+// Return optional double upper bound
+std::optional<double> Number::doubleMax()
```
trisyoungs: We don't use these in the GUI yet but can we just have generic functions to return `std::optional<NumberVariant> max()`? This conveys all the information the GUI widget needs to know - whether there is a defined bound, as well as the type if required.

src/nodes/number.cpp
```diff
@@ -6,13 +6,35 @@
...
+    {
+        max_ = max;
+    }
+}
 
 Number &Number::operator=(const Number &other)
 {
-    value_ = other.value_;
+    set(other);
     return *this;
```
trisyoungs, line 16: If we rewrote the `limit()` function to be more operator-like - e.g. `NumberVariant &limit(NumberVariant newValue)` - then all of the code...
```
    set(other);   // or set(value_);
    return *this;
```
...could be replaced with...
```
    return limit(other);   // or return limit(value_);
```
Just a thought!

src/nodes/number.cpp
```diff
@@ -57,44 +82,134 @@ Number Number::operator/(const Number &other) const
...
+}
+
+bool Number::operator<=(const Number &other) const
+{
+    return std::visit([](auto a, auto b) -> bool { return {a <= b}; }, value_, other.value_);
+}
+
+bool Number::operator>(const Number &other) const
+{
+    return std::visit([](auto a, auto b) -> bool { return {b < a}; }, value_, other.value_);
```
trisyoungs: Well, this confused me at first and I wondered how your unit tests were passing! For clarity, can we write `a > b` instead to match the actual operator? Same for `>=` below.

tests/nodes/number.cpp
```diff
@@ -157,6 +354,24 @@ TEST(NumberTest, MultiplyAssignment)
...
+    EXPECT_EQ(a.asInteger(), 2);
+    EXPECT_TRUE((a *= 2.0).isInteger());
+    EXPECT_DOUBLE_EQ(a.asInteger(), 2);
+
+    // Multiplication of a double always results in a double
+    Number b(1.0, {}, 2.0);
+    EXPECT_TRUE((b *= 2).isDouble());
+    EXPECT_DOUBLE_EQ(b.asDouble(), 2.0);
+    EXPECT_TRUE((b *= 2.0).isDouble());
+    EXPECT_DOUBLE_EQ(b.asDouble(), 2.0);
```
trisyoungs, line 370: This doesn't actually test whether the limiting works, since you're multiplying exactly up to the upper bound.  Need a bigger operand!

trisyoungs (commented):
Some comprehensive unit testing here - excellent!
Have made a few suggestions on how a few things might be tweaked.

tests/nodes/number.cpp
RobBuchananCompPhys: Good catch!

src/nodes/number.cpp
RobBuchananCompPhys: Yep, this is probably preferable.

PR #2156, opened 2025-05-12
feat: Update node type checking, add node registry unit test

### LINKED ISSUE #2091: Unit test for `NodeRegistry`

trisyoungs:
A small PR to add a bit more unit test coverage, and start thinking about more relaxed determination of valid node type names.

Closes #2091.

### DIFF
--- a/src/nodes/graph.cpp
+++ b/src/nodes/graph.cpp
@@ -35,9 +35,28 @@ std::string Graph::uniqueNodeName(const Node *node, std::string_view baseName) c
 }
 
 // Create node of specified type with the name provided
-Node *Graph::createNode(std::string_view type, std::string_view name)
+Node *Graph::createNode(std::string_view nodeType, std::string_view nodeName, bool strictTypeName)
 {
-    return addNode(NodeRegistry::produce(this, type), name.empty() ? type : name);
+    std::string newNodeType{nodeType};
+    if (strictTypeName)
+    {
+        if (!NodeRegistry::hasNodeType(nodeType))
+        {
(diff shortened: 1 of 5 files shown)


rprospero:
@trisyoungs Did you mean to merge this into `develop` or should this merge into `develop2`

trisyoungs:
@rprospero Curses - I keep doing that!  Meant for `develop2`. Happy for you to merge if you approve.

src/nodes/graph.cpp
```diff
@@ -29,21 +29,42 @@ std::string Graph::uniqueNodeName(const Node *node, std::string_view baseName) c
...
-        newName = std::format("{}%02d", baseName, count++);
+        newName = std::format("{}{}", baseName, count++);
 
     return newName;
 }
 
-// Add nodes
-Node *Graph::addNode(std::string_view type, std::string_view name)
+// Create node of specified type with the name provided
+Node *Graph::createNode(std::string_view nodeType, std::string_view nodeName, bool strictTypeName)
```
rprospero, line 38: I think that, in the long run, we'll want this function to remain always strict.  What we'll want to do is have the registry be able to return a sorted list of node types based on the user input nd the user can select one of those (like any other combo box entry).  It's this exact value that would then be passed to `createNode`.  This both simplifies the logic of this function and allows for matches with multiple interpretations.

I'll add this on for a future issue - it shouldn't stop this PR from being merged.

rprospero (approved):
I'll put in an issue on the fuzzy matching, but otherwise things look good

PR #2158, opened 2025-05-13
feat: Graph Node GUI

rprospero:
This PR merges in the beginning of the Graph Node GUI.  The following functionality is currently implemented:

gui

- Right click anywhere in the window to bring up a menu to add new nodes
- Node names can be seen and editted
- Nodes can be moved by clicked on the icon and dragging
- Nodes can be deleted by clicking on the X button
- Node Inputs and Outputs are visibly enumerated
- Node Options are enumerated
- The values of Node options are visible for Number and Bool types

The following *bugs* are know

- For some reason, there is always and extra empty node.  This needs fixing
- Most nodes are missing icons.  For some nodes, this will consist of renaming a file, but other nodes may require whole new icons

The following basic functionality is needed, but outside the scope of this PR:

- Edges are not currently displayed
- Edges cannot be created or destroyed
- Parameter options are not properly editted
- Parameters of *all* types should be properly displayed and editted (esp. enum types)
- Either remove the toolbar or replace it with something relevant


CMakeLists.txt
```diff
@@ -285,6 +285,23 @@ if(GUI)
...
+  )
+  foreach(icon ${icons})
+    if(${icon} MATCHES "\\.\\..*")
+
+    else()
+      set_source_files_properties(${SRCS}/gui/icons/${icon} PROPERTIES QT_RESOURCE_ALIAS icons/${icon})
+      list(APPEND REAL_ICONS "${SRCS}/gui/icons/${icon}")
+    endif()
+  endforeach()
+
```
RobBuchananCompPhys: Wondering if this recursive search for icon files is strictly necessary now we have a QML `icons` module that performs the same

(see: https://github.com/disorderedmaterials/dissolve/pull/2059/files#:~:text=GLOB%20basic_icons-,RELATIVE%20%24%7BCMAKE_CURRENT_SOURCE_DIR%7D,-%22./*.svg%22)

RobBuchananCompPhys:
Just a thought - do we have a `.qmlformat.ini` file for dissolve2? It might offer greater flexibility over exactly how we format QML going ahead.

rprospero:
@RobBuchananCompPhys I wouldn't be opposed to explicitly declaring the qml format.  The one issue that we ran into in the past was that different version of `qmlformat` (e.g.  6.2 vs. 6.5) would format the same code differently, even with the same `.qmlformat.ini` file, so it wasn't worthwhile at the time.

src/gui/models/dissolveModel.cpp
```diff
@@ -2,11 +2,20 @@
...
+#include "nodes/dissolve.h"
 
 /*
  * Data
  */
 
+Dissolve &DissolveModel::dissolve()
+{
+    if (!dissolve_)
+        Messenger::exception("DissolveModel is laching a backend.  This should *never* happen, so please contact the "
```
trisyoungs:
```suggestion
        Messenger::exception("DissolveModel is lacking a backend.  This should *never* happen, so please contact the "
```

src/gui/models/nodeGraph/CMakeLists.txt
trisyoungs, line 1: Since we use `automoc` and friends we shouldn't need to use the horrid `qt6_wrap_cpp` etc. any more.

src/gui/models/nodeGraph/graphEdgeModel.cpp
```diff
@@ -0,0 +1,94 @@
...
+        case 0:
+            return QString(edge.source.c_str());
+        case 1:
+            return edge.sourceIndex;
+        case 2:
+            return QString(edge.destination.c_str());
+        case 3:
+            return edge.destinationIndex;
+        default:
+            return {};
```
trisyoungs, line 82: Definitely needs its own `enum`!

src/gui/models/nodeGraph/graphEdgeModel.h
```diff
@@ -0,0 +1,79 @@
...
+    std::string source;
+    // The index of the specific data *within* the source node
+    int sourceIndex;
+    // The node index of the destination node
+    std::string destination;
+    // The index of the specific data *within* the destination node
+    int destinationIndex;
+    // Equality comparison (we get these for free in C++20
+    bool operator==(const GraphRawEdge &other) const;
+};
```
trisyoungs, line 23: Were it not for the integer indices, this is our `EdgeDefinition` structure.

src/gui/models/nodeGraph/graphModel.cpp
```diff
@@ -0,0 +1,93 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#include "graphModel.h"
+#include "graphEdgeModel.h"
+#include <iostream>
+#include <qvariant.h>
```
trisyoungs:
```suggestion
#include <QVariant>
```

trisyoungs (approved):
Only had a brief look through, but this all looks good.  I'll get into the machinations of the QML in my own time because I want to understand better how it all works, but this can definitely go in when ready.

src/gui/models/nodeGraph/graphEdgeModel.cpp
rprospero: Fully agreed!  However, getting the edges to display is resulting in a full rewrite of `GraphEdgeModel`, so I'll leave that for the Edge PR.

src/gui/models/nodeGraph/graphEdgeModel.h
rprospero: I had the exact same realisation once I started on my Edge PR.

CMakeLists.txt
rprospero: Thank you for the suggestion on that.  I knew that we had the icons module, but I hadn't been quite sure how to put it in before.  Using that makes the configuration much cleaner.

PR #2159, opened 2025-05-14
feat: make the epsrManager count concencentration products factor instead

Danbr4d:
Need to make the count that we divide by the concentration product of the atom types in the configuration to see if it makes a difference for one of the Moonshine configurations.

### DIFF
--- a/src/modules/epsr/epsr.h
+++ b/src/modules/epsr/epsr.h
@@ -109,7 +109,7 @@ class EPSRModule : public Module
      * Functions
      */
     private:
-    // Target Configuration (determined from target modules)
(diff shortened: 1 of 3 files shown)


src/modules/epsr/epsr.h
```diff
@@ -109,7 +109,7 @@ class EPSRModule : public Module
      * Functions
      */
     private:
-    // Target Configuration (determined from target modules)
+    // Target Configuration (determined from target modules) and a function to return this for EPSRManager
```
trisyoungs, line 112: Don't really need this additional comment:
```suggestion
    // Target Configuration (determined from target modules)
```

trisyoungs (commented):
I think this all looks OK. Let's see if it works!

PR #2160, opened 2025-05-14
feat: AtomicMC node

RobBuchananCompPhys:


src/nodes/atomicMC/process.cpp
```diff
@@ -23,10 +23,10 @@ NodeConstants::ProcessResult AtomShakeNode::process()
     const auto rRT = 1.0 / (.008314472 * targetConfiguration_->temperature());
 
     // Print argument/parameter summary
-    Messenger::print("AtomShake: Performing {} shake(s) per Atom\n", nShakesPerAtom);
-    Messenger::print("AtomShake: Step size for adjustments is {:.5f} Angstroms (allowed range is {} <= delta <= {}).\n",
+    Messenger::print("AtomicMC: Performing {} shake(s) per Atom\n", nShakesPerAtom);
```
trisyoungs: Actually, here is another new rule for all of us moving forward.  The module type prefixes in all output messages will become redundant because we now store output on the `Node` itself.  By extension, we need to get into the habit of calling the local node messaging functions and not using `Messenger` anymore.

trisyoungs (approved):
All good, but as my comment suggests we should all start using the `Node`-centric messaging calls from this point forward, otherwise there will be a big (and tedious) refactoring headache later down the line!

PR #2161, opened 2025-05-15
feat: species unit cells for xyx import and bond tolerance adjustment

### LINKED ISSUE #1798: SpeciesEditor misses O-H bonds when importing
**Describe the bug**
When importing an xyz file created from external software (e.g. Aten) into Dissolve, occasionally certain bonds are not detected correctly as they are slightly too far apart. We really need to add an option to recalculate bonding with an adjusted tolerance.

aboswel:
Add and remove unit cell button, bond tolerance spin box.


src/gui/speciesCellDialog.cpp
```diff
@@ -0,0 +1,23 @@
...
+    ui_.setupUi(this);
+    sp_ = sp;
+}
+
+void SpeciesCellDialog::on_buttonBox_accepted()
+{
+    // Generate a box for the species
+    Vec3<double> axisLengths{ui_.BoxASpin->value(), ui_.BoxBSpin->value(), ui_.BoxCSpin->value()};
+    Vec3<double> axisAngles{ui_.BoxAlphaSpin->value(), ui_.BoxBetaSpin->value(), ui_.BoxGammaSpin->value()};
+    sp_->createBox(axisLengths, axisAngles);
```
trisyoungs: You can avoid the two temporary variables here and do this (check the formatting afterwards!):
```suggestion
    sp_->createBox({ui_.BoxASpin->value(), ui_.BoxBSpin->value(), ui_.BoxCSpin->value()}, {ui_.BoxAlphaSpin->value(), ui_.BoxBetaSpin->value(), ui_.BoxGammaSpin->value()});
```

src/gui/speciesCellDialog.h
```diff
@@ -0,0 +1,31 @@
...
+    Q_OBJECT
+
+    public:
+    SpeciesCellDialog(QWidget *parent, Species *sp);
+    ~SpeciesCellDialog() = default;
+
+    private:
+    // Main Dissolve pointer
+    Ui::SpeciesCellDialog ui_;
+    QDialog *view_{nullptr};
```
trisyoungs: Not needed
```suggestion
```

src/gui/speciesCellDialog.ui
trisyoungs, line 1: Should have some "sensible" defaults for the cell lengths - 20.0 as opposed to 1.0.

src/gui/speciesEditor.cpp
```diff
@@ -187,8 +189,17 @@ void SpeciesEditor::on_ToolsCalculateBondingButton_clicked(bool checked)
...
-    // Calculate missing bonds
-    sp->addMissingBonds();
+    // Now tolerance can be changed, clear bonds before updating (can't iterate directly over bonds...)
+    auto &bonds = sp->bonds();
+    std::vector<std::pair<int, int>> idxs;
+    if (bonds.size())
+        for (const auto &bond : bonds)
+            idxs.push_back({bond.indexI(), bond.indexJ()});
+    for (const auto &[i, j] : idxs)
+        sp->removeBond(i, j);
```
trisyoungs: Ah, so I totally understand why you had to do this, but we should definitely make a function in the `Species` class itself to handle this kind of thing.  That function can look a little like this:
```
void Species::recalculateIntramolecularTerms(double tolerance)
{
    bonds_.clear();
    removeHigherOrderIntramolecularTerms();

    addMissingBonds(tolerance);
    updateIntramolecularTerms();
}
```

src/gui/speciesEditor.cpp
```diff
@@ -259,3 +270,22 @@ void SpeciesEditor::on_ToolsMinimiseButton_clicked(bool checked)
     speciesViewer()->postRedisplay();
     speciesViewer()->notifyDataModified();
 }
+
+void SpeciesEditor::on_ToolsAddCellButton_clicked(bool checked)
+{
+    SpeciesCellDialog dialog(this, speciesViewer()->species());
+    dialog.exec();
```
trisyoungs: Once I've opened the add cell dialog I'm basically committed to adding a cell - we should have a Cancel button on the dialog so the user can bail out of the choice, and check for whether the dialog was accepted here.

trisyoungs: After the cell has been added we would also need to recalculate bonding, so a call to our new function in the `Species` class.

src/gui/speciesEditor.cpp
```diff
@@ -259,3 +270,22 @@ void SpeciesEditor::on_ToolsMinimiseButton_clicked(bool checked)
...
+    speciesViewer()->postRedisplay();
+    speciesViewer()->notifyDataModified();
+}
+
+void SpeciesEditor::on_ToolsRemoveCellButton_clicked(bool checked)
+{
+    speciesViewer()->species()->removeBox();
+    speciesViewer()->view().showAllData();
+    speciesViewer()->postRedisplay();
+    speciesViewer()->notifyDataModified();
```
trisyoungs, line 282: As with adding a cell, once you remove the cell the bonding also needs to be recalculated.

src/gui/speciesEditor.ui
trisyoungs, line 1: Few things to tweak in the UI:
1) Move the BondTolerance label and spinbox to immediately after the Rebond button
2) I fell into the trap of typing in a number into the bond tolerance spin and pressing enter, the result of which was the dialog was accepted and closed! There is a "Default Button" or similar defined somewhere which controls this behaviour, but for sure we don't want the dialog to close in this situation.
3) Change the step size on the bond tolerance spin to 0.01.

trisyoungs (changes requested):
Excellent stuff. Some things to modify, but basically there.

src/gui/icons/remove_box.svg
aboswel, line 1: Didn't realise there was already an axis button!

src/classes/species_intra.cpp
```diff
@@ -154,6 +154,17 @@ void Species::removeHigherOrderIntramolecularTerms()
     ++version_;
 }
 
+// Clear and recalculate all bonds on the species
+void Species::recalculateIntermolecularTerms(double tolerance)
+{
+    // Need to detach().
+    while (bonds_.size())
+        removeBond(bonds_[0].i(), bonds_[0].j());
```
trisyoungs, line 162: :+1:

src/gui/speciesEditor.cpp
```diff
@@ -259,3 +260,34 @@ void SpeciesEditor::on_ToolsMinimiseButton_clicked(bool checked)
...
+}
+
+void SpeciesEditor::on_ToolsBondToleranceSpin_valueChanged(double value)
+{
+    if (tolerance_ != value) // Segfault without
+    {
+        tolerance_ = value;
+        speciesViewer()->species()->recalculateIntermolecularTerms(tolerance_);
+    }
```
trisyoungs: Just need to protect against the widget value being set when there is no current species, as other functions do.  You can also do away with the `tolerance_` local and just have the default set to 1.1 in the UI file.  Have also added a call to `postRedisplay()` so the view updates when you change the bond tolerance.
```suggestion
    // Get displayed Species
    const auto *sp = speciesViewer()->species();
    if (!sp)
        return;

    speciesViewer()->species()->recalculateIntermolecularTerms(value);
    speciesViewer()->postRedisplay();
```

trisyoungs (approved):
One final suggestion, then good to go.

PR #2163, opened 2025-05-15
feat: GR node

RobBuchananCompPhys:


src/nodes/gr/gr.cpp
```diff
@@ -0,0 +1,41 @@
...
+                                           targetConfigurations_)
+        ->setFlags({ParameterBase::Required, ParameterBase::ClearData});
+    addOption<Number>("BinWidth", "Bin width (spacing in r) to use", binWidth_);
+    addOption<std::optional<Number>>("Range", "Maximum r to calculate g(r) out to", requestedRange_);
+    addOption<std::optional<Number>>("Averaging", "Number of historical partial sets to combine into final partials",
+                                     averagingLength_);
+    addOption<Averaging::AveragingScheme>("AveragingScheme", "Weighting scheme to use when averaging partials",
+                                          averagingScheme_);
+    addOption<Function1DWrapper>("IntraBroadening", "Type of broadening to apply to intramolecular g(r)", intraBroadening_);
+    addOption<std::optional<Number>>("Smoothing", "Specifies the degree of smoothing to apply to calculated g(r)", nSmooths_);
```
trisyoungs, line 18: This is the prime use case of the bounded / optional, integer `Number`.  Feels like this should be one of our next bits of work. There is also the `Averaging` option above.

src/nodes/gr/gr.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "nodes/gr/gr.h"
+
+GRNode::GRNode(Graph *parentGraph) : Node(parentGraph)
+{
+    addInput<std::vector<Configuration *>>("Configurations", "Set target configuration(s) for the module",
+                                           targetConfigurations_)
+        ->setFlags({ParameterBase::Required, ParameterBase::ClearData});
+    addOption<Number>("BinWidth", "Bin width (spacing in r) to use", binWidth_);
+    addOption<std::optional<Number>>("Range", "Maximum r to calculate g(r) out to", requestedRange_);
```
trisyoungs, line 12: An example of why we added `addBoundedOptionalInput()` to `Node()`, and needs reimplementation in our `Number` class.

src/nodes/gr/gr.h
```diff
@@ -0,0 +1,109 @@
...
+
+    private:
+    // Target configurations
+    std::vector<Configuration *> targetConfigurations_;
+    // Number of historical partial sets to combine into final partials
+    std::optional<Number> averagingLength_{5};
+    // Weighting scheme to use when averaging partials
+    Averaging::AveragingScheme averagingScheme_{Averaging::LinearAveraging};
+    // Bin width (spacing in r) to use
+    Number binWidth_{0.025};
```
trisyoungs: This could have a sensible lower limit applied - perhaps `0.001`.

src/nodes/gr/process.cpp
```diff
@@ -0,0 +1,129 @@
...
+        if (!requestedRange_)
+            message("Maximal cutoff used for Configuration '{}' ({} Angstroms).\n", cfg->niceName(), rdfRange);
+        else
+        {
+            if (requestedRange_.value_or(Number(0.0)) > rdfRange)
+            {
+                error("Specified RDF range of {} Angstroms is out of range for Configuration "
+                      "'{}' (max = {} Angstroms).\n",
+                      requestedRange_.value().asDouble(), cfg->niceName(), rdfRange);
+                return NodeConstants::ProcessResult::Failed;
```
trisyoungs, line 64: In the same way that `Messenger::error` returned `false` so we could do `return Messenger::error()` to bail out of a function and indicate failure, we could do the same with the local `message()` function, but return `NodeConstants::ProcessResult::Failed` instead.

trisyoungs (approved):
Looking good. There are some strong suggestions that we need to continue work on `Number` before we go much further with module -> node conversion - see the comments.

src/math/function1D.cpp
```diff
@@ -533,3 +547,34 @@ double Function1DWrapper::normalisation(double omega) const
...
+            if constexpr (HasEnumOptions<Functions1D::Form>)
+            {
+                setForm(arg);
+            }
+            else if constexpr (std::is_same_v<T, toml::array>)
+            {
+                setParameters(std::get<std::vector<double>>(arg));
+            }
+        },
+        node);
```
trisyoungs: Isn't the use of `toml::visit` only necessary when we don't explicitly know what data type is stored there (e.g. as for `Number`)?  Here we do - we have "form" and "parameters" so I think you can just directly access those entries, no?

src/math/function1D.h
```diff
@@ -155,4 +158,13 @@ class Function1DWrapper
     double yFT(double x, double omega = 0.0) const;
     // Return normalisation factor at specified omega
     double normalisation(double omega = 0.0) const;
+
+    /*
+     * Serialisable
```
trisyoungs, line 162: I have been adding headers like this called "I / O" rather than "serialisable", but I think we should go and convert all of them to "Serialisation" at some point in the future.

src/math/averaging.h
```diff
@@ -21,6 +21,8 @@ enum AveragingScheme
 // Return enum options for AveragingScheme
 EnumOptions<Averaging::AveragingScheme> averagingSchemes();
 
+EnumOptions<Averaging::AveragingScheme> getEnumOptions(Averaging::AveragingScheme);
+
```
trisyoungs, line 25: Don't think you need this declaration - just the definition in the `cpp` file.

src/math/function1D.h
```diff
@@ -93,6 +94,7 @@ class Functions1D
     };
     // Return enum options for form
     static EnumOptions<Form> forms();
+    static EnumOptions<Form> getEnumOptions(Form);
```
trisyoungs: Same here.

src/nodes/gr/process.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+// Run main processing
+NodeConstants::ProcessResult GRNode::process()
+{
+    /*
+     * Calculate standard partial g(r)
+     *
+     * This is a serial routine, with each process constructing its own copy of the data.
+     * Partial calculation routines called by this routine are parallel.
+     */
```
trisyoungs, line 20: I think this kind of comment can be pruned out as we go - this is a throwback to the MPI days!

src/nodes/node.h
```diff
@@ -81,12 +81,14 @@ class Node : public Serialisable<>
...
     // Node error format
-    template <typename... Args> void error(std::format_string<Args...> format, Args... args)
+    template <typename... Args> bool error(std::format_string<Args...> format, Args... args)
     {
         messages_.emplace_back(std::make_pair(MessageStatus::Error, std::format(format, std::forward<Args>(args)...)));
 
         if (echo_)
             echo();
+
+        return false;
```
trisyoungs: Not `ProcessResult` here instead of `bool`? It depends which way you want to play it: either you make it possible for the helper functions to return in a one-liner (as you have here) or you make it so that the main `process()` routine can error and return in one line (if you change the return type to `ProcessResult`.  My feeling is that there will be more nodes which don't use helper functions than those that do, so perhaps returning `ProcessResult` is the better option.

trisyoungs: Helper functions then must just use the `error()` function and `return false;`.

src/nodes/parameter.h
```diff
@@ -306,7 +307,33 @@ template <typename T> class PointerParameter : public Parameter<T>
 
     public:
     // Set the object
-    void set(const T &value) override{};
+    void set(const T &value) override {};
     // Assign the value of another parameter to this one.
     bool assign(ParameterBase *other) override { return false; }
 };
+
+// Template specialisation for non-dedfaulted type Function1DWrapper
```
```suggestion
// Template specialisation for non-defaulted type Function1DWrapper
```

src/nodes/node.h
RobBuchananCompPhys: Sounds like a way forward

trisyoungs (approved):

PR #2164, opened 2025-05-20
feat: SubGraphs

trisyoungs:
This PR intends to introduce the idea of "subgraphs" (graphs within graphs) and provide the necessary connective edges between the different "layers" of graph we now have. I have attempted to do this in such a way that the UI doesn't have to do any checking on what it is actually trying to display.

## TODO
- [x] Handle the creation of different types of inputs/outputs. At present only `Number` is supported, but some larger mechanism is required moving forward.
- [x] Fix serialisation!

Important changes:
- `InputNode` and `OutputNods` has been introduced to handle connections in to and out of `Graph`s. Their function is to provide the Node representation of inputs and outputs when looking "inside" the Graph. In order is to allow the creation of a typed `Parameter` input mapped to the same typed output, but this requires the use of a nasty factory-style construct since there is no way to go back from a `type_id` to a _bona fide_ class type. Why do we need that? to provide a suitable proxy object to hold the value of the incoming / outgoing parameter data. The alternative was to have `Edge`s connecting directly between different layers of `Graph`s, and this didn't seem very nice at all!
- `Graph`s self-create an `InputNode` and an `OutputNode` - these are owned locally by the `Graph` just like any other child node, and so are displayed in the UI / connected to in the standard way.
- Creating `Edge` connections in to or out of a `Graph` auto-creates an input or output (in the relevant `ParameterMappingNode` object), so serialising and deserialising `Edge`s handles neatly the creation of the custom inputs and outputs on `Graph` nodes.
- Save for sharing the same data object there is no actual link (`Edge` or otherwise) between the input parameters of a `Graph` and the outputs of the `InputNode` (nor the outputs of a `Graph` and the inputs of an `OutputNode`). The necessary update / pull logic is handled by `Graph::process()` for instance.
- Setting a parameter further upstream now correctly propagates the `upToDate_` flag to all dependent downstream nodes (prior to this PR only immediate neighbours were aware of changes).
- Created `Edge`s still retain a pointer to the source or target `Graph` rather than the Inputs or Outputs node to which it is really connected. This is necessary for both UI and serialisation to work properly, but creates a small problem in the general running in that edges trying to pull data will target the `Graph` rather than the Outputs node. To work around that, `Edge::pull()` checks what the `sourceNode_` actually is, and redirects its focus to the "Output" node if its a `Graph`.
- Another sprinkling of `dynamic_cast`s were necessary - apologies!
- Unit test added, of course.


src/nodes/graph.h
```diff
@@ -30,6 +44,33 @@ class Graph : public Node
...
+    // Mapped input and output nodes
+    InputsNode *mappedInputs_{nullptr};
+    OutputsNode *mappedOutputs_{nullptr};
+
+    public:
+    // Create and return mapped input
+    std::shared_ptr<ParameterBase> createMappedInput(std::string_view inputName, std::type_index typeIndex);
+    // Create mapped output, returning the relevant input (rather than the output)
+    std::shared_ptr<ParameterBase> createMappedOutput(std::string_view outputName, std::type_index typeIndex);
+
```
RobBuchananCompPhys, line 59: Interested in renaming these methods - how does the following idea sound `createMappedInput/Output` -> **`createProxyInput/Output`** ? I feel that the reference to mapping gives the impression that we are actually creating a `std::map` of proxy parameter inputs/outputs, when we are actually appending to a `std::vector`.

src/nodes/graph.h
```diff
@@ -13,6 +13,20 @@
...
 class Edge;
 class EdgeDefinition;
+class InputsNode;
+class OutputsNode;
+
+// Parameter Holder Base
+class ParameterHolderBase
+{
+};
+
```
RobBuchananCompPhys: Do we actually need a `ParmeterHolderBase` class, if it's empty?

src/nodes/graph.cpp
```diff
@@ -17,6 +23,123 @@ std::string_view Graph::type() const { return "Graph"; }
...
+    // TODO Convert to Factory
+    if (typeIndex == std::type_index(typeid(Number)))
+    {
+        // Create a parameter holder object with the correct type
+        auto proxy = std::make_shared<ParameterHolder<Number>>();
+        parameterHolders_.emplace_back(proxy);
+
+        // Create an input on ourself, linked to the proxy data
+        inputParameter = addInput(inputName, "", proxy->data);
+
```
RobBuchananCompPhys: Thought incoming - is there some way we could leverage `PointerParameter` and `addPointerInput` (instead of `addInput`) to get around the empty serialisation problem with `InputNode`? We know that pointer serialisation will work for parameters and we could dereference every data member coming out of the sub graph, converting them to mere `Parameter`s, added with `addOutput`. Of course, there's a question of how to get the interior sub-graph nodes the dereferenced data.

src/nodes/graph.h
trisyoungs: I think this is sensible.  Given the changes I've just made in the "no factory" follow-on, I'll make these modifications there instead.

src/nodes/graph.h
trisyoungs: We did in order to store the derived class pointers as a series of base class pointers in `Node`, but actually now this is not necessary at all in the "no factory" work.  I'll remove it there.

src/nodes/graph.cpp
trisyoungs: As discussed, probably yes, but worth understanding the limitations of returning `{}` before we try to work around it.

rprospero (approved):

rprospero:
@trisyoungs One question that does come up to me as I look through the factory PR is whether NodeParameterMap really needs `std::shared_ptr<ParameterBase>`.  Do we have a true ownership relationship here?  I thought that the Parameter was owned somewhere else and we could just use `ParameterBase*`.

trisyoungs:
> @trisyoungs One question that does come up to me as I look through the factory PR is whether NodeParameterMap really needs `std::shared_ptr<ParameterBase>`. Do we have a true ownership relationship here? I thought that the Parameter was owned somewhere else and we could just use `ParameterBase*`.

The `Node`s are the true owners of the `Parameter`s so should have the map of `std::shared_ptr`s - they are not stored anywhere else.  Perhaps my `ParameterLink` class in the "no factory" PR muddied the water a bit since it contains two shared pointers to the input and output sides.  However, this is just a container to allow the return of the necessary info and its subsequent storage elsewhere, and itself is not meant to be persistent. It contained more than just the two pointers in the past but this was refined down, so a `class` might now be overkill and we could revert to a simple `std::pair` at the cost of explicit data labelling.

PR #2167, opened 2025-05-23
fix: Overload edge add/remove for MSVC unit tests

RobBuchananCompPhys:


trisyoungs (commented):
Interesting!  This really shouldn't be necessary, so I'm keen to learn what MSVC's problem is with it.  Might be that we need to have two constructors on `EdgeDefinition` - one that takes no parameters, and one that takes explicitly the four string arguments...

PR #2169, opened 2025-05-23
feat: Added forcefield and Function1D for MCM-41 bulk

aboswel:
Implemented a modified buckingham potential Function1D form for pair potential overrides, skeletal forcefield with just charges for Si, O.


src/math/function1D.cpp
```diff
@@ -286,6 +286,44 @@ const std::map<Functions1D::Form, Function1DDefinition> &functions1D()
...
+         */
+        functions[Functions1D::Form::CamelloneBKS] =
+            Function1DDefinition({"A", "B", "C", "D", "E"},
+                                 [](double x, double omega, const std::vector<double> &params)
+                                 {
+                                     auto B = exp(-params[1] * x);
+                                     auto C = params[2] / pow(x, 6.0);
+                                     auto D = params[3] / pow(x, 12.0);
+                                     auto E = params[4] / pow(x, 8.0);
+                                     return (params[0] * B - C + D - E) * 96.48533; // Paper gives in eV
```
trisyoungs: In the interests of "Dissolve Consistency" and not mix units in our input files or forcefields we should take the energetic parameters in kJ/mol as standard and not convert them here.  It's up to the user to make those calculations.

src/math/function1D.cpp
```diff
@@ -397,6 +435,7 @@ EnumOptions<Functions1D::Form> Functions1D::forms()
                                            {Functions1D::Form::GaussianC2, "GaussianC2", 2},
                                            {Functions1D::Form::LennardJones126, "LennardJones126", 2},
                                            {Functions1D::Form::Buckingham, "Buckingham", 3},
+                                           {Functions1D::Form::CamelloneBKS, "CamelloneBKS", 5},
```
trisyoungs: Perhaps a more general name and `enum` - "Buckingham12-8" to reflect the powers of the additional terms?

src/data/ff/camelloneBKS/camelloneBKS.h
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#pragma once
+
+#include "data/ff/ff.h"
+
+// SPC/Fw Forcefield
```
trisyoungs:
```suggestion
// Camellone BKS Forcefield
```

src/data/ff/camelloneBKS/camelloneBKS.cpp
```diff
@@ -0,0 +1,55 @@
...
+/*
+ * Set Up
+ */
+
+// Set up / create all forcefield terms
+bool Forcefield_CamelloneBKS::setUp()
+{
+    // Atom types
+    addAtomType(Elements::O, 1, "O", "", "Silica oxygen", -1.2, "");
+    addAtomType(Elements::Si, 2, "Si", "", "Silica silicon", 2.4, "");
```
trisyoungs, line 33: Now, I wonder how we should handle the definition of the short-range Buckingham parts. They will be presumably defined per atom- pair in the paper, so we may need to extend the `Forcefield` class a bit to include the ability to add `PairPotentialOverride`s...

trisyoungs (changes requested):
Excellent start.  Need to add the new Buckingham style potential to the unit tests in "tests/pairPotentials/analytic.cpp" so we can ensure that the implemented form is correct / consistent between energy and force.

aboswel:
Sorry this has taken a while - was more fiddley than I expected...

src/data/ff/ff.h
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "classes/atomType.h"
+#include "classes/pairPotentialOverride.h"
```
trisyoungs, line 7: You don't need this include as your forward-declaring the `PairPotentialOverride` class.
```suggestion
```

src/main/pairPotentials.cpp
```diff
@@ -161,6 +161,12 @@ bool Dissolve::updatePairPotentials(std::optional<bool> useCombinationRulesHint)
 
     // Third step - apply any overrides
     Messenger::print("Applying pair potential overrides...\n");
+
+    // Check for forcefield specified overrides
+    for (const auto &sp : coreData_.species())
+        if (sp->forcefield())
+            sp->forcefield()->applyPairPotentialOverrides(coreData_);
+
```
trisyoungs: Shouldn't do this here - overrides will either exist or not, and this function is not responsible for adding / applying more.

src/gui/models/addForcefieldDialogModel.cpp
```diff
@@ -304,6 +304,11 @@ void AddForcefieldDialogModel::finalise()
             }
         }
     }
+
+    // Associate the forcefield to the species
+    if (ff_)
+        species_->setForcefield(ForcefieldLibrary::forcefield(ff_->name()));
+
```
trisyoungs: This is where the overrides from the forcefield can be retrieved and owned.

src/data/ff/ff.h
```diff
@@ -191,6 +193,23 @@ class Forcefield
...
+     */
+    public:
+    // Associate an override with a forcefield
+    void addPairPotentialOverride(std::string_view matchI, std::string_view matchJ,
+                                  PairPotentialOverride::PairPotentialOverrideType overrideType,
+                                  const InteractionPotential<Functions1D> potential);
+    // Return overrides for current forcefield
+    const std::vector<std::unique_ptr<PairPotentialOverride>> &pairPotentialOverrides() const;
+    // Apply overrides to coreData
+    bool applyPairPotentialOverrides(CoreData &coreData) const;
```
trisyoungs: I wonder if we can simplify things a bit here.  Rather than create the local overrides in setUp() and store them in the forcefield object, why don't we have a function which simply creates and returns the overrides all in one go. This removes `addPairPotentialOverride` and `applyPairPotentialOverrides()` and the functionality in the latter can be moved to `addForcefieldDialogModel::finalise()`.

src/data/ff/ff.cpp
```diff
@@ -539,3 +539,70 @@ bool Forcefield::assignIntramolecular(Species *sp, int flags) const
...
+
+            if (!exists)
+            {
+                coreData.addPairPotentialOverride(override->matchI(), override->matchJ(), override->type(),
+                                                  override->interactionPotential());
+                Messenger::print("Override applied");
+                newAdditions = true;
+            }
+        }
+    }
```
trisyoungs: This is a very correct way of doing things, so well done for that, but I think in the grand scheme of things we should just copy over any defined overrides to `CoreData`. See later comments.

trisyoungs (changes requested):
I'm suggesting a small refactor to make things a bit cleaner and more in keeping with how forcefields are applied in other respects.  Your reading of the code is correct (esp. the `Species` own `forcefield_` variable) but we currently don't use it as you might think! Another thing to clean up in Dissolve2...

src/data/ff/camelloneBKS/camelloneBKS.h
```diff
@@ -0,0 +1,47 @@
...
+                                                      {133996.2240, 2.76, 16884.9331, 17367.3598, 2315.6480});
+        InteractionPotential<Functions1D> siOPotential(Functions1D::Form::Buckingham128,
+                                                       {1737098.491, 4.8732, 12884.4679, 1929.7066, 578.912});
+
+        std::vector<PairPotentialOverride> overrides{
+            {"O", "O", PairPotentialOverride::PairPotentialOverrideType::Add, oOPotential},
+            {"Si", "O", PairPotentialOverride::PairPotentialOverrideType::Add, siOPotential}};
+
+        return overrides;
+    }
```
trisyoungs: Yes, basically this, but don't implement this in the header! Also I don't see why you should't be able to use brace-initialisation for the `InteractionPotential` part itself (this will need formatting properly):
```suggestion
    // Return available pair potential overrides
    std::vector<PairPotentialOverride> pairPotentialOverrides() override
    {
       return {
            {"O", "O", PairPotentialOverride::PairPotentialOverrideType::Add, {Functions1D::Form::Buckingham128,
                                                      {133996.2240, 2.76, 16884.9331, 17367.3598, 2315.6480}}},
            {"Si", "O", PairPotentialOverride::PairPotentialOverrideType::Add, {Functions1D::Form::Buckingham128,
                                                       {1737098.491, 4.8732, 12884.4679, 1929.7066, 578.912}}}
                   };
    }
```

src/data/ff/ff.h
```diff
@@ -191,6 +192,13 @@ class Forcefield
     void assignAtomType(const ForcefieldAtomType &ffa, SpeciesAtom &i, CoreData &coreData, bool setSpeciesAtomCharges) const;
     // Assign intramolecular parameters to the supplied Species
     bool assignIntramolecular(Species *sp, int flags = Forcefield::GenerateImpropersFlag) const;
+
+    /*
+     * Pair Potential Overrides
+     */
+    public:
+    // Add or Return overrides for current forcefield
+    virtual std::vector<PairPotentialOverride> pairPotentialOverrides() { return {}; };
```
trisyoungs: Tweaking the comment a bit, and I'd rather see the function definition itself in the cpp file, even though it is a trivial one.
```suggestion
    // Return available pair potential overrides
    virtual std::vector<PairPotentialOverride> pairPotentialOverrides();
```

src/gui/models/addForcefieldDialogModel.cpp
```diff
@@ -304,6 +304,24 @@ void AddForcefieldDialogModel::finalise()
...
+        {
+            auto present = false;
+            for (const auto &orr : dissolve_->coreData().pairPotentialOverrides())
+                if (orr.get()->matchI() == override.matchI() && orr.get()->matchJ() == override.matchJ())
+                    present = true;
+
+            if (!present)
+                dissolve_->coreData().addPairPotentialOverride(override.matchI(), override.matchJ(), override.type(),
+                                                               override.interactionPotential());
+        }
```
trisyoungs: There's a case for a std algorithm here (this will definitely need reformatting!). Also need to check the reverse case for name matching (matchI == matchJ and matchJ == matchI). Third thing (!) I had envisaged that the unique pointers returned by the `Forcefield` would just become owned by the CoreData, but this might just be a unnecessarily messy. In that case, there's no point returning unique pointers from the `Forcefield` itself and it can just be a straight `std::vector<PairPotentialOverride>` instead.
```suggestion
    auto &currentOverrides = dissolve_->coreData().pairPotentialOverrides();
    for (const auto &ffOverride : ff_->pairPotentialOverrides())
    {
        auto it = std::find_if(currentOverrides.begin(), currentOverrides.end(), [&](const auto &ppOverride) {
            return (ppOverride->matchI() == ffOverride.matchI() && ppOverride.get()->matchJ() == ffOverride.matchJ()) || (ppOverride->matchI() == ffOverride.matchJ() && ppOverride.get()->matchJ() == ffOverride.matchI());
        });
        if (it == currentOverrides.end())
        {
            dissolve_->coreData().addPairPotentialOverride(override.matchI(), override.matchJ(), override.type(),
                                                               override.interactionPotential());
        }
    }
```

src/main/pairPotentials.cpp
```diff
@@ -161,6 +161,7 @@ bool Dissolve::updatePairPotentials(std::optional<bool> useCombinationRulesHint)
 
     // Third step - apply any overrides
     Messenger::print("Applying pair potential overrides...\n");
+
```
trisyoungs:
```suggestion
```

trisyoungs (changes requested):
Yes, this is what I had in mind.  But of tidying up and we should be good to go.

src/data/ff/ff.cpp
```diff
@@ -539,3 +539,6 @@ bool Forcefield::assignIntramolecular(Species *sp, int flags) const
 
     return true;
 }
+
+// Return avaliable pair potential overrides
```
trisyoungs:
```suggestion
// Return available pair potential overrides
```

src/data/ff/ff.h
```diff
@@ -191,6 +192,13 @@ class Forcefield
     void assignAtomType(const ForcefieldAtomType &ffa, SpeciesAtom &i, CoreData &coreData, bool setSpeciesAtomCharges) const;
     // Assign intramolecular parameters to the supplied Species
     bool assignIntramolecular(Species *sp, int flags = Forcefield::GenerateImpropersFlag) const;
+
+    /*
+     * Pair Potential Overrides
+     */
+    public:
+    // Return avaliable pair potential overrides
```
trisyoungs:
```suggestion
    // Return available pair potential overrides
```

trisyoungs (approved):
Two typos to fix, then looks good to merge in.

src/data/ff/ff.cpp
aboswel: It's one of those words...

Issue #2152, opened 2025-05-07
[Epic] Dissolve2Do


trisyoungs:
### The Dissolve2 ToDo!

A dynamic list of small jobs that will need doing in Dissolve2 towards the end of the transition to `Node`s, but that we don't want to (or can't) do while the old system is still present.

- [ ] #2352
- [ ] #2353
- [x] #2354
- [ ] #2355
- [x] `PartialSet` averaging is performed (i.e. `GRNode::process`) by obtaining processing module data from a `GenericList` item. Instead do this work on an owned `PartialSet` object.
- [ ] #2356
            GenericItem::ProtectedFlag) = true;``` ) - how do we rework this in the graph/node framework? ([see](https://github.com/disorderedmaterials/dissolve/pull/2182#discussion_r2142024382))
- [ ] #2357
- [x] Nodes should pass on their configuration after processing for use by subsequent nodes, therefore the addition of a `addPointerOutput<configuration *>` is required for relevant nodes, where they don't already do this
- [ ] #2358
- [ ] #2359
- [ ] #2360
- [ ] #2361
- [x] #2203 
- [x] #2242
- [ ] #2362
- [ ] #2363
- [ ] [When and where to generate cells as required](https://github.com/disorderedmaterials/dissolve/pull/2215#discussion_r2244817006l)
- [ ] #2364
- [x] Call `CellArray::generate()` from node process only when the box has fundamentally changed (excluding scalings) - [see](https://github.com/disorderedmaterials/dissolve/pull/2215#discussion_r2306652351)
- [x] Refactoring the `ScatteringMatrix` disabled saving of estimated S(Q) and g(r) to the restart file since their storage moved to `DoubleKeyedMap`. This needs to be reinstated.
- [ ] #2365
- [ ] #2366
- [ ] #2367
- [ ] #2368
- [ ] #2370
- [ ] #2371
- [ ] #2372
- [ ] #2373
- [ ] #2374
- [ ] #2375
- [ ] #2376
- [ ] #2377
- [ ] #2378
- [ ] #2379
- [ ] #2380
- [ ] #2381
- [x] #2382
- [ ] #2383
- [ ] #2384
- [ ] #2385
- [ ] #2386
- [ ] #2387
- [ ] #2388
- [ ] #2389

PR #2170, opened 2025-05-29
feat: SQ Node

RobBuchananCompPhys:


src/nodes/gr/gr.cpp
```diff
@@ -25,6 +25,8 @@ GRNode::GRNode(Graph *parentGraph) : Node(parentGraph)
         internalTest_);
     addOption<GRNode::PartialsMethod>("Method", "Calculation method for partial radial distribution functions",
                                       partialsMethod_);
+    addPointerOutput<PartialSet *>("UnweightedGR", "Unweighted partials for target configuration", unweightedgr_);
+    addOutput<std::optional<Number>>("Rho", "Effective density", rho_);
```
trisyoungs: Strikes me that this is such a key bit of information required by the `PartialSet` and any subsequent use of it that it should be included within the `PartialSet` class itself. This will greatly simplify the links between GR, SQ, NeutronSQ etc.

src/nodes/gr/helpers.cpp
```diff
@@ -321,15 +321,19 @@ std::vector<std::pair<const Species *, double>> GRNode::speciesPopulations() con
...
     // Does a PartialSet already exist for this Configuration?
+    // UNUSED
     auto originalGRObject = processingData.realiseIf<PartialSet>(std::format("{}//OriginalGR", cfg->niceName()), name(),
                                                                  GenericItem::InRestartFileFlag);
     auto &originalgr = originalGRObject.first;
     if (originalGRObject.second == GenericItem::ItemStatus::Created)
         originalgr.setUp(cfg->atomTypePopulations(), rdfRange, rdfBinWidth);
+    // UNUSED
+
+    gr.setUp(cfg->atomTypePopulations(), rdfRange, rdfBinWidth);
```
trisyoungs: One benefit of the old way of storing data was that it was easy to see if the data needed creating, and if it did you could call, e.g., `setUp()` on the `PartialSet`. Here we don't have that info any more, and we don't want to be setting the partial set up every single time. I see two options moving forward: 1) We have / call functions in each class to determine if they have already been primed and are ready for use, or; 2) We store this kind of data in `std::optional` so we definitely know that it exists (and is implicitly valid) or not.  A drawback with 2) is that it may complicate our `Parameter` (-> `OptionalPointerParameter`) flow, but it is quite neat. A small bonus in doing it this way is that a node's `reset()` function can just nullify any `std::optional`s.

src/nodes/gr/helpers.cpp
```diff
@@ -321,15 +321,19 @@ std::vector<std::pair<const Species *, double>> GRNode::speciesPopulations() con
 }
 
 // Calculate unweighted partials for the specified Configuration
-bool GRNode::calculateGR(GenericList &processingData, const ProcessPool &procPool, Configuration *cfg,
+bool GRNode::calculateGR(GenericList &processingData, const ProcessPool &procPool, Configuration *cfg, PartialSet &gr,
```
trisyoungs: In the original function declaration the `PartialSet` argument was called "originalgr" but this has changed to "gr" here. We also have a local "originalgr" variable and a member "originalgr_" variable, so I think this is getting a little confusing!

src/nodes/sq/process.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    // Set fingerprint
+    unweightedsq_.setFingerprint(
+        std::format("{}/{}", dissolve().processingModuleData().version("UnweightedGR", "UNUSED"),
+                    nullptr ? dissolve().processingModuleData().version("Reflections", "UNUSED") : -1));
+
+    // Save data if requested
+    // UNUSED
+    if (save_ && !MPIRunMaster(moduleContext.processPool(), unweightedsq.save(name_, "UnweightedSQ", "sq", "Q, 1/Angstroms")))
+        return NodeConstants::ProcessResult::Failed;
```
trisyoungs: Ick.  Not unused, but just involving a check for MPI that we don't use any more.  We have always left the door open to bring it back, but increasingly I feel that we should remove this kind of check especially.  I would bet money that we don't consistently use `MPIRunMaster` across all the things we export!

src/nodes/sq/sq.h
```diff
@@ -0,0 +1,72 @@
...
+    std::string_view summary() const override;
+
+    /*
+     * Definition
+     */
+    private:
+    // Unweighted g(r)
+    PartialSet *unweightedgr_{nullptr};
+    // Unweighted S(Q)
+    PartialSet unweightedsq_;
```
trisyoungs: Naming suggestion, here and throughout:
```suggestion
    PartialSet *unweightedGR_{nullptr};
    // Unweighted S(Q)
    PartialSet unweightedSQ_;
```

trisyoungs (commented):
Looking good. Some comments on naming, data passing etc.

src/classes/partialSet.h
```diff
@@ -91,6 +93,9 @@ class PartialSet
     // Return total bound function
     Data1D &boundTotal();
     const Data1D &boundTotal() const;
+    // Effective density
+    void setEffectiveDensity(std::optional<double> rho);
+    std::optional<double> effectiveDensity() const;
```
trisyoungs: Looking at the code I think this can / should be just `double` rather than optional.  Reason being is that if there is no way to calculate and set the effective density, this implies there is no valid `Configuration` to work from, and so there is no valid `PartialSet`.

src/nodes/gr/gr.h
```diff
@@ -47,6 +47,12 @@ class GRNode : public Node
     private:
     // Target configurations
     std::vector<Configuration *> targetConfigurations_;
+    // Original g(r)
+    std::optional<PartialSet> originalgr_;
+    // Unweighted g(r)
+    std::optional<PartialSet> unweightedgr_;
+    // Summed unweighted g(r)
+    std::optional<PartialSet> summedunweightedgr_;
```
trisyoungs: Mr Picky Names here again:
```suggestion
    std::optional<PartialSet> originalGR_;
    // Unweighted g(r)
    std::optional<PartialSet> unweightedGR_;
    // Summed unweighted g(r)
    std::optional<PartialSet> summedUnweightedGR_;
```

src/nodes/gr/process.cpp
```diff
@@ -74,56 +75,47 @@ NodeConstants::ProcessResult GRNode::process()
...
         bool alreadyUpToDate;
-        calculateGR(dissolve().processingModuleData(), processPool(), cfg, partialsMethod_, rdfRange, binWidth_.asDouble(),
-                    alreadyUpToDate);
-        auto &originalgr =
-            dissolve().processingModuleData().retrieve<PartialSet>(std::format("{}//OriginalGR", cfg->niceName()), name());
+        calculateGR(processPool(), cfg, originalGR(cfg, rdfRange, binWidth_.asDouble()), partialsMethod_, rdfRange,
+                    binWidth_.asDouble(), alreadyUpToDate);
 
         // Perform averagingLength_ of unweighted partials if requested, and if we're not already up-to-date
+        /*
```
trisyoungs, line 82: Need to add the job of reworking `Averaging` into its new, non-`GenericList` form onto the Dissolve2TODO (#2152).

src/nodes/md/process.cpp
```diff
@@ -98,9 +100,8 @@ NodeConstants::ProcessResult MDNode::process()
     // Initialise the random number buffer for all processes
     RandomBuffer randomBuffer(processPool(), ProcessPool::PoolProcessesCommunicator);
 
-    // Read in or assign random velocities
-    auto [velocities, status] = dissolve().processingModuleData().realiseIf<std::vector<Vector3>>(
-        std::format("{}//Velocities", targetConfiguration_->niceName()), name(), GenericItem::InRestartFileFlag);
+    std::vector<Vector3> velocities;
```
trisyoungs: Here, `velocities` should become a member variable, and `std::optional`.

src/nodes/sq/sq.h
trisyoungs:
```suggestion
    PartialSet *unweightedGR_{nullptr};
    // Unweighted S(Q)
    PartialSet unweightedSQ_;
```

trisyoungs (commented):
Lots here!  Few comments sprinkled throughout, but looking very good.  Proof will be in the testing (which we should start thinking about quite soon).

PR #2171, opened 2025-05-30
fix: Windows build and test issues

trisyoungs:
This PR fixes issues on the Windows build covering `EdgeDefinition`, the `Parameter` template specialisation for `Function1DWrapper`, and fixes unit tests.


src/nodes/edge.h
```diff
@@ -12,9 +12,12 @@
...
 
 // Edge Definition
-struct EdgeDefinition : public Serialisable<>
+class EdgeDefinition : public Serialisable<>
 {
-    EdgeDefinition(std::string srcNode = "", std::string srcOutput = "", std::string tgtNode = "", std::string tgtInput = "");
+    public:
+    EdgeDefinition() = default;
+    EdgeDefinition(std::string srcNode, std::string srcOutput, std::string tgtNode, std::string tgtInput);
+    EdgeDefinition(const EdgeDefinition &other);
```
RobBuchananCompPhys, line 20: Wow 3 constructors were needed in the end!

src/nodes/parameter.h
```diff
@@ -381,4 +381,9 @@ class Parameter<Function1DWrapper> : public ParameterBase, public std::enable_sh
     bool assign(ParameterBase *other) override { return false; }
     // Return whether the contained data represents the default value
     bool isDefault() const override { return false; }
+    // Create a parameter link (input - data proxy - output) for the derived class type
+    ParameterLink createParameterLink(std::string_view newName, std::string_view newDescription = "") const override
+    {
+        throw(std::runtime_error("Can't create a ParameterLink for a Function1DWrapper.\n"));
+    }
```
RobBuchananCompPhys, line 388: Interesting - how is this supposed to be used? And how did it compile without returning a `ParameterLink`..

src/nodes/edge.h
trisyoungs: _Only_ 3! :)

src/nodes/parameter.h
trisyoungs: Short answer - it is not.  I didn't want to spend the time looking in to how a `ParameterLink` would work with the edge case that `Function1DWrapper` has become, so I reasoned that we'd just throw here for now.  I added a note to the Dissolve2 TODO to remove `Function1DWrapper` in favour of something else, as I'm not sure it's even necessary.

PR #2172, opened 2025-05-30
build: Fix MacOS installer for the nth time

trisyoungs:
A PR to address failing MacOS packages. The culprit was the Freetype library: the Homebrew-derived version was being included rather than the one we deliberately build ourselves.


Danbr4d (approved):
From what little I understand, I think this looks fine.

PR #2174, opened 2025-06-03
feat: Optional pointer parameter

RobBuchananCompPhys:


src/nodes/parameter.h
```diff
@@ -352,11 +352,47 @@ template <typename T> class PointerParameter : public Parameter<T>
...
+    void set(const T &value) override {};
     // Assign the value of another parameter to this one.
     bool assign(ParameterBase *other) override { return false; }
 };
 
+// OptionalPointerParameter, returning a pointer from a target object rather than the object itself
+template <typename T> class OptionalPointerParameter : public Parameter<T>
+{
+    public:
+    OptionalPointerParameter(Node *parent, std::string_view name, std::string_view description, std::optional<T *> &object)
```
trisyoungs: The `Node` where the object is will have a `std::optional<MyObject>` member rather than `std::optional<MyObject *>`.

src/nodes/parameter.h
```diff
@@ -352,11 +352,47 @@ template <typename T> class PointerParameter : public Parameter<T>
...
+        : Parameter<T>(parent, name, description, object_)
```
trisyoungs: The underlying `Parameter<T>` will still need to be given a local `pointer_` object (as with `PointerParameter<T>`) since that's the data type it's representing. This amounts to being basically a dummy variable for `OptionalPointerParameter` since `get()` is redefined, but we will need it to satisfy `Parameter<T>`.

src/nodes/parameter.h
```diff
@@ -352,11 +352,47 @@ template <typename T> class PointerParameter : public Parameter<T>
...
+        object_ = object;
+    }
+    virtual ~OptionalPointerParameter() = default;
+
+    /*
+     * Data
+     */
+    protected:
+    // Optional object
+    std::optional<T *> &object_;
```
```suggestion
    std::optional<T> &object_;
```

src/nodes/parameter.h
```diff
@@ -352,11 +352,47 @@ template <typename T> class PointerParameter : public Parameter<T>
...
+
+    public:
+    // Return the parameter value
+    T &get()
+    {
+        if (object_.has_value())
+            return object_.value();
+
+        return {};
```
trisyoungs: I think here we can do:
```suggestion
        return object_.value_or(nullptr);
```

src/nodes/parameter.h
```diff
@@ -352,11 +352,47 @@ template <typename T> class PointerParameter : public Parameter<T>
...
+    /*
+     * Data
+     */
+    protected:
+    // Optional object
+    std::optional<T *> &object_;
+
+    public:
+    // Return the parameter value
+    T &get()
```
```suggestion
    T* &get()
```

src/nodes/node.h
```diff
@@ -233,6 +233,21 @@ class Node : public Serialisable<>
         param->setFlags(ParameterBase::ParameterFlags::Output);
         return param;
     }
+    // Add optional pointer input parameter
+    template <typename T>
+    std::shared_ptr<ParameterBase> addOptionalPointerInput(std::string_view inputName, std::string_view description,
+                                                           std::optional<T *> &object)
+    {
```
trisyoungs: Output!

src/nodes/node.h
```diff
@@ -233,6 +233,21 @@ class Node : public Serialisable<>
         param->setFlags(ParameterBase::ParameterFlags::Output);
         return param;
     }
+    // Add optional pointer input parameter
```
trisyoungs:
```suggestion
    // Add optional pointer output parameter
```

src/nodes/node.h
```diff
@@ -233,6 +233,21 @@ class Node : public Serialisable<>
...
+    template <typename T>
+    std::shared_ptr<ParameterBase> addOptionalPointerOutput(std::string_view outputName, std::string_view description,
+                                                            std::optional<T *> &object)
```
trisyoungs: The `Node` will be storing, e.g., `std::optional<PartialSet>` so this just needs to be `std::optional<T> &object`.

trisyoungs (approved):

PR #2181, opened 2025-06-09
build: Fix bad syntax in release pipeline.

trisyoungs:
Our `detect_release.yml` currently does not run - potentially the culprit is a recent change to the `if:` test in the workflow.  Until it is merged and I try and push a new release it's hard to say for certain though!


Danbr4d (approved):
Sorry it's late been in meetings all day. Hopefully this fix works!

PR #2185, opened 2025-06-11
feat: Draw edges between nodes

rprospero:
Not perfect, but a first attempt at drawing the edges between nodes.

The PR currently has the following known issues:

- The outgoing edge occurs at the edge of the name and not the edge of the node
- A magic number is needed to adjust for the height of the header on the node box.  This should be calculated directly, but the QML keeps fighting me on it.
- The graph seems to contain an extra edge that I do not understand.

20250611_14h24m10s_grim


src/gui/models/nodeGraph/graphEdgeModel.h
```diff
@@ -74,6 +51,17 @@ class GraphEdgeModel : public QAbstractListModel
...
     private:
-    std::vector<GraphRawEdge> edgeCache_;
+    // The graph whose edges we model
+    Graph *&graph_;
+    // The owner of this edge model
+    GraphModel *parent_;
+
+    // The edges of the graph
+    Graph::Edges &edges_();
+    const Graph::Edges &edges_() const;
```
trisyoungs:
```suggestion
    Graph::Edges &edges();
    const Graph::Edges &edges() const;
```

src/gui/models/nodeGraph/graphModel.cpp
```diff
@@ -3,10 +3,15 @@
 
 #include "graphModel.h"
 #include "graphEdgeModel.h"
+#include "graphNodeModel.h"
 #include <QVariant>
 #include <iostream>
+#include <qabstractitemmodel.h>
```
trisyoungs:
```suggestion
#include <QAbstractItemModel>
```

trisyoungs (commented):
This is a lot cleaner in places for sure.  Does the big picture plan allow all of the known issues to be properly resolved? Or will we be forever bending QML too far to get what we want?

PR #2187, opened 2025-06-23
feat: Better fuzzy search

### LINKED ISSUE #2157: Better fuzzy matching in Dissolve 2.0
Currently, we have the option to allow fuzzy searching to find the exact type while calling `createNode`.  Hopwever, this limits our fuzzy searching to cases where there is no ambiguity (e.g. case sensitivity).  It also adds extra logic to `createNode`.

A better setup would be for `createNode` to only accept exact values.  Instead, the registry should provide a fuzzy matching function that returns a *list* of names, sorted by relevance.  Then, the user can select the *exact* value from the list and pass it on to `createNode`

rprospero:
This PR improves the handling of fuzzy search for node names.  First, the `createNode` function now requires an exact value and has no logic to perform fuzzy searching.  Instead, the `getNodeTypeFuzzy` method has been upgraded to return a list of *all* of the matching node names (to eventually allow the user to select a name from the list).  It also does a slightly more advanced search allowing for substrings and case insensitivity.

I used the new C++20 ranges to handle collecting the list of names.  If this is too confusing, I can switch it over to a `for` loop.

Closes #2157

### DIFF
--- a/src/nodes/registry.cpp
+++ b/src/nodes/registry.cpp
@@ -19,6 +19,8 @@
 #include "nodes/vec3Decomposition.h"
 #include <memory>
 
+using namespace std::string_literals;
+
 // Static Singletons
 std::map<std::string_view, NodeProducer> NodeRegistry::producers_;
 
@@ -66,7 +68,7 @@ std::vector<std::string_view> NodeRegistry::getNodeTypeFuzzy(std::string_view we
 
     std::vector<std::string_view> result;
 
-    auto fuzzy = std::string("*") + std::string(weakNodeType) + std::string("*");
+    auto fuzzy = "*"s + std::string(weakNodeType) + "*"s;
 
     // Case insensitive search for now - fuzzy search to be implemented at a later date
     for (auto &&[nodeType, _] : producers_)
--- a/src/nodes/registry.cpp
+++ b/src/nodes/registry.cpp
@@ -66,10 +66,12 @@ std::vector<std::string_view> NodeRegistry::getNodeTypeFuzzy(std::string_view we
 
     std::vector<std::string_view> result;
 
+    auto fuzzy = std::string("*") + std::string(weakNodeType) + std::string("*");
+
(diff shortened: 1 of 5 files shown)


src/nodes/registry.h
```diff
@@ -23,7 +24,7 @@ class NodeRegistry
     // Check whether the supplied node type is known
     static bool hasNodeType(std::string_view nodeType);
     // Search for the supplied node type, returning strict node type if found
-    static std::string_view getNodeTypeFuzzy(std::string_view weakNodeType);
+    static std::vector<std::string_view> getNodeTypeFuzzy(std::string_view weakNodeType);
```
trisyoungs: Naming suggestion. Also the comment on the line above (which I can't attach a comment on) should be rewritten.
```suggestion
    static std::vector<std::string_view> getNodeTypesFuzzy(std::string_view weakNodeType);
```

src/nodes/registry.cpp
```diff
@@ -60,18 +61,22 @@ bool NodeRegistry::hasNodeType(std::string_view nodeType)
...
+    using namespace std::string_literals;
 
-    return {};
+    auto predicate = [weakNodeType](const auto nodeType)
+    { return DissolveSys::sameWildString("*"s + std::string(weakNodeType) + "*"s, nodeType); };
+
+    // Iterate over the keys of the map and include only the names
+    // which match our fuzzy match
+    using namespace std::views;
+    auto range = producers_ | keys | filter(predicate);
```
trisyoungs: I like this, but part of me wants to remove the `using namespace` and write it explicitly on the relevant objects...

trisyoungs (approved):
:+1: Couple of small comments only.

src/nodes/registry.cpp
rprospero: I fully agree!  I'd pulled in the namespace because it was the first `views` pr and I wanted to make it as clear as possible, but I do think that keeping the explicit names is better here.

PR #2188, opened 2025-06-23
feat: Nanotube species creator

trisyoungs:
This PR adds the ability to create H-terminated or periodic nanotubes directly within Dissolve.
image

### DIFF
--- a/src/gui/createNanotubeSpeciesDialog.cpp
+++ b/src/gui/createNanotubeSpeciesDialog.cpp
@@ -28,15 +28,21 @@ CreateNanotubeSpeciesDialog::CreateNanotubeSpeciesDialog(QWidget *parent, Dissol
  */
 
 // Plot an AB atom layer
-void CreateNanotubeSpeciesDialog::plotLayer(double z, double tubeRadius, double radialStep, double radialOffset)
+void CreateNanotubeSpeciesDialog::plotLayer(double z, double tubeRadius, double radialStep, double radialOffset,
+                                            Elements::Element zA, Elements::Element zB)
 {
     for (auto radial = 0; radial < ui_.RadialRingSizeSpin->value(); ++radial)
     {
         auto angle = radial * radialStep + radialOffset;
-        species_.addAtom(zA_, {tubeRadius * cos(angle), tubeRadius * sin(angle), z});
+
+        if (zA != Elements::Unknown)
+            species_.addAtom(zA, {tubeRadius * cos(angle), tubeRadius * sin(angle), z});
+
         angle += radialStep * 0.5;
-        species_.addAtom(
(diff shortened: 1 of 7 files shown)


src/gui/createNanotubeSpeciesDialog.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+        if (zA != Elements::Unknown)
+            species_.addAtom(zA, {tubeRadius * cos(angle), tubeRadius * sin(angle), z});
+
+        angle += radialStep * 0.5;
+
+        if (zB != Elements::Unknown)
+            species_.addAtom(
+                zB, {tubeRadius * cos(angle), tubeRadius * sin(angle), z + ui_.BondLengthSpin->value() * cos(60.0 / DEGRAD)});
+    }
```
RobBuchananCompPhys, line 46:
```suggestion
    auto angles = std::views::iota(0, ui_.RadialRingSizeSpin->value()) |
                  std::views::transform([=](const auto i) { return i * radialStep + radialOffset; });
    if (zA != Elements::Unknown)
        std::for_each(angles.cbegin(), angles.cend(),
                      [&](auto angle) { species_.addAtom(zA, {tubeRadius * cos(angle), tubeRadius * sin(angle), z}); });

    if (zB != Elements::Unknown)
        std::for_each(angles.cbegin(), angles.cend(),
                      [&](auto angle)
                      {
                          species_.addAtom(zB, {tubeRadius * cos(angle + radialStep * 0.5),
                                                tubeRadius * sin(angle + radialStep * 0.5),
                                                z + ui_.BondLengthSpin->value() * cos(60.0 / DEGRAD)});
                      });
```

Just a suggestion (and I actually think this looks *worse*), but I noticed some new features in C++ `ranges` such as pipe operators could help generate the angles.

src/gui/createNanotubeSpeciesDialog.cpp
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// Regenerate species
+void CreateNanotubeSpeciesDialog::regenerate()
+{
+    species_.clear();
+
+    const auto r = ui_.BondLengthSpin->value();
+    const auto axialN = ui_.AxialRingLengthSpin->value();
```
RobBuchananCompPhys: What is `axialN`? Could this variable be better named?

RobBuchananCompPhys (approved):
Approved 👍  (with a couple of comments)

rprospero (approved):

src/gui/createNanotubeSpeciesDialog.cpp
trisyoungs: Agreed. Renamed to `nAxialRings`.

src/gui/createNanotubeSpeciesDialog.cpp
trisyoungs: Nice use of `iota` and `transform` but I would agree that this rather complicates the working of the function so will stick with the original dumb loop.

PR #2189, opened 2025-06-27
feat: Multiple inputs to vector-type parameters

trisyoungs:
This PR implements handling of `std::vector`s in `Parameter`s, allowing them to be set via either another vector or to be constructed piecewise from multiple single data item inputs. The nice side effect is that there is now a single `Parameter<T>` class with no derived specialisations (save for `Function1DWrapper` which could also be accounted for but would pollute the class a little bit, plus it's slated for addressing at a later stage).

I'll be honest - I went around the houses on this one! No amount of template mangling, overloading, or C++20 sugar would ultimately allow me to follow my initial route which was to move to a `Parameter<DataClass,StorageClass>` where the storage type was explicitly known and could thus be accounted for with `constexpr` conditionals, but it turns out that this is (probably) impossible as one or the other template classes is "lost" when you work from functions in `ParameterBase` (which we necessarily do). However, the C++20 `requires` keyword came to the rescue, and the end result is something not too far from what I had originally tried to create.

The base `upcast()` function is gone and the `ParameterBase<T>::set()` and `get()` functions are now very type explicit, e.g. passing an `int` or a `double` when setting a `Number` will not work. However, we only do this in the unit tests so no real biggie.


src/nodes/parameter.h
```diff
@@ -90,18 +106,45 @@ class ParameterBase : public Serialisable<>
...
+        if (std::type_index(typeid(DataClass)) != storedDataType_)
             return nullptr;
-        auto cast1 = dynamic_cast<PointerParameter<T> *>(this);
-        if (cast1)
-            return cast1->shared_from_this();
-        auto cast2 = static_cast<Parameter<T> *>(this);
-        return cast2->shared_from_this();
+        auto cast1 = static_cast<Parameter<DataClass> *>(this);
+        return cast1->shared_from_this();
+    }
```
rprospero: Since you have the templated `get` method, I think that we can remove this method enirely.  The only place it's currenly being used are in some of the unit tests and I think tha most of those calls we could either replace with calls to `get` or just remove entirely.

src/nodes/parameter.h
```diff
@@ -17,6 +17,11 @@ class Node;
 class ParameterBase;
 template <typename T> class Parameter;
 template <typename T> class PointerParameter;
+template <typename T> class OptionalPointerParameter;
+
+// Template functions to determine if a given derives from a specific base
+template <class T, template <class...> class U> inline constexpr bool is_instance_of_v = std::false_type{};
+template <template <class...> class U, class... Vs> inline constexpr bool is_instance_of_v<U<Vs...>, U> = std::true_type{};
```
rprospero: This is very nice, but feels like it should be both in a separate header so that we can reuse it again and also namespaced.

src/nodes/parameter.h
```diff
@@ -116,84 +159,185 @@ class ParameterBase : public Serialisable<>
...
+        : ParameterBase(parent, name, description, std::type_index(typeid(DataClass))), data_(localPointer_), default_(nullptr),
+          dataGetter_([&]() { return &value; }), dataSetter_([](const DataClass &value) { return false; })
+    {
+    }
+    Parameter(Node *parent, std::string_view name, std::string_view description,
+              std::optional<std::remove_pointer_t<DataClass>> &targetData)
+        requires(std::is_pointer_v<DataClass>)
+        : ParameterBase(parent, name, description, std::type_index(typeid(DataClass))), data_(localPointer_), default_(nullptr),
+          dataGetter_([&]() { return targetData.has_value() ? &targetData.value() : nullptr; }),
+          dataSetter_([](const DataClass &value) { return false; })
```
rprospero, line 196: Is this just a temp value until we fix it at a later date or am I missing a reason why we can't have a setter here?  Somethin along the lines of:

```c++
if (value == nullptr && targetData.has_value())
{
    targetData = {};
    return true;
}
if (value != nullptr && (!targetData.has_value() || *targetData != value))
{
    targetData = value;
    return true;
}
return false;
```

src/nodes/parameter.h
```diff
@@ -116,84 +159,185 @@ class ParameterBase : public Serialisable<>
...
-        : ParameterBase(parent, name, description, std::type_index(typeid(T))), data_(value), default_(value)
+    Parameter(Node *parent, std::string_view name, std::string_view description, DataClass &value)
+        requires(is_instance_of_v<DataClass, std::vector>)
+        : ParameterBase(parent, name, description, std::type_index(typeid(DataClass))), data_(value), default_(value)
     {
     }
-    Parameter(Node *parent, std::string_view name, std::string_view description, std::shared_ptr<ParameterProxy<T>> &proxy)
-        : ParameterBase(parent, name, description, std::type_index(typeid(T))), data_(proxy->data), default_(proxy->data)
+    Parameter(Node *parent, std::string_view name, std::string_view description, std::remove_pointer_t<DataClass> &value)
+        requires(std::is_pointer_v<DataClass>)
```
rprospero, line 186: The `requires` statement does make these methods *much* nicer to implement

src/nodes/parameter.h
trisyoungs: Agreed!

src/nodes/parameter.h
trisyoungs: I confess it's a bit of a copy/paste, but definitely does the job.

src/nodes/parameter.h
trisyoungs: One of the reasons against all this `requires` usage is the fact that it can obfuscate the intent of things a bit, or at least require you to balance the context with another four overloads at the same time!  Here the source data object is a `std::optional` and basically constitutes a read-only `Parameter`, hence the setter returns a hard fail.

src/nodes/parameter.h
trisyoungs: With the caveat above!

src/nodes/parameter.h
```diff
@@ -48,6 +50,13 @@ class ParameterBase : public Serialisable<>
         Input,     /* Indicates that the parameter is meant to be a sink for data and not a source */
         Output,    /* Indicates that the parameter is meant to be a source of data and not a sink */
     };
+    // Allowed Edge Connections
+    enum EdgeLimit
+    {
+        None, /* No edges are allowed */
+        One,  /* Exactly one edge is allowed */
+        Many  /* Many edges are allowed */
+    };
```
RobBuchananCompPhys: `EdgeLimit::None` is feels uncomfortably close to "there are no limits on this edge!" - if I'm not mistaken the real meaning is basically the opposite, an edge with no connections. How about the following
```suggestion
enum class EdgePopulation
{
   None, /* No edges are allowed */
   One, /* Exactly one edge is allowed */
   Many /* Many edges are allowed */
}
```
If "Many" is too imprecise to constitute a Population, we could also call the enum `ExpectedEdges`, or `HasEdges`

rprospero (approved):

PR #2190, opened 2025-06-27
feat: Drag and drop edges

rprospero:
This PR contains multiple updates to the view of nodes and edges in the QML GUI.

20250627_15h11m25s_grim

- Create edges by dragging between input and output parameters.  Works in both directions and uses the proper QML Drag and Drop framework.
- Fix display of parameters to be less ugly
- Fix position of edges with respect to nodes.  There *may* be an issue with lwo DPI displays, but there *absolutely* was an issue before, so I will call "may" an improvement.
- Add more icons for different node types

This might look like a huge PR, but most of the lines of code are just from my copying icons from the "icons/modules" directory into "icons/nodes"


src/gui/models/nodeGraph/graphModel.cpp
```diff
@@ -46,17 +47,17 @@ int GraphModel::count() { return nodes_.rowCount(); }
 void GraphModel::addInput(int nodeIndex, QString paramName, double x, double y)
 {
     auto &node = wrapped_[nodeIndex];
-    x -= node.posx;
-    y -= node.posy;
+    x += 16;
+    y += 64;
```
trisyoungs, line 51: Aside from these being magic numbers (which I will absolutely forgive given the relatively early nature of the encompassing feature!) I'm not quite sure what they're accounting for, especially given that the offsets in x and y are the same for both inputs (on the left) and outputs (on the right)?

trisyoungs (approved):
Looking way more solid.  I had a comment about me not understanding the offsets applied to inputs and outputs, but apart from that this looks good to go in.

A comment on the graphic (because I haven't compiled / tested this myself) but the spinbox delegates for the number inputs are missing / hidden, whereas they should be displayed (with the current values) but disabled for user interaction.  Perhaps this is jumping the gun, but wanted to mention it!

src/gui/models/nodeGraph/graphModel.cpp
rprospero: The short version is that it accounts for the header on the group box.  In the GroupBox, position (0, 0) counts is the top left corner for the border.  Outside the group box, position (0, 0) counts as the top left corner of the icon.  The magic numbers will eventually be solved by getting this distance directly from the program, but that's proven trickier that expected.

rprospero:
@trisyoungs I agree with you about the delegrates.  My plan for when I come back next week is to go more work on the delegates for both the inputs and the options.

PR #2192, opened 2025-07-02
chore: Remove MPI

### LINKED ISSUE #562: Clarify broadcasting code in RDFModule::calculateGR
calculateGR performs the calculations in a triple loop over all the atom pairs in all the molecules, then broadcasts the resulting histograms in a double loop over the atom types.  There's a weird interaction between these loop indices where striding over the initial loop works but cutting the range doesn't.  This code, at the very least, needs to be simplified so that it's possible to understand why the two different looping methods produce different results.

trisyoungs:
I don't really know how you go about reviewing this, but this PR removes all trace of MPI-related parallelism from the code. The various `Distributor` classes as well as `ChangeStore` are now basically redundant, but the effort to remove usage of them is left for when relevant modules are converted to `Node`s.

## TODO
- [x] Check usage of `Distributor` and associated classes
- [x] `ChangeStore`
- [x] `ModuleContext`
- [x] Sweep for `#ifdef PARALLEL` and `Parallel Comms`
- [x] Sweep for `commsTimer`
- [x] Benchmarks
- [x] Test for procedures??
- [x] nix

Closes #562.


rprospero (commented):
I've pushed out a PR that should fix the nix flake.  You covered most of it - just needed to delete a few more lines we didn't need any more.

rprospero:
As I'm reviewing this, it looks like we can get rid of the `chop_range` method and the terribly named `tests/algorthims/aragorn.cpp`, as well as two comments in the `gr` node warning note to use `chop_range` there.

src/nodes/gr/helpers.cpp
```diff
@@ -395,19 +384,15 @@ bool GRNode::calculateGR(const ProcessPool &procPool, Configuration *cfg, Partia
...
-    auto nChunks = (method == PartialsMethod::TestMethod ? 1 : procPool.interleavedLoopStride(ProcessPool::PoolStrategy));
-
     timer.start();
 
     // Loop over molecules...
     // NOTE: If you attempt to use chop_range for this loop, instead of stride, it will fail.
     // The problem does not seem to be in chop_range, but rather in how the loops are merged.
     // This is GitHub issue #562
-    for (auto it = cfg->molecules().begin() + offset; it < cfg->molecules().end(); it += nChunks)
+    for (auto &mol : cfg->molecules())
```
rprospero: This single change was worth reviewing 233 files.

rprospero (approved):
Everything looks much cleaner than before.  I've made some comments above on extra code that we could delete, but that might just need to be a new issue.  I've already added a couple of other issues from items I noticed during this review.

trisyoungs:
> As I'm reviewing this, it looks like we can get rid of the `chop_range` method and the terribly named `tests/algorthims/aragorn.cpp`, as well as two comments in the `gr` node warning note to use `chop_range` there.

Now removed in this PR.

PR #2193, opened 2025-07-02
feat: Clustering improvements

aboswel:
- Addition of more controls for analysis
- Improvements to the size distribution display
- Changes to export behaviour
- Renamed `coordination` to `contact` and fixed issue with values being double when using self clustering options
- Fixed plot display such that maxima display values are determined correctly


src/modules/clustering/clustering.cpp
```diff
@@ -16,28 +16,39 @@ ClusteringModule::ClusteringModule() : Module(ModuleTypes::Clustering)
...
         "Cut-off", "The maximum distance between sites for them to be considered part of the same cluster", cutoff_);
-    keywords_.add<BoolKeyword>("Self-clustering", "If SiteA and SiteB are different, A-A and B-B will contribute to clusters",
-                               selfClustering_);
-
-    keywords_.setOrganisation("Options", "Directional Hydrogen Bonding");
     keywords_.add<BoolKeyword>(
-        "DirectionalHBonding",
+        "Self-clusteringA", "If SiteA and SiteB are different, A-A interactions will contribute to clusters", selfClusteringA_);
+    keywords_.add<BoolKeyword>(
+        "Self-clusteringB", "If SiteA and SiteB are different, B-B interactions will contribute to clusters", selfClusteringB_);
```
trisyoungs: Do you think that calling these "IncludeAA" and "IncludeBB" would be better keyword names?

src/modules/clustering/clustering.cpp
```diff
@@ -16,28 +16,39 @@ ClusteringModule::ClusteringModule() : Module(ModuleTypes::Clustering)
     keywords_.addTarget<ConfigurationKeyword>("Configuration", "Set target configuration for the module", targetConfiguration_)
         ->setEditSignals({KeywordBase::ClearModuleData, KeywordBase::RecreateRenderables});
 
-    keywords_.setOrganisation("Options", "Sites");
+    keywords_.setOrganisation("Options", "Analysis", "Hover for more details on settings");
     keywords_.add<SpeciesSiteKeyword>("SiteA", "Choose the first site for cluster definition", a_);
     keywords_.add<SpeciesSiteKeyword>("SiteB", "Choose the second site for cluster definition", b_);
     keywords_.add<DoubleKeyword>(
         "Cut-off", "The maximum distance between sites for them to be considered part of the same cluster", cutoff_);
```
trisyoungs:
```suggestion
        "Cutoff", "The maximum distance between sites for them to be considered part of the same cluster", cutoff_);
```

web/docs/userguide/modules/analysis/clustering/_index.md
trisyoungs, line 1: Might be the best documented thing in the whole of Dissolve! :)

src/modules/clustering/process.cpp
```diff
@@ -91,39 +92,75 @@ Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
...
 
-    // Decide if we're filtering everything by everything or one by the other
-    if (a_ == b_ && selfClustering_)
+    // Decide how to filter the site maps
+    if (a_ == b_ && (selfClusteringA_ || selfClusteringB_))
     {
         Messenger::error("Sites are the same! Disabling self-clustering...");
-        selfClustering_ = false;
+        selfClusteringA_ = false;
+        selfClusteringB_ = false;
```
trisyoungs: I think it would be safer to hard error here (i.e. `return ExecutionResult::Failed;`) rather than disable the self-clustering and carry on.  The user would potentially never see the error and receive "unexpected" results.

src/modules/clustering/process.cpp
```diff
@@ -91,39 +92,75 @@ Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
...
-                                                                         100); // min of 0.01 to avoid self-selection
+    // min of 0.01 to avoid self-selection, max neighbours 100 (should be enough!)
+    SiteFilter base(targetConfiguration_, baseSiteVector);
+    std::tie(std::ignore, neighbourMapA) = base.filterBySiteProximity(filterSiteVector, Range(0.001, cutoff_), 1, 100);
 
-    // If we're dealing with a definition between the same site, it's already symmetric. Running twice will add dupes
-    if (a_ != b_)
+    // NeighbourMap needs to by symmetric (Every site has a key)
+    // Need to be careful about duplicated entries when we come to combine the maps later
+    // If the inital Vecs are the same, the map is already symmetric
```
trisyoungs:
```suggestion
    // If the initial Vecs are the same, the map is already symmetric
```

src/modules/clustering/process.cpp
```diff
@@ -91,39 +92,75 @@ Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
...
+    if (baseVec != filterVec)
```
trisyoungs: Is there a cheaper way of knowing whether the two vectors are going to contain the same elements, rather than checking every element against the other?

src/modules/clustering/process.cpp
```diff
@@ -207,6 +242,7 @@ Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
 
     // Molecule cluster map - ensures metrics calculated correctly when a molecule has multiple sites
     molClusterMap_.clear();
+    Messenger::print("Mol Map sizes");
```
trisyoungs:
```suggestion
```

src/modules/clustering/process.cpp
```diff
@@ -221,10 +257,47 @@ Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
...
+    if (status == GenericItem::ItemStatus::Created)
+        histSizeData.initialise(0.5, targetConfiguration_->nMolecules() + 0.5, 1.0);
+
+    histSizeData.zeroBins();
+
+    // Figure out how many molecules of interest are in the configuration
+    int interestingMols;
+    a_->parent() == b_->parent() ? interestingMols = targetConfiguration_->speciesPopulation(a_->parent())
+                                 : interestingMols = targetConfiguration_->speciesPopulation(a_->parent()) +
+                                                     targetConfiguration_->speciesPopulation(b_->parent());
```
trisyoungs:
```suggestion
    auto interestingMols =
    a_->parent() == b_->parent() ? targetConfiguration_->speciesPopulation(a_->parent())
                                 : targetConfiguration_->speciesPopulation(a_->parent()) +
                                                     targetConfiguration_->speciesPopulation(b_->parent());
```

src/modules/clustering/process.cpp
```diff
@@ -221,10 +257,47 @@ Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
...
+
+    // Find the number of molecules not in clusters
+    int totalMolsClustered = 0;
+    for (const auto &[cluster, memVec] : molClusterMap_)
+        totalMolsClustered += memVec.size();
+
+    molsNotClustered_ = interestingMols - totalMolsClustered;
+
+    // Fill histogram with number of mols per cluster size
+    for (int i = 0; i < molsNotClustered_; i++)
```
trisyoungs:
```suggestion
    for (auto i = 0; i < molsNotClustered_; i++)
```

src/modules/clustering/process.cpp
```diff
@@ -221,10 +257,47 @@ Module::ExecutionResult ClusteringModule::process(ModuleContext &moduleContext)
...
+
+    // Figure out how many molecules of interest are in the configuration
+    int interestingMols;
+    a_->parent() == b_->parent() ? interestingMols = targetConfiguration_->speciesPopulation(a_->parent())
+                                 : interestingMols = targetConfiguration_->speciesPopulation(a_->parent()) +
+                                                     targetConfiguration_->speciesPopulation(b_->parent());
+
+    // Find the number of molecules not in clusters
+    int totalMolsClustered = 0;
+    for (const auto &[cluster, memVec] : molClusterMap_)
```
trisyoungs:
```suggestion
    auto totalMolsClustered = 0;
    for (const auto &[_, memVec] : molClusterMap_)
```

trisyoungs (approved):
Looking really good.  I haven't tested it, but perhaps it would be nice to write a short user example on a system that you already have to illustrate the numbers? I appreciate time is short, but even a rough version that I can write up later would be awesome.

src/modules/clustering/process.cpp
aboswel: These vectors will have a maximum of two elements each so shouldn't be too bad?

trisyoungs: I guess not, so I suppose I misunderstood the precise contents of these two vectors. Maybe their naming and commenting could better describe that they actually contain?

PR #2194, opened 2025-07-02
fix: No more custom close buttons on tabs

trisyoungs:
This PR removes the custom close buttons from main tabs for two reasons:

1. They were too easy to accidentally click on and panic the user that they were about to close a tab, when in fact this is quire a rare request (tabs can still be removed via the main menu).
2. Because they were custom tabs and not the Qt "built-in" close buttons, adding them on OSX would hide the tab title for some reason. The alternative is to add built-in close buttons to _all_ tabs, but this would then create a bit of an issue with the messages and forcefield tabs which can never be closed.


RobBuchananCompPhys (approved):
Sounds like a good reason to chop out some code!

PR #2195, opened 2025-07-08
feat: Chiral nanotube and graphene sheet creator

trisyoungs:
This PR replaces the previous simplistic nanotube creation with one which implements the proper (n,m) system. It will also generate plain graphene sheets, periodic or not.

## TODO
- [x] Rename dialog
- [x] H-termination
- [x] TIdy up tube ends in non-periodic generation
- [x] Extension of tubes / sheets along Y in integer multiples of _c_
- [x] Fix - generating one layer of overlapping atoms?

### DIFF
--- a/src/gui/createGrapheneSpeciesDialog.cpp
+++ b/src/gui/createGrapheneSpeciesDialog.cpp
@@ -99,14 +99,14 @@ std::map<SpeciesAtom *, std::vector<SpeciesAtom *>> CreateGrapheneSpeciesDialog:
 
 // Get vector of atom keys from atom/neighbour map, sorted by position along indicated direction
 std::vector<SpeciesAtom *>
-CreateGrapheneSpeciesDialog::getSorted(const std::map<SpeciesAtom *, std::vector<SpeciesAtom *>> &atoms, int dir) const
+CreateGrapheneSpeciesDialog::getSortedByY(const std::map<SpeciesAtom *, std::vector<SpeciesAtom *>> &atoms) const
 {
     std::vector<SpeciesAtom *> sorted;
     for (auto &&[i, _] : atoms)
         sorted.push_back(i);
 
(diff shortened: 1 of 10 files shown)


src/gui/createGrapheneSpeciesDialog.cpp
```diff
@@ -0,0 +1,521 @@
...
+    std::vector<SpeciesAtom *> sorted;
+    for (auto &&[i, _] : atoms)
+        sorted.push_back(i);
+
+    std::sort(sorted.begin(), sorted.end(),
+              [dir](const SpeciesAtom *a, const SpeciesAtom *b) { return a->r().get(dir) < b->r().get(dir); });
+
+    return sorted;
+}
+
```
rprospero: Small thing, but, since this is only ever used in one place, we could just make this `getSortedByY`

```suggestion
// Get vector of atom keys from atom/neighbour map, sorted by position along indicated direction
std::vector<SpeciesAtom *>
CreateGrapheneSpeciesDialog::getSortedByY(const std::map<SpeciesAtom *, std::vector<SpeciesAtom *>> &atoms) const
{
    std::vector<SpeciesAtom *> sorted;
    for (auto &&[i, _] : atoms)
        sorted.push_back(i);

    std::sort(sorted.begin(), sorted.end(),
              [dir](const SpeciesAtom *a, const SpeciesAtom *b) { return a->r().y < b->r().y; });

    return sorted;
}

```

rprospero (approved):
One minor suggestion that could result a negligible performance gain, but everything else looks good.

src/gui/createGrapheneSpeciesDialog.cpp
trisyoungs: Ah, yes.  This was introduced thinking there would be more trials of sorting along different directions, but as it turned out not so much!

PR #2201, opened 2025-07-10
refactor:Fewer for each pair early

### LINKED ISSUE #2199: Refactor for_each_pair_early calls
Multiple different calls to `for_each_pair_early` do not actually use
the early return functionality.  These instances should be replaced
with a regular `for_each_pair`.  This can have significant performance
benefits, since `for_each_pair_early` cannot run in parallel, but
`for_each_pair` can.

rprospero:
There were a couple of instances of `for_each_pair_early` that never had an early return, so I conveted them over to `for_each_pair`.  With the addition of a mutex that is only triggered during the *creation* of new pair potentials, it is now possible to update the pair potentials in parallel.

Closes #2199

### DIFF
--- a/src/main/pairPotentials.cpp
+++ b/src/main/pairPotentials.cpp
@@ -119,39 +119,35 @@ bool Dissolve::updatePairPotentials(std::optional<bool> useCombinationRulesHint)
                           pairPotentials_.end());
 
     // Second step - add or update tabulated pair potentials defined by the parameters and form of the associated atom types
-    if (!for_each_pair_early(coreData_.atomTypes(),
-                             [&](int typeI, const auto &at1, int typeJ, const auto &at2) -> EarlyReturn<bool>
-                             {
-                                 // Try to locate existing pair potential between these atom types
-                                 auto *pot = pairPotential(at1, at2);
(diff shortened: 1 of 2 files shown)


src/main/pairPotentials.cpp
```diff
@@ -118,40 +119,39 @@ bool Dissolve::updatePairPotentials(std::optional<bool> useCombinationRulesHint)
...
+                            [&](int typeI, const auto &at1, int typeJ, const auto &at2)
+                            {
+                                // Try to locate existing pair potential between these atom types
+                                auto *pot = pairPotential(at1, at2);
+
+                                // If it doesn't exist we create it
+                                if (!pot)
+                                {
+                                    // Don't let multiple threads try to simultaneously add new potentials
+                                    std::lock_guard<std::mutex> guard(addingPotential);
```
trisyoungs: Is this necessary since the policy is `ParallelPolicies::seq`?

trisyoungs (approved):

src/main/pairPotentials.cpp
rprospero: I just did a second check and I was *not* convinced that the parallel code was safe, so I've removed all of the mutex code and kept everything sequential for now.

PR #2203, opened 2025-07-22
feat: Generalise method for node data serialisation

trisyoungs:
Just a draft for now so you can see where this is going and comment before it goes too far!  I took inspiration from my recent use of `requires` in the vector parameter PR to implement a general way of flagging data which should be serialised because it is "restart" or "persistent" data.  Reasons for doing it in this way are:

1. It makes it trivially easy to flag what data get saved / restored from a `Node` - just call `addSerialisable()` with the name (key) of the data as it appears in the TOML and the object you want serialised.
2. Different types of data - PODs, classes derived from `Serialisable`, `Serialisable`s contained within `std::optional`s - can all be handled without requiring multiple function definitions or overloads.
3. (De)serialisation can be handled automatically from within the `Node`.
4. Keys for data are defined once, rather than requiring it to be written e.g. once in `serialise()` and once in `deserialise()` is one was going to go the manual route.

This approach will require that all classes to be serialised derive from `Serialisable`, but many currently don't (e.g. `PartialSet`) so this will encompass most of the work past this point.

## Dependents
- [x] #2206
- [x] #2209
- [x] #2213
- [x] #2216

### DIFF
--- a/src/nodes/atomicMC/atomicMC.h
+++ b/src/nodes/atomicMC/atomicMC.h
@@ -45,4 +45,13 @@ class AtomicMCNode : public Node
     private:
     // Run main processing
     NodeConstants::ProcessResult process() override;
+
+    /*
+     * Serialisation
+     */
+    protected:
+    // Express persistent data within the supplied serialisable value
+    virtual void serialisePersistentData(SerialisedValue &value) const;
+    // Retrieve persistent data from the supplied serialisable value
+    virtual void deserialisePersistentData(const SerialisedValue &value);
 };
--- a/src/nodes/atomicMC/atomicMC.h
+++ b/src/nodes/atomicMC/atomicMC.h
@@ -45,13 +45,4 @@ class AtomicMCNode : public Node
     private:
     // Run main processing
     NodeConstants::ProcessResult process() override;
-
-    /*
-     * Serialisation
-     */
-    protected:
-    // Express persistent data within the supplied serialisable value
-    virtual void serialisePersistentData(SerialisedValue &value) const;
-    // Retrieve persistent data from the supplied serialisable value
-    virtual void deserialisePersistentData(const SerialisedValue &value);
 };
--- a/src/nodes/atomicMC/process.cpp
+++ b/src/nodes/atomicMC/process.cpp
@@ -124,16 +124,3 @@ NodeConstants::ProcessResult AtomicMCNode::process()
 
     return NodeConstants::ProcessResult::Success;
 }
-
-/*
- * Serialisation
- */
-
-// Express persistent data within the supplied serialisable value
-void AtomicMCNode::serialisePersistentData(SerialisedValue &value) const { value["stepSize"] = stepSize_; }
-
-// Retrieve persistent data from the supplied serialisable value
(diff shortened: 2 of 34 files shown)


src/nodes/atomicMC/process.cpp
```diff
@@ -124,3 +124,16 @@ NodeConstants::ProcessResult AtomicMCNode::process()
 
     return NodeConstants::ProcessResult::Success;
 }
+
+/*
+ * Serialisation
+ */
+
+// Express persistent data within the supplied serialisable value
+void AtomicMCNode::serialisePersistentData(SerialisedValue &value) const { value["stepSize"] = stepSize_; }
```
rprospero: Is there a particular reason that method modifies the pass data instead of taking no parameters and returning a SerialisedValue?

src/nodes/node.cpp
```diff
@@ -350,3 +353,30 @@ void Node::deserialise(const SerialisedValue &node)
...
+SerialisedValue Node::serialiseData() const
+{
+    SerialisedValue result;
+    result["timing"] = timing_.serialise();
+
+    for (auto &[key, serialisable] : serialisables_)
+    {
+        if (serialisable->canSerialise())
+            result[key] = serialisable->serialise();
+    }
```
rprospero: Stupid pedantic stylistic issue, but dropping these braces leads to a beautiful double indent that I believe our formatting settings should allow.

rprospero (approved):
It mostly looks good.  I had a design question, but I'm suspecting that there's some limitations that I'm not seeing.

src/nodes/atomicMC/atomicMC.h
```diff
@@ -45,4 +45,13 @@ class AtomicMCNode : public Node
...
     NodeConstants::ProcessResult process() override;
+
+    /*
+     * Serialisation
+     */
+    protected:
+    // Express persistent data within the supplied serialisable value
+    virtual void serialisePersistentData(SerialisedValue &value) const;
+    // Retrieve persistent data from the supplied serialisable value
+    virtual void deserialisePersistentData(const SerialisedValue &value);
```
RobBuchananCompPhys: Why are these methods defined as virtuals at the level of `AtomicMCNode`? If they have any general usage, should we not be simply overriding them here?

src/nodes/atomicMC/process.cpp
trisyoungs: Nope. None at all. I will change to be consistent with our other TOML functions.

src/nodes/atomicMC/atomicMC.h
trisyoungs: Good catch - in fact these are leftovers from testing I was doing while implementing this, and are no longer relevant.

src/nodes/node.cpp
trisyoungs: Nope, I agree. Extraneous braces to be removed.

src/nodes/atomicMC/process.cpp
trisyoungs: Ah, in fact I'm already doing this! This was a test function I introduced to get the idea rolling, and is now removed.  The bona fide `serialiseData()` function returns a `SerialisedValue`.

PR #2204, opened 2025-07-22
refactor: Single configuration GR node and updated associates

trisyoungs:
Ahead of #2203 it was worthwhile taking a step back and refactoring the `GRNode` to take a single `Configuration` input rather than a `std::vector`.  Combining `GR` results is a job for a separate node in the new scheme of things.

At the same time I have updated `SQNode` and `NeutronSQNode` and clarified the two separate "species population" maps we have - one with only ever integer populations coming from a `Configuration`, and one with only ever real populations coming from a `PartialSet` (which will occur when we start mixing `GR` results in the future).

### DIFF
--- a/src/nodes/gr/gr.cpp
+++ b/src/nodes/gr/gr.cpp
@@ -5,8 +5,7 @@
 
 GRNode::GRNode(Graph *parentGraph) : Node(parentGraph)
 {
-    addInput<std::vector<Configuration *>>("Configurations", "Set target configuration(s) for the module",
-                                           targetConfigurations_)
+    addInput<Configuration *>("Configurations", "Set target configuration(s) for the module", targetConfiguration_)
         ->setFlags({ParameterBase::Required, ParameterBase::ClearData});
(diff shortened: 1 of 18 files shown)


src/nodes/gr/gr.cpp
```diff
@@ -5,8 +5,7 @@
 
 GRNode::GRNode(Graph *parentGraph) : Node(parentGraph)
 {
-    addInput<std::vector<Configuration *>>("Configurations", "Set target configuration(s) for the module",
-                                           targetConfigurations_)
+    addInput<Configuration *>("Configurations", "Set target configuration(s) for the module", targetConfiguration_)
```
RobBuchananCompPhys:
```suggestion
    addInput<Configuration *>("Configuration", "Set target configuration for the module", targetConfiguration_)
```
Just for now!

RobBuchananCompPhys (approved):

src/nodes/gr/gr.cpp
trisyoungs: Oops. Good catch!

PR #2205, opened 2025-07-23
refactor: Retrofit AtomicMC MPI

RobBuchananCompPhys:


src/nodes/atomicMC/process.cpp
```diff
@@ -30,101 +30,61 @@ NodeConstants::ProcessResult AtomicMCNode::process()
...
+            er = kernel->totalEnergy(*i);
+            currentEnergy = er.totalUnbound();
+            currentIntraEnergy = er.geometry() * termScale;
 
-            // Get Molecule pointer
-            std::shared_ptr<Molecule> mol = targetConfiguration_->molecule(molId);
+            // Loop over number of shakes per Atom
+            for (auto n = 0; n < nShakesPerAtom; ++n)
+            {
+                // Create a random translation vector
```
trisyoungs, line 63: Apart from building up a list of positional changes to send to all MPI processes, the other useful function of `ChangeStore` was to track and revert positional changes when moves were rejected.  That is now missing from the routine, so we don't really have MC any more! At the beginning of this inner loop you should store the current coordinates of `i` in a temporary var, and revert back to those coordinates if the move is not accepted (`if` on line 76). You'll also need to call `targetConfiguration_->updateAtomLocation(i);` after reverting the position.

trisyoungs (approved):
Incredible how much code fluff that was adding to the routine, eh!  One critical thing to add back in, then good to go I think.

trisyoungs (approved):
:+1:

PR #2206, opened 2025-07-23
feat: Data history storage (for averaging)

trisyoungs:
### Follows #2204 and must be merged before #2203.

This PR introduces a new `History` template class to generalise (and simplify) the functionality of the old `Averaging` namespace. The idea is straightforward - introduce a sort of managed `std::vector` of data which can be easily formed into an average as well as serialised if required (e.g. as node data will require).

### DIFF
--- a/src/math/history.h
+++ b/src/math/history.h
@@ -5,12 +5,13 @@
 
 #include <memory>
 #include <vector>
+#include "base/serialiser.h"
 
-// Data Averaging
-template <class T> class Averager
+// Data History
+template <class T> class History
 {
     private:
-    // Historical data
+    // Stored historical data
     std::vector<std::unique_ptr<T>> history_;
 
     public:
@@ -22,7 +23,7 @@ template <class T> class Averager
 
         // Prune old data to get to the averagingLength
         while (history_.size() > averagingLength)
-            history_.pop_front();
+            history_.erase(history_.begin());
 
         // Perform averaging of the datasets that we have
         T averaged;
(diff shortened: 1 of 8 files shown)


src/math/history.h
```diff
@@ -0,0 +1,45 @@
...
+
+        // Prune old data to get to the averagingLength
+        while (history_.size() > averagingLength)
+            history_.erase(history_.begin());
+
+        // Perform averaging of the datasets that we have
+        T averaged;
+        auto weight = 1.0 / history_.size();
+        for (auto &data : history_)
+            averaged += *data * weight;
```
rprospero, line 32:
```suggestion
        // How many items to average
        auto length = history.size() < averagingLength ? history.size() : averagingLength;

        // Perform averaging of the datasets that we have
        T averaged;
        auto weight = 1.0 / length;
        for (auto &data : std::span(history_.rbegin(), ristory_.rbegin()+length))
            averaged += *data * weight;
```

This is an alternate implementation that doesn't lose history.  This would allow us to look at multiple averaging lengths simultaneously (the current implementation essentially locks you into the shortest length).  The disadvantage is that it can grow without bound, though that could be fixed by replacing the vector with a ring buffer.

This isn't a necessary change - just a suggestion.

rprospero (commented):
Looks good.  Also gives us a nice framework for other stats (e.g. standard deviation, median, etc.)

One small suggested change, but, otherwise, things look good.

src/math/history.h
trisyoungs: Interesting thought. While I like the idea, my concern is that this could add a lot of bloat as many of the objects being stored here are pretty chunky (I'm looking at you, `PartialSet`).  As such, I would prefer to stick to the ultra-simple, shortest-length version for the time being.

Issue #2207, opened 2025-07-23
AtomType Refactoring 1


trisyoungs:
One of the most fundamental objects within Dissolve is `AtomType` - in itself it is just a storage object for potential interaction information for a specific type of atom, but is used as a means to index a vast array of other quantities. For example, when calculating g(r) (or S(Q)) one requires a set of partials _between atom types_, leading to the commonplace use of `Array2D`s whose elements are keyed by the indices of `AtomTypes`.

But where do the indices come from?  It was historically assumed that there would be a master vector of `AtomType`s living in `CoreData` which would not change order and could reliably be used as a source of "index truth". While to-date this has proven successful, it does represent a rather fragile assumption, easily broken with poor code or innocent user action.

A further issue comes when considering serialisation. The old non-TOML serialisation in Dissolve allows ready access to `CoreData` and hence to current `AtomType` information, but moving to the node graphs of Dissolve2 this is not the case.  As such, (de)serialisation of `PartialSet` (which depends on `AtomTypeMix`, which depends on `AtomTypeData`, which contains an `AtomType` pointer to store & retrieve) is not possible.

### Solution
Assuming that `AtomType` data becomes delocalised in Dissolve2, but that an `Atom` in a `Configuration` still knows which `AtomType` is relevant for it (via it's knowledge of it's related `SpeciesAtom`) we can consider storing only atom type _names_ rather than their pointers. This will drive us away from indexed `Array2D` structures for e.g. `PartialSet` and towards a "double-keyed" `std::map` where the key is the combination of two atom type names.

### Tasks
- [x] Introduce templated `TypeMap` to provide a double-keyed, optionally "half-matrix"-style alternative to the use of `Array2D` in e.g. `PartialSet`.
- [x] Refactor storage in `AtomTypeMix` to use a `std::map` keyed by `AtomType` name.
- [ ] `Atom` still contains `masterTypeIndex_`, and it is still this which is used in energy routines.

### Follow-On
[...]

PR #2209, opened 2025-07-25
refactor: Sweep through underlying AtomType-using classes

trisyoungs:
As lengthily detailed in #2207, this PR attempts to mitigate the prevalent use of (de)serialising data requiring lookup of `AtomType` names in `CoreData` which, in Dissolve2, we will not have the luxury of. Most changes revolve around deciding what really needs to be (de)serialised vs what can be cheaply recalculated, but also separates out functionality from (in particular) `AtomTypeMix` into classes with more specific use cases.

For the record, many storage instances of `std::shared_ptr<AtomType>` have changed to the bare pointer since there is no associated ownership.

Changes to classes:
- `KeyedVector` template introduced to maintain a `std::vector` of `std::pair`s with some `std::map`-like functionality.
- `IsotopologueWeight` is removed, and `Isotopologues` storage changed to a `KeyedVectorr<Isotopologue *, double>`.

### TODO
- [ ] ~Check / change behaviour of `IsotopologueSetModel` since it allows changing of the `Isotopologue` to another `Isotopologue` for a given weight. When we had a vector of `IsotopologueWeights` this made sense, but now we have a map it can lead to unintentional user error.~  Moved to #2152


src/classes/atomTypeMix.cpp
```diff
@@ -69,9 +69,9 @@ void AtomTypeMix::add(const AtomTypeMix &source)
 }
 
 // Add/increase this AtomType/Isotope pair
-void AtomTypeMix::addIsotope(std::shared_ptr<AtomType> atomType, Sears91::Isotope tope, double popAdd)
+void AtomTypeMix::addIsotope(const AtomType *atomType, Sears91::Isotope tope, double popAdd)
 {
-    auto &atd = std::get<0>(add(std::move(atomType)));
+    auto &atd = std::get<0>(add(atomType));
```
RobBuchananCompPhys: As we are moving away from smart pointers for this, do we need a purpose-built move constructor for `AtomType`, to ensure safety? (And if we did have one, would not `std::move` then invoke it, meaning no change here?)

src/templates/keyedVector.h
```diff
@@ -0,0 +1,119 @@
...
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#pragma once
+
+#include "templates/optionalRef.h"
+#include <functional>
+#include <vector>
+
+// Keyed Vector
+template <typename KeyClass, typename ValueClass> class KeyedVector
```
RobBuchananCompPhys: Interesting - how exactly is this preferable to a `std::map` (I think I found the paragraph where you mentioned why we need this, but can't locate it now!)

RobBuchananCompPhys (approved):

src/classes/atomTypeMix.cpp
trisyoungs: I don't think so. We're now just storing raw pointers in these kinds of classes, so no actual moving needs to (or should ever) happen (also, `AtomType`s are arguably better off existing as `std::unique_ptr`s).

src/templates/keyedVector.h
trisyoungs: Now, there is no 100% convincing answer to this as it turns out. I wrote `KeyedVector` to offer convenience of `std::map`'s key access to elements but with the rigid order of a `std::vector`.  At this point, after doing a bunch of other work, I'm not so sure it is really required, so I will attempt to remove it and see how bad it is further on up the PR stack.

PR #2210, opened 2025-07-25
feat:Display enum parameters

rprospero:
This PR presents the user with a drop down for selecting enum values.  This requires that the enum is included in the `EnumRegistry`.

As a side benefit of this PR, the options are now editable (they were previously read only).  Additionally, it disables building the legacy GUI, though the tests remain.


src/gui/models/nodeGraph/enumOptionsModel.cpp
```diff
@@ -1,13 +1,14 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // Copyright (c) 2025 Team Dissolve and contributors
 
-#include "gui/models/enumOptionsModel.h"
+#include "gui/models/nodeGraph/enumOptionsModel.h"
 #include "base/enumOptionsBase.h"
 
 // Set source AtomType data
```
trisyoungs, line 7: No prizes for guessing where this code originated! :)

src/gui/models/nodeGraph/enumRegistry.cpp
```diff
@@ -0,0 +1,43 @@
...
+void EnumRegistry::instantiateOptions()
+{
+    if (!options_.empty())
+        return;
+
+    auto wrap = [](auto &&x) -> std::shared_ptr<EnumOptionsModel>
+    {
+        std::shared_ptr<EnumOptionsModel> result = std::make_shared<EnumOptionsModel>();
+        std::shared_ptr<EnumOptionsBase> copy{std::make_shared<std::remove_reference_t<decltype(x)>>(x)};
+        result->setData(copy);
```
trisyoungs: If you wanted to:
```suggestion
        auto result = std::make_shared<EnumOptionsModel>();
        result->setData(std::make_shared<std::remove_reference_t<decltype(x)>>(x));
```

src/gui/models/nodeGraph/parameterModel.cpp
```diff
@@ -11,8 +14,11 @@ enum Roles
     DESCRIPTION,
     TYPE,
     DATA,
+    MODEL,
 };
 
+class EnumModel;
+
```
trisyoungs: Not needed.
```suggestion
```

src/nodes/parameter.h
```diff
@@ -109,6 +112,11 @@ class ParameterBase : public Serialisable<>
     virtual bool assign(ParameterBase *other) = 0;
     // Return whether this parameter accepts the output type of the other
     virtual bool acceptsOutput(ParameterBase *other) const = 0;
+    // The type's representation as a raw int (only valid for int and enum)
+    virtual int asInt() const { return -1; }
+    // Set type's repreentation as a raw int (only valid for int and enum)
+    virtual void asInt(int value) { return; }
```
trisyoungs: I am slightly perturbed that the "rigidity" of the new `ParameterBase` implementation means that we have to do this, but I don't see a way around it at present (at least not without rewriting `EnumOptionsBase`). I would suggest the following name changes though for clarity.
```suggestion
    // The type's representation as a raw int (only valid for int and enum)
    virtual int getAsInt() const { return -1; }
    // Set type's representation as a raw int (only valid for int and enum)
    virtual void setFromInt(int value) { return; }
```

src/nodes/parameter.h
```diff
@@ -224,6 +232,27 @@ template <typename DataClass> class Parameter : public ParameterBase, public std
...
+    {
+        if constexpr (std::is_integral_v<DataClass>)
+            return data_;
+        if constexpr (std::is_enum_v<DataClass>)
+            return static_cast<int>(data_);
+
+        Messenger::exception("Tried to extract int value from non integral type {}", storedDataType_.name());
+    }
+    // Set type's repreentation as a raw int (only valid for int and enum)
+    void asInt(int value) override
```
```suggestion
    // Set type's representation as a raw int (only valid for int and enum)
    void setFromInt(int value) override
```

src/nodes/parameter.h
```diff
@@ -224,6 +232,27 @@ template <typename DataClass> class Parameter : public ParameterBase, public std
         else
             return AllowedEdgeCount::One;
     }
+    // The type's representation as a raw int (only valid for int and enum)
+    int asInt() const override
```
```suggestion
    // The type's representation as a raw int (only valid for int and enum)
    int getAsInt() const override
```

trisyoungs (approved):
This has come together really nicely, and I kinda like the centralised storage of models for different `EnumOptions`.  Couple of minor suggesetions / comments, but otherwise all good.

rprospero:
@trisyoungs I thought it might be worth a second review before I merge to get a glimpse of the changes I made to disable the packaging.

trisyoungs (approved):

PR #2214, opened 2025-07-28
feat: Introduce DoubleKeyedMap for more generalised storage of pair-related data

trisyoungs:
## Follows #2209

This PR replaces the `Array2D` storage of RDFs and histograms, used extensively by `PartialSet`, with one based on `std::map`.  The principal reason for doing so is detailed in #2207 and removes any dependence of calculated partials on the order of the underlying `AtomType`s.  A new class `DoubleKeyedMap` is introduced to implement this, and allows e.g. entries to be retrieved based on a key generated from two atom type names, also taking into account whether the terms i-j and j-i should be treated as equivalent (`Array2D::half_ == true`).


src/templates/doubleKeyedMap.h
```diff
@@ -0,0 +1,140 @@
...
+    // Iterators
+    std::map<std::string, ValueClass>::iterator begin() { return data_.begin(); }
+    std::map<std::string, ValueClass>::const_iterator begin() const { return data_.begin(); }
+    std::map<std::string, ValueClass>::iterator end() { return data_.end(); }
+    std::map<std::string, ValueClass>::const_iterator end() const { return data_.end(); }
+    // Return the underlying map
+    std::map<std::string, ValueClass> &map() { return data_; }
+    const std::map<std::string, ValueClass> &map() const { return data_; }
+    // Return number of data in map
+    int size() const { return data_.size(); }
```
RobBuchananCompPhys, line 119: Since we have the possibility that `A//B` == `B//A`, do we need something like a `countUnique()` method to give the effective size of the data in terms of keys that have this mirror form (thus exluding the equivalent keys)? Would that be useful information to access from the point of view of the simulation?

RobBuchananCompPhys (approved):

src/templates/doubleKeyedMap.h
trisyoungs: Good question.  You might expect this to be quite a useful number ( the total number of unique pairs is (N * (N+1)) / 2)) but if you're not storing it in a 2D matrix and are just looking up numbers then you don't need to know, which is kinda nice.

Issue #2196, opened 2025-07-10
Move to Project Folders


trisyoungs:
Currently we have an almost non-existent way of managing the data that Dissolve generates and the data which Dissolve requires, leading to a flat file structure which can be confusing to the users.  This issue proposes to move to a project folder-style management, with Dissolve taking more responsibility in organising restart files / checkpoints, data output, and external data import.

This may look a little like:
```
MyProjectName/
├─ checkpoints/                # Restart points, one folder per point
│  ├─ equilibration/
│  │  ├─ checkpoint.restart
│  │  ├─ dangle.txt
│  ├─ first_test/
│  │  ├─ checkpoint.restart
│  ├─ refinement/
│  │  ├─ checkpoint.restart
├─ data/                        # Reference data, e.g. experimental data
│  ├─ NIMROD00095952.mint01
│  ├─ NIMROD00095953.mint01
│  ├─ xray.dat
├─ run/                         # Output from the current run
│  ├─ checkpoint.restart
│  ├─ dangle.txt
│  ├─ rdfbc.txt
├─ dissolve.dat                 # Project information
├─ input.txt                    # Main input file
```

PR #2215, opened 2025-07-29
refactor: Argon nodes testing

RobBuchananCompPhys:


src/nodes/atomicMC/atomicMC.cpp
```diff
@@ -12,6 +12,7 @@ AtomicMCNode::AtomicMCNode(Graph *parentGraph) : Node(parentGraph)
     addOption<Number>("TargetAcceptanceRate", "Target acceptance rate for Monte Carlo moves", targetAcceptanceRate_);
     addOption<Number>("StepSizeMax", "Maximum allowed value for step size, in Angstroms", stepSizeMax_);
     addOption<Number>("StepSizeMin", "Minimum allowed value for step size, in Angstroms", stepSizeMin_);
+    addOutput<Configuration *>("Configuration", "Output configuration for the module", targetConfiguration_);
```
trisyoungs: Also needs changing in the `addInput()` line.
```suggestion
    addOutput<Configuration *>("Configuration", "Output configuration", targetConfiguration_);
```

src/nodes/atomicSpecies.cpp
```diff
@@ -1,13 +1,15 @@
 #include "atomicSpecies.h"
+#include "dissolve.h"
 
 AtomicSpeciesNode::AtomicSpeciesNode(Graph *parentGraph, Elements::Element Z) : Node(parentGraph)
 {
     // Create atom and AtomType
-    auto at = atomTypes_.emplace_back(std::make_shared<AtomType>(Z));
+    auto &at = DissolveGraph::addAtomType(atomTypes_, Z);
```
trisyoungs: Ah, now, here this is conforming to the old adage of "Mr Dissolve owns it all", whereas now I think `AtomType`s need to be stored as local objects on the `Species` itself (meaning that `DissolveNode` owns no `AtomType`s at all).  This way, a `Species` truly owns all of its data (we'll also need to the same with `MasterTerm`s at some point).  How do we share `AtomType`s between `Species` if we do this?  We'd need some way of configuring that at the point of creation of the `Species`, but this is an issue for another day.

src/nodes/atomicSpecies.cpp
```diff
@@ -1,13 +1,15 @@
 #include "atomicSpecies.h"
+#include "dissolve.h"
```
trisyoungs, line 5:
```suggestion
// SPDX-License-Identifier: GPL-3.0-or-later
// Copyright (c) 2025 Team Dissolve and contributors

#include "nodes/atomicSpecies.h"
#include "nodes/dissolve.h"
```

src/nodes/dissolve.cpp
```diff
@@ -20,3 +20,111 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+    // Create a suitable unique name
+    newAtomType->setName(DissolveSys::uniqueName(Elements::symbol(Z), atomTypes,
+                                                 [&](const auto &at) { return newAtomType == at ? "" : at->name(); }));
+
+    // Set data
+    newAtomType->setZ(Z);
+    newAtomType->setIndex(atomTypes.size() - 1);
+
+    return newAtomType;
+}
```
trisyoungs: As per the comment above, this function would become redundant, but we would need equivalent functions in `Species`.

src/nodes/dissolve.cpp
```diff
@@ -20,3 +20,111 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+{
+    auto pop = configuration->atomTypePopulations();
+    std::vector<std::shared_ptr<AtomType>> atomTypes;
+    atomTypes.resize(pop.nItems());
+
+    for (int i = 0; i < pop.nItems(); i++)
+        atomTypes[i] = pop.atomType(i);
+
+    return atomTypes;
+}
```
trisyoungs: This is really a local function within `Configuration`, rather than a function in `DissolveNode`.

src/nodes/dissolve.cpp
```diff
@@ -20,3 +20,111 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+
+bool DissolveGraph::updatePairPotentials(Dissolve &dissolve, const std::vector<std::shared_ptr<AtomType>> &atomTypes,
```
trisyoungs: So here is where a bit of the unknown comes in, for me at least. I think that, rather than storing a global `PotentialMap` within `DissolveNode`, and retrieving it for use when required, we should instead be thinking about having a function which _returns_ a `PotentialMap` based on a `Configuration`s used `AtomType`s (via its `SpeciesPopulations`).  Where should this function live? Arguably, here in `DissolveNode` where you have it, since it will need to access (when we reimplement them) any defined `PairPotentialOverrides`, plus we could practically store generated potentials in a sort of cache to save regenerating them each and every time if no changes have been made.

src/nodes/energy/energy.cpp
```diff
@@ -15,6 +15,7 @@ EnergyNode::EnergyNode(Graph *parentGraph) : Node(parentGraph)
                    stabilityWindow_);
 
     addOption<bool>("Save", "Save calculated energies to disk, one file per targeted configuration", save_);
+    addOutput<Configuration *>("Configuration", "Output configuration for the module", targetConfiguration_);
```
trisyoungs:
```suggestion
    addOutput<Configuration *>("Configuration", "Output configuration", targetConfiguration_);
```

src/nodes/insert.cpp
```diff
@@ -5,13 +5,15 @@
...
 #include "classes/species.h"
+#include "dissolve.h"
 #include "math/mathFunc.h"
 
 InsertNode::InsertNode(Graph *parentGraph) : Node(parentGraph)
 {
     addInput<Configuration *>("Configuration", "Target configuration to insert into", configuration_);
     addOutput<Configuration *>("Configuration", "Modified configuration", configuration_);
     addInput<const Species *>("Species", "Species to add - all resulting molecules will have identical geometry", species_);
+    addInput<const std::vector<std::shared_ptr<AtomType>> *>("AtomTypes", "AtomTypes owned by the node", atomTypes_);
```
trisyoungs: As discussed, this should be unnecessary since the `Species` contains all the information we need.

src/nodes/md/md.cpp
```diff
@@ -28,6 +28,7 @@ MDNode::MDNode(Graph *parentGraph) : Node(parentGraph)
     addOption<bool>("IntraOnly",
                     "Only forces arising from intramolecular terms (including pair potential contributions) will be calculated",
                     intramolecularForcesOnly_);
+    addOutput<Configuration *>("Configuration", "Output configuration for the module", targetConfiguration_);
```
trisyoungs:
```suggestion
    addOutput<Configuration *>("Configuration", "Output configuration", targetConfiguration_);
```

src/nodes/insert.cpp
```diff
@@ -147,6 +149,10 @@ void InsertNode::scaleVolume(int nAtomsToAdd, double massToAdd) const
 // Run main processing
 NodeConstants::ProcessResult InsertNode::process()
 {
+    // Add atom type and update pair potentials
+    if (!DissolveGraph::updatePairPotentials(dissolve(), *atomTypes_))
+        return NodeConstants::ProcessResult::Failed;
+
```
trisyoungs: Ah, OK, I see why you wanted to add this now.  The good news is that it shouldn't be necessary, and a simple call to `configuration_->updateObjectRelationships();` at the end of the function is all that is required (this is all the `AddGeneratorNode` did.  The call to `cells().generate(...)` is necessary but I think should not be the responsibility of `InsertNode`, nor any other, to call once it has changed the composition. Rather, I think that this should be called at the point of a node wanting to use a `Configuration` for something when the cells are genuinely required (e.g. `Energy`, `RDF`, etc.). Trouble is, it's an expensive function to call and we don't want to regenerate cells unless the `Box` itself has been fundamentally changed (scaling doesn't count, btw). For now we can add this to the Dissolve2Do and just call it when its needed, and worry about optimising in a separate PR.

src/nodes/insert.cpp
```diff
@@ -197,6 +203,8 @@ NodeConstants::ProcessResult InsertNode::process()
         }
     }
 
+    configuration_->cells().generate(box, configuration_->requestedCellDivisionLength(), dissolve().potentialMap().range());
```
trisyoungs: As per previous comment, this ugly function call can be refactored into something a bit nicer, particularly when `PotentialMap`s become more local.

trisyoungs (commented):
Lots happening here!  This is probably going to turn into a fairly big PR, but I think it needs to be to capture all the bits that are being changed coherently in one place.

src/classes/configuration_contents.cpp
```diff
@@ -52,6 +52,21 @@ void Configuration::empty()
...
+    std::vector<std::shared_ptr<AtomType>> atomTypes;
+    atomTypes.resize(pop.nItems());
+
+    for (int i = 0; i < pop.nItems(); i++)
+        atomTypes[i] = pop.atomType(i);
+
+    return atomTypes;
+}
+*/
+
```
trisyoungs: Redundant now thanks to `Configuration::atomTypeIndexMap()` and friends.
```suggestion
```

src/nodes/insert.cpp
```diff
@@ -197,6 +200,17 @@ NodeConstants::ProcessResult InsertNode::process()
...
+    std::vector<std::shared_ptr<AtomType>> atomTypes;
+    auto atomTypePop = species_->atomTypePopulations();
+    atomTypes.reserve(atomTypePop.size());
+    for (const auto &at : atomTypePop)
+        atomTypes.push_back(std::make_shared<AtomType>(at.first->Z(), at.first->name()));
+
+    auto kernel = DissolveGraph::prepareEnergyCalculation(dissolve(), {}, configuration_);
+
+    configuration_->updateCells(kernel.get()->potentialMap().range());
+
```
trisyoungs: None of this should be required here for the `InsertNode`, but something like this is exactly what needs to happen at the start of the `AtomicMCNode`.

src/nodes/dissolve.h
```diff
@@ -35,4 +38,10 @@ class DissolveGraph : public Graph
     public:
     // Return dissolve
     Dissolve &dissolve() const override;
+
+    /*
+     * Functions
+     */
+    public:
+    static std::unique_ptr<EnergyKernel> prepareEnergyCalculation(Dissolve &dissolve, const std::vector<std::shared_ptr<AtomType>> &atomTypes, Configuration *cfg, std::optional<double> energyCutoff = {});
```
trisyoungs: Should this function be `static`? We have gone to some effort to make `DissolveNode` discoverable by other nodes anywhere in the graphs so we can have functions like this available.

src/nodes/dissolve.cpp
```diff
@@ -20,3 +21,24 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
 
 // Return dissolve
 Dissolve &DissolveGraph::dissolve() const { return dissolve_; }
+
+/*
+ * Functions
+ */
+
+std::unique_ptr<EnergyKernel> DissolveGraph::prepareEnergyCalculation(Dissolve &dissolve,
```
trisyoungs: There's quite a lot to think about with this function......

src/nodes/dissolve.cpp
```diff
@@ -20,3 +21,24 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+                                                                      const std::vector<std::shared_ptr<AtomType>> &atomTypes,
+                                                                      Configuration *cfg, std::optional<double> energyCutoff)
+{
+	// Update atom type indexing
+	cfg->updateTypeIndexing();
+
+	// Generate configuration potential map
+    PotentialMap potentialMap;
+    potentialMap.initialise(atomTypes, dissolve.pairPotentials(), dissolve.pairPotentialRange());
```
trisyoungs: We shouldn't take a vector of `AtomType`s as an argument - rather, we call `Configuration::atomTypeIndexMap()` or similar to get it, since this is the source of truth we want to use.

src/nodes/dissolve.cpp
```diff
@@ -20,3 +21,24 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+
+	// Regenerate cells
+    cfg->cells().generate(cfg->box(), cfg->requestedCellDivisionLength(), potentialMap.range());
```
trisyoungs, line 54: We certainly shouldn't do this every call - I think previously this was done at the end of `Configuration::generate()` in Dissolve1.  We will have to put in some sort of check to determine whether the `CellArray` needs regenerating.

src/nodes/dissolve.cpp
trisyoungs: Of course it doesn't need to take `Dissolve&` as an argument since all the data it needs will be owned by `DissolveNode` (the generated `PairPotential`s store, defined `PairPotentialOverrides` etc.) or retrieved via the `Configuration` (used `AtomType`s).

src/classes/potentialMap.cpp
```diff
@@ -60,6 +60,92 @@ bool PotentialMap::initialise(const std::vector<std::shared_ptr<AtomType>> &mast
...
+    range_ = pairPotentialRange;
+
+    return true;
+}
+
+bool PotentialMap::initialise(const std::vector<const AtomType *> &masterAtomTypes,
+                              const DoubleKeyedMap<PairPotential*> &pairPotentials, double pairPotentialRange)
+{
+    // Clear old data first
+    clear();
```
trisyoungs: Similar to how I'm approaching my refactor of `ScatteringMatrix`, it feels preferable here to have a constructor which takes the same arguments, rather than an `initialise()` function. Then the call to `clear()` can be removed.

src/classes/potentialMap.cpp
```diff
@@ -60,6 +60,92 @@ bool PotentialMap::initialise(const std::vector<std::shared_ptr<AtomType>> &mast
...
+        }
+    }
+
+    // Store potential range
+    range_ = pairPotentialRange;
+
+    return true;
+}
+
+bool PotentialMap::initialise(const std::vector<const AtomType *> &masterAtomTypes,
```
trisyoungs: Rather than `masterAtomTypes` this can now be just `atomTypes`.

src/classes/potentialMap.cpp
```diff
@@ -60,6 +60,92 @@ bool PotentialMap::initialise(const std::vector<std::shared_ptr<AtomType>> &mast
...
+    potentialMatrix_.initialise(nTypes_, nTypes_);
+
+    dissolve::for_each_pair(
+        ParallelPolicies::seq, masterAtomTypes,
+        [&](int i, const auto &atI, int j, const auto &atJ)
+        {
+            if (i == -1)
+                return Messenger::error("Couldn't find AtomType '{}' in typeIndex.\n", atI->name());
+            if (j == -1)
+                return Messenger::error("Couldn't find AtomType '{}' in typeIndex.\n", atJ->name());
```
trisyoungs: Don't need to check the indices. They are always valid.
```suggestion
```

src/classes/potentialMap.cpp
```diff
@@ -60,6 +60,92 @@ bool PotentialMap::initialise(const std::vector<std::shared_ptr<AtomType>> &mast
...
+                Messenger::print("Linking self-interaction PairPotential for '{}' (index {},{} in matrix).\n", atI->name(), i,
+                                 j);
+                potentialMatrix_[{i, j}] = pp;
+            }
+            else
+            {
+                Messenger::print("Linking PairPotential between '{}' and '{}' (indices {},{} and {},{} in matrix).\n",
+                                 atI->name(), atJ->name(), i, j, j, i);
+                potentialMatrix_[{i, j}] = pp;
+                potentialMatrix_[{j, i}] = pp;
```
trisyoungs: Interesting - I don't remember storing the full matrix here (i.e. i,j and j,i) since they are very obviously pointing to the same pairpotential.  Still valid, however, as the code to lookup an element in a full array is simpler / faster, and this will be called a lot in the energy functions.

src/nodes/atomicMC/process.cpp
```diff
@@ -30,8 +31,11 @@ NodeConstants::ProcessResult AtomicMCNode::process()
     message("Target acceptance rate is {}.\n", targetAcceptanceRate);
     message("\n");
 
-    auto kernel =
-        KernelProducer::energyKernel(targetConfiguration_, dissolve().potentialMap(), dissolve().pairPotentialRange());
+    // Prepare for energy calculation, generate kernel
+    auto dissolveNode = dissolveGraph();
+    auto kernel = static_cast<DissolveGraph*>(dissolveNode)->prepareEnergyCalculation(targetConfiguration_);
+
+    targetConfiguration_->updateCells(kernel.get()->potentialMap().range());
```
trisyoungs: Another function that could be moved to `prepareEnergyCalculation()`.

src/nodes/node.cpp
```diff
@@ -302,6 +302,9 @@ Graph *Node::parentGraph() const { return parentGraph_; }
 // Return the Dissolve reference
 Dissolve &Node::dissolve() const { return parentGraph_->dissolve(); }
 
+// Return the DissolveGraph reference
+Node *Node::dissolveGraph() const { return parentGraph_->node("Dissolve"); }
```
```suggestion
Node *Node::dissolveGraph() const { return parentGraph_->dissolveGraph(); }
```

src/nodes/dissolve.h
```diff
@@ -6,6 +6,10 @@
 #include "main/dissolve.h"
 #include "nodes/edge.h"
 #include "nodes/graph.h"
+#include "templates/doubleKeyedMap.h"
+
+class EnergyKernel;
+class PotentialMap;
 
 // Main Dissolve Node
 class DissolveGraph : public Graph
```
trisyoungs, line 17: `DissolveGraph` needs to implement the `dissolveGraph()` function and return itself.

src/nodes/dissolve.cpp
```diff
@@ -20,3 +21,43 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+ * Functions
+ */
+
+// Return maximum distance for tabulated PairPotentials
+const double DissolveGraph::pairPotentialRange() const { return pairPotentialRange_; }
+
+// Return energy kernel containing potential map
+std::unique_ptr<EnergyKernel> DissolveGraph::prepareEnergyCalculation(Configuration *cfg, std::optional<double> energyCutoff)
+{
+	// Update atom type indexing
```
trisyoungs: Since we'll use it a few times, and it is calculated every time, it makes sense to grab this once and re-use it.
```suggestion
    auto atomTypes = cfg->atomTypeVector();

	// Update atom type indexing
```

src/nodes/dissolve.cpp
```diff
@@ -20,3 +21,43 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+	cfg->updateTypeIndexing();
+
+	// Generate configuration potential map
+    PotentialMap potentialMap;
+    potentialMap.initialise(cfg->atomTypeVector(), pairPotentialStore(), pairPotentialRange());
```
```suggestion
    potentialMap.initialise(atomTypes, pairPotentialStore(), pairPotentialRange());
```

src/nodes/dissolve.cpp
```diff
@@ -20,3 +21,43 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+
+// Return maximum distance for tabulated PairPotentials
+const double DissolveGraph::pairPotentialRange() const { return pairPotentialRange_; }
+
+// Return energy kernel containing potential map
+std::unique_ptr<EnergyKernel> DissolveGraph::prepareEnergyCalculation(Configuration *cfg, std::optional<double> energyCutoff)
+{
+	// Update atom type indexing
+	cfg->updateTypeIndexing();
+
```
trisyoungs, line 45: Here should be the double loop over `AtomType`s in the vector, updating and storing the `PairPotentials` (`updatePairPotential`).

src/nodes/dissolve.h
```diff
@@ -31,8 +35,25 @@ class DissolveGraph : public Graph
...
+    DoubleKeyedMap<PairPotential*> pairPotentialStore_{true};
+    // Pair potential range
+    double pairPotentialRange_{0.0};
 
     public:
     // Return dissolve
-    Dissolve &dissolve() const override;
+    Dissolve &dissolve() const;
+    // Return pair potential store
+    const DoubleKeyedMap<PairPotential*> &pairPotentialStore();
```
trisyoungs:
```suggestion
    const DoubleKeyedMap<PairPotential> &pairPotentialStore();
```

src/nodes/dissolve.h
```diff
@@ -31,8 +35,25 @@ class DissolveGraph : public Graph
     private:
     // Dissolve reference
     Dissolve &dissolve_;
+    // Pair potential store
+    DoubleKeyedMap<PairPotential*> pairPotentialStore_{true};
```
trisyoungs:
```suggestion
    DoubleKeyedMap<PairPotential> pairPotentialStore_{true};
```

src/nodes/dissolve.cpp
```diff
@@ -20,3 +21,43 @@ std::string_view DissolveGraph::summary() const { return "Parent node of all sim
...
+void DissolveGraph::updatePairPotentials(const AtomType &i, const AtomType &j)
+{
+    auto nameI = i.name(), nameJ = j.name();
+    if (pairPotentialStore_.contains(nameI, nameJ))
+        return;
+
+    auto interactionPotential = ShortRangeFunctions::combine(i.interactionPotential(), j.interactionPotential());
+    pairPotentialStore_.set(
+        nameI, nameJ,
+        interactionPotential.has_value() ? new PairPotential(nameI, nameJ, *interactionPotential) : new PairPotential(nameI, nameJ)
```
```suggestion
        interactionPotential.has_value() ? PairPotential(nameI, nameJ, *interactionPotential) : PairPotential(nameI, nameJ)
```

trisyoungs: Would be nice here to split out the logic from the ternary operator into a standard `if` so that we can push an error message if the parameters fail to combine.

src/nodes/dissolve.h
```diff
@@ -31,8 +35,25 @@ class DissolveGraph : public Graph
...
+    /*
+     * Functions
+     */
+    public:
+    // Return maximum distance for tabulated PairPotentials
+    const double pairPotentialRange() const;
+    // Return energy kernel containing potential map
+    std::unique_ptr<EnergyKernel> prepareEnergyCalculation(Configuration *cfg, std::optional<double> energyCutoff = {});
+    // Update pair potential store
+    void updatePairPotentials(const AtomType &i, const AtomType &j);
```
trisyoungs, line 66: Can be private.

src/nodes/dissolve.h
RobBuchananCompPhys: Will be mutable too, so the underlying virtual getter needs to be non-const

src/classes/partialSet.cpp
```diff
@@ -248,9 +284,13 @@ bool PartialSet::save(std::string_view prefix, std::string_view tag, std::string
...
 
-            parser.openOutput(filename, true);
+            auto cwd = std::filesystem::current_path();
+            auto path = cwd.parent_path().parent_path() / "tests" / "nodes" / "output" / filename;
+            auto fullPath = path.string();
+
+            parser.openOutput(fullPath, true);
             if (!parser.isFileGoodForWriting())
-                return Messenger::error("Couldn't open file '{}' for writing.\n", filename);
+                return Messenger::error("Couldn't open file '{}' for writing.\n", fullPath);
```
trisyoungs:
```suggestion
            parser.openOutput(fullPath, true);
            if (!parser.isFileGoodForWriting())
                return Messenger::error("Couldn't open file '{}' for writing.\n", filename);
```

src/classes/potentialMap.cpp
```diff
@@ -76,9 +153,12 @@ double PotentialMap::energy(const Atom &i, const Atom &j, double r) const
     // Check to see whether Coulomb terms should be calculated from atomic charges, rather than them being included in the
     // interpolated potential
     auto *pp = potentialMatrix_[{i.masterTypeIndex(), j.masterTypeIndex()}];
-    return pp->energy(r) + (PairPotential::includeCoulombPotential()
+    auto energy = pp->energy(r);
+    auto coulombEnergy = (PairPotential::includeCoulombPotential()
                                 ? 0
                                 : pp->analyticCoulombEnergy(i.speciesAtom()->charge() * j.speciesAtom()->charge(), r));
+
+    return energy + coulombEnergy;
```
trisyoungs: Curious as to why this got changed to use temporary vars?

src/io/import/data1D.cpp
```diff
@@ -64,8 +65,11 @@ bool Data1DImportFileFormat::importData(Data1D &data)
 {
     // Open file and check that we're OK to proceed importing from it
     LineParser parser;
-    if ((!parser.openInput(filename_)) || (!parser.isFileGoodForReading()))
-        return Messenger::error("Couldn't open file '{}' for loading Data1D data.\n", filename_);
+    auto cwd = std::filesystem::current_path();
+    auto path = cwd.parent_path().parent_path() / filename_;
```
trisyoungs: We should remove this kind of path-hacking in the code, but I understand why it's here.  Since the present example is created programmatically there is no context of a source input file and hence no reference directory. This may be another thing that we have to bring on to the table to discuss sooner rather than later - in #2196 I detailed my thoughts on moving to project folders rather than "bare input files in a random directory" that we have now.

src/math/history.h
```diff
@@ -16,7 +18,7 @@ template <class T> class History
 
     public:
     // Update history with supplied data and return current average
-    T average(const T &currentData, int averagingLength)
+    T average(const T &currentData, int averagingLength, std::optional<std::function<T()>> initialiser = {})
```
trisyoungs: You can do away with the `std::optional` and just have `std::function<T()> initialised = {}` I believe.  You can check whether a function was actually passed in with `if (initialiser)`

src/nodes/gr/gr.h
```diff
@@ -96,6 +96,12 @@ class GRNode : public Node
     bool testReferencePartial(const PartialSet &partials, double testThreshold, const Data1D &testData,
                               std::string_view typeIorTotal, std::string_view typeJ = "", std::string_view target = "");
 
+    private:
+    // Return value of unweighted GR, emplacing if optional not initialised
+    PartialSet &unweightedGR();
+    // Return value of raw GR, emplacing if optional not initialised
+    PartialSet &rawGR();
+
```
trisyoungs: I'm not convinced these operations need to be functionalised since they only ever have an effect at the beginning of the function on the first call, making subsequent calls unnecessary.

src/nodes/gr/helpers.cpp
```diff
@@ -412,35 +412,35 @@ bool GRNode::calculateRawGR(const double grRange, bool &alreadyUpToDate)
 // Calculate smoothed/broadened partial g(r) from supplied partials
 bool GRNode::calculateUnweightedGR()
 {
-    *unweightedGR_ = *rawGR_;
+    unweightedGR() = rawGR();
```
trisyoungs: Assigning the result of a function call to the result of a function call - this line makes me wonder what's going on behind the scenes!

src/nodes/md/process.cpp
```diff
@@ -24,28 +25,30 @@ NodeConstants::ProcessResult MDNode::process()
...
-        Messenger::print("MD: Summary will be written every {} step(s).\n", outputFrequency);
+        message("MD: Summary will be written every {} step(s).\n", outputFrequency);
     else
-        Messenger::print("MD: Summary will not be written.\n");
+        message("MD: Summary will not be written.\n");
     if (!restrictToSpecies_.empty())
-        Messenger::print("MD: Calculation will be restricted to species: {}\n",
-                         joinStrings(restrictToSpecies_, "  ", [](const auto &sp) { return sp->name(); }));
-    Messenger::print("\n");
+        message("MD: Calculation will be restricted to species: {}\n",
```
trisyoungs:
```suggestion
    message("Number of steps = {}\n", nSteps);
    message("Timestep type is '{}'\n", timestepType().keyword(timestepType_));
if (onlyWhenEnergyStable_)
        message("Only perform MD if target Configuration energies are stable.\n");
if (trajectoryFrequency > 0)
        message("Trajectory file will be appended every {} step(s).\n", trajectoryFrequency);
else
        message("Trajectory file off.\n");
if (capForces_)
        message("Forces will be capped to {:10.3e} kJ/mol per atom per axis.\n", maxForce / 100.0);
if (energyFrequency > 0)
        message("Energy will be calculated every {} step(s).\n", energyFrequency);
else
        message("Energy will be not be calculated.\n");
if (outputFrequency > 0)
        message("Summary will be written every {} step(s).\n", outputFrequency);
else
        message("Summary will not be written.\n");
if (!restrictToSpecies_.empty())
        message("Calculation will be restricted to species: {}\n",
```

src/nodes/neutronSQ/helpers.cpp
```diff
@@ -60,30 +60,31 @@ bool NeutronSQNode::calculateWeightedSQ()
                                 auto boundWeight = weights_.intramolecularWeights().get(key);
 
                                 // Bound (intramolecular) partial (multiplied by the bound term weight)
-                                weightedSQ_->boundPartials().get(key).copyArrays(unweightedSQ_->boundPartials().get(key));
-                                weightedSQ_->boundPartials().get(key) *= boundWeight;
+                                weightedSQ().boundPartials().get(key).copyArrays(unweightedSQ_->boundPartials().get(key));
+                                weightedSQ().boundPartials().get(key) *= boundWeight;
```
trisyoungs: Again, this feels like something hidden is happening as we're not referencing a variable...

src/nodes/sq/sq.h
```diff
@@ -49,7 +52,7 @@ class SQNode : public Node
     // Minimum Q for calculated S(Q)
     Number qMin_{0.01};
     // Whether to save partials to disk after calculation
-    bool save_{false};
+    bool save_{true};
```
trisyoungs:
```suggestion
    bool save_{false};
```

src/nodes/dissolve.h
```diff
@@ -6,6 +6,10 @@
 #include "main/dissolve.h"
 #include "nodes/edge.h"
 #include "nodes/graph.h"
+#include "templates/doubleKeyedMap.h"
+
+class EnergyKernel;
```
trisyoungs, line 13:
```suggestion
// Forward Declarations
class EnergyKernel;
```

src/nodes/node.cpp
```diff
@@ -302,6 +302,9 @@ Graph *Node::parentGraph() const { return parentGraph_; }
 // Return the Dissolve reference
 Dissolve &Node::dissolve() const { return parentGraph_->dissolve(); }
 
+// Return the DissolveGraph reference
+DissolveGraph *Node::dissolveGraph() { return static_cast<DissolveGraph *>(parentGraph_->dissolveGraph()); }
```
trisyoungs: This shouldn't need a cast since the function being called already returns a pointer of the correct type, no?

.gitignore
```diff
@@ -39,6 +39,7 @@ changes
 dependencies
 msvc-env
 aqtinstall.log
+tests/nodes/output
```
trisyoungs: Very specific path to exclude?

src/classes/potentialMap.cpp
RobBuchananCompPhys: Ah, this was temporarily changed for debugging ease, it will be reverted

.gitignore
RobBuchananCompPhys: I was thinking we could place output from the graph unit tests into this folder during development, without commiting them to the remote.

src/math/history.h
```diff
@@ -26,7 +28,14 @@ template <class T> class History
...
         // Perform averaging of the datasets that we have
-        T averaged;
+        T averaged = [&]()
+        {
+            if (initialiser)
+                return initialiser();
+            else
+                return T();
+        }();
+
```
trisyoungs: Not sure why this is written inside a lambda?

```suggestion
        T averaged = initialiser ? initialiser() : T();

```

trisyoungs (approved):
One final suggested change, then I think we're good to go!

PR #2216, opened 2025-07-30
refactor: Separate histogram calculation from PartialSet

trisyoungs:
This PR splits the histograms for radial distribution function calculation out of `PartialSet` into their own `HistogramSet` since this represented a very specific set of data contained in a quite general-use class otherwise.

Required by #2203.

### DIFF
--- a/src/classes/histogramSet.cpp
+++ b/src/classes/histogramSet.cpp
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#include "classes/histogramSet.h"
+#include "base/lineParser.h"
+#include "classes/atomType.h"
+#include "items/deserialisers.h"
+#include "math/mathFunc.h"
+#include "templates/algorithms.h"
+
+HistogramSet::~HistogramSet()
+{
+    fullHistograms_.clear();
+    boundHistograms_.clear();
+    unboundHistograms_.clear();
+}
+
+/*
+ * Data
+ */
+
+// Initialise histograms
+void HistogramSet::initialise(const AtomTypeMix &atomTypeMix, double rdfRange, double binWidth)
+{
+    auto nTypes = atomTypeMix_.nItems();
+
+    fullHistograms_.initialise(nTypes, nTypes, half_);
+    boundHistograms_.initialise(nTypes, nTypes, half_);
+    unboundHistograms_.initialise(nTypes, nTypes, half_);
+
+    dissolve::for_each_pair(
+        ParallelPolicies::par, nTypes,
+        [&](int i, int j)
+        {
+            fullHistograms_[{i, j}].initialise(0.0, rdfRange, binWidth);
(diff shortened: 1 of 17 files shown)


src/classes/histogramSet.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+// Zero histogram bins
+void HistogramSet::zeroBins()
+{
+    for (auto &histo : fullHistograms_.linearArray())
+        histo.zeroBins();
+    for (auto &histo : boundHistograms_.linearArray())
+        histo.zeroBins();
+    for (auto &histo : unboundHistograms_.linearArray())
+        histo.zeroBins();
```
RobBuchananCompPhys, line 60: What do you think of a static method in `Array2D`, `flatten`, that takes a couple of `Array2D<T>`'s (like the above `full/bound/unboundHistograms_`) and flattens their `linearArray()`s into a single vector?
```suggestion
   for (auto &histo : Array2D::flatten({fullHistograms_, boundHistograms_, unboundHistograms_}))
        histo.zeroBins();
```

I believe the helper method could be implemented like:
```
static std::vector<A> flatten(const std::vector<Array2D<A>> arrays)
{
    std::vector<A> flat;
    for (const auto &a : arrays)
          flat.insert(flat.end(), a.begin(), a.end());
    return flat;
} 
```

src/classes/histogramSet.h
```diff
@@ -0,0 +1,64 @@
...
+    // Set up histograms
+    void initialise(const AtomTypeMix &atomTypeMix, double rdfRange, double binWidth);
+    // Clear all histogram data
+    void clear();
+    // Zero histogram bins
+    void zeroBins();
+    // Return atom types mis
+    const AtomTypeMix &atomTypeMix() const;
+    // Set new fingerprint
+    void setFingerprint(std::string_view fingerprint);
```
RobBuchananCompPhys, line 45: Might be something to make a record of in [TODO](https://github.com/disorderedmaterials/dissolve/issues/2152) since I think in a discussion recently the usefulness of `fingerprint` was brought into question.

RobBuchananCompPhys (approved):
Looks good! Just a couple of comments including a suggestion that may/may not be worth implementing.

src/classes/histogramSet.cpp
trisyoungs: I see what you mean - in fact Adam's `zip()` operator would allow us to do this in one loop, albeit with three variables.  Your suggestion makes a copy of the data, so unfortunately wouldn't have the intended result. I think I will leave this as-is for now, but your suggestion makes me think that some kind of function which takes a lambda and operates on every element in the `Array2D` would be useful to implement in the future (e.g. `fullHistograms_.operate([](auto &histo) { histo.zeroBins(); })`)

src/classes/histogramSet.h
trisyoungs: I think fingerprints will disappear in one of my upcoming PRs, but I will certainly capture it in the 2Do to check later.

PR #2217, opened 2025-08-06
build:Upgrade clang-format version

rprospero:
This PR accomplishes the following chores

- [x] Upgrade the nixpkgs version to 25.05 (from 24.05)
- [x] Remove the outdated container builder
- [x] Upgrade clang-format to version 20.1.6
- [x] Reformat source to match new clang format

As a quick reminder on the container builder, there was a bug in singularity that caused it to fail on the CI.  We had worked around it by using an old version of singularity from before when the bug was created.  However, I am now checking whether the bug has been cleared and we can use a more recent singularity version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ stamp-h1
 ui_*
 moc_*
 *.kdev4
+/.cache/
 build/
 build-*/
 *.exe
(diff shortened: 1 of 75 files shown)


trisyoungs (approved):
👍

PR #2220, opened 2025-08-13
refactor: Remove Configuration AtomTypeMix

trisyoungs:
## Follows #2214

This PR removes the `AtomTypeMix` from `Configuration`, moving towards a dynamic calculation of atom type populations etc. from stored `Species` populations. The effects are quite wide-ranging, so this is the only thing addressed here.


src/modules/neutronSQ/functions.cpp
```diff
@@ -53,13 +54,14 @@ bool NeutronSQModule::calculateWeightedGR(const PartialSet &unweightedgr, Partia
 bool NeutronSQModule::calculateWeightedSQ(const PartialSet &unweightedsq, PartialSet &weightedsq, NeutronWeights &weights,
                                           StructureFactors::NormalisationType normalisation)
 {
-    dissolve::for_each_pair(ParallelPolicies::seq, unweightedsq.atomTypeMix(),
-                            [&](int typeI, const AtomTypeData &atd1, int typeJ, const AtomTypeData &atd2)
+    dissolve::for_each_pair(ParallelPolicies::seq, unweightedsq.atomTypeFractions(),
+                            [&](int indexI, const auto &popI, int indexJ, const auto &popJ)
                             {
-                                auto key = DoubleKeyedMapKey{atd1.atomTypeName(), atd2.atomTypeName()};
+                                auto key = DoubleKeyedMapKey{popI.first->name(), popJ.first->name()};
```
RobBuchananCompPhys, line 60: On one hand I do quite like the idea of self-contained class we can instantiate to store these more abstract keys - however, I worry it is a bit onerous to fully initialise a key in this way. Is it not possible to overload an operator on `DoubleKeyedMap` to handle it all-in-one?

src/modules/sq/process.cpp
```diff
@@ -172,28 +160,29 @@ Module::ExecutionResult SQModule::process(Dissolve &dissolve)
...
+                                [&](auto indexI, auto &popI, int indexJ, auto &popJ)
                                 {
                                     // Note: Intramolecular broadening will not be applied to bound terms within the
                                     // calculated Bragg scattering
-                                    auto key = DoubleKeyedMapKey{atd1.atomTypeName(), atd2.atomTypeName()};
+                                    auto key = DoubleKeyedMapKey{popI.first->name(), popJ.first->name()};
+
                                     auto &bound = unweightedsq.boundPartials().get(key);
                                     auto &unbound = unweightedsq.unboundPartials().get(key);
                                     auto &partial = unweightedsq.partials().get(key);
```
RobBuchananCompPhys, line 184: Further to the previous comment, is:
`auto &bound = unweightedsq.boundPartials[popI.first->name(), popJ.first->name()]` where `& operator[]`
not possible?

RobBuchananCompPhys (approved):
Looks good 👍

src/modules/sq/process.cpp
trisyoungs: The answer is "Yes, but...".  You could absolutely have an `operator[]` which takes a `std::pair` of the names and does the lookup (you can't have multiple args to `operator[]` so you have to wrap them in something like a `std::pair`). The "but" is that  we typically only want to use it as a `const` getter, while `std::map`'s `operator[]` is fully non-`const` as it creates new values if the key is not present. So, in the interest of maintaining "semantic compatibility" between `DoubleKeyedMap` and `std::map`, I elected to implement `get()` and `at()` instead.

PR #2221, opened 2025-08-13
refactor: Reinvent AtomTypeMix into IsotopeMix

trisyoungs:
## Follows #2219

This PR focuses on the `AtomTypeMix` and associated classes `AtomTypeData` and `IsotopeData` and simplify / combining them into a single class `IsotopeMix` providing all functionality. `IsotopeMix` stores a lot less data than `AtomTypeMix` at the expense of having to calculate values when required, but since these values are not needed in hot loops IMO this is a reasonable compromise.

I have tried to minimised impact in other classes, but `XRayWeights` warranted a partial refactor here. Also, the unit test associated to `AtomTypeMix` is removed as functionality of `IsotopeMix` is well-covered in tests of the `NeutronWeights` class (the principal user of `IsotopeMix`).


RobBuchananCompPhys (commented):
Review in progress! Will try and get to the end after lunch.

RobBuchananCompPhys (approved):

PR #2222, opened 2025-08-15
refactor: Convert NeutronWeights to use keying

trisyoungs:
## Follows #2221

This PR focuses on the `NeutronWeights` class, moving it to use `DoubleKeyedMap`s for data storage rather than `Array2D`s.

### DIFF
--- a/src/modules/neutronSQ/process.cpp
+++ b/src/modules/neutronSQ/process.cpp
@@ -195,11 +195,8 @@ Module::ExecutionResult NeutronSQModule::process(Dissolve &dissolve)
     if (wSQstatus == GenericItem::ItemStatus::Created)
         weightedSQ.initialise(unweightedSQ);
 
-    printf("HERE3\n");
-
     // Calculate weighted S(Q)
     calculateWeightedSQ(unweightedSQ, weightedSQ, weights, normaliseTo_);
-    printf("HERE4\n");
 
     // Save data if requested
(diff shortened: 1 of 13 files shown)


src/modules/neutronSQ/process.cpp
```diff
@@ -195,8 +195,11 @@ Module::ExecutionResult NeutronSQModule::process(Dissolve &dissolve)
     if (wSQstatus == GenericItem::ItemStatus::Created)
         weightedSQ.initialise(unweightedSQ);
 
+    printf("HERE3\n");
```
RobBuchananCompPhys: Do we want this print statement to go in?

src/modules/neutronSQ/process.cpp
```diff
@@ -195,8 +195,11 @@ Module::ExecutionResult NeutronSQModule::process(Dissolve &dissolve)
...
+
     // Calculate weighted S(Q)
     calculateWeightedSQ(unweightedSQ, weightedSQ, weights, normaliseTo_);
+    printf("HERE4\n");
```
RobBuchananCompPhys: Same here.

RobBuchananCompPhys (approved):

src/modules/neutronSQ/process.cpp
trisyoungs: OG debugging right there...

PR #2224, opened 2025-08-15
refactor: Convert XRayWeights to use keying

trisyoungs:
## Follows #2222

This PR moves `XRayWeights` to use `DoubleKeyedMap`, but retains some `AtomType`-specific accessors because of the nature of the class.  Since we never (de)serialise `XRayWeights`, we can do this safely.

### DIFF
--- a/src/classes/xRayWeights.cpp
+++ b/src/classes/xRayWeights.cpp
@@ -148,7 +148,7 @@ double XRayWeights::boundCoherentAverageOfSquares(double Q) const
                            [&, Q](auto acc, const auto &typePop)
                            {
                                auto mag = formFactorData_.at(typePop.first).get().magnitude(Q);
-                               return acc + typePop.second * mag * mag;
+                               return acc + typePop.second * pow(mag, 2);
                            });
 }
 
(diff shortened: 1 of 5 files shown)


src/classes/xRayWeights.cpp
```diff
@@ -161,8 +144,12 @@ std::vector<double> XRayWeights::boundCoherentSquareOfAverage(const std::vector<
 // Calculate and return Q-dependent squared average scattering (<b**2>) for supplied Q value
 double XRayWeights::boundCoherentAverageOfSquares(double Q) const
 {
-    return std::inner_product(concentrations_.begin(), concentrations_.end(), formFactorData_.begin(), 0.0, std::plus<>(),
-                              [Q](auto con, auto form) { return con * form.get().magnitude(Q) * form.get().magnitude(Q); });
+    return std::accumulate(typeFractions_.begin(), typeFractions_.end(), 0.0,
+                           [&, Q](auto acc, const auto &typePop)
+                           {
+                               auto mag = formFactorData_.at(typePop.first).get().magnitude(Q);
+                               return acc + typePop.second * mag * mag;
```
RobBuchananCompPhys:
```suggestion
                               return acc + typePop.second * pow(mag, 2);
```

RobBuchananCompPhys (changes requested):

RobBuchananCompPhys (approved):

PR #2226, opened 2025-08-22
feat:Inner graphs

rprospero:
When a node contains an inner graph, the user can press a button in the node to descend into the inner graph.  There is a location bar at the top of the window that states what node the user is currently in with an `up` button to return back to the parent graph.

I'm submitting this with the following known issue:  when you leave a graph, all of the nodes lose their positions.  This is because the positions are only stored in the graph model, which only looks at the current node.  It will be a larger change to start saving this information, but I thought that I should make this available before I go on leave next week.

This PR also includes a change to the Mac build on github actions.  GitHub has upgraded to XCode 17, but XCode 16 is the **last** version supported by Conan 1.  I've pinned XCode to version 16, but we need to migrate to Conan 2

### DIFF
--- a/.github/workflows/build/osx/action.yml
+++ b/.github/workflows/build/osx/action.yml
@@ -29,6 +29,11 @@ runs:
   # Setup / Install Dependencies
   #
 
+  - name: Choose XCode version
+    shell: bash
+    run: |
+      sudo xcode-select -s /Applications/Xcode_16.2.app
+
   - name: Manage Homebrew Dependencies
     shell: bash
     run: |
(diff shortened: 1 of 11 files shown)


trisyoungs (approved):
Tried and tested - I think this mechanism will work well.

I may also take back what I said this morning about node positioning not being important - it will become a proper mess in the UI once we have a few nodes populated, so perhaps making the positioning persistent is more of a priority than I suggested! Not ultra-urgent, but probably near the top of the list when we come to getting the unit tests ported over.

PR #2230, opened 2025-09-04
refactor: Move ScatteringMatrix to use keyed mappings where required

trisyoungs:
This PR moves `ScatteringMatrix` over to use keyed data storage where appropriate, in line with recent work on other classes. While this does not necessarily help the mathematical representation of the matrix, it does make data tracking a lot clearer. None of the functions are called in hot loops during running, so I think this is a good compromise.

One thing I'd like an opinion on from everybody is the new requirement to loop over a linear index in a pairwise loop - e.g. I use `for_each_pair` on a vector of `AtomType`s in order to generate the columns of a 2D matrix, but then later need to reference the column by its linear index:
```
        // Loop over columns (atom-atom partials)
        auto partialIndex = 0;
        dissolve::for_each_pair(ParallelPolicies::seq, atomTypes_,
                                [&](int indexI, const auto &typeI, int indexJ, const auto &typeJ)
                                {
                                    ... interesting code ...
                                    ++partialIndex;
                                });
```
Obviously this absolutely has to be `ParallelPolicies::seq`, but it still feels a little dangerous.

### DIFF
--- a/src/classes/scatteringMatrix.cpp
+++ b/src/classes/scatteringMatrix.cpp
@@ -37,14 +37,8 @@ Array2D<double> ScatteringMatrix::A() const
     return result;
 }
 
-// Return number of atom types involved
-int ScatteringMatrix::nAtomTypes() const { return atomTypes_.size(); }
-
 // Return atom types
 const std::vector<const AtomType *> &ScatteringMatrix::atomTypes() const { return atomTypes_; }
-
-// Return atom type at index specified
-const AtomType *ScatteringMatrix::atomType(int index) const { return atomTypes_[index]; }
 
 // Generate matrices
 void ScatteringMatrix::generateMatrices(const std::vector<double> &qValues)
--- a/src/classes/scatteringMatrix.cpp
+++ b/src/classes/scatteringMatrix.cpp
@@ -11,12 +11,7 @@
 #include <algorithm>
 #include <utility>
 
-ScatteringMatrix::ScatteringMatrix(const std::vector<const AtomType *> &atomTypes)
-{
-    atomTypes_ = atomTypes;
-    rows_.clear();
-    matricesValid_ = false;
-}
+ScatteringMatrix::ScatteringMatrix(const std::vector<const AtomType *> &atomTypes) : atomTypes_(atomTypes) { rows_.clear(); }
 
 /*
  * Data
@@ -53,10 +48,13 @@ const std::vector<const AtomType *> &ScatteringMatrix::atomTypes() const { retur
 const AtomType *ScatteringMatrix::atomType(int index) const { return atomTypes_[index]; }
 
 // Generate matrices
-void ScatteringMatrix::generateMatrices()
+void ScatteringMatrix::generateMatrices(const std::vector<double> &qValues)
 {
-    if (matricesValid_)
+    // If the supplied and current qValue vectors are identical, nothing to do...
+    if (qValues_ == qValues)
         return;
+
+    qValues_ = qValues;
 
     // We always generate the matrices for Q = 0
     Messenger::printVerbose("Generating Q = 0.0 matrix and inverse.\n");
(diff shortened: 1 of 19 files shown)


src/classes/scatteringMatrix.cpp
```diff
@@ -18,51 +19,33 @@
...
-{
-    return {indexOf(typeI), indexOf(typeJ)};
-}
-
-// Return column index of specified AtomType pair
-int ScatteringMatrix::columnIndex(const AtomType *typeI, const AtomType *typeJ) const
-{
-    auto index = 0;
-    for (auto [i, j] : typePairs_)
+    Array2D<double> result(rows_.size(), (atomTypes_.size() * (atomTypes_.size() + 1)) / 2, false);
```
RobBuchananCompPhys: Could we make a private getter for `(atomTypes_.size() * (atomTypes_.size() + 1)) / 2`? It seems to crop up every so often in this class, but quite long winded to duplicate the expression everytime. Perhaps it could be implemented as
```
double ScatteringMatrix::rowLimit()
{
    return (atomTypes_.size() * (atomTypes_.size() + 1)) / 2;
}

File: src/classes/scatteringMatrix.cpp [hunk 2]
```diff
@@ -249,14 +234,14 @@ void ScatteringMatrix::printInverse(double q) const
...
+// Generate estimated partials from reference data using the inverse coefficients matrix
+DoubleKeyedMap<Data1D> ScatteringMatrix::generateEstimatedPartials() const
 {
     // Check that we have the correct number of reference data to be able to invert the matrix
-    if (data_.size() < A_.nColumns())
+    if (rows_.size() < (atomTypes_.size() * (atomTypes_.size() + 1)) / 2)
         return Messenger::error("Can't finalise this scattering matrix, since there are not enough reference data ({}) "
                                 "compared to rows in the matrix ({}).\n",
-                                data_.size(), A_.nColumns());
+                                rows_.size(), (atomTypes_.size() * (atomTypes_.size() + 1)) / 2);
```
RobBuchananCompPhys: As previously mentioned, a getter for this value could be interesting to have.

src/templates/doubleKeyedMap.h
```diff
@@ -94,24 +94,8 @@ template <typename ValueClass> class DoubleKeyedMap
...
-    {
-        // Need to split key back into its constituent parts and find()
-        auto keys = DissolveSys::splitString(key, separator_);
-        if (keys.size() != 2)
-            throw(std::runtime_error(std::format("Invalid key '{}' given to DoubleKeyedMap::at()\n", key)));
-
-        return get(keys[0], keys[1]);
-    }
+    ValueClass &get(const std::string_view key) { return get(keyPair(key)); }
+    const ValueClass &get(const std::string_view key) const { return get(keyPair(key)); }
```
rprospero, line 98: If nothing else, I love how much simpler this is.

src/classes/scatteringMatrix.cpp
```diff
@@ -98,34 +76,34 @@ const Array2D<double> &ScatteringMatrix::qZeroMatrixInverse() const { return qZe
...
+            auto normFactor = 1.0;
+            if (rowData.xRayNormalisation == StructureFactors::AverageOfSquaresNormalisation)
+                normFactor = xRayWeights.boundCoherentAverageOfSquares(q);
+            else if (rowData.xRayNormalisation == StructureFactors::SquareOfAverageNormalisation)
+                normFactor = xRayWeights.boundCoherentSquareOfAverage(q);
+
+            // Loop over columns and get weights for columns according to the elements of the atom type pairs
+            std::vector<double> weights;
+            dissolve::for_each_pair(ParallelPolicies::seq, atomTypes_, [&](int i, auto &atI, int j, auto &atJ)
+                                    { weights.push_back(xRayWeights.formFactorProduct(atI, atJ, q) / normFactor); });
```
rprospero, line 100: If this ever becomes a performance issue, I might know a way to do this in parallel.  C++20 unlocked some new iterator stuff that might eventually lead to replacing `for_each_pair` with a more composable method that's easier to run in parallel, but this may not be the right time and place for it.

rprospero (approved):

src/classes/scatteringMatrix.cpp
trisyoungs: Yes, we could. This is (and I don't know a better way to call it) the number of unique elements of a square half-matrix (i.e. where i,j == j,i), so should be implemented as a general math function.

RobBuchananCompPhys (approved):
Namespace math function looks good, seems sensible if we are using the equation across multiple classes.

PR #2233, opened 2025-09-11
refactor: SpeciesParticle base class

BTSpeake:
Add a base class for an abstract particle from which SpeciesAtom will inherit. This will provide the base for adding in a SpeciesBead class for coarse grained workflows utilising several common properties which facilitate position/force evaluation. All SpeciesIntra classes have also been made to work with the new SpeciesParticle base class requiring dynamic casting in places where retrieving a SpeciesAtom object from a bond is necessary.

### DIFF
--- a/src/classes/CMakeLists.txt
+++ b/src/classes/CMakeLists.txt
@@ -59,6 +59,7 @@ add_library(
   speciesAngle.cpp
   speciesAtom.cpp
   speciesBond.cpp
+  speciesParticle.cpp
   species.cpp
   species_atomic.cpp
   species_ff.cpp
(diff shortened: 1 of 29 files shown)


trisyoungs (commented):
First off, this is all excellent and exactly what I had in mind.

Second, however, seeing the effects of these changes propagate through into other bits of the code makes me consider whether this is entirely the right approach!  In particular I don't like the need for all the `dynamic_cast`s everywhere.

These changes had to be attempted to see what trajectory it would put us on, but I think I need to assess what we're doing a little bit before moving on.

PR #2237, opened 2025-09-15
chore: Upgrade develop branch to conan 2

rprospero:
This PR upgrades the develop branch to Conan 2.  This should fix the issues with MacOS and the cmake version.

This had the following side-effects

1. To get the latest conan, we had to upgrade our nixpkg version (matching develop2)
2. Step 1 necessitated upgrading clang-format
3. Step 2 necessitated reformatting the all the c++ source.  This is all contained in the last commit, which does not particularly need review.
4. Some dependencies had their version bumped, as the old versions required ancient CMake version.
5. I moved to the Python based `conanfile` to prevent the dev dependencies (ie. gtest and benchmark) from being built on Mac or Windows.  This both speeds up builds on those platforms and skips a weird issue with gtest failing to build on Windows


develop.ps1
```diff
@@ -420,9 +459,12 @@ try {
...
 }
 
-& $conan profile new default --detect
-& $conan profile update settings.compiler="Visual Studio" default
-& $conan profile update settings.compiler.version=17 default
+if  (-not ($conanVersion -eq 2))
+{
+    & $conan profile detect
+    # & $conan profile update settings.compiler="Visual Studio" default
+    # & $conan profile update settings.compiler.version=17 default
```
RobBuchananCompPhys:
```suggestion

```

I think we could probably remove these lines

RobBuchananCompPhys (changes requested):
Just need to remove some unused (commented out) lines from `develop.ps1` - otherwise, looks great.

PR #2238, opened 2025-09-16
fix: Consistent object naming

trisyoungs:
This PR forces the use of "nice names" (i.e. with spaces and other annoying characters removed) in `Configuration`s, `Species`, `SpeciesSite`s, and `Isotopologues`. This came about from a bug report where a species containing a space in its name (which was previously allowed) was written without quotes by `SpeciesVectorKeyword` / `SpeciesKeyword` and subsequently broke the input file on re-reading.

Here, using `setName()` on any of the above objects calls `DissolveSys::niceName()` automatically - this was already done for `Configuration`, but the original unchanged `name_` was also stored - this is now removed.  To prevent these changes from breaking existing input files, I/O routines have been modified to convert any read-in name to its nice equivalent before searching for the object. I have left the TOML-related I/O largely untouched since it is not the prinicipal I/O stream at present.

### DIFF
--- a/src/classes/configuration.cpp
+++ b/src/classes/configuration.cpp
@@ -40,19 +40,10 @@ void Configuration::clear()
  */
 
 // Set name of the Configuration
-void Configuration::setName(std::string_view name)
-{
-    name_ = name;
-
-    // Generate a nice name (i.e. no spaces, slashes etc.)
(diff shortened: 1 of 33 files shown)


rprospero (approved):
Looks good.  I'm hoping that the TOML won't have these issues, but that requires a *full* move to TOML, including the restart file.

PR #2239, opened 2025-09-22
fix: Save node coordinates

rprospero:
This PR performs two major updates.  First, it performs some long ignored fixes to the TOML serialisation and deserialisation to bring them back in line with modern development.  Second, it adds the ability to save files from the new QML GUI and load them back in again.  This should hopefully ease the process of making some example files for testing larger structures.


trisyoungs (approved):
Yep - looks good to me!

PR #2242, opened 2025-09-29
feat: Node data serialisation

trisyoungs:
This PR implements serialisation of restart file data for the current complement of nodes, building off #2203.

## TODO
- [x] Reading in a serialisable based on `std::optional` needs to emplace the data first.
- [x] ~Template args to `is_instance_of<>` are reversed c.f. comparable stdlib functions~ Moved to #2260
- [x] DoubleKeyedMap serialisation test
- [x] Histogram1D serialisation test (inc `long`s for bin counts)
- [x] SampledDouble, SampledVector, and SampledData1D serialisation tests.
- [x] `History` serialisation test, especially with custom initialiser function.
- [x] ~Move simple test functions out of `DissolveSystemTest`~ Moved to Dissolve2Do


src/base/serialiser.h
```diff
@@ -159,14 +168,15 @@ template <typename... Contexts> class Serialisable
...
-
     // A helper function to add the elements of a map to a node under a name
     template <typename K, typename V> static void fromMap(const std::map<K, V> &map, std::string name, SerialisedValue &node)
     {
         SerialisedValue result;
         for (auto &[key, value] : map)
             if constexpr (serialisablePointer<V>)
                 result[std::string(key)] = value->serialise();
+            else if constexpr (std::is_base_of_v<Serialisable, V>)
+                result[std::string(key)] = value.serialise();
```
rprospero, line 179: I stole these lines to clean up some bits in my TOML PR as well.

src/templates/doubleKeyedMap.h
```diff
@@ -156,4 +162,33 @@ template <typename ValueClass> class DoubleKeyedMap
...
+        {
+            if constexpr (std::is_same_v<ValueClass, double>)
+                data_[mapKey] = value.as_floating();
+            else if constexpr (std::is_same_v<ValueClass, int>)
+                data_[mapKey] = value.as_integer();
+            else if constexpr (std::is_same_v<ValueClass, bool>)
+                data_[mapKey] = value.as_boolean();
+            else
+                data_[mapKey].deserialise(value);
+        }
```
rprospero, line 192: We could use `toMap` here for symmetry, but I'm not sure that we get that much out of it.

```suggestion
        toMap(node["map"], [this](const auto &mapKey, const auto &value)
        {
            if constexpr (std::is_same_v<ValueClass, double>)
                data_[mapKey] = value.as_floating();
            else if constexpr (std::is_same_v<ValueClass, int>)
                data_[mapKey] = value.as_integer();
            else if constexpr (std::is_same_v<ValueClass, bool>)
                data_[mapKey] = value.as_boolean();
            else
                data_[mapKey].deserialise(value);
        });
```

rprospero (approved):
One minor suggestion, but things otherwise look good.

src/templates/doubleKeyedMap.h
trisyoungs: I was all for this change as it would remove the ugly `toml::find` call, but alas because `toMap()` takes a `const` node pointer, using the nice map-style `node["map"]` lookup isn't possible.  So, leaving as-is.

PR #2244, opened 2025-09-30
doc: Update compilation instructions.

trisyoungs:
Compilation instructions were out-of-date, so here they are updated and clarified.


RobBuchananCompPhys (approved):
Good to see it looking up to date!

PR #2246, opened 2025-10-01
build: Don't downgrade MSVC

trisyoungs:


RobBuchananCompPhys:
Looks good, hope we won't need it again!

PR #2247, opened 2025-10-02
feat: Data import nodes

RobBuchananCompPhys:


src/nodes/data1DImport.h
```diff
@@ -0,0 +1,54 @@
...
+    // Column index of data Y values
+    Number yColumn_{2};
+    // Column index of data error values
+    Number errorColumn_{0};
+    // Remove average from X
+    std::optional<Number> removeAverageFromX_{0.0};
+    // Minimum X value
+    std::optional<Number> xMin_{0.0};
+    // Maximum X value
+    std::optional<Number> xMax_{0.0};
```
trisyoungs: Do you really want to supply default numbers to these optionals?

src/nodes/data1DImport.h
```diff
@@ -0,0 +1,54 @@
...
+    // Index of Error column
+    Number nPointsToRemove_{0};
+    // Imported data
+    Data1D data_{};
```
trisyoungs:
```suggestion
    Data1D data_;
```

src/nodes/data1DImport.h
```diff
@@ -0,0 +1,54 @@
...
+    Number yColumn_{2};
+    // Column index of data error values
+    Number errorColumn_{0};
+    // Remove average from X
+    std::optional<Number> removeAverageFromX_{0.0};
+    // Minimum X value
+    std::optional<Number> xMin_{0.0};
+    // Maximum X value
+    std::optional<Number> xMax_{0.0};
+    // Index of Error column
```
trisyoungs:
```suggestion
    // Number of points to remove from beginning of data
```

src/nodes/data1DImport.cpp
```diff
@@ -0,0 +1,38 @@
...
+std::string_view Data1DImportNode::type() const { return "Data1DImport"; }
+
+std::string_view Data1DImportNode::summary() const { return "Import 1D data."; }
+
+NodeConstants::ProcessResult Data1DImportNode::process()
+{
+    auto removeAverageFromX = removeAverageFromX_.value().asDouble(), xMin = xMin_.value().asDouble(),
+         xMax = xMax_.value().asDouble();
+    auto xColumn = xColumn_.asInteger(), yColumn = yColumn_.asInteger(), errorColumn = errorColumn_.asInteger(),
+         nPointsToRemove = nPointsToRemove_.asInteger();
```
trisyoungs: None of these variable retrievals should be necessary as you're just passing them into your new constructor.  Also, by assuming default values (in the header initialisations) you're changing the behaviour of the data processing in `Data1DImportFileFormat` which checks for all the optional values itself.

src/io/import/data1D.h
```diff
@@ -24,6 +24,8 @@ class Data1DImportFileFormat : public FileAndFormat
 
     Data1DImportFileFormat(std::string_view filename = "", Data1DImportFormat format = Data1DImportFormat::XY, int xColumn = 1,
                            int yColumn = 2, int errorColumn = 0);
+    Data1DImportFileFormat(std::string_view filename, Data1DImportFileFormat::Data1DImportFormat format, std::optional<double> removeAverageFromX, std::optional<double> xMin, std::optional<double> xMax,
+                           int nPointsToRemove, int xColumn = 1, int yColumn = 2, int errorColumn = 0);
```
trisyoungs: Your new constructor might as well demand all the parameters at once.
```suggestion
                           int nPointsToRemove, int xColumn, int yColumn, int errorColumn);
```

src/io/import/data1D.h
```diff
@@ -24,6 +24,8 @@ class Data1DImportFileFormat : public FileAndFormat
 
     Data1DImportFileFormat(std::string_view filename = "", Data1DImportFormat format = Data1DImportFormat::XY, int xColumn = 1,
                            int yColumn = 2, int errorColumn = 0);
+    Data1DImportFileFormat(std::string_view filename, Data1DImportFileFormat::Data1DImportFormat format, std::optional<double> removeAverageFromX, std::optional<double> xMin, std::optional<double> xMax,
```
trisyoungs: For consistency with the original constructor I would reorder the common parameters to match, and have all the optional stuff afterwards. If you make all args mandatory (as per next comment) the new order shouldn't cause any clashes between the two constructors if only the bare minimum parameters (i.e. those of the original) are passed.

trisyoungs (changes requested):

src/nodes/importConfigurationCoordinates.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "nodes/importConfigurationCoordinates.h"
+#include "nodes/dissolve.h"
+
+ImportConfigurationCoordinatesNode::ImportConfigurationCoordinatesNode(Graph *parentGraph) : Node(parentGraph)
+{
+    addInput<Configuration *>("Configuration", "Configuration to which coordinates will be imported", configuration_);
+    addOption<std::string>("FilePath", "File path", filePath_);
+    addOption<CoordinateImportFileFormat::CoordinateImportFormat>("FileFormat", "File format", format_);
+    addOutput<Configuration *>("Configuration", "Output configuration", configuration_);
```
trisyoungs, line 17: Just a bit of formatting for consistency with other changes I'm adding in.
```suggestion
    // Inputs
    addInput<Configuration *>("Configuration", "Configuration to which coordinates will be imported", configuration_);

    // Options    
    addOption<std::string>("FilePath", "File path", filePath_);
    addOption<CoordinateImportFileFormat::CoordinateImportFormat>("FileFormat", "File format", format_);

    // Outputs    
    addOutput<Configuration *>("Configuration", "Output configuration", configuration_);
```

src/nodes/data1DImport.cpp
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#include "nodes/data1DImport.h"
+#include "nodes/dissolve.h"
+
+Data1DImportNode::Data1DImportNode(Graph *parentGraph) : Node(parentGraph)
+{
+    addOption<std::string>("Filepath", "File path", filePath_);
+    addOption<Data1DImportFileFormat::Data1DImportFormat>("ImportFormat", "Format of imported data", format_);
```
trisyoungs, line 11: One for the Dissolve2do - a file selector "type".

src/nodes/data1DImport.cpp
```diff
@@ -0,0 +1,35 @@
...
+    addOption<std::optional<Number>>("RemoveAverageFromX", "Remove average from X values", removeAverageFromX_);
+    addOption<std::optional<Number>>("XMin", "Minimum X value", xMin_);
+    addOption<std::optional<Number>>("XMax", "Maximum X value", xMax_);
+    addOption<Number>("NPointsToRemove", "Number of points to remove", nPointsToRemove_);
+    addOption<Number>("XColumn", "Column index of data X values", xColumn_);
+    addOption<Number>("YColumn", "Column index of data Y values", yColumn_);
+    addOption<Number>("ErrorColumn", "Column index of data error values", errorColumn_);
+    addOutput<Data1D>("Data", "Imported data", data_);
```
trisyoungs:
```suggestion
    // Options
    addOption<std::string>("Filepath", "File path", filePath_);
    addOption<Data1DImportFileFormat::Data1DImportFormat>("ImportFormat", "Format of imported data", format_);
    addOption<std::optional<Number>>("RemoveAverageFromX", "Remove average from X values", removeAverageFromX_);
    addOption<std::optional<Number>>("XMin", "Minimum X value", xMin_);
    addOption<std::optional<Number>>("XMax", "Maximum X value", xMax_);
    addOption<Number>("NPointsToRemove", "Number of points to remove", nPointsToRemove_);
    addOption<Number>("XColumn", "Column index of data X values", xColumn_);
    addOption<Number>("YColumn", "Column index of data Y values", yColumn_);
    addOption<Number>("ErrorColumn", "Column index of data error values", errorColumn_);

    // Outputs
    addOutput<Data1D>("Data", "Imported data", data_);
```

trisyoungs (approved):
Couple of small comments.  I think this all looks good otherwise - we'll find out when we put it in a graph!

src/nodes/data1DImport.cpp
```diff
@@ -0,0 +1,38 @@
...
+    addOption<std::optional<Number>>("RemoveAverageFromX", "Remove average from X values", removeAverageFromX_);
+    addOption<std::optional<Number>>("XMin", "Minimum X value", xMin_);
+    addOption<std::optional<Number>>("XMax", "Maximum X value", xMax_);
+    addOption<Number>("NPointsToRemove", "Number of points to remove", nPointsToRemove_);
+    addOption<Number>("XColumn", "Column index of data X values", xColumn_);
+    addOption<Number>("YColumn", "Column index of data Y values", yColumn_);
+    addOption<Number>("ErrorColumn", "Column index of data error values", errorColumn_);
+
+    // Outputs
+    addOutput<std::optional<Data1D>>("Data", "Imported data", data_);
```
trisyoungs, line 21: I understand why you've declared this as `std::optional<Data1D>` but I'm in two minds as to whether this is the right choice (vs just plain `Data1D`).  It depends on how the node actually handles loading from the file, in particular when it encounters an error. There's also a question on how all import nodes like this one should reference files in the first place - if we do indeed move to a project folder structure (as I think we should) then data has to be manually imported in the first instance, so a Data1D node would just reference this "internal" data. Anyway, a problem and discussion for another time!

trisyoungs (approved):
We'll see how the edges behave passing the `std::optional<Data1D>` once you link these nodes into your argon test!

PR #2248, opened 2025-10-03
refactor: Make Interpolator class const and immutable.

trisyoungs:
This PR makes some changes to `Interpolator`, making it an immutable object with all-`const` functions. It also discovers and fixes an age-old bug whereby requesting interpolation of a point with x value beyond the data space lower x limit would return an incorrectly interpolated value.


src/math/interpolator.h
```diff
@@ -48,14 +45,16 @@ class Interpolator
...
-    void interpolate(InterpolationScheme scheme = SplineInterpolation);
     // Return interpolated y value for supplied x
-    double y(double x);
-    // Return interpolated y value for supplied x, specifying containing interval
-    double y(double x, int interval);
+    double y(double x) const;
+    // Return interpolated y value for supplied x from the specified interval
+    double y(double x, int interval) const;
+    // Return interpolated y values for supplied, sequentially increasing x values
+    std::vector<double> y(const std::vector<double> &xs) const;
```
rprospero, line 57: This overload is such a nice improvement.

rprospero (approved):

PR #2249, opened 2025-10-07
feat: Argon nodes with test data

RobBuchananCompPhys:


src/math/function1D.h
```diff
@@ -36,6 +36,15 @@ class Function1DDefinition
...
 
+    bool operator==(const Function1DDefinition &other)
+    {
+        return parameterNames_ == other.parameterNames_ && properties_ == other.properties_;
+    }
+    bool operator!=(const Function1DDefinition &other)
+    {
+        return parameterNames_ != other.parameterNames_ || properties_ != other.properties_;
+    }
+
```
trisyoungs: Now, since we chopped and changed `Parameter` a bit this is redundant and can be removed _if_ the `ParameterBase::isDefault()` virtual is removed (also in `SerialisableParameter`). They aren't used anywhere now.

src/math/function1D.h
```diff
@@ -116,6 +125,15 @@ class Function1DWrapper : public Serialisable<>
...
 
+    bool operator==(const Function1DWrapper &other)
+    {
+        return form_ == other.form_ && parameters_ == other.parameters_ && internalParameters_ == other.internalParameters_;
+    }
+    bool operator!=(const Function1DWrapper &other)
+    {
+        return form_ != other.form_ || parameters_ != other.parameters_ || internalParameters_ != other.internalParameters_;
+    }
+
```
trisyoungs: And this?

src/nodes/gr/process.cpp
```diff
@@ -78,9 +78,11 @@ NodeConstants::ProcessResult GRNode::process()
     message("Cutoff (snapped to bin width) is {} Angstroms.\n", grRange);
 
     // Convert configuration species populations into real species populations
+    /*
     std::map<const Species *, double> realSpeciesPopulations;
     for (auto &[sp, iPop] : targetConfiguration_->speciesPopulations())
         realSpeciesPopulations[sp] = iPop;
+    */
```
trisyoungs: Can all be removed.

src/nodes/neutronSQ/neutronSQ.cpp
```diff
@@ -39,6 +39,8 @@ NeutronSQNode::NeutronSQNode(Graph *parentGraph) : Node(parentGraph)
     addOption<bool>("SaveRepresentativeGR", "Save representative G(r), obtained from Fourier transform of the calculated F(Q)",
                     saveRepresentativeGR_);
     addOption<bool>("SaveSQ", "Save weighted partial and total structure factors", saveSQ_);
+    addOptionalPointerOutput<PartialSet>("WeightedSQ", "Weighted partials for target configuration", weightedSQ_);
+    addOptionalPointerOutput<PartialSet>("WeightedGR", "Weighted partials for target configuration", weightedGR_);
```
trisyoungs:
```suggestion
    addOptionalPointerOutput<PartialSet>("WeightedSQ", "Weighted partial structure factors for target configuration", weightedSQ_);
    addOptionalPointerOutput<PartialSet>("WeightedGR", "Weighted partial radial distribution functions for target configuration", weightedGR_);
```

src/nodes/data1DImport.cpp
```diff
@@ -27,9 +27,12 @@ std::string_view Data1DImportNode::summary() const { return "Import 1D data."; }
...
 NodeConstants::ProcessResult Data1DImportNode::process()
 {
+    auto xMin = xMin_ ? std::make_optional<double>(xMin_->asDouble()) : std::nullopt;
+    auto xMax = xMax_ ? std::make_optional<double>(xMax_->asDouble()) : std::nullopt;
+    auto removeAverageFromX = removeAverageFromX_ ? std::make_optional<double>(removeAverageFromX_->asDouble()) : std::nullopt;
+
     Data1DImportFileFormat dataImport(filePath_, format_, xColumn_.asInteger(), yColumn_.asInteger(), errorColumn_.asInteger(),
-                                      removeAverageFromX_.value().asDouble(), xMin_.value().asDouble(),
-                                      xMax_.value().asDouble(), nPointsToRemove_.asInteger());
+                                      removeAverageFromX, xMin, xMax, nPointsToRemove_.asInteger());
```
trisyoungs, line 35: Well, this is all a bit awkward isn't it? No way around it unless we make `Data1DImportFormat` take `std::optional<Number>` instead of `std::optional<double>`.  Perhaps note this as a future thing in Dissolve2Do.

src/nodes/registry.cpp
```diff
@@ -43,25 +45,29 @@ void NodeRegistry::instantiateNodeProducers()
...
+        {"MD", makeDerivedNode<MDNode>()},
+        {"Multiply", makeDerivedNode<MultiplyNode>()},
+        {"NeutronSQ", makeDerivedNode<NeutronSQNode>()},
+        {"Number", makeDerivedNode<NumberNode>()},
+        {"SQ", makeDerivedNode<SQNode>()},
+        {"Subtract", makeDerivedNode<SubtractNode>()},
+        {"Vec3Assembly", makeDerivedNode<Vec3AssemblyNode>()},
+        {"Vec3Decomposition", makeDerivedNode<Vec3DecompositionNode>()},
+        {"ImportConfigurationCoordinates", makeDerivedNode<ImportConfigurationCoordinatesNode>()},
+        {"Data1DImport", makeDerivedNode<Data1DImportNode>()},
```
trisyoungs: Alphabetical order for the fussy PM please!

tests/.gitignore
```diff
@@ -4,6 +4,7 @@
 *.out
 **/*.rdf
 **/*.sq
+!data/dissolve2/argon/*.sq
```
trisyoungs: I might suggest instead that the ignore entries for both `.rdf` and `.sq` are instead removed, rather than adding on an explicit rule for your new data.  These are quite old rules!

trisyoungs (approved):
Last thing for me to do is sanity check the SQ comparison, but I guess what you meant at stand-up was that you had to up the tolerance from the generic `5.0e-3` that the `checkData1D` function uses, so I think it should all be fine.

I believe this is ready to go in, notwithstanding the few little suggestions. :+1:

src/nodes/data1DImport.cpp
RobBuchananCompPhys: Good point, but do we want `src` code to have a dependency on `nodes` ?

PR #2250, opened 2025-10-09
feat: Introduce SerialisableParameter.

trisyoungs:
To date we had made the assumption (willingly or not) that any `Parameter` must be serialisable to/from an input file, since `ParameterBase` inherits `Serialisable`. However, there are certain use cases of parameters, as well as certain data types, that never need to be / can't be serialised in the first place, leading to lots of boilerplate `constexpr` and compile-time type-checking.

This PR clears things up considerably by stating the following rules:
1) Any option in a `Node` (via `addOption`) is _always_ potentially serialised, but conversely always represents a simple type (`int`, `float`, `Number`, `EnumOptions` etc.) that are easily serialisable.
2) Node outputs (via `addOutput`) should _never_ need to be explicitly serialised as they represent either access to local complex data variables (the results of the node's processing, and which should be covered by specific `addSerialisable` calls sending data to the restart file) or are straight pointer variables which are set via `Edge` connections and should not be written out explicitly.
3) Node inputs (via `addInput`) are generally not serialisable as they typically represent data which is never locally available and delivered via an `Edge`, hence there is no need for serialisation. However, some node inputs _do_require serialisation - e.g. `Number`s provided to the `AddNode` and other math functions - _if_ they do not have a current `Edge` connection defining their value. Detecting whether this is actually the case is difficult at the point of serialising the data, so current values should be written to the input file regardless. Node inputs are therefore not serialisable _by default_.

To achieve this neatly a new `template` `SerialisableParameter<T>` which inherits from `Parameter<T>` is introduced, and simply defines the (virtual) `serialise()` and `deserialise()` functions. Calling `addOption` always creates a `SerialisableParameter<T>`,  `addOutput` always creates a non-serialisable `Parameter<T>`, and `addInput` creates a non-serialisable `Parameter<T>` by default, but which can be flagged to be generated as a `SerialisableParameter<T>` with the new `addSerialisableInput()`.

### DIFF
--- a/src/nodes/add.cpp
+++ b/src/nodes/add.cpp
@@ -5,8 +5,8 @@
 
 AddNode::AddNode(Graph *parentGraph) : Node(parentGraph)
 {
-    addInput<Number>("A", "First operand to the addition", a_);
-    addInput<Number>("B", "Second operand to the addition", b_);
+    addSerialisableInput<Number>("A", "First operand to the addition", a_);
+    addSerialisableInput<Number>("B", "Second operand to the addition", b_);
     addOutput<Number>("Result", "The sum of the operands", result_);
 }
 
--- a/src/nodes/dotProduct.cpp
+++ b/src/nodes/dotProduct.cpp
@@ -2,8 +2,8 @@
 
 DotProductNode::DotProductNode(Graph *parentGraph) : Node(parentGraph)
 {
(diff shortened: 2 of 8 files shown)


RobBuchananCompPhys (approved):
Nice formalisation! Good to see it can be cleanly applied to existing nodes too.

PR #2252, opened 2025-10-10
fix: Allow integers to change to doubles

trisyoungs:
This PR addresses a poor design decision on my part related to the `Number` class. Previously, all assignment-based mathematical operators - `+=`, `-=`, `*=`, and `/=` - would preserve a starting `int` type regardless of the operand. In other words, adding a `double` to an `int` _a la._ `myInt += 2.5` still resulted in an `int` contained in `myInt` rather than the `double` value one might expect.  I don't think this is the behaviour one would expect from the class, and it breaks my upcoming `History` unit test (because the `Number` variable created in order to compute the average steadfastly remains an `int` despite having a load of `double`s added to it) so here it is changed to have the more natural behaviour.

### DIFF
--- a/src/nodes/number.cpp
+++ b/src/nodes/number.cpp
@@ -23,8 +23,7 @@ Number Number::operator+(const Number &other) const
 
 Number &Number::operator+=(const Number &other)
 {
-    std::visit([](auto &a, auto b) { a += b; }, value_, other.value_);
-    set(value_);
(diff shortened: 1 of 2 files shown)


RobBuchananCompPhys (approved):
I think further to the discussion earlier, it makes a lot of sense to do this. The only concern I had at first was that we mutate the value into something `double` when the user expected the Number to hold an `int` at all times. But then I realised we don't actually ever specify that the Number's value must remain a certain type!

rprospero (approved):

PR #2255, opened 2025-10-15
chore:Fix submodule hack on CI

rprospero:
At the base level, this PR just updates the nix version that we use on the CI.  However, that nix version includes support for flakes in submodules, so that is added to the flake.  In turn, since flakes are support in submodules, we can remove the submodule hack on the CI.


trisyoungs (approved):
Goodbyeeeee unholy hack!

PR #2256, opened 2025-10-15
chore:Nix flake whitelist

rprospero:
This PR sets a specific whitelist of the files and folders needed to build Dissolve within the flake.  Anything *not* on this list is invisible to the flake.

Previously, we allowed the flake to see the whole repo.  However, this meant that *any* changes to the folder meant that nix would perform a full rebuild, often resulting in double compilation.  For example, we currently build the executable and then build the singularity image.  However, building the executable puts the resulting binary in the folder.  Nix sees that the folder has changed and, just to be safe, recompiles Dissolve when we go to build the singularity image.  With the whitelist, nix will know that the binary output of the build process isn't part of the compilation, so it will just use the previously cached output to build the singularity image.

The most interesting question to review is where there are any other files or folders that we truly require.

### DIFF
--- a/.github/workflows/build/linux/action.yml
+++ b/.github/workflows/build/linux/action.yml
@@ -72,33 +72,33 @@ runs:
       mkdir packages-binary
       cp -v binary packages-binary/${{ inputs.target }}-${{ inputs.currentVersion }}
 
-  - name: Run benchmarks
-    if: inputs.benchmark == 'true'
-    shell: bash
-    run: |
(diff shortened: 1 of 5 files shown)


trisyoungs (approved):
Nice bit of detective work here.  Very happy to see this fix go in!

trisyoungs:
And just to follow-up, I don't think there's anything else that needs adding to the whiltelist.

PR #2258, opened 2025-10-15
refactor: Const PairPotential

trisyoungs:
## Precedes #2242

Following #2248 this PR makes the `PairPotential` class completely `const` in its usage, and moves its storage in the new `DissolveNode` map to the bona fide object rather than a `std::shared_ptr`.


rprospero (approved):
Having the const operations is an obvious boon.  One of my favourite bits of Rust is the fact that members functions are const by default.

PR #2261, opened 2025-10-16
build: Fix continuous2 CI

trisyoungs:
Just noticed that our continuous Dissolve2 builds were broken - apparently my previous self had also noticed this on the 8th of October, made the fix, and neglected to push the branch.


Danbr4d (approved):
Looks good

PR #2262, opened 2025-10-17
feat: LoopGraph

trisyoungs:
This PR implements a `LoopGraph` node - a simple `Graph`-type node which iterates over its children a number of times. It changes the `InputsNode` to mirror its proxy output parameters to identically-named input parameters, allowing edges to be created to transfer data back to the beginning of the loop.

The actual data flow in this PR is worth explicitly stating.  In the unit test, when the node `y_` is pulled to start the processing the `LoopGraph` is activated and processes it's input edges (only one, from the node `i_`). It then proceeds to iterate its subgraph for the required number of cycles. Loopback edges are not pulled on the first iteration, only on every subsequent iteration in order to provide any new input data. This means that, unless the node `i_` is modified, calling the graph again (as is performed in the unit test) will result in it being `Unchanged`.

The loopback of data is achieved in practice by essentially disconnecting the `InputsNode` from natural node behaviour, principally overriding it's `run()` function to not automatically pull input edges (which would cause an infinite loop) and let `LoopGraph` handle this when necessary. Furthermore, propagation of "update required" flags is deliberately stopped at edges connecting to an input on an `InputsNode` as again this would cause an infinite loop.

Complicated, but hopefully a solid solution.  Thoughts please!

### DIFF
--- a/src/nodes/loop.cpp
+++ b/src/nodes/loop.cpp
@@ -8,8 +8,6 @@
 LoopGraph::LoopGraph(Graph *parentGraph) : Graph(parentGraph)
 {
     addOption<Number>("Iterations", "Number of iterations to perform", iterations_);
-    addOption<bool>("LoopbackInvalidates", "Whether loopback edges cause iterations to be rerun on next call",
-                    loopbackInvalidates_);
 }
 
 /*
@@ -20,7 +18,7 @@ LoopGraph::LoopGraph(Graph *parentGraph) : Graph(parentGraph)
 std::string_view LoopGraph::type() const { return "Loop"; }
 
 // Return short summary of the node's purpose
-std::string_view LoopGraph::summary() const { return "Loop the contained graph"; }
+std::string_view LoopGraph::summary() const { return "A graph which iterates"; }
 
 /*
  * Processing & Validity
@@ -29,8 +27,7 @@ std::string_view LoopGraph::summary() const { return "Loop the contained graph";
 // Perform processing
 NodeConstants::ProcessResult LoopGraph::process()
 {
-    // Could we have a map of inputs to the graph and when they were updated, which we maintain between runs? We can then decide
-    // whether to use current parameter values or pull from loopback edges.
+    //
     for (auto n = 0; n < iterations_.asInteger(); ++n)
     {
         std::cout << std::format("\nLoopGraph::process() - iteration {}\n\n", n);
@@ -62,17 +59,6 @@ NodeConstants::ProcessResult LoopGraph::process()
         if (result == NodeConstants::ProcessResult::Failed || result == NodeConstants::ProcessResult::InputsNotSatisfied)
             return NodeConstants::ProcessResult::Failed;
     }
-
-    // // If allowing loopback data to invalidate our status, check any loopback edges for changed data
-    // if (loopbackInvalidates_)
-    // {
-    //     for (auto &[inputName, edges] : proxyInputs_->inputEdges())
-    //     {
-    //         for (const auto edge : edges)
-    //             if (edge->requiresPull())
-    //                 return NodeConstants::ProcessResult::SuccessAndNotUpdated;
-    //     }
-    // }
 
     return NodeConstants::ProcessResult::Success;
 }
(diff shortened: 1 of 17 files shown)


tests/nodes/loop.cpp
```diff
@@ -0,0 +1,99 @@
...
+         *                           \   -    |     ------------------    /    |-----/    o-B (0)           |
+         *                            +-o-->>>-o---o-A         result-o--+     |           -----------------/
+         *                               -    |    o-B (1)           |    \    |-LB--\
+         *                               -----|     -----------------/     \   |     |
+         *                                    |                             +-o-->IN |
+         *                                    |                                |-----/
+         *                                    ----------------------------------
+         */
+        root_.x = 0;
+        root_.y = 0;
```
RobBuchananCompPhys, line 37: Is this something we need to do often when setting up test graph? It may be good to have a base class for these tests which takes ownership of `root_` and zeroes the coordinates behind the scenes, so we don't have  to think about it.

```
namespace UnitTest
{
class LoopGraphTest :  public GraphTest // derives from public ::testing::Test
{
    public:
    LoopGraphTest() : GraphTest(), dissolve_(coreData_) {} // GraphTest could have default xtor args like centered = true //which give us a centered (x0, y0) graph

    // Create a graph for testing
    void createGraph()
```

RobBuchananCompPhys (approved):
Good progress! I think it would be worth investing some time while we're developing more tests to try and make test graph initialisation more efficient i.e. via some sort of TestGraph class hierarchy as commented above.

src/nodes/node.h
```diff
@@ -48,52 +48,63 @@ class Node : public Serialisable<>
...
-
+    public:
     // Enumeration for message status
     enum class MessageStatus
     {
         Info,
         Warn,
-        Error
+        Error,
+        Debug
```
rprospero, line 60: Watching this grow, I'm wondering if we might be better off with a third party logging library.  We've already written half of one and we might get some free functionality from one that is pre-built

src/nodes/loop.cpp
```diff
@@ -0,0 +1,67 @@
...
+                    }
+                }
+            }
+        }
+
+        // Process child nodes, returning early if we encounter an error
+        auto result = Graph::process();
+        if (result == NodeConstants::ProcessResult::Failed || result == NodeConstants::ProcessResult::InputsNotSatisfied)
+            return NodeConstants::ProcessResult::Failed;
+    }
```
rprospero, line 64: When we had discussed your proposal during the stand-up, I had misunderstood and thought that you were putting *this* logic into the input nodes.  Having this in the node answers 95% of my concerns.

tests/nodes/loop.cpp
rprospero: My suspicion would be that, in the longer term, most of these test graphs would be loaded from a TOML file instead of crafting the graphs by hand.

rprospero (commented):
This looks like a solid design to me and answers most of the questions that I had in the stand-up.  I'm particularly please that we know that the loop won't have to be re-run if the user changes and downstream analysis nodes.

src/nodes/loop.cpp
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+// Perform processing
+NodeConstants::ProcessResult LoopGraph::process()
+{
+    // Before arriving here we will have pulled in any external input edges, and this is the data we should use on the
+    // first iteration of the loop. So, we only pull from inputs to our InputsNode (proxyInputs_) after the first
+    // iteration. Furthermore, we do this manually here rather than let InputsNode pull its own edges as otherwise we
+    // get into an infinite loop when the graph tries to run. In this sense the
+
```
RobBuchananCompPhys, line 34: Seems like there was more to come here!

RobBuchananCompPhys (commented):
My feeling is that there are some similarities we have converged on, but also differences in the complexity and robustness of the looping strategies. Personally, my preference would be to implement something simpler, which the other [PR](https://github.com/disorderedmaterials/dissolve/pull/2264) demonstrates (in about half the lines of code)! This economy, I find, is easier to reason about. However, there are short comings there which I think you have covered here particularly with regards to robustness of version indexing, and as you mentioned in the stand up, ensuring that the loop graph isn't re-traversed when an external node is changed, which I hadn't taken into account. So, ideally I'd like to go with the simpler option and see how it holds up for now, but if you and Adam think this is the one that's fine also.

PR #2264, opened 2025-10-21
Dissolve2/loop subgraph 2

RobBuchananCompPhys:
LoopGraph

The idea here is to use the original structure of a separate proxy on the subgraph outputs side for the `LoopBacks`. These are edge connections that are populated according to the available incoming proxy inputs, such as a numerical value. However, the `LoopBacks` do not output _from_ the graph like the normal outputs, but rather the special `LoopEdge` objects (which are edges connected from an internal node output to a `LoopBack`) feedback to the proxy inputs during the active iterations of the `LoopGraph`. So, after the initial traversal of the subgraph, when a `LoopEdge` is pulled, it sets the corresponding `analogue` parameter (i.e. the original incoming numerical value) to the result of the internal computation.


src/nodes/edge.h
```diff
@@ -83,3 +85,34 @@ class Edge : public Serialisable<>
     // Read values from a serialisable value
     void deserialise(const SerialisedValue &node) override;
 };
+
+// Loop edge
+class LoopEdge : public Edge
```
rprospero, line 90: One concern I have about this is serialisation.  During since this uses the same `serialise` and `deserialise` methods as `Edge`, how does Dissolve know to create a `LoopEdge` instead of just an `Edge` when it loads from a file?

I think that we're okay because the deserialisation calls the overloaded `addEdge` method on the `Loop` node, which then makes the conversion for us, but I am curious if you've tested this functionality.

rprospero (commented):
Honestly, things look much clearer than before and I do like the proxy node setup for other actions we might choose.  I do have one concern listed below about having multiple edge types and how that interacts with serialisation.

Beyond that and cleaning up the conflicts with `develop2` I'm much happier with this code than before.

src/nodes/edge.h
RobBuchananCompPhys: I haven't yet tested that, so thanks for the suggestion. I think there are one or two other things that need to be tested such as automatic removal of any loop edges corresponding to removal of an external input to the `LoopGraph`

src/nodes/loopBack.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "nodes/outputs.h"
+
+LoopBacksNode::LoopBacksNode(Graph *parentGraph) : Node(parentGraph) {}
+
+/*
+ * Definition (Virtuals)
+ */
+
+// Return type of the node
+std::string_view LoopBacksNode::type() const { return "Outputs"; }
```
trisyoungs: LoopBack?

src/nodes/loopBack.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+// Return short summary of the node's purpose
+std::string_view LoopBacksNode::summary() const { return "Maps local inputs to graph outputs"; }
```
trisyoungs: Also needs changing.

src/nodes/loopBack.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+    auto loopGraph = static_cast<LoopGraph *>(parentGraph_);
+
+    if (loopGraph->loopEdges().empty())
+        return status;
+
+    for (const auto &edge : loopGraph->loopEdges())
+        status = static_cast<LoopEdge *>(edge.get())->pull();
+
+    return status;
```
trisyoungs, line 46: As written this would not fail correctly if, say there are two `LoopEdge`s defined where the first pull fails and the second pull succeeds. Returning Failure on the first pull failure is a safer way to go, and is in line with other parts of the graph system.

src/nodes/loopBack.h
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2025 Team Dissolve and contributors
+
+#pragma once
+
+#include "nodes/node.h"
+
+// Outputs Node
```
trisyoungs:
```suggestion
// LoopBacks Node
```

src/nodes/loopGraph.h
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public:
+    // Return node name
+    std::string_view name() const override;
+    // Return type of the node
+    std::string_view type() const override;
+    // Return short summary of the node's purpose
+    std::string_view summary() const override;
+    // Increment loop counter
+    void increment();
```
trisyoungs: Not a virtual, so should be moved.  Should it be `private`?

src/nodes/loopGraph.h
```diff
@@ -0,0 +1,78 @@
...
+    // Number of loops (iterations) to perform
+    const int nLoops() const;
+    // Loop backs
+    LoopBacksNode *loopBacks();
+    // Loop edges
+    Edges &loopEdges();
+    // Current loop iteration
+    int loopCount();
+    // Reset the loop counter to zero
+    void resetLoopCounter();
```
trisyoungs: Similar thing - how much of this really needs to be `public` since it's all related to internal operation?

tests/nodes/loop.cpp
```diff
@@ -50,18 +49,22 @@ class LoopGraphTest : public ::testing::Test
...
         EXPECT_TRUE(loop_->addEdge({"x", "Result", "Outputs", "C"}));
-        EXPECT_TRUE(loop_->addEdge({"x", "Result", "Inputs", "I"}));
+        EXPECT_TRUE(loop_->addEdge({"x", "Result", "LoopBacks", "I"}));
         // - The output "C" of the loop graph then goes to input "A" of Add 'y'
-        EXPECT_TRUE(root_.addEdge({"loop", "C", "y", "A"}));
+        EXPECT_TRUE(root_.addEdge({"Loop", "C", "y", "A"}));
+        /*
+         * TODO There is an obvious, smaller unit test to write here which tests the validity of trying to make a loopback
+         * connection to a named Input which doesn't exist.
+         */
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
Overall I think this a nice compromise between code complexity and developer readability!  I think the unit test should include some kind of check of the `version_`s of the various nodes as this is really important for quantifying "up-to-datedness" when traversing the graph and deciding whether nodes need to be rerun or not. See one of the other graph-related unit tests for examples.

src/nodes/loopBack.h
```diff
@@ -0,0 +1,49 @@
...
+     */
+    protected:
+    // Perform processing
+    NodeConstants::ProcessResult process() override;
+
+    public:
+    // Run the node, retrieving dependent inputs as necessary
+    NodeConstants::ProcessResult run() override;
+
+    public:
```
trisyoungs, line 35:
```suggestion
```

src/nodes/loopGraph.h
```diff
@@ -0,0 +1,78 @@
...
+    // Return node name
+    std::string_view name() const override;
+    // Return type of the node
+    std::string_view type() const override;
+    // Return short summary of the node's purpose
+    std::string_view summary() const override;
+
+    /*
+     * Feedback
+     *
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
Couple of small formatting things.  Otherwise, let's run with this and see how we get on!

PR #2265, opened 2025-10-21
refactor: WIP Serialise refactor

rprospero:
This refactors the TOML serialisation code.  Previously, `serialise` returned a `SerialisedValue` for all objects.  However, this might be an invalid, empty object and it was the caller's responsibility to check for this case.  Now, `serialise` takes a `SerialisedValue` reference and a label string and serialises the object onto the passed node under the given label.  This allows the `serialise` method to choose *not* to make any changes in the case where the result would be an empty object.

This is currently a work in progress with a large number of debugging printfs that need cleaning.  If you're seeing this message, don't bother trying to review it in its current state.

Also, this will eventually want to be merged onto `develop2`, but it needs the other branch merged in first.


rprospero:
This was surpassed by #2268

PR #2268, opened 2025-11-10
refactor:change serialisation

rprospero:
TOML officially does not support null values.  Instead, the recommended method of handling nulls is to not include the associated key.  This is a problem for our current setup, where the `serialise` method returns a value that is then added to a key.

This PR changes `serialise` so that it is passed in a node onto which to serialise the data as well as a key for where to serialise the information.  In other words, what had once been

```cpp
result[key] = data.serialise()
```

is now

```cpp
data.serialise(key, result)
```

The primary advantage of this scenario is that the `serialise` method can now detect when it would return a null value and choose to do nothing.  This is preferable to *every call site* needing to perform testing for null values.

Honestly, the best solution would have been for the original function to return a `Option<SerialisedValue>` and for the TOML library to correctly handle assignment from an option, but there's a reason that the library is named toml11 and not toml17

### DIFF
--- a/src/classes/speciesImproper.cpp
+++ b/src/classes/speciesImproper.cpp
@@ -219,9 +219,12 @@ double SpeciesImproper::force(double phi) const
 }
 
 // Express as a serialisable value
-SerialisedValue SpeciesImproper::serialise() const
+void SpeciesImproper::serialize(std::string tag, SerialisedValue &target) const
 {
-    auto improper = SpeciesIntra<SpeciesImproper, TorsionFunctions>::serialise();
+    // if (i_ == nullptr && j_ == nullptr && k_ == nullptr && l_ == nullptr)
+    //     return;
+    SpeciesIntra<SpeciesImproper, TorsionFunctions>::serialize(tag, target);
+    auto &improper = target.at(tag);
     if (i_ != nullptr)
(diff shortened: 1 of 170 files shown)


src/classes/speciesImproper.cpp
```diff
@@ -219,9 +219,12 @@ double SpeciesImproper::force(double phi) const
 }
 
 // Express as a serialisable value
-SerialisedValue SpeciesImproper::serialise() const
+void SpeciesImproper::serialise(std::string tag, SerialisedValue &target) const
 {
-    auto improper = SpeciesIntra<SpeciesImproper, TorsionFunctions>::serialise();
+    // if (i_ == nullptr && j_ == nullptr && k_ == nullptr && l_ == nullptr)
+    //     return;
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
Looking good.  One deletion in the review, but nothing else major.  Could / should we make the function signature `serialise(const std::string &tag, SerialisedValue &target)` though?

This is probably already on your list of ideas for future improvement, but it would be worthwhile making the various `from*` functions local to `SerialisableValue` rather than `static` so we can remove the copy from the "inner" temporary.

PR #2270, opened 2025-11-17
fix: Iterator errors and string matching (Debug mode)

trisyoungs:
This PR makes a few fixes based on running the test suite in Debug mode on Windows - the same errors are not picked up by GCC.


RobBuchananCompPhys (approved):
Great to see this fixed, thank you spending time delving in!

Issue #2277, opened 2025-11-26
Make species node editable


rprospero:
**Describe the feature / issue**
Currently, once a species is created, it is entirely immutable. This means that a user who realises that they made a mistake during the creation will need to recreate an entirely new node from scratch.  It would be better if the node was mutable and the node's output would then update that its output had changed.

**Additional details**
It should be possible to reuse the SpeciesDialog and SpeciesModel to handle this.

PR #2281, opened 2025-12-04
refactor:Direct species node

### LINKED ISSUE #2280: Hide certain parameters in nodes
Currently, all Options on a node are displayed in the GUI.  However, we might have content in a node that needs to be saved in the TOML file, but doesn't make sense to present to the user.  We could add a `hidden` flag to a parameter that would tell the GUI not to display the parameter, but would ensure that the data is still properly serialised.

**Additional details**
Add any other context or screenshots relevant to the feature request here.

rprospero:
The SpeciesNode has been modified to output an actual "Species*" instead of "std::shared_ptr<Species>".  This more closely matches the behaviour of the configuration node.  Furthermore, the actual species is contained in a hidden parameter of the node, so there is no vestigial option that might confuse the user.  This hidden parameter functionality can be used on other node types as well in the future for information that must be serialised but is not necessarily user facing.

This closes #2280 and #2277

### DIFF
--- a/src/gui/models/speciesModel.cpp
+++ b/src/gui/models/speciesModel.cpp
@@ -44,7 +44,6 @@ GraphModel *SpeciesModel::graphModel() { return graphModel_; }
 // Produce this species node on the given graph
 void SpeciesModel::assignModel(GraphModel *graphModel)
 {
-
     graphModel_ = graphModel;
     node_ = std::make_unique<SpeciesNode>(graphModel_->graph());
     auto &species = node_->species();
--- a/src/gui/models/speciesModel.cpp
+++ b/src/gui/models/speciesModel.cpp
@@ -7,21 +7,20 @@
 #include "speciesAtomModel.h"
 #include <memory>
 
-SpeciesModel::SpeciesModel()
-{
-    species_ = std::make_unique<Species>();
-    atoms_.setSpecies(species_.get());
-    bonds_.setBonds(species_->bonds());
-    angles_.setAngles(species_->angles());
-    torsions_.setTorsions(species_->torsions());
-    impropers_.setImpropers(species_->impropers());
-}
+SpeciesModel::SpeciesModel() : node_(nullptr) {}
 
-QString SpeciesModel::name() { return QString::fromStdString(std::string(species_->name())); }
+QString SpeciesModel::name()
+{
+    if (node_ == nullptr)
+        return "";
+    return QString::fromStdString(std::string(node_->species().name()));
+}
 
 void SpeciesModel::setName(QString name)
 {
-    species_->setName(name.toStdString());
+    if (node_ == nullptr)
+        return;
+    node_->species().setName(name.toStdString());
     Q_EMIT(nameChanged());
 }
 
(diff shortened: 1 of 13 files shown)


src/gui/models/speciesModel.cpp
```diff
@@ -40,40 +39,58 @@ SpeciesImproperModel *SpeciesModel::impropers() { return &impropers_; }
...
 SpeciesAtomModel *SpeciesModel::atoms() { return &atoms_; }
 
+GraphModel *SpeciesModel::graphModel() { return graphModel_; }
+
 // Produce this species node on the given graph
-void SpeciesModel::create(QVariant graphModel)
+void SpeciesModel::assignModel(GraphModel *graphModel)
+{
+
+    graphModel_ = graphModel;
```
RobBuchananCompPhys, line 47:
```suggestion
{
    graphModel_ = graphModel;
```

src/nodes/node.h
```diff
@@ -312,8 +312,14 @@ class Node : public Serialisable<>
     protected:
     // Persistent data serialisables
     std::map<std::string, std::shared_ptr<SerialisableData>> serialisables_;
+    // Serialise any hidden content
+    virtual void hidden_serialise(SerialisedValue &target) const {}
+    // Deserialise any hidden content
+    virtual void hidden_deserialise(const SerialisedValue &node) {}
```
RobBuchananCompPhys:
```suggestion
    // Serialise any hidden content
    virtual void hiddenSerialise(SerialisedValue &target) const {}
    // Deserialise any hidden content
    virtual void hiddenDeserialise(const SerialisedValue &node) {}
```

src/nodes/species.cpp
```diff
@@ -3,19 +3,29 @@
...
 
 std::string_view SpeciesNode::summary() const { return "Produce a species"; }
 
 NodeConstants::ProcessResult SpeciesNode::process() { return NodeConstants::ProcessResult::Unchanged; }
+
+Species &SpeciesNode::species() { return species_; }
+const Species &SpeciesNode::species() const { return species_; }
+
+// Serialise any hidden content
+void SpeciesNode::hidden_serialise(SerialisedValue &target) const { species_.serialise("species", target); }
```
RobBuchananCompPhys:
```suggestion
void SpeciesNode::hiddenSerialise(SerialisedValue &target) const { species_.serialise("species", target); }
```

src/nodes/species.cpp
```diff
@@ -3,19 +3,29 @@
...
+
+// Deserialise any hidden content
+void SpeciesNode::hidden_deserialise(const SerialisedValue &node)
```
RobBuchananCompPhys:
```suggestion
void SpeciesNode::hiddenDeserialise(const SerialisedValue &node)
```

src/nodes/species.h
```diff
@@ -25,14 +25,27 @@ class SpeciesNode : public Node
...
      * Processing
      */
     private:
     // Run main processing
     NodeConstants::ProcessResult process() override;
+
+    // Serialise any hidden content
+    void hidden_serialise(SerialisedValue &target) const override;
+    // Deserialise any hidden content
+    void hidden_deserialise(const SerialisedValue &target) override;
```
RobBuchananCompPhys:
```suggestion
    void hiddenSerialise(SerialisedValue &target) const override;
    // Deserialise any hidden content
    void hiddenDeserialise(const SerialisedValue &target) override;
```

RobBuchananCompPhys (changes requested):

src/nodes/node.h
rprospero: I've made these changes in a separate commit to ensure that all of the call sites also got updated.

RobBuchananCompPhys (approved):

src/nodes/node.h
```diff
@@ -312,8 +312,14 @@ class Node : public Serialisable<>
     protected:
     // Persistent data serialisables
     std::map<std::string, std::shared_ptr<SerialisableData>> serialisables_;
+    // Serialise any hidden content
+    virtual void hiddenSerialise(SerialisedValue &target) const {}
+    // Deserialise any hidden content
+    virtual void hiddenDeserialise(const SerialisedValue &node) {}
```
trisyoungs: I might suggest that "internal" is a better definition than "hidden":

```suggestion
    // Serialise any required internal content
    virtual void serialiseInternal(SerialisedValue &target) const {}
    // Deserialise any required internal content
    virtual void deserialiseInternal(const SerialisedValue &node) {}
```

trisyoungs (commented):
This all looks good, but I have one question - why does the `SpeciesModel` act as (from what I see) a producer of `SpeciesNode`s? I guess it depends on how we are utilising `SpeciesModel` in the UI, but In my head it should only ever be receiving the pointer to a `SpeciesNode` (or even just a `Species`...) from which it should be setting / getting data. I'm probably misunderstanding the intent here, so I had to ask!

PR #2282, opened 2025-12-11
test: Dissolve2 AtomShake unit test

RobBuchananCompPhys:
This PR will probably need to encompass a few developments in order to get a successful porting of the `atomShake` test into Dissolve2 node-graph framework.

- [ ] Definitive unit test for configuration generator nodes (`ConfigurationNode`, `InsertNode`) to ensure that downstream changes such as resetting the temperature of the configuration do not result in the generator steps re-running
- [ ] Replicate `atomShake` using a combination of `SpeciesNode`, and `AtomicMCNode` inside a `LoopGraph`


src/nodes/configuration.cpp
```diff
@@ -5,11 +5,16 @@
 
 ConfigurationNode::ConfigurationNode(Graph *parentGraph) : Node(parentGraph)
 {
+    addOption<Number>("Temperature", "Configuration temperature", temperature_);
```
trisyoungs:
```suggestion
    addOption<Number>("Temperature", "Configuration temperature (K)", temperature_);
```

tests/nodes/dissolve/atomShake.cpp
trisyoungs, line 1: Putting this in a `dissolve/` subdirectory feels a little arbitrary. We might want to organise the tests in this dir as the number grows for sure, but for now it should live in the parent `nodes/` dir.

tests/nodes/dissolve/atomShake.cpp
```diff
@@ -0,0 +1,222 @@
...
+    AtomShakeTest() : dissolve_(coreData_), root_(dissolve_) { Node::echo_ = true; }
+
+    void createGraph(const int nLoops)
+    {
+        coreData_.clear();
+
+        // Create loop graph containing an AtomicMCNode
+        loop_ = dynamic_cast<LoopGraph *>(root_.createNode("Loop", "Loop"));
+        auto nLoopsOpt = loop_->findOption("NLoops");
+        nLoopsOpt->set<Number>(nLoops);
```
trisyoungs:
```suggestion
        ASSERT_TRUE(loop_->setOption<Number>("NLoops", nLoops));
```

trisyoungs: Also, should this parameter be just `N` for simplicity?

tests/nodes/dissolve/atomShake.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+        auto atomicMCNode = std::make_unique<AtomicMCNode>(static_cast<Graph *>(loop_));
+        auto nShakes = atomicMCNode->findOption("ShakesPerAtom");
+        nShakes->set<Number>(10);
```
trisyoungs:
```suggestion
        ASSERT_TRUE(atomicMCNode->setOption<Number>("ShakesPerAtom", 10));
```

tests/nodes/dissolve/atomShake.cpp
```diff
@@ -0,0 +1,222 @@
...
+        atomicMCNode_ = atomicMCNode.get();
+        ASSERT_TRUE(loop_->addNode(std::move(atomicMCNode), "AtomicMC"));
+
+        // Water species
+        auto speciesNode = std::make_unique<SpeciesNode>(&root_);
+        speciesNode_ = speciesNode.get();
+        auto species = makeH2O(&(speciesNode->species()), {hA_, o_, hB_}, terms_);
+        ASSERT_TRUE(root_.addNode(std::move(speciesNode), "Water"));
```
trisyoungs: This kind of thing we will be doing a fair amount, so I think this should be compressed into a line like:
```
auto species = createWater(root_);
```
This function would live in `testData.h` and create the water `SpeciesNode` in the supplied `root_` graph, returning the `Species*` for any further required manipulation. On that note, passing the parameters in from this test is an unnecessary data flow - this should all be contained within the `createWater` function. We will use the water molecule for a whole bunch of other (analysis) tests, so it should be the "correct" molecule - from the perspective of the unit test here this means adjusting the bond and angle parameters via the returned `Species*` pointer to some local target values we decide, then `AtomShake`ing to them.

tests/nodes/dissolve/atomShake.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+        // Water configuration
+        configurationNode_ = dynamic_cast<ConfigurationNode *>(root_.createNode("Configuration", "Bulk"));
+        ASSERT_TRUE(configurationNode_);
+        auto temp = configurationNode_->findOption("Temperature");
+        temp->set<Number>(temperature_);
```
trisyoungs:
```suggestion
        ASSERT_TRUE(configurationNode_->setOption<Number>("Temperature", 300.0));
```

tests/nodes/dissolve/atomShake.cpp
```diff
@@ -0,0 +1,222 @@
...
+                    auto form = std::get<AngleFunctions::Form>(std::get<0>(attrs));
+                    auto indices = std::get<1>(attrs);
+                    auto params = std::get<2>(attrs);
+                    assert(indices.size() == 3);
+                    auto [i, j, k] = std::array<int, 3>{indices[0] - 1, indices[1] - 1, indices[2] - 1};
+                    species->addAngle(i, j, k).setInteractionFormAndParameters(form, params);
+                }
+                continue;
+            }
+        }
```
trisyoungs: This is a lot of unnecessary code when one could just call `addBond()` twice and `addAngle()` once with the explicit definitions - there's no need for them to be flexibly changed!

tests/nodes/dissolve/atomShake.cpp
```diff
@@ -0,0 +1,222 @@
...
+        auto population = node->findInput("Population");
+        population->set<Number>(n);
+
+        // Set density
+        auto density = node->findInput("Density");
+        density->set<Number>(rho);
+
+        // Set density units
+        ASSERT_TRUE(node->setOption<Units::DensityUnits>("DensityUnits", units));
+    }
```
trisyoungs: Again I think this can be done in-line in the test body. We don't even really need to use `TEST_F` here since we're not reusing anything.

tests/nodes/dissolve/atomShake.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+        // Set connections
+        EXPECT_TRUE(root_.addEdge({"Water", "Species", "Insert", "Species"}));
+        EXPECT_TRUE(root_.addEdge({"Bulk", "Configuration", "Insert", "Configuration"}));
+        EXPECT_TRUE(root_.addEdge({"Insert", "Configuration", "Loop", "Configuration"}));
+        EXPECT_TRUE(loop_->addEdge({"Inputs", "Configuration", "AtomicMC", "Configuration"}));
+        // TODO: Output parameter name cannot be the same as loopback (although loopback param CAN match input).
+        // May need to enforce this.
+        EXPECT_TRUE(loop_->addEdge({"AtomicMC", "Configuration", "Outputs", "OutputConfiguration"}));
+        EXPECT_TRUE(loop_->addEdge({"AtomicMC", "Configuration", "LoopBacks", "Configuration"}));
```
trisyoungs: These two lines don't cause a test failure, but do raise Dissolve ERRORs in the output:
```
***  ERROR
***  ERROR Tried to unlink an outgoing edge from source node 'AtomicMC' which knew nothing about it.
***  ERROR


***  ERROR
***  ERROR Tried to unlink an incoming edge to target node 'Inputs' which knew nothing about it.
***  ERROR
```

trisyoungs (commented):
One other general comment - for consistence with writing these sorts of tests in code we should use atom indices from 0 - (N-1) and avoid having to subtract one (this might be irrelevant, assuming that the data blocks disappear....)

tests/nodes/atomShake.cpp
```diff
@@ -0,0 +1,104 @@
...
+        atomicMCNode_ = atomicMCNode.get();
+        ASSERT_TRUE(loop_->addNode(std::move(atomicMCNode), "AtomicMC"));
+
+        // Water species
+        auto species = createWater(&root_);
+
+        // Water configuration
+        configurationNode_ = dynamic_cast<ConfigurationNode *>(root_.createNode("Configuration", "Bulk"));
+        ASSERT_TRUE(configurationNode_);
+        ASSERT_TRUE(configurationNode_->setOption<Number>("Temperature", 300.0));
```
trisyoungs: Ah, temperature - the bugbear in this PR!  For a "normal" simulation you'd absolutely want this to be 300 K, but for this specific test it has to be zero.  This is referred to as "zero K optimisation" in some simulation circles, and essentially AtomShake then doesn't explore any geometry other than the equilibrium distances and angles (i.e. the minima in those potential energy functions) defined in the molecule. Otherwise, it's exploring around those minima, as it should do, and so the test would likely never pass.

tests/nodes/atomShake.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+        insertNode_ = dynamic_cast<InsertNode *>(root_.createNode("Insert", "Insert"));
+        ASSERT_TRUE(insertNode_);
+
+        auto population = insertNode_->findInput("Population");
+        population->set<Number>(1);
```
trisyoungs: Feels like we really should have an analogous `setInput<T>()` function.

tests/nodes/atomShake.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+        auto density = insertNode_->findInput("Density");
+        density->set<Number>(9.99999642e-05);
```
trisyoungs: As part of the tidy up of our tests, I think we can safely revert this back to a value of 0.1 - that should be sufficient!

tests/testData.h
```diff
@@ -703,6 +705,46 @@ const Species &benzeneSpecies()
...
+
+    auto hW1 = std::make_shared<AtomType>("HW");
+    hW1->setZ(Elements::Element::H);
+    species->addAtom(Elements::Element::H, Vector3{-0.257130, 0.637822, 0.198948}, 0.41, hW1);
+
+    auto hW2 = std::make_shared<AtomType>("HW");
+    hW2->setZ(Elements::Element::H);
+    species->addAtom(Elements::Element::H, Vector3{0.100387, -0.485320, -0.516624}, 0.41, hW2);
+
+    species->simplifyAtomTypes();
```
trisyoungs: I'd suggest we clean this up and use a bit of trig (as in the function below). Also, single atom type for both H to start with:

```suggestion
    auto oW = std::make_shared<AtomType>(Elements::Element::O, "OW");
    species->addAtom(Elements::Element::O, {}, -0.82, oW);

    auto hW = std::make_shared<AtomType>(Elements::Element::H, "HW");
    species->addAtom(Elements::Element::H, {1, 0, 0}, 0.41, hW);
    species->addAtom(Elements::Element::H, {cos(DissolveMath::toRadians(113.24)), sin(DissolveMath::toRadians(113.24)), 0.0},
                     0.41, hW);
```

tests/testData.h
```diff
@@ -703,6 +705,46 @@ const Species &benzeneSpecies()
...
+
+    assert(species->atom(0).Z() == Elements::Element::O);
+    assert(species->atom(1).Z() == Elements::Element::H);
+    assert(species->atom(2).Z() == Elements::Element::H);
+
+    // Apply intramolecular terms
+    species->addBond(0, 1).setInteractionFormAndParameters(BondFunctions::Form::Harmonic, "k=4431.53 eq=1.012");
+    species->addBond(0, 2).setInteractionFormAndParameters(BondFunctions::Form::Harmonic, "k=4431.53 eq=1.012");
```
trisyoungs: Just reverting to SPC/Fw default values:
```suggestion
    species->addBond(0, 1).setInteractionFormAndParameters(BondFunctions::Form::Harmonic, "k=4431.53 eq=1.0");
    species->addBond(0, 2).setInteractionFormAndParameters(BondFunctions::Form::Harmonic, "k=4431.53 eq=1.0");
```

tests/nodes/atomShake.cpp
```diff
@@ -0,0 +1,104 @@
...
+    // Check the geometry of the molecule
+    auto cfg = configurationNode_->getOutputValue<Configuration *>("Configuration");
+    auto *box = cfg->box();
+    auto &atoms = cfg->atoms();
+    auto rMin01 = box->minimumDistance(atoms[0].r(), atoms[1].r());
+    auto rMin02 = box->minimumDistance(atoms[0].r(), atoms[2].r());
+    auto angle102 = box->angleInDegrees(atoms[1].r(), atoms[0].r(), atoms[2].r());
+    EXPECT_NEAR(rMin01, 1.012, 1.0e-4);
+    EXPECT_NEAR(rMin02, 1.012, 1.0e-4);
+    EXPECT_NEAR(angle102, 113.24, 1.7e-2);
```
trisyoungs: Here we should test for the default forcefield geometry as defined:

```suggestion
    EXPECT_NEAR(rMin01, 1.0, 1.0e-4);
    EXPECT_NEAR(rMin02, 1.0, 1.0e-4);
    EXPECT_NEAR(angle102, 113.24, 1.7e-3);
```

tests/nodes/atomShake.cpp
```diff
@@ -0,0 +1,104 @@
...
+    protected:
+    // We need a CoreData and Dissolve definition to properly instantiate DissolveGraph at present.
+    CoreData coreData_;
+    Dissolve dissolve_;
+    DissolveGraph root_;
+    ConfigurationNode *configurationNode_{nullptr};
+    SpeciesNode *speciesNode_{nullptr};
+    AtomicMCNode *atomicMCNode_{nullptr};
+    InsertNode *insertNode_{nullptr};
+    LoopGraph *loop_{nullptr};
```
trisyoungs: Related to my comment about using `TEST_F`, since this is a specific test of one Node and there are no variations we could do without these private vars completely and just store temporary vars when we need to.

tests/nodes/atomShake.cpp
```diff
@@ -0,0 +1,104 @@
...
+};
+
+TEST_F(AtomShakeTest, Water)
+{
+    auto nLoops = 100;
+    createGraph(nLoops);
+    ASSERT_TRUE(loop_->run() == NodeConstants::ProcessResult::Success);
```
trisyoungs: With the suggestions to this PR as-is there is no work for the AtomShake to do since the molecule is created with the perfect geometry and the test checks for that geometry.  What we could do is run a single loop iteration of the simulation at 300 K to "rough up" the molecule a bit, and then run the loop for 100 iterations at 0 K to minimise it. I think this makes for a nicer test!

src/nodes/atomicMC/process.cpp
```diff
@@ -22,7 +22,7 @@ NodeConstants::ProcessResult AtomicMCNode::process()
 
     // Retrieve control parameters from Configuration
     const auto termScale = 1.0;
-    const auto rRT = 1.0 / (.008314472 * targetConfiguration_->temperature());
+    const auto rRT = 1.0 / (.008314472 * temperature_.asDouble());
```
trisyoungs, line 25: I meant to mention this but forgot - if T is zero we obviously get a `nan` here, and it's always been like that without causing a problem.  However, I think I would prefer to avoid introducing `nan`s wherever we can, so would be tempted to change this line:
```suggestion
    const auto rRT = temperature_.asDouble() > 0.0 ? 1.0 / (.008314472 * temperature_.asDouble()) : 0.0;
```

src/nodes/dissolve.cpp
```diff
@@ -50,12 +50,11 @@ std::unique_ptr<EnergyKernel> DissolveGraph::prepareEnergyCalculation(Configurat
     // Generate configuration potential map
     PotentialMap potentialMap(atomTypes, pairPotentialStore(), pairPotentialRange());
 
-    // Regenerate cells
-    cfg->cells().generate(cfg->box(), cfg->requestedCellDivisionLength(), potentialMap.range());
+    // Regenerate cells in the configuration if necessary
+    // TODO
```
```suggestion
```

trisyoungs (approved):
Two minor suggestions (first one might break the formatting, so best check if you include it) but looks good to go!

PR #2287, opened 2026-01-06
feat: LJ+q parameters for SPC/Fw water.

trisyoungs:


RobBuchananCompPhys (approved):
Looks good! Thought crossed my mind we may want to modularise this in some way to limit repetitive calls to these methods, but as its only for two atoms this should be fine.

trisyoungs:
As we move on and introduce other test species this is probably a wise move, and we can refactor this one at the same time.

PR #2288, opened 2026-01-06
feat: Stop regenerating cells all the time

trisyoungs:


src/classes/cellArray.cpp
```diff
@@ -190,11 +190,16 @@ bool CellArray::minimumImageRequired(const Cell &a, const Cell &b) const
 // Generate Cells for Box
 bool CellArray::generate(const Box *box, double cellSize, double pairPotentialRange)
 {
+    // We need to regenerate the cell array only if it is currently empty or the pair potential range has changed
+    if (!cells_.empty() && pairPotentialRangeCreatedAt_ && pairPotentialRangeCreatedAt_.value() == pairPotentialRange)
+        return true;
+
```
RobBuchananCompPhys, line 196:
```suggestion
    if (!cells_.empty() &&  pairPotentialRangeCreatedAt_.value_or(static_cast<double>(std::nan())) == pairPotentialRange)
        return true;

```
I think this could be more concise using `value_or` to compare with a sort of "null" value (if no `pairPotentialRangeCreatedAt_`) i.e. `std::nan`

RobBuchananCompPhys (approved):
Approved with suggested change.

src/classes/cellArray.cpp
trisyoungs: I was trying to avoid using a magic number (`-1` obviously!) in this check, hence the initial test for the valid optional.  Does using `std::nan()` still count as a magic number?

RobBuchananCompPhys: Not quite sure how to answer that! But if -1 is not considered "in range", why not use that as a default?

trisyoungs: For a while we've been trying to remove uses of magic `-1` to mean "not a sensible value because it must be a positive number if it has been set". Rather than testing against some obviously wrong number - `-1` or `nan` - I just reasoned that checking for any number at all explicitly first was a more correct approach!

PR #2289, opened 2026-01-07
fix: AtomicMC energy updates

trisyoungs:
Fixes an issue when running `AtomicMC` with more than one shake per atom and updates random unit vector generation.

### DIFF
--- a/src/math/vector3.cpp
+++ b/src/math/vector3.cpp
@@ -405,14 +405,13 @@ void Vector3::orthogonalise(const Vector3 &reference1, const Vector3 &reference2
 // Prints the contents of the vector
 void Vector3::print() const { std::cout << std::format("{} {} {}", x, y, z) << std::endl; }
 
-// Generate random unit vector
-void Vector3::randomUnit()
+// Generate random unit vector (on a unit sphere)
+Vector3 Vector3::randomUnit()
 {
-    // Generates a random unit vector
-    x = DissolveMath::random() - 0.5;
-    y = DissolveMath::random() - 0.5;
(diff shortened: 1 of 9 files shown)


src/math/vector3.cpp
```diff
@@ -405,14 +405,13 @@ void Vector3::orthogonalise(const Vector3 &reference1, const Vector3 &reference2
...
-    y = DissolveMath::random() - 0.5;
-    z = DissolveMath::random() - 0.5;
-    normalise();
+// Generate random unit vector (on a unit sphere)
+Vector3 Vector3::randomUnit()
+{
+    auto y = 2.0 * (DissolveMath::random() - 0.5);
+    auto r = sqrt(1 - y * y);
+    auto lambda = 2.0 * M_PI * (DissolveMath::random() - 0.5);
+    return {r * sin(lambda), y, r * cos(lambda)};
```
RobBuchananCompPhys, line 414: Interesting use of variable name `lambda` for an angle increment in spherical coords. To me it implies possibly a distance, or, wavelength - which may make sense in the latter case. Dunno, I'd feel more at home with perhaps `dAngle`, or `theta`/`phi` even.

trisyoungs: From the site I reworked this code from the third parameter here is longitude which is apparently given the symbol `lambda`.  I agree that I'd expect to see `theta` or `phi` here because you'd assume the use of spherical coordinates, but not this time!

RobBuchananCompPhys (approved):
Approved with a suggestion on mathematical notation.

PR #2290, opened 2026-01-07
refactor: Loop graph should count from 0 (no runs performed)

RobBuchananCompPhys:
The main change here is the refactor of `LoopGraph` to take parameter `N` equal to zero to mean "no run", and `N` < 1 to be feedback (loops) of the graph. As well as this, renaming of variables inside `LoopGraph` to clarify intention.


src/nodes/iterableGraph.cpp
```diff
@@ -5,7 +5,7 @@
 
 LoopGraph::LoopGraph(Graph *parentGraph) : Graph(parentGraph)
 {
-    addOption<Number>("NLoops", "Number of loops (iterations) to perform", nLoops_);
+    addOption<Number>("NLoops", "Number of loops (iterations) to perform", nIterations_);
```
trisyoungs, line 8: Didn't we talk about renaming `NLoops` to just `N` at some point in the past, or was that a xmas fever dream? Also, the description refers still to loops _and_ iterations - I think we should say "Number of times to iterate the graph" or similar.

src/nodes/loopGraph.cpp
```diff
@@ -5,7 +5,7 @@
 
 LoopGraph::LoopGraph(Graph *parentGraph) : Graph(parentGraph)
```
trisyoungs: Given the move towards "iterations" over "loops", would we want to rename the node itself?  `IterableGraph` for instance?  The user-friendly name could be "Iterator".

src/nodes/loopGraph.cpp
```diff
@@ -23,10 +23,10 @@ std::string_view LoopGraph::type() const { return "Loop"; }
 std::string_view LoopGraph::summary() const { return "Loop the contained graph"; }
 
 // Increment the loop counter
-void LoopGraph::increment() { loopCounter_++; }
+void LoopGraph::increment() { ++i_; }
```
trisyoungs: Such a simple function, and only called internally, so I would suggest removing it.

src/nodes/loopGraph.cpp
```diff
@@ -118,7 +118,7 @@ LoopEdge *LoopGraph::findLoopEdge(const EdgeDefinition &definition) const
 }
 
 // Reset the loop counter to zero
-void LoopGraph::resetLoopCounter() { loopCounter_ = 0; }
+void LoopGraph::reset() { i_ = 0; }
```
trisyoungs: Same here.

src/nodes/iterableGraph.cpp
```diff
@@ -127,27 +127,26 @@ void LoopGraph::resetLoopCounter() { loopCounter_ = 0; }
...
+            return NodeConstants::ProcessResult::Failed;
 
         increment();
     }
 
-    resetLoopCounter();
+    reset();
 
-    return Graph::process();
+    return NodeConstants::ProcessResult::Success;
```
trisyoungs, line 133: This feels like it can be more concisely written as an `for` loop to me, e.g.:
```
auto N = nIterations_.asInteger();
for (i_ == 0; i < N; ++i_)
{
    graphStatus = Graph::process();
    if (Graph::process() == NodeConstants::ProcessResult::Failed)

    // If there is an iteration after this one, push loopbacks
    if (i_ < (N - 1))
    {
        if (loopBacks_->run() == NodeConstants::ProcessResult::Failed)
    }
}
```
I also think it's best not to immediately reset the loop counter at the end of the loop, since it could be useful to know at which iteration a loop might fail.

src/nodes/loopGraph.cpp
```diff
@@ -127,27 +127,26 @@ void LoopGraph::resetLoopCounter() { loopCounter_ = 0; }
...
-        auto graphStatus = NodeConstants::ProcessResult::Unchanged;
+    const auto maxIters = nIterations_.asInteger();
 
-        // Don't run loopbacks on 0th iteration
-        if (loopCounter_ >= 1)
+    // Max iterations equal to zero constitutes no run of the loop graph
+    while (maxIters > 0 && i_ <= maxIters)
+    {
+        // Current iteration i greater than one consitutes a feedback loop
+        if (i_ > 1)
```
trisyoungs: Shouldn't this run on i_ == 1 as well?

src/nodes/iterableGraph.cpp
RobBuchananCompPhys: This has already been done in the `atomShake-unittest` PR, I will keep that current change (relative to that branch) when rebasing

src/nodes/loopGraph.cpp
RobBuchananCompPhys: Don't think so, iteration `i_ == 1` would constitute a standard traversing of the internal nodes in the graph, using the external inputs from the parent graph. On subsequent iterations, we want to pull in the loop backs and feed those into the inputs.

src/nodes/iterableGraph.cpp
RobBuchananCompPhys: Hmm, not sure this set of changes makes thing more clear. 
For instance, regarding the reset, it may be preferable to record a failed iteration in a separate variable, I think we do need to reset the counter in order to get the correct behaviour where:
i = 0: no run
i = 1: standard run with external inputs
i > 1: run with inputs from loopbacks

This could also be fixed by resetting the counter at the beginning of the loop..

src/nodes/loopGraph.cpp
RobBuchananCompPhys: I think - yes you are correct, this renaming should take place. However this implies we might need to rename alot of assets inside the class (i.e. all the various members referring to "Loop..."). I think many of those member names already work well, for instance `LoopEdge` is preferable to `IterableEdge` for some reason! I'd say we keep those, but rename the class as you have suggested.

trisyoungs: Agreed.

src/nodes/iterableGraph.cpp
trisyoungs: But I think the use of the `for` loop handles all this seamlessly with `i_` containing the last iteration run, unless I'm missing something.   If `N` is zero, `i_` gets initialised to zero, no loopbacks pulled. If `N` is 1 then `i_` ends up as `1` after the graph is processed (but loopbacks are not used) and for `N >= 2` you get `i_` as the number of successful iterations and loopback inputs for everything after the first (zeroth) iteration.

src/nodes/loopGraph.cpp
trisyoungs: But `i_` starts from zero, so this would only be true after two iterations / processings of the graph (at `i_ == 0` and `i_ == 1`) had been performed, and meaning loopbacks were not used for one iteration where they should have been (`i_ == 1`).

tests/nodes/loop.cpp
```diff
@@ -105,11 +140,11 @@ TEST_F(LoopGraphTest, NoFeedback)
     EXPECT_EQ(loop_->proxyOutputs().versionIndex(), 0);
     EXPECT_EQ(y_->versionIndex(), 0);
 
-    // Loopbacks node only runs on iteration > 0
+    // Loopbacks node only runs on iteration i > 0
     EXPECT_EQ(loop_->loopBacks()->versionIndex(), -1);
```
trisyoungs:
```suggestion
    EXPECT_EQ(loop_->loopBacks()->versionIndex(), NodeConstants::InvalidVersion);
```

tests/nodes/loop.cpp
```diff
@@ -92,10 +92,45 @@ TEST_F(LoopGraphTest, NoFeedback)
...
+
+    // Check node versioning
+    EXPECT_EQ(i_->versionIndex(), 0);
+    EXPECT_EQ(loop_->proxyInputs().versionIndex(), -1);
+    EXPECT_EQ(x_->versionIndex(), -1);
+    EXPECT_EQ(loop_->proxyOutputs().versionIndex(), -1);
+    EXPECT_EQ(y_->versionIndex(), 0);
+
+    // Loopbacks node only runs on iteration i > 0
+    EXPECT_EQ(loop_->loopBacks()->versionIndex(), -1);
```
trisyoungs:
```suggestion
    EXPECT_EQ(loop_->proxyInputs().versionIndex(), NodeConstants::InvalidVersion);
    EXPECT_EQ(x_->versionIndex(), NodeConstants::InvalidVersion);
    EXPECT_EQ(loop_->proxyOutputs().versionIndex(), NodeConstants::InvalidVersion);
    EXPECT_EQ(y_->versionIndex(), 0);

    // Loopbacks node only runs on iteration i > 0
    EXPECT_EQ(loop_->loopBacks()->versionIndex(), NodeConstants::InvalidVersion);
```

trisyoungs (approved):

PR #2294, opened 2026-01-08
chore:Fix nix bundle version and fix copyright dates

rprospero:
This PR fixes the bundling issue on the Linux build and should prevent the problem for recurring again in the future.  These changes are in `flake.nix` and `.github/workflows/build/linux/action.yml`

It also updates the copyright dates.

### Short version
While the nix flake ensures that all of our dependencies are locked (preventing them from changing beneath us), that does not help if the dependencies for nix itself decide to break.  

I've now added a fixed bundler to the flake.  This ensures that the nix bundler will always use a known working version.

### Long version

The nix bundler is in a weird area where it is both a part of nix and also a library for nix.  Since it's a part of nix, it uses the latest version of all dependencies by default.  Since it's just a library, there's no check whether the bundler dependencies are valid before a new nix version is released.

In this particular case, it was the *arx* dependency, which packs Dissolve's dependencies into a single compressed file, that was broken.  That break occurred because the haskell compiler team released a new version and the bytestring-num library used a feature that became deprecated.

The solution was to add an explicit bundler to the flake.  By doing this, we fix the versions of all the dependencies of the bundler and don't have to worry about versions changing out from beneath us in the future.


trisyoungs (approved):
I haven't look through all the copyright line changes, but when I did it myself I found a couple of 2024/2025 ones in a few files here and there (qml I think) that aren't picked up by the CI checks.  Saying this for information rather than blocking the merge of the PR!

PR #2301, opened 2026-01-23
feat: NeutronSQNode unit tests

trisyoungs:
## Based on #2300 - do not merge until #2300 is resolved

This PR adds unit tests for the `NeutronSQNode`. One relatively significant change is the way that we now must handle our lists of exchangeable atom types. Previously this was defined as a very literal vector of `AtomType*` but here the preference is to move to a simple list of atom type names. The reason for that is that multiple species may contain atom types of the same name while not explicitly sharing data through the pointer - a name will capture all such occurrences of a given atom type.  This may of course be subject to change as we move forward!


src/base/serialiser.h
```diff
@@ -188,6 +188,13 @@ template <typename... Contexts> class Serialisable
         std::transform(vector.begin(), vector.end(), std::back_inserter(result), toSerial);
         return result;
     }
+    // A helper function to add the elements of a ranged object to a node under a name
+    template <typename Range, typename Lambda> static SerialisedValue fromRange(const Range &range, Lambda toSerial)
+    {
+        SerialisedValue result = SerialisedValue::array_type{};
+        std::transform(range.begin(), range.end(), std::back_inserter(result), toSerial);
+        return result;
+    }
```
rprospero:
```suggestion
    template <std::ranges::input_range Range, typename Lambda> static SerialisedValue fromRange(const Range &range, Lambda toSerial)
    {
        SerialisedValue result = SerialisedValue::array_type{};
        std::ranges::copy(std::ranges::views::transform(range, toSerial), std::back_inserter(result));
        return result;
    }
```

A couple of small changes:

1. Specify that the `Range` template type must satisfy the `input_range` concept, which helps narrow down template errors.
2. Use range based transform and copy (mostly to avoid explicit `begin` and `end`

This suggestion is entirely optional.

rprospero (approved):
One optional, stylistic suggestion, but otherwise looks good

PR #2305, opened 2026-01-28
build: Fix Windows aqt failure

trisyoungs:


RobBuchananCompPhys (approved):
Sounds like this works then? I'll note this change and investigate whether any thing else needs to be done for the separate MSVC workflow and the `develop.ps1` script.

trisyoungs:
> Sounds like this works then? I'll note this change and investigate whether any thing else needs to be done for the separate MSVC workflow and the `develop.ps1` script.

Yes, I imagine your `develop.ps1` script will need to forcibly install the correct version of `py7zr` too.

PR #2306, opened 2026-01-28
feat:Add forcefield node

rprospero:
This PR adds a forcefield node, as well as a GUI to add the node and a test.  I've also independently tested that the forcefield nodes can be saved to a file.

### DIFF
--- a/tests/gui/forcefieldModel.cpp
+++ b/tests/gui/forcefieldModel.cpp
@@ -11,7 +11,6 @@ namespace UnitTest
 {
 TEST(ForcefieldModel, Basic)
 {
-
     // Stage 1: Simulate GUI interactions to create species
     Graph graph(nullptr);
     GraphModel graphModel;
--- a/tests/gui/forcefieldModel.cpp
+++ b/tests/gui/forcefieldModel.cpp
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#include "data/ff/library.h"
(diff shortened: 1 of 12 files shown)


tests/gui/forcefieldModel.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "gui/models/simpleForcefieldModel.h"
+#include "gui/models/nodeGraph/graphModel.h"
+#include "nodes/forcefield.h"
+#include <gtest/gtest.h>
+
+namespace UnitTest
+{
+TEST(ForcefieldModel, Basic)
+{
+
```
trisyoungs:
```suggestion
```

tests/gui/forcefieldModel.cpp
```diff
@@ -0,0 +1,34 @@
...
+    model.assignModel(&graphModel);
+    model.setName("Kulmala2010");
+    model.create(0, 0);
+
+    ASSERT_EQ(graph.nodes().size(), 3);
+    auto node = dynamic_cast<ForcefieldNode*>(graph.node("Kulmala2010"));
+    ASSERT_TRUE(node);
+    EXPECT_EQ(node->forcefield().get(), ForcefieldLibrary::forcefield("Kulmala2010").get());
+    
+
```
trisyoungs:
```suggestion
```

trisyoungs (approved):
All looks good to me - spotted a couple of blank lines (the police are back! :) ). I wonder where we end up with forcefield handling in Dissolve2, but this will certainly get us moving in the short term.

PR #2307, opened 2026-01-30
refactor: Add HistogramND constructors taking Vector3

trisyoungs:


rprospero (approved):
Approved with hesitation.  I'm not particularly fond of the interface, since it is not obvious that the y component of the vector will correspond with the maximum histogram value.  To be honest, my first assumption was that x was the minimum, z was the maximum, and y was the step size.

However, the only real solution would be creating a new HistogramParams struct with `min`, `max`, and `binWidth`.  I briefly considered just passing in the individual values, but that just silly when we hit the 3D histograms.  I'm not sure how much it is worth the overhead of the extra struct.

trisyoungs:
@rprospero I hear you. This was just a little convenience PR to stop passing huge argument lists of `double`s.  I fully agree that we should think about implementing some of general `QuantityRange` class with specifically-named `_min`, `_max`, and `_delta` variables, as it would get used a lot in the analysis modules options. My new `RangedVector3` I guess is kind of a halfway house...

PR #2308, opened 2026-01-30
feat: Add RangedVector3 class

trisyoungs:
This PR adds in a `RangedVector3` class for use in node options. Previously, limit handling of `Vector3` keywords in Dissolve1 was performed by the `Vector3Keyword` handler on a target `Vector3`, but in DIssolve2 we don't have this structure and all inputs, options must "do their own thing" in toto.

So, here is a small class to allow limiting of a `Vector3`.


tests/math/rangedVector3.cpp
```diff
@@ -0,0 +1,48 @@
...
+    EXPECT_DOUBLE_EQ(v.x(), -1.0);
+    EXPECT_DOUBLE_EQ(v.y(), -2.0);
+    EXPECT_DOUBLE_EQ(v.z(), -3.0);
+
+    v.set({2.0, 1.0, 6.0});
+    EXPECT_DOUBLE_EQ(v.x(), 1.0);
+    EXPECT_DOUBLE_EQ(v.y(), 1.0);
+    EXPECT_DOUBLE_EQ(v.z(), 3.0);
+}
+} // namespace UnitTest
```
RobBuchananCompPhys, line 48: It would be interesting to see a few test cases of `RangedVector3::value()` being used, since it is a public member. I guess my concern is whether we want to keep the return type as `const Vector3`, in case the value is used in a context where strict range is expected, but it could theoretically be expanded beyond as a `Vector3`.

RobBuchananCompPhys (approved):

tests/math/rangedVector3.cpp
trisyoungs: Good catch.  `const &Vector3` is what it should return.

PR #2310, opened 2026-01-30
feat: Add ImportConfigurationTrajectory node

trisyoungs:
As part of work towards reimplementing analysis modules as nodes, this PR adds the ability to import a trajectory (frame) using a node.  No unit test is added, as testing is performed implicitly by the analysis node testing.

### DIFF
--- a/src/nodes/importConfigurationTrajectory.cpp
+++ b/src/nodes/importConfigurationTrajectory.cpp
@@ -21,7 +21,10 @@ ImportConfigurationTrajectoryNode::ImportConfigurationTrajectoryNode(Graph *pare
 
 std::string_view ImportConfigurationTrajectoryNode::type() const { return "ImportTrajectory"; }
 
-std::string_view ImportConfigurationTrajectoryNode::summary() const { return "Import Trajectory from a file."; }
+std::string_view ImportConfigurationTrajectoryNode::summary() const
+{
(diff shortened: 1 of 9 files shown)


src/nodes/importConfigurationTrajectory.cpp
```diff
@@ -0,0 +1,62 @@
...
+    // Outputs
+    addOutput<Configuration *>("Configuration", "Output configuration", configuration_);
+
+    // Serialisable data
+    addSerialisable("filePosition", filePosition_);
+}
+
+std::string_view ImportConfigurationTrajectoryNode::type() const { return "ImportTrajectory"; }
+
+std::string_view ImportConfigurationTrajectoryNode::summary() const { return "Import Trajectory from a file."; }
```
RobBuchananCompPhys: Is this description totally accurate, given that we are in particular reading a single frame (I think) here? Should we expand it a little, as it's not entirely clear until you read the code.

RobBuchananCompPhys (approved):

src/nodes/importConfigurationTrajectory.cpp
trisyoungs: Fair comment.  How about "Import configuration coordinates from sequential frames of a trajectory"?

PR #2311, opened 2026-01-30
feat: Dissolve2 XRaySQNode

RobBuchananCompPhys:


src/nodes/xRaySQ/helpers.cpp
trisyoungs, line 1: Not one to address in this PR, but I think in general I would push to have each node contained in a single header and source file pair, rather than splitting out functions as we used to do for modules. For me it hinders code navigation when I'm looking for the functions within a module/node, and the monolithic source file shouldn't be too absurdly long. Anyway, one for us to discuss at some point.

trisyoungs (approved):

PR #2312, opened 2026-01-30
feat: Angle analysis node

trisyoungs:
This PR implements `AngleNode`, replacing `AngleModule`. A `SpeciesSItes` class is introduced to handle resolution of sites at input load time, and is very similar to the recently updated `IsotopologueSet` class.


src/nodes/angle.cpp
```diff
@@ -0,0 +1,248 @@
...
+    if (!DataExporter::exportData(angleABC_, exportFileAndFormatAngle_))
+        return NodeConstants::ProcessResult::Failed;
+
+    // Save DAngle((A-B)-C) data?
+    if (!DataExporter::exportData(dAngleAB_, exportFileAndFormatDAngleAB_))
+        return NodeConstants::ProcessResult::Failed;
+
+    // Save DAngle(A-(B-C)) data?
+    if (!DataExporter::exportData(dAngleBC_, exportFileAndFormatDAngleBC_))
+        return NodeConstants::ProcessResult::Failed;
```
RobBuchananCompPhys: This is something that I've also questioned while working on the `SiteRDFNode` - do we not want this to be an option that is satisfied depending on the value of some `save_` member? Right now, there is no way to disable exporting of the data even though the comment suggests there might be.

tests/speciesData.h
```diff
@@ -69,6 +69,16 @@ inline SpeciesNode *createWater(Graph *parentGraph)
...
+    // Create sites
+    species->addSite("Origin")->setStaticIndices({1}, {0, 2}, {2});
+    species->addSite("O")->setStaticIndices({1}, {}, {});
+    auto hSite = species->addSite("H");
+    hSite->setType(SpeciesSite::SiteType::Dynamic);
+    hSite->setDynamicElements({Elements::Element::H});
+    auto comSite = species->addSite("COM");
+    comSite->setStaticIndices({0, 1, 2}, {}, {});
+    comSite->setOriginMassWeighted(true);
+
```
RobBuchananCompPhys, line 81: We may want to give `createWater` a `std::optional<SpeciesSites>` argument, from which different analysis sites can be set, while the above could be a default setup.

RobBuchananCompPhys (approved):

PR #2313, opened 2026-02-04
fix: Iterable graph bug on Graph::process empty/non-empty edges

RobBuchananCompPhys:
This PR addresses the seemingly odd behaviour of `IterableGraph`, in which a series of nodes internal to the graph and connected _only_ to the graph's proxy inputs, leads to the execution of all but the last node in the series. 

Turns out this was caused by the requirement for a node's [output edges to be empty](https://github.com/disorderedmaterials/dissolve/pull/2313/files#:~:text=if%20(!node%2D%3E,node%2D%3EoutputEdges().empty())) in order to process the node (akin to traversing the node graph until a node is found that is not a dependency for another node), which was actually incorrectly coded and itself has been addressed in this PR.

The previous state of the logic meant that the final node in the chain was never processed. Additionally, the `Graph::outputEdges` method had to be overridden for the `LoopBacksNode`, since the loopBacks do not show up in the `outputEdges_` member, which caused nodes in connected `IterableGraph`s to fail in the same way.

### DIFF
--- a/src/nodes/loopBack.cpp
+++ b/src/nodes/loopBack.cpp
@@ -65,10 +65,6 @@ Node::EdgeMap &LoopBacksNode::outputEdges()
         loopEdges_.insert_or_assign(output, edges);
     }
 
-    std::vector<Edge *> edges;
-    for (const auto &edge : loop->loopEdges())
-        edges.push_back(edge.get());
-
     return loopEdges_;
 }
 
--- a/src/nodes/loopBack.cpp
+++ b/src/nodes/loopBack.cpp
@@ -46,6 +46,32 @@ NodeConstants::ProcessResult LoopBacksNode::run()
     return status;
 }
 
+// Get the outgoing edges from this node
+Node::EdgeMap &LoopBacksNode::outputEdges()
+{
+    auto loop = static_cast<IterableGraph *>(parentGraph_);
+
+    for (const auto &edge : loop->loopEdges())
+    {
+        auto output = edge->sourceOutput().name();
+        auto it = loopEdges_.find(output);
+
+        std::vector<Edge *> edges;
+        if (it != loopEdges_.end())
+            edges = it->second;
+
+        edges.push_back(edge.get());
+
+        loopEdges_.insert_or_assign(output, edges);
+    }
+
+    std::vector<Edge *> edges;
+    for (const auto &edge : loop->loopEdges())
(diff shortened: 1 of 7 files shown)


tests/nodes/loop.cpp
```diff
@@ -73,6 +73,65 @@ class IterableGraphTest : public ::testing::Test
...
+    EXPECT_TRUE(b->versionIndex() == 1);
+    EXPECT_TRUE(c->versionIndex() == 0);
+    ASSERT_EQ(res3, 4);
+
+    // Run 100 times
+    ASSERT_TRUE(loop->setOption<Number>("N", 100));
+    EXPECT_EQ(loop->run(), NodeConstants::ProcessResult::Success);
+    ASSERT_EQ(a->getOutputValue<Number>("Result").asInteger(), 2);
+    ASSERT_EQ(b->getOutputValue<Number>("Result").asInteger(), 3);
+    ASSERT_EQ(c->getOutputValue<Number>("Result").asInteger(), 4);
```
trisyoungs, line 144: Worth repeating the `versionIndex()` checks here too.

src/nodes/loopBack.cpp
```diff
@@ -46,6 +46,32 @@ NodeConstants::ProcessResult LoopBacksNode::run()
     return status;
 }
 
+// Get the outgoing edges from this node
```
trisyoungs, line 49: I _think_ I understand why this is necessary, but a couple of comments in the function describing what's going on will help when our future selves revisit the code!

src/nodes/loopBack.cpp
```diff
@@ -46,6 +46,32 @@ NodeConstants::ProcessResult LoopBacksNode::run()
...
+            edges = it->second;
+
+        edges.push_back(edge.get());
+
+        loopEdges_.insert_or_assign(output, edges);
+    }
+
+    std::vector<Edge *> edges;
+    for (const auto &edge : loop->loopEdges())
+        edges.push_back(edge.get());
```
trisyoungs: Does this bit of code have a purpose? Just seems to be creating a local vector...

trisyoungs (approved):

src/nodes/loopBack.cpp
RobBuchananCompPhys: It's unused, I just removed it.

PR #2319, opened 2026-02-11
refactor: Forcefield unit tests

trisyoungs:
This PR moves the forcefield unit tests over to Dissolve2. In reality this is nothing to do with Graphs/Nodes and more to do with removing our dependence on old-style input files as sources of input.  Here those input files are converted to the relevant XYZ coordinates files and removed, with the tests refactored accordingly.

It was incredibly tedious to write, and it will be incredibly tedious to review. Advise checking explicitly the first three commits as the rest are just repetition.


src/classes/species.cpp
```diff
@@ -205,13 +205,54 @@ void Species::print() const
...
+
+    // Assign atom types to the species
+    if (!ff->assignAtomTypes(this, coreData, Forcefield::TypeAll, false).empty())
+        return false;
+
+    // Assign intramolecular terms
+    if (!ff->assignIntramolecular(this))
+        return false;
+
+    return true;
```
RobBuchananCompPhys, line 226:
```suggestion
    if (!(ff && ff->assignAtomTypes(this, coreData, Forcefield::TypeAll, false).empty() && ff->assignIntramolecular(this)))
        return Messenger::error("Failed to apply forcefield terms for {(ff ? ff->name() : "Null")}\n");
    return true;
```

RobBuchananCompPhys (changes requested):

tests/ff/assignment-SPCFw.cpp
```diff
@@ -1,28 +1,32 @@
...
-    auto *sp = systemTest.coreData().findSpecies("Water");
-    ASSERT_TRUE(sp);
-    systemTest.checkSpeciesAtomType(sp, 0, "OW");
-    systemTest.checkSpeciesAtomType(sp, 1, "HW");
-    systemTest.checkSpeciesAtomType(sp, 2, "HW");
-    systemTest.checkSpeciesIntramolecular(sp, {0, 1}, {BondFunctions::Form::Harmonic, "k=4431.53 eq=1"});
-    systemTest.checkSpeciesIntramolecular(sp, {1, 0, 2}, {AngleFunctions::Form::Harmonic, "k=317.5656 eq=113.24"});
+    systemTest.checkSpeciesAtomType(&species, 0, "OW");
+    systemTest.checkSpeciesAtomType(&species, 1, "HW");
+    systemTest.checkSpeciesAtomType(&species, 2, "HW");
```
RobBuchananCompPhys: Wondering if we could save lines of C++ here and in similar tests, by utilising a `std::map<int, std::string>` like 
```
    systemTest.checkSpeciesAtomType(&species, { {0, "OW"}, {1, "HW"}, {2, "HW"});
```

src/classes/species.cpp
trisyoungs: I completely see why you've suggested this change - I would argue that it reduces the readability of the code more than it neatens things up?

tests/ff/assignment-SPCFw.cpp
trisyoungs: Yes, this is a good idea!

RobBuchananCompPhys (approved):

PR #2321, opened 2026-02-13
refactor: Regression function naming and unit test

trisyoungs:
A quick PR to rename some functions in the `Regression` namespace (they're not returning the linear regression, just the slope / gradient), add a function to perform gradient calculation on a vector of (assumed) equally-spaced values which is useful, and add a unit test to cover it.


src/math/regression.cpp
```diff
@@ -44,7 +37,42 @@ double linear(const Data1D &data, int nSamples, double &yBar)
...
+    for (auto n = y.size() - nSamples; n < y.size(); ++n)
+        yBar += y[n];
+    yBar /= nSamples;
+
+    // Determine Sx, Sy, and Sxy
+    double dx, dy;
+    for (auto n = 0; n < nSamples; ++n)
+    {
+        dx = n - xBar;
+        dy = y[y.size() - nSamples + n] - yBar;
```
RobBuchananCompPhys:
```suggestion
        auto i = y.size() - nSamples + n;
        dy = y[i] - yBar;
```
I think that making this into a local variable could be useful for debugging.

RobBuchananCompPhys (approved):
Looks good! 👍

PR #2322, opened 2026-02-13
feat: Species site node

rprospero:
This adds a node for SpeciesSite, since it will be needed for AvgMol.  I considered just having the AvgMol node construct the site itself, but I figured that it would be better to make the node once instead of writing that boilerplate for every node which needs a species site.


rprospero:
I'm rescinding this PR because we decided that SpeciesSite should not be a *node*, but rather an option that is selected.

PR #2323, opened 2026-02-19
feat: SiteRDF Node (VERSION 2)

RobBuchananCompPhys:


src/nodes/siteRDF.cpp
```diff
@@ -0,0 +1,158 @@
...
+    addOption("ExcludeSameMolecule", "Whether to exclude correlations between sites on the same molecule",
+              excludeSameMolecule_);
+    addOption("RangeAEnabled", "Whether calculation of the second coordination number is enabled", rangeEnabled_[0]);
+    addOption("RangeA", "Distance range for first coordination number", rangeA_);
+    addOption("RangeBEnabled", "Whether calculation of the second coordination number is enabled", rangeEnabled_[1]);
+    addOption("RangeB", "Distance range for second coordination number", rangeB_);
+    addOption("RangeCEnabled", "Whether calculation of the third coordination number is enabled", rangeEnabled_[2]);
+    addOption("RangeC", "Distance range for third coordination number", rangeC_);
+    addOption("Instantaneous", "Whether to calculate instantaneous coordination numbers rather than forming an average",
+              instantaneous_);
```
trisyoungs, line 38: It's this kind of option that we will wrap into our "accumulation or not" framework, whatever it may be!

tests/nodes/siteRDF.cpp
```diff
@@ -0,0 +1,306 @@
...
+    // Return pointer to DissolveGraph
+    DissolveGraph *root() { return &(testData_->graphRoot); }
+    // create graph
+    void createGraph(const std::string &trajectoryFilename, TrajectoryImportFileFormat::TrajectoryImportFormat format,
+                     std::vector<std::string> siteNames)
+    {
+        // Create graph test data, resetting unique pointer if necessary
+        if (testData_.get())
+            testData_.reset();
+
```
trisyoungs: Should never be needed.
```suggestion
```

trisyoungs (changes requested):
This all looks great, but I do have a bit of a gripe with the way the unit tests are structured.  There are several functions which perform tests on the presence of objects and then just return data from within the graph - I'm not sure who this is protecting! I think I would prefer a single function doing all the setup, rather than separating it out like this. The partial creation of the `SiteRDF` nodes also feels a bit half-way with a lot of setting going on in the individual tests.  What could work well is a single function in the fixture called `createSiteRDFNode` which takes as arguments the two site names (thus also giving you a unique name for the node), the distance range, and the exclude same molecule flag, and returns the pointer to the node (which you then have in-hand for retrieval of the data for testing. The "RangeBEnabled" flag is set a lot but is only relevant in a couple of the tests, so this could be set as an option using the returned pointer only as and when required.

trisyoungs (approved):

PR #2324, opened 2026-02-24
refactor: Pair Potential Core Definitions

trisyoungs:
This PR moves a few core pair potential-related properties from `Dissolve` into the `PairPotential` class itself as static members. There was no real reason to have them separated, and it cleans up a bunch of little bits of code.  The "Force Choice" option is gone because, frankly, it was annoying to basically be asked twice if that was really what you wanted to do!


src/main/keywords_pairPotentials.cpp
```diff
@@ -74,21 +75,21 @@ bool PairPotentialsBlock::parse(LineParser &parser, Dissolve *dissolve)
...
+                PairPotential::setRange(PairPotential::range(), parser.argd(1));
                 break;
             case (PairPotentialsBlock::EndPairPotentialsKeyword):
                 Messenger::print("Found end of {} block.\n",
                                  BlockKeywords::keywords().keyword(BlockKeywords::PairPotentialsBlockKeyword));
                 blockDone = true;
                 break;
             case (PairPotentialsBlock::ForceChargeSourceKeyword):
-                dissolve->setForceChargeSource(parser.argb(1));
+                Messenger::warn("The ForceChargeSource keyword is now redundant.\n");
```
rprospero, line 86: Should we just drop the ForceCharge keyword from the enum entirely?  I don't think that there should be legacy files that would contain the keyword.

rprospero (approved):
I'm largely happy with this.  The most actionable comment I have below.  The only other concern that I have is a bit more abstract.  A few years ago, we'd discussed removing static members so that you could have multiple, independent `Dissolve` objects operating in the same application.  We used this functionality briefly in the loading of species from XML.  Specifically, the species are loaded into a separate `Dissolve` versions that can then ensure that they support various forcefields.  Since this doesn't apply to PairPotential, it's likely not an issue, but I thought it was worth bringing up.

src/main/keywords_pairPotentials.cpp
trisyoungs: Left in for now to protect any existing input files related to test data, but this entire routine will disappear in the near future.

PR #2330, opened 2026-03-05
feat: Add SpeciesKernel

trisyoungs:
This PR adds a `SpeciesKernel` specifically for the calculation of energies and forces within a `Species`, leaving the larger `EnergyKernel` and `ForceKernel` classes to focus exclusively on `Configuration`s. This involved a lot of (beneficial) refactoring along the way, so not the most interesting PR in the world, but will make subsequent required efforts on `EnergyNode` and the intended absence of `ForcesNode` way smaller.

Oh, and we also remove `GeometryOptimisationModule` - never used AFAIK.


src/kernels/common.h
```diff
@@ -0,0 +1,102 @@
...
+        : bondEnergy(bond), angleEnergy(angle), torsionEnergy(torsion), improperEnergy(improper)
+    {
+    }
+
+    // Energy values
+    double bondEnergy{0.0};
+    double angleEnergy{0.0};
+    double torsionEnergy{0.0};
+    double improperEnergy{0.0};
+
```
RobBuchananCompPhys, line 33:
```suggestion
    GeometryEnergyValue(double bond = 0.0, double angle = 0.0, double torsion = 0.0, double improper = 0.0)
        : bondEnergy_(bond), angleEnergy_(angle), torsionEnergy_(torsion), improperEnergy_(improper)
    {
    }
    
    private:
    // Energy values
    double bondEnergy_{0.0};
    double angleEnergy_{0.0};
    double torsionEnergy_{0.0};
    double improperEnergy_{0.0};

    public:
```
The public members seem atypical for a class - should it be a `struct GeometryEnergyValue`?

src/kernels/species.cpp
```diff
@@ -0,0 +1,247 @@
...
+                                                    return 0.0;
+
+                                                // Get intramolecular scaling of atom pair
+                                                auto &&[scalingType, elec14, vdw14] = i.scaling(&j);
+                                                if (scalingType == SpeciesAtom::ScaledInteraction::NotScaled)
+                                                    return potentialMap_.energy(&i, &j, r);
+                                                else if (scalingType == SpeciesAtom::ScaledInteraction::Scaled)
+                                                    return potentialMap_.energy(&i, &j, r, elec14, vdw14);
+
+                                                return 0.0;
```
RobBuchananCompPhys:
```suggestion
                                                switch(scalingType) {
                                                    case (SpeciesAtom::ScaledInteraction::NotScaled):
                                                        return potentialMap_.energy(&i, &j, r);
                                                    case (SpeciesAtom::ScaledInteraction::Scaled):
                                                        return potentialMap_.energy(&i, &j, r, elec14, vdw14);
                                                    default:
                                                        return 0.0;
                                                }
```

src/kernels/species.cpp
```diff
@@ -0,0 +1,247 @@
...
+            return;
+
+        auto &&[scalingType, elec14, vdw14] = i.scaling(&j);
+        if (scalingType == SpeciesAtom::ScaledInteraction::Excluded)
+            return;
+
+        // Determine final forces
+        auto vecij = box_->minimumVector(i.r(), j.r());
+        auto magjisq = vecij.magnitudeSq();
+        if (magjisq > cutoffDistanceSquared_)
```
RobBuchananCompPhys:
```suggestion
        auto sqMagJI = minVecIJ.magnitudeSq();
        if (sqMagJI > cutoffDistanceSquared_)
```
Perhaps this is cosmetic, but I sense some slightly obscure variable names here - not sure I could glance at this code and feel confident about what these are. May I suggest some other names?

src/kernels/species.cpp
```diff
@@ -0,0 +1,247 @@
...
+            return;
+
+        auto &&[scalingType, elec14, vdw14] = i.scaling(&j);
+        if (scalingType == SpeciesAtom::ScaledInteraction::Excluded)
+            return;
+
+        // Determine final forces
+        auto vecij = box_->minimumVector(r[indexI], r[indexJ]);
+        auto magjisq = vecij.magnitudeSq();
+        if (magjisq > cutoffDistanceSquared_)
```
RobBuchananCompPhys: as above

RobBuchananCompPhys (approved):
Looking good - a couple of suggestions from me.

src/kernels/common.h
trisyoungs: Yeah, fair.  I went backwards and forwards a bit on this one, but `struct` is probably the better choice. Maintaining the energy values as public members is most important, so this makes sense from the `class` vs `struct` argument.

src/kernels/species.cpp
trisyoungs: I quite agree!  Changed to `magSq` throughout.

PR #2334, opened 2026-03-11
refactor: Node ancestor search

rprospero:
This PR adds the ability for a node to recursively examine its inputs.  A template parameter filters for only nodes of the given type.

### DIFF
--- a/src/nodes/node.cpp
+++ b/src/nodes/node.cpp
@@ -395,7 +395,7 @@ void Node::deserialiseData(const SerialisedValue &node)
         }
 }
 
-std::set<const Node *> Node::ancestors() const
+std::set<const Node *> Node::ancestors_() const
 {
     std::set<const Node *> result{}, stack{this};
     while (!stack.empty())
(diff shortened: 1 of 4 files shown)


src/nodes/node.cpp
```diff
@@ -394,3 +394,21 @@ void Node::deserialiseData(const SerialisedValue &node)
             serialisable->resolve(reachableSpecies);
         }
 }
+
+std::set<const Node *> Node::ancestors_() const
```
```suggestion
// Get all nodes that lead into this node
std::set<const Node *> Node::ancestors_() const
```

src/nodes/node.h
```diff
@@ -92,10 +93,20 @@ class Node : public Serialisable<>
 
         return NodeConstants::ProcessResult::Failed;
     }
+    // Get all nodes that lead into this node
+    std::set<const Node *> ancestors_() const;
```
trisyoungs: This would be the sole example of a function with this `name_` convention in the codebase I think - not a problem per se, but is there an alternative name we could give it to clarify its intent? Perhaps rename the `public` function to `getAncestors()` (which makes it consistent with e.g. `getNode()`) and rename this to `ancestors()`?

trisyoungs (approved):

PR #2335, opened 2026-03-13
refactor: Parameter context

rprospero:
Allows types to define a context type.  Declaring the parameter then requires that the context be passed in, if and only if there *is* a context.

I went with `std::monostate` for two reasons.  First, it's the "official" type of empty data.  As such, it includes such niceties as default and copy constructors.  Secondly, as the official empty data type, it will take up zero bytes the moment that that is possible.  It also gives the compiler the best opportunity to optimise out the context call in the parameter.

### DIFF
--- a/src/classes/speciesSite.h
+++ b/src/classes/speciesSite.h
@@ -15,6 +15,7 @@
 #include <vector>
 
 // Forward Declarations
+class Configuration;
 class LineParser;
 class Site;
 class Species;
@@ -213,8 +214,6 @@ class SpeciesSite : public Serialisable<CoreData &>
     void serialise(std::string tag, SerialisedValue &target) const override;
     void deserialise(const SerialisedValue &node, CoreData &coreData) override;
 };
-
-class Configuration;
 
 template <> struct Context<const SpeciesSite *>
 {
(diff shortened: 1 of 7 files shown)


src/classes/speciesSite.h
```diff
@@ -213,3 +213,10 @@ class SpeciesSite : public Serialisable<CoreData &>
     void serialise(std::string tag, SerialisedValue &target) const override;
     void deserialise(const SerialisedValue &node, CoreData &coreData) override;
 };
+
+class Configuration;
```
trisyoungs: Perhaps move this up so its with the other fwd decs above?

src/nodes/parameter.h
```diff
@@ -134,6 +135,21 @@ class ParameterBase : public Serialisable<>
             throw(std::runtime_error(std::format("ParameterBase::get() failed to cast, name = {}.\n", name_)));
         return cast->getData();
     }
+    // Get the parameter's value
```
```suggestion
    // Get the parameter's context
```

src/nodes/parameter.h
```diff
@@ -269,6 +299,8 @@ template <typename DataClass> class Parameter : public ParameterBase, public std
     protected:
     // Reference to target data
     DataClass &data_;
+    // Reference to target data
```
```suggestion
    // Context for target data
```

trisyoungs (approved):

PR #2339, opened 2026-03-17
feat: CIF handling

RobBuchananCompPhys:
This work aims to encompass much of the existing functionality of the `CIFHandler` and `ImportCIFDialog` interfaces. It does this via passing a pointer to a `CIFContext` (which is a node type alias for `CIFHandler`) between various nodes.
Together these form a sort of CIF-node "family".

- `CIFLoaderNode`: Initialises a `CIFContext` object and reads into it the contents of a .cif file.
- `CIFMolecularSpecies`: Recieves and stores a `CIFContext *`, generating from it the detected molecular species and supercell configuration as output parameters
- `CIFSuperMolecule`: Recieves and stores a `CIFContext *`, generating from it the supercell species
- `CIFPeriodicFramework`: Recieves and stores a `CIFContext *`, generating from it the unit cell species and configuration as output parameters
- `CIFAtomicOptionsNode`, `CIFBondingOptionsNode`, and `CIFNETAOptionsNode` operate on the context pointer and apply various settings to it (for instance `removeAtomics`, `useCIFBondingDefinitions`, and `setMoietyRemovalNeta` etc.`
- `SetCIFAtomGroupActivity` takes the `CIFContext *` and allows for activation/deactivation of chosen `CIFAssembly` `CIFAtomGroup` instances.


src/nodes/cifAtomicOptions.cpp
```diff
@@ -0,0 +1,39 @@
...
+    addInput<CIFLoaderNode::CIFContext *>("CIFContext", "CIF handling context derived from parsing of CIF file", context_);
+
+    // Outputs
+    addOutput<CIFLoaderNode::CIFContext *>("CIFContext", "CIF handling context derived from parsing of CIF file", context_);
+
+    // Options
+    addOption<Number>("OverlapTolerance", "Tolerance for removal of overlapping atoms", overlapTolerance_);
+    addOption<bool>("RemoveAtomics", "Whether to remove free atomic moieties in clean-up", removeAtomics_);
+    addOption<bool>("RemoveWaterAndCoordinateOxygens", "Whether to remove water and coordinated oxygen atoms in clean-up",
+                    removeWaterAndCoordinateOxygens_);
```
trisyoungs: Should be "Coordinated":
```suggestion
    addOption<bool>("RemoveWaterAndCoordinatedOxygens", "Whether to remove water and coordinated oxygen atoms in clean-up",
                    removeWaterAndCoordinatedOxygens_);
```

src/nodes/cifConfiguration.cpp
trisyoungs, line 1: As briefly discussed, it is probably worth having a separate `Configuration`-maker node for each type of crystal cell we're generating (supercell, non-periodic, full species etc.).

src/nodes/cifConfiguration.cpp
```diff
@@ -0,0 +1,51 @@
...
+std::string_view CIFConfigurationNode::summary() const { return "Generate a supercell configuration from a CIF context"; }
+
+// Run main processing
+NodeConstants::ProcessResult CIFConfigurationNode::process()
+{
+    if (!context_)
+    {
+        error("CIF context is null");
+        return NodeConstants::ProcessResult::Failed;
+    }
```
trisyoungs: I suspect that checks like this should be rendered irrelevant with "required" parameter inputs.

src/nodes/cifConfiguration.h
```diff
@@ -0,0 +1,43 @@
...
+    // CIF handler context
+    CIFLoaderNode::CIFContext *context_{nullptr};
+    // Supercell configuration
+    Configuration *supercellConfiguration_{nullptr};
+    // Unit cell species
+    const Species *unitCellSpecies_{nullptr};
+    // Supercell species
+    const Species *supercellSpecies_{nullptr};
+    // Detected molecular species
+    const std::vector<CIFMolecularSpecies> *molecularSpecies_{nullptr};
```
trisyoungs: Should this really be a pointer to a `std::vector`? Why not a plain `std::vector`?

src/nodes/cifStructureCleanup.cpp
trisyoungs, line 1: Here I am in a bit of a quandry.  I wanted to suggest that this node, along with `CIFAtomicOptionsNode`, should be combined into a `CIFCleanupNode`. But, on reflection I think that actually the separation is better like this. However, I would suggest renaming this one (involving NETA) to `CIFCleanupNode` and rename/split `CIFAtomicOptionsNode` into `CIFWaterRemover` and `CIFAtomicRemover` perhaps?  I think this is on the edge of overnodification, but it makes sense to me to do it.

src/nodes/cifAtomicOptions.cpp
```diff
@@ -0,0 +1,39 @@
...
+CIFAtomicOptionsNode::CIFAtomicOptionsNode(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput<CIFLoaderNode::CIFContext *>("CIFContext", "CIF handling context derived from parsing of CIF file", context_);
+
+    // Outputs
+    addOutput<CIFLoaderNode::CIFContext *>("CIFContext", "CIF handling context derived from parsing of CIF file", context_);
+
+    // Options
+    addOption<Number>("OverlapTolerance", "Tolerance for removal of overlapping atoms", overlapTolerance_);
```
trisyoungs: This is a more general option related to the generation of the atomic coordinates from the data in the CIF, so this isn't the best place for it.

src/nodes/cifConfiguration.h
RobBuchananCompPhys: I coded these to be in line with how species generation is handled elsewhere in Dissolve2 nodes, i.e. as `const` pointers. Since in this case, we are not dealing with a `const Species *`, but something quite different - a `std::vector<CIFMolecularSpecies>` - so yes, perhaps in this case, we should just output the container.

src/nodes/cifAtomicOptions.cpp
RobBuchananCompPhys: Good to know - we can find somewhere else for it to be set, when the need arises

src/nodes/cifConfiguration.h
trisyoungs: Ah, I see. I think this needs a bit of thought, but certainly until they become bona-fide `Species` then the full-on vector could be passed as output, rather than the pointer.  The alternative would be to have dynamically-created `const Species *` outputs on the CIF configuration nodes, based on the content of the CIF. Until we try one route out it's hard to say which route is best.

RobBuchananCompPhys: Also, there is the same issue we ran into with requiring the `BraggReflectionVectorNode` - i.e. vector serialisation

trisyoungs: True, but this should never be serialised - its just an output for internal node data.

RobBuchananCompPhys: Sounds good, let's go with a `std::vector<CIFMolecularSpecies>` for now

src/data/spaceGroups.cpp
```diff
@@ -703,4 +703,542 @@ std::string formattedInformation(SpaceGroupId sgid)
...
+                                                      {SpaceGroup_224_2, "SpaceGroup_224_2"},
+                                                      {SpaceGroup_225, "SpaceGroup_225"},
+                                                      {SpaceGroup_226, "SpaceGroup_226"},
+                                                      {SpaceGroup_227_1, "SpaceGroup_227_1"},
+                                                      {SpaceGroup_227_2, "SpaceGroup_227_2"},
+                                                      {SpaceGroup_228_1, "SpaceGroup_228_1"},
+                                                      {SpaceGroup_228_2, "SpaceGroup_228_2"},
+                                                      {SpaceGroup_229, "SpaceGroup_229"},
+                                                      {SpaceGroup_230, "SpaceGroup_230"},
+                                                      {nSpaceGroupIds, "nSpaceGroupIds"}});
```
rprospero, line 1240: I really hope that you didn't have to type this all out by hand!

src/nodes/cifLoader.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <vector>
+
+CIFLoaderNode::CIFLoaderNode(Graph *parentGraph) : Node(parentGraph)
+{
+    // Outputs
+    addPointerOutput<CIFContext>("CIFContext", "CIF handling context derived from parsing of CIF file", context_)
+        ->setFlags({ParameterBase::Required});
+
+    // Option
+    addOption<std::string>("FilePath", "File path", filePath_);
```
rprospero, line 16: I don't want to stop the PR on this, but this will eventually need to change types.  Specifically, we need an official FilePath type for options that are file paths.  That way, instead of the user needing to type in a string, the GUI can show a file dialog.

src/nodes/cifMolecularSpecies.cpp
```diff
@@ -0,0 +1,49 @@
...
+    context_->generate();
+
+    // Get supercell configuration
+    supercellConfiguration_ = context_->generatedConfiguration();
+    supercellConfiguration_->setName(context_->chemicalFormula());
+
+    // Get detected molecular species
+    auto &cifMols = context_->molecularSpecies();
+    molecularSpecies_.clear();
+    std::copy(cifMols.begin(), cifMols.end(), std::back_inserter(molecularSpecies_));
```
rprospero:
```suggestion
    std::ranges::copy(cifMols, std::back_inserter(molecularSpecies_));
```

Possibly a bit cleaner using Ranges.

src/nodes/cifRemoveWater.h
```diff
@@ -0,0 +1,35 @@
...
+    std::string_view summary() const override;
+
+    /*
+     * Definition
+     */
+    private:
+    // CIF handler context
+    CIFLoaderNode::CIFContext *context_{nullptr};
+    // Whether to remove water and coordinated oxygen atoms in clean-up
+    bool RemoveWaterAndCoordinatedOxygens_{false};
```
rprospero:
```suggestion
    bool removeWaterAndCoordinatedOxygens_{false};
```

src/nodes/cifRemoveWater.cpp
```diff
@@ -0,0 +1,31 @@
...
+    // Inputs
+    addInput<CIFLoaderNode::CIFContext *>("CIFContext", "CIF handling context derived from parsing of CIF file", context_)
+        ->setFlags({ParameterBase::Required});
+
+    // Outputs
+    addOutput<CIFLoaderNode::CIFContext *>("CIFContext", "CIF handling context derived from parsing of CIF file", context_);
+
+    // Options
+    addOption<bool>("RemoveWaterAndCoordinatedOxygens", "Whether to remove water and coordinated oxygen atoms in clean-up",
+                    RemoveWaterAndCoordinatedOxygens_);
```
rprospero:
```suggestion
                    removeWaterAndCoordinatedOxygens_);
```

I've added a couple of suggestions to make this variable lower case.

src/nodes/cifRemoveWater.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+std::string_view CIFRemoveWaterNode::type() const { return "CIFRemoveWater"; }
+
+std::string_view CIFRemoveWaterNode::summary() const { return "Remove water from a CIF context"; }
+
+// Run main processing
+NodeConstants::ProcessResult CIFRemoveWaterNode::process()
+{
+
+    context_->setRemoveWaterAndCoordinateOxygens(RemoveWaterAndCoordinatedOxygens_);
```
rprospero:
```suggestion
    context_->setRemoveWaterAndCoordinateOxygens(removeWaterAndCoordinatedOxygens_);
```

rprospero (approved):
Looks clearer than I'd expected from anything with CIF attached to it.  I have some general comments and a few minor suggestions, but the only major concern is establishing a FilePath type in the long term.

src/nodes/cifMolecularSpecies.cpp
RobBuchananCompPhys: Thanks!

src/nodes/cifLoader.cpp
RobBuchananCompPhys: Added to the Dissolve2 TODO

src/data/spaceGroups.cpp
RobBuchananCompPhys: I scripted it - don't worry!

src/nodes/cifLoader.cpp
```diff
@@ -0,0 +1,41 @@
...
+    addOption<SpaceGroups::SpaceGroupId>("SpaceGroupID", "Set space group from index", spaceGroup_);
+}
+
+std::string_view CIFLoaderNode::type() const { return "CIFLoader"; }
+
+std::string_view CIFLoaderNode::summary() const
+{
+    return "Load a CIF file and apply contained crystallographic data to a target configuration";
```
trisyoungs, line 24:
```suggestion
    return "Load and parse a Crystallographic Information File (CIF)";
```

src/nodes/cifPeriodicFramework.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+CIFPeriodicFrameworkNode::CIFPeriodicFrameworkNode(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput<CIFLoaderNode::CIFContext *>("CIFContext", "CIF handling context derived from parsing of CIF file", context_)
+        ->setFlags({ParameterBase::Required});
+
+    // Outputs
+    addOutput<const Species *>("FrameworkSpecies", "Species", frameworkSpecies_);
+    addOutput<Configuration *>("SuperCellConfiguration", "Supercell configuration pointer", supercellConfiguration_);
```
trisyoungs, line 14: Convention dictates that "supercell" is one word:
```suggestion
    addOutput<Configuration *>("SupercellConfiguration", "Supercell configuration pointer", supercellConfiguration_);
```

src/nodes/cifPeriodicFramework.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    // Options
+    addOption<Vector3i>("SuperCellRepeat", "Supercell repeat", supercellRepeat_);
```
trisyoungs, line 17:
```suggestion
    addOption<Vector3i>("SupercellRepeat", "Supercell repeat", supercellRepeat_);
```

trisyoungs (approved):
I think this is a great start for the new CIF handling in Dissolve2. From looking through the nodes as they stand I think we should move away from operating on the context and passing it as a pointer, and move towards passing a proper `CIFContext` object around, each node making a copy of the input context, modifying it, and sending its own CIFContext as an output. If we do this then we can also have a pseudo-Configuration generated within each CIFContext at each stage of the graph.

PR #2341, opened 2026-03-31
refactor: Easier test graph construction

trisyoungs:
This PR adds in a set of functions allowing piecewise construction of graphs for unit tests in a consistent way. All existing custom graph functions are removed, including MgO, and replaced with the new system.

There is an obvious follow-on which is to bring all of the functions to create nodes in the test graphs into the `GraphTestData` class which I will do as a follow-up PR.


tests/graphData.h
```diff
@@ -27,393 +29,148 @@ class GraphTestData
...
-            CoordinateImportFileFormat::coordinateImportFileFormat().enumerationByIndex(initialCoordinates.formatIndex())));
-        ASSERT_TRUE(root->addEdge({"InsertO", "Configuration", "Import", "Configuration"}));
+        // Find the named species node
+        auto speciesNode = dynamic_cast<SpeciesNode *>(root->findNode(speciesName));
+        if (!speciesNode)
+        {
+            std::cout << std::format("No species named '{}' exists in the graph - can't construct IsotopologueSet\n",
+                                     speciesName);
+            return nullptr;
+        }
```
rprospero, line 122:
```suggestion
        ASSERT_TRUE(speciesNode) << std::format("No species named '{}' exists in the graph - can't construct IsotopologueSet\n",                                  speciesName);
```

Feels like the test should just fail if it's called for a non-existent species.

tests/graphData.h
```diff
@@ -27,393 +29,148 @@ class GraphTestData
...
+        auto &species = speciesNode->species();
+        auto isotopologue = species.findIsotopologue(isotopologueName);
+        if (!isotopologue)
+        {
+            std::cout << std::format("No isotopologue named '{}' exists in species '{}' - can't construct IsotopologueSet\n",
+                                     isotopologueName, speciesName);
+            return nullptr;
+        }
```
rprospero, line 130:
```suggestion
        ASSERT_TRUE(isotopologue) << std::format("No isotopologue named '{}' exists in species '{}' - can't construct IsotopologueSet\n",
                                     isotopologueName, speciesName);
```

rprospero (approved):
This is much cleaner and more composable than what we had before.  I added a couple of suggestions on error handling, but it looks good.

tests/graphData.h
trisyoungs: I fully agree, however this is within a function which returns a value, and `ASSERT_TRUE` can't be used in a non-`void` function context, annoyingly. So, the best thing we can do is `ASSERT_TRUE` on the returned value.

tests/nodes/bragg.cpp
```diff
@@ -27,34 +27,48 @@ class BraggNodeTest : public ::testing::Test
...
-        braggNode_ = static_cast<BraggNode *>(testData_.graphRoot.findNode("Bragg01"));
 
-        // Set cell dimensions
-        auto setCellNode = testData_.graphRoot.findNode("Box");
-        ASSERT_TRUE(setCellNode->setOption<Vector3>("Lengths", {21.085, 21.085, 21.085}));
-
-        // Set options on GR node
-        auto grNode = testData_.graphRoot.findNode("GRs");
+        // Add correlation function nodes
+        auto &&[grNode, sqNode] = appendGRSQ(&testData_.graphRoot, lastNode, false, true);
```
RobBuchananCompPhys, line 47: Hoping calls like the above won't feel too awkward when writing tests. What happens when I only want a GR/SQ node and not the other? I can disregard one or the other like `auto &&[grNode, _]`, but it appears from the reference  `&testData_.graphRoot` passed in that the root graph data will be mutated anyway. Additionally, passing in `lastNode` doesn't tell me easily what that last node was, and whether it will still be the last one after this operation.

tests/nodes/bragg.cpp
```diff
@@ -27,34 +27,48 @@ class BraggNodeTest : public ::testing::Test
...
     void createGraph()
     {
-        // Create MgO graph
-        createMgOGraph(
-            &testData_.graphRoot, 500, 500,
+        // Set up the test graph
+        auto lastNode = createConfiguration(
+            &testData_.graphRoot, "Crystal",
+            {{[] { return createAtomic(Elements::Mg); }, 500}, {[] { return createAtomic(Elements::O); }, 500}}, 0.1,
+            Units::DensityUnits::AtomsPerAngstromUnits, InsertNode::BoxActionStyle::None);
```
RobBuchananCompPhys, line 41: This feels a bit busy, in particular the multiple lambdas being passed in, which all do the same thing for different species. Unless the lambda is necessary as a fallback in order to perform any additional setup logic, I wonder if this would be better served by creating a dynamic `std::map<Element::Elements, int>`, with the `createAtomic` called on each contained pair. Of course, this would require some sort of persistent object to which the map is being added.

RobBuchananCompPhys (commented):
Mainly regarding the changes to `graphData.h` - while I do very much like the idea of functions dealing with standard setup for nodes (since we have a good idea what kind of options and edges should be created with them), I wonder if this could be tied together with a relatively simple test class like `GraphConstructor` or similar, which would allow us to build up the graph within a single object where we could better keep track of its progress, as well as making it more editable. I've encountered situations writing test graphs where it's been necessary to dive in and replace an existing edge with multiple edges corresponding to a small group of newly added nodes for instance. As far as I am aware there isn't currently a quick way to do this. Operations that alter the graph during the test execution probably change what it means to be a `lastNode`, and I think an object might be better placed to keep track of that. Furthermore, we could possibly map node types to the methods here in `graphTest.h`, so that the `lastNode` could be handled internally, maybe by a method `head()` which will be automatically updated to point to the _current_ last, offsetting any possible confusion about which node is being referred to.

PR #2343, opened 2026-03-31
refactor: TestGraph object

trisyoungs:
Following on from #2341 this PR creates a more useful `TestGraph` object (previously `GraphTestData`) moving the functions inside the class and publicly deriving from `DissolveGraph` to expose functionality where needed.

The Mermaid generation is also moved into `TestGraph` with a mermaid file saved automatically just prior to test destruction.


tests/graphData.h
```diff
@@ -21,156 +21,181 @@
 namespace UnitTest
 {
 // Basic object setup for any Graph-based test
-class GraphTestData
+class TestGraph : public DissolveGraph
 {
     public:
-    GraphTestData() : dissolve(coreData), graphRoot(dissolve) { Node::echo_ = true; }
+    TestGraph() : DissolveGraph(dissolve), dissolve(coreData) { Node::echo_ = true; }
+    ~TestGraph() { exportMermaidGraph(); }
```
RobBuchananCompPhys, line 28: Nice!

tests/graphData.h
```diff
@@ -21,156 +21,181 @@
...
+        std::ofstream myfile;
+        myfile.open(filename);
+        myfile << graph;
+        myfile.close();
     }
-
-    return xRaySQNode;
-}
+    // Save the current graph in Mermaid format to a file named after the unit test
+    void exportMermaidGraph(std::optional<std::string> context = {}) { exportMermaidGraph(*this, context); }
```
RobBuchananCompPhys:
```suggestion
    // Save the current graph in Mermaid format to a file named after the unit test
    void exportMermaidGraph() { exportMermaidGraph(*this, context_.value_or("")); }
```
Do we really need the optional context arg or could this be set at construction time? The context string could then best tested for content via `.empty()`

RobBuchananCompPhys (approved):
Overall, it looks very neat as a next step from the graph construction 👍 
I am planning to experiment on a separate branch regarding the node tracking we discussed earlier.

tests/graphData.h
trisyoungs: To be honest I didn't really pay much attention to the context arg - it isn't used at all it seems, so I think I may just remove it for now.

PR #2344, opened 2026-04-01
feat: Refactor BraggNode test to get MgO from CIF file

RobBuchananCompPhys:


src/io/import/cif.h
```diff
@@ -118,6 +118,8 @@ class CIFHandler
     bool useCIFBondingDefinitions_{false};
     // Bonding tolerance, if calculating bonding rather than using CIF definitions
     double bondingTolerance_{1.1};
+    // Whether to ignore all bonds
+    bool preventAllBonds_{false};
```
trisyoungs, line 122: The new `preventAllBonds_` option can be combined with `useCIFBondingDefinitions_` into a useful enumeration when we start doing the secondary splitting / refactoring of this stuff, so the options for bonding will be something like "None", "Calculate", "CIFDefinitions"

trisyoungs (changes requested):
Looks good - tested and working! The actual unit cell is basically the same as from the previous old input file, so no need to rerun the EPSR tests as it turns out.  There are some test data files to be removed (the whole `dissolve2/bragg` directory) and the PR itself can now be rebased onto `develop2` and the tests themselves refactored a little to use the new graph creation functions.

PR #2346, opened 2026-04-01
refactor: Benzene energy and force tests

trisyoungs:
This reimplements the benzene energy/force tests as far as is currently possible - testing the individual components of the intermolecular energy requires setting of `AtomType` data from the `Species`, and is left for a later date.


RobBuchananCompPhys (approved):
Looks good!

PR #2347, opened 2026-04-08
refactor: Cells unit tests

trisyoungs:
Reimplements cells-based unit tests as a side-quest (relevant to ongoing Species/CoreData-centric work).


RobBuchananCompPhys (approved):
This looks generally fine 👍 
One question I have is: we've stripped a fair bit of code from cells1-4.cpp - is there any reason we can't bounce them into a single file (`cells.cpp`), containing all the `TEST` and classes + `TEST_F` calls? This might be a nice way to organise these tests.

rprospero (approved):

trisyoungs:
> This looks generally fine 👍 One question I have is: we've stripped a fair bit of code from cells1-4.cpp - is there any reason we can't bounce them into a single file (`cells.cpp`), containing all the `TEST` and classes + `TEST_F` calls? This might be a nice way to organise these tests.

Taking a look at the four files side-by-side this would be sensible, but there are enough subtle differences between the four files to make it not a quick operation, so I will elect to bump this to a later date.

PR #2349, opened 2026-04-08
refactor: Species AtomTypes

trisyoungs:
This PR is the first significant step towards removing the old `Dissolve` and `CoreData` classes, removing the storage of `AtomType`s in `CoreData`. It breaks a couple of things related to the `AddForcefieldDialog` but the justification for doing so is that this is likely to be completely rewritten (most likely in a more `Node`-y form) for Dissolve2.

It is about as small as I could make it - a few follow-up PRs will address the issues listed below.

## Blockers
(In order to fully remove `AtomType`s from `CoreData`)
- [x] ~`PotentialSet` stores `AtomType` pointers and needs to serialise them, so must be converted to use a `DoubleKeyedMap` of resolvables~ Moved to issue #2407
- [x] ~`AtomTypeVectorKeyword` stores `AtomType` pointers and needs to serialise them, however is only used in a single place in the old `NeutronSQModule`.~ Moved to issue #2406
- [x] `EPSRModule::readPCof` converts names into `AtomType`s using `CoreData`.
- [x] `ForcefieldSortFilterModel::filterAcceptsRow` uses a temporary `Dissolve` and `CoreData` to assign atom types and check for forcefield validity.
- [x] `DissolveModel` exposes `AtomType`s through its `AtomTypeModel`, but this is no longer relevant as they are local to `Species` (same for master terms).
- [x] `AddForcefieldDialogModel` uses temporary `Dissolve` and `CoreData` extensively, including but not limited to `AtomType`s. Since `AtomType` and master term data are to be local to a `Species`, much of the existing checking is irrelevant.
- [x] `EPSRModuleTest` accesses the test object's `CoreData` to get the number of atom types which (currently) is both unnecessary and incorrect!
- [x] `AtomTypeModel` (used in lots of places) uses `CoreData::atomTypes_` to check for unique naming when renaming an atom type.


tests/pairPotentials/scaleFactors.cpp
```diff
@@ -56,27 +56,24 @@ class PairPotentialsScaleFactorsTest : public ::testing::Test
...
-        auto *pp22 = std::get<2>(pairPotentials_.emplace_back(atC2_, atC2_,
-                                                              std::make_unique<PairPotential>(atC2_->name(), atC2_->name())))
-                         .get();
-        pp22->setInteractionPotential(interactionPotential_);
-        pp22->setLocalChargeProduct(atC2_->charge() * atC2_->charge());
-        pp22->tabulate();
+        // auto *pp11 = std::get<2>(pairPotentials_.emplace_back(atC1_, atC1_,
+        //                                                       std::make_unique<PairPotential>(atC1_->name(), atC1_->name())))
+        //                  .get();
+        // pp11->setInteractionPotential(interactionPotential_);
```
RobBuchananCompPhys:
```suggestion

```

tests/pairPotentials/scaleFactors.cpp
```diff
@@ -191,9 +187,8 @@ TEST_F(PairPotentialsScaleFactorsTest, MoleculeEnergyWithSpeciesCharges)
     auto &i = molecule_.localAtoms()[0];
     auto &j = molecule_.localAtoms()[3];
 
-    testScalings(i, j, (j.r() - i.r()).magnitude(), fourSixthsBenzene_.atom(0).charge() * fourSixthsBenzene_.atom(3).charge());
-    testAnalyticScalings(i, j, (j.r() - i.r()).magnitude(),
-                         fourSixthsBenzene_.atom(0).charge() * fourSixthsBenzene_.atom(3).charge());
+    testScalings(i, j, (j.r() - i.r()).magnitude(), species_.atom(0).charge() * species_.atom(3).charge());
+    testAnalyticScalings(i, j, (j.r() - i.r()).magnitude(), species_.atom(0).charge() * species_.atom(3).charge());
```
RobBuchananCompPhys, line 187: Interesting to think about this from a vector algebra point of view here - I believe that the line

 `(j.r() - i.r()).magnitude()`

is equivalent to the magnitude of the separation vector defined by **j** and **i**, which you might call "distance".

In case we end up calling this operation frequently, we could compact the code a little with a math helper that looks something like
```
Vector3 separation(Vector3 v, Vector3 u = {0, 0, 0})
{
     return v.r() - u.r();
}

double distance(Vector3 v, Vector3 u = {0, 0, 0})
{
     return separation(v, u).magnitude();
}
```
so this line becomes
`testAnalyticScalings(i, j, distance(j, i), species_.atom(0).charge() ...`

RobBuchananCompPhys (approved):
Looks good 👍 A couple of suggested changes

tests/pairPotentials/scaleFactors.cpp
trisyoungs: This is a valid point, although I would probably opt to have a function on `Vector3` such as `Vector3::distanceTo(const Vector3 &other) const;`

My biggest reservation, however, is that while this kind of distance calculation is used in loads of places, some of them I would be hesitant to change without a lot of benchmarking. I'm specifically thinking of the distance calculations in the energy and force kernels.  I remember spending a lot of time experimenting with using temporary variables, not using temporary variables, doing the distance check in parentheses, doing it stepwise in code, etc. all with a view to getting the most performance out of some of the hottest functions in the code. I do recall that, while inline functions for this sort of thing made sense, they definitely had a performance impact. Perhaps compilers have moved on since then, but regardless this is a suggestion to consider as a future PR IMO.

PR #2350, opened 2026-04-08
chore: Disable Moscito energy/force tests

trisyoungs:
Without the ability to read in TOML-based species definitions (which depends on #2349) the Moscito energy/force unit tests are an absolute nightmare to reimplement, hence we're disabling them for now. Once #2349 is completed we can address their reinclusion.

Note added to #2152.


RobBuchananCompPhys (approved):
Sounds like a necessary intervention!

PR #2403, opened 2026-04-10
refactor: Species common terms

### LINKED ISSUE #2394: Move master terms from CoreData
In line with having `AtomType`s be local to specific `Species` we should do the same for master terms in `CoreData`.

trisyoungs:
This PR moves master terms from `CoreData` to `Species`, refactors lots of stuff, and renames them to `CommonBond`, `CommonAngle` etc. It was absolutely horrible to do, and not easy to break up into smaller bits.  However, a significant chunk of the changed lines are editing input files to retain compatibility with the old input files we still depend on.

Once again the `AddForcefieldDialog` took a beating and is now even more broken.

Closes #2394.

## TODO
- [x] ~Do we really need `shared_ptr`s to common terms? A `std::map` to standard `SpeciesIntra` terms would be simpler to manage.~
- [x] Can remove `CoreData` context from `deserialise()` calls to intramolecular classes, and fwd decl from header files.
- [x] Rename `MasterTermModel` source files
- [x] Need to set `parent_` pointer in `SpeciesIntra` (!IMPORTANT! :smile: :top:)


src/gui/models/addForcefieldDialogModel.cpp
```diff
@@ -124,19 +124,21 @@ void AddForcefieldDialogModel::setDissolve(Dissolve &dissolve)
...
-                                       { return dissolve_->coreData().getMasterImproper(name).has_value(); });
+    // commons_->setBondQueryFunction([this](std::string_view name)
+    //                                { return dissolve_->coreData().getCommonBond(name).has_value(); });
+    // commons_->setAngleQueryFunction([this](std::string_view name)
+    //                                 { return dissolve_->coreData().getCommonAngle(name).has_value(); });
+    // commons_->setTorsionQueryFunction([this](std::string_view name)
+    //                                   { return dissolve_->coreData().getCommonTorsion(name).has_value(); });
+    // commons_->setImproperQueryFunction([this](std::string_view name)
+    //                                    { return dissolve_->coreData().getCommonImproper(name).has_value(); });
+    // TODO DISSOLVE2
```
rprospero:
```suggestion
    commons_->setBondQueryFunction([this](std::string_view name, Species *sp)
                                   { return sp->getCommonBond(name).has_value(); });
    commons_->setAngleQueryFunction([this](std::string_view name, Species *sp)
                                    { return sp->getCommonAngle(name).has_value(); });
    commons_->setTorsionQueryFunction([this](std::string_view name, Species *sp)
                                      { return sp->getCommonTorsion(name).has_value(); });
    commons_->setImproperQueryFunction([this](std::string_view name, Species *sp)
                                       { return sp->getCommonImproper(name).has_value(); });
```

This should fix up the query functions.

src/gui/models/commonTermModel.h
```diff
@@ -26,29 +26,27 @@ enum Roles
...
-    protected:
-    // CoreData object containing the master terms
-    CoreData &coreData_;
+    CommonTermModel(Species *species);
 
     protected:
+    // Source Species
+    Species *species_{nullptr};
     // Icon return function
     std::function<bool(std::string_view termName)> queryFunction_;
```
rprospero:
```suggestion
    std::function<bool(std::string_view termName, Species* sp)> queryFunction_;
```

We now pull the common terms directly from the species, instead of from CoreData.

src/gui/models/commonTermModel.cpp
```diff
@@ -42,45 +42,45 @@ QVariant MasterTermModel::headerData(int section, Qt::Orientation orientation, i
...
     if (!index.isValid())
         return {};
 
     if (index.row() < 0 || index.row() >= rowCount())
         return {};
 
-    if (role == MasterTermModelData::Roles::Query && queryFunction_)
-        return queryFunction_(getTermData(index.row(), MasterTermModelData::DataType::Name).toString().toStdString());
+    if (role == CommonTermModelData::Roles::Query && queryFunction_)
+        return queryFunction_(getTermData(index.row(), CommonTermModelData::DataType::Name).toString().toStdString());
```
rprospero:
```suggestion
        return species_ && queryFunction_(getTermData(index.row(), CommonTermModelData::DataType::Name).toString().toStdString(), species_);
```

Pass the species into the queryFunction, since we now pull the common terms from the species instead of CoreData

rprospero (approved):
I've included some (*untested*) suggestions that should clear up *some* of the changes to the add forcefield dialog.

trisyoungs:
@rprospero - I'll check these changes out before merging. I confess I had been rather blase about breaking the forcefield terms dialog as I was convinced that we probably wouldn't need it going forward. This morning's chat maybe says that is not necessarily the case, so maintaining it's "health" for now seems reasonable.

PR #2404, opened 2026-04-10
test: TestGraph track internals

RobBuchananCompPhys:
The idea here is to allow `TestGraph` to keep track of the last node added to the graph (mostly useful when the nodes are being added in a broadly linear sequence), stored internally in `head_` allowing the developer to get either a `Node *` via a `fetchHead()` call, or a pointer to the desired node type via `head<NodeType>`. The latter is useful for compile time checking that the node is of the expected type.

nextnode

### DIFF
--- a/tests/classes/cells.cpp
+++ b/tests/classes/cells.cpp
@@ -16,12 +16,11 @@ TEST(CellsTest, Basic)
 {
     // Set up the test graph
     TestGraph testGraph;
-    auto lastNode =
-        testGraph.createConfiguration("Box", {{[]() { return createAtomic(Elements::Ar); }, 1}, {createWaterDLPoly, 267}}, 0.1,
-                                      Units::DensityUnits::AtomsPerAngstromUnits);
-    auto importNode = testGraph.appendImportCoordinates(
-        lastNode, CoordinateImportFileFormat("dlpoly/solvated_atom/solvated-argon-rcut5.CONFIG",
-                                             CoordinateImportFileFormat::CoordinateImportFormat::DLPOLY));
+    ASSERT_TRUE(testGraph.createConfiguration("Box",
+                                              {{[]() { return createAtomic(Elements::Ar); }, 1}, {createWaterDLPoly, 267}}, 0.1,
+                                              Units::DensityUnits::AtomsPerAngstromUnits));
+    ASSERT_TRUE(testGraph.appendImportCoordinates(CoordinateImportFileFormat(
+        "dlpoly/solvated_atom/solvated-argon-rcut5.CONFIG", CoordinateImportFileFormat::CoordinateImportFormat::DLPOLY)));
 
     // Set all charge and short-range interaction potentials to zero
     auto arSpeciesNode = dynamic_cast<SpeciesNode *>(testGraph.findNode("Ar"));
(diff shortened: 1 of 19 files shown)


tests/graphData.h
```diff
@@ -29,16 +30,44 @@ class TestGraph : public DissolveGraph
...
+        return true;
+    }
+
+    public:
+    // Returns pointer to current top node in graph
+    Node *fetchHead() const { return head_; }
+    // Returns reference to current top node in graph, cast to the known node type
+    template <class NodeType> NodeType *head() const { return static_cast<NodeType *>(head_); }
+    // Add next (or, current "last") node in graph
+    bool nextNode(const std::string &nodeType, const std::optional<std::string> &name = {})
```
trisyoungs: I'd suggest a rename here to bring it in line with the more specific functions (e.g. `appendGRSQ`). Also, I think we may as well return the node (which can be used for the truth check) and then the calling function has direct access to the new node if it wants it.
```suggestion
    // Append new node to the graph
    Node *appendNode(const std::string &nodeType, const std::optional<std::string> &name = {})
```

tests/graphData.h
```diff
@@ -29,16 +30,44 @@ class TestGraph : public DissolveGraph
...
      * Graph Creation Helpers
      */
     private:
+    // Update pointer to current top node in graph
+    bool updateHead(Node *node)
+    {
+        head_ = node;
+        return true;
+    }
+
```
trisyoungs: Only used once and doesn't do a huge amount, so I'd suggest removing it.

tests/graphData.h
```diff
@@ -29,16 +30,44 @@ class TestGraph : public DissolveGraph
...
+    template <class NodeType> NodeType *head() const { return static_cast<NodeType *>(head_); }
+    // Add next (or, current "last") node in graph
+    bool nextNode(const std::string &nodeType, const std::optional<std::string> &name = {})
+    {
+        auto node = name.has_value() ? createNode(nodeType, *name) : createNode(nodeType);
+
+        if (!node)
+            return false;
+
+        return updateHead(node);
```
trisyoungs: Wrapping in the `updateHead()` function here.
```suggestion
        head_ = node;
        return node;
```

tests/graphData.h
```diff
@@ -49,20 +78,16 @@ class TestGraph : public DissolveGraph
...
+            EXPECT_TRUE(nextNode("Insert", insertNodeName));
+            EXPECT_TRUE(fetchHead()->setInput<Number>("Population", population));
+            EXPECT_TRUE(fetchHead()->setInput<Number>("Density", rho));
+            EXPECT_TRUE(fetchHead()->setOption("BoxAction", boxActionStyle));
+            EXPECT_TRUE(fetchHead()->setOption<Units::DensityUnits>("DensityUnits", rhoUnits));
             EXPECT_TRUE(addEdge({std::string(speciesNode.name()), "Species", insertNodeName, "Species"}));
-            EXPECT_TRUE(addEdge({std::string(lastNode->name()), "Configuration", insertNodeName, "Configuration"}));
-
-            lastNode = insertNode;
+            EXPECT_TRUE(addEdge({std::string(cfgSourceNode->name()), "Configuration", insertNodeName, "Configuration"}));
```
trisyoungs: The logic here has become broken - the `Insert` nodes should all be chained one after another, not all linked from the current head (in `cfgSourceNode`).

tests/classes/cells.cpp
```diff
@@ -48,11 +47,11 @@ TEST(CellsTest, Basic)
                                        {Functions1D::Form::LennardJones126, "epsilon=0.35 sigma=2.166"});
 
     // Run the graph from the Import node to set up the configuration
```
trisyoungs: This comment appears 14 times but is now slightly wrong with the new head node tracking. Suggest a search/replace to:
```suggestion
    // Run the graph from the head node to set up the configuration
```

trisyoungs (approved):
One final comment, and then good to go.  This tidies things up really nicely IMO.

PR #2408, opened 2026-04-15
chore: More CoreData work after species-atomtypes

trisyoungs:
This PR removes a bunch of other `CoreData` uses highlighted after #2349.

### DIFF
--- a/src/gui/models/atomTypeModel.h
+++ b/src/gui/models/atomTypeModel.h
@@ -45,11 +45,6 @@ class AtomTypeModel : public QAbstractListModel
 
     public:
     AtomTypeModel() = default;
-    explicit AtomTypeModel(const CoreData &coreData);
-
-    private:
-    // Optional CoreData reference
-    OptionalReferenceWrapper<const CoreData> coreData_;
 
     private:
     ModelUpdater modelUpdater;
@@ -64,8 +59,7 @@ class AtomTypeModel : public QAbstractListModel
 
     public:
     // Set source AtomType data
-    void setData(const std::vector<std::shared_ptr<AtomType>> &atomTypes,
-                 OptionalReferenceWrapper<const CoreData> coreData = std::nullopt);
+    void setData(const std::vector<std::shared_ptr<AtomType>> &atomTypes);
     // Set function to return QIcon for item
     void setQueryFunction(std::function<bool(const std::shared_ptr<AtomType> &atomType)> func);
     // Set vector containing checked items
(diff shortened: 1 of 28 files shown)


src/gui/models/atomTypeModel.h
```diff
@@ -45,11 +45,6 @@ class AtomTypeModel : public QAbstractListModel
 
     public:
     AtomTypeModel() = default;
-    explicit AtomTypeModel(const CoreData &coreData);
```
rprospero, line 48: QML does not link non-default constructors, so getting rid of this makes me very happy

src/gui/models/ffSortFilterModel.cpp
```diff
@@ -23,19 +23,22 @@ bool ForcefieldSortFilterModel::filterAcceptsRow(int sourceRow, const QModelInde
...
+    // modifiedSpecies_->copyBasic(species_);
+    // // // originalAtomTypeNames_.clear();
+    //
+    // // modifiedSpecies_->clearAtomTypes();
+    // // temporaryDissolve.clearAtomTypes();
+    //
+    // ff->assignAtomTypes(modifiedSpecies_, temporaryCoreData, Forcefield::TypeAll, true);
+    // Messenger::unMute();
+    // return temporaryDissolve.coreData().atomTypes().size() > 0;
+    // TODO DISSOLVE2
```
rprospero, line 39: Feels like this is another model that's going to be a complete rewrite

src/modules/epsr/io.cpp
```diff
@@ -142,13 +145,19 @@ bool EPSRModule::readPCof(Dissolve &dissolve, std::string_view filename)
             return Messenger::error("Failed to read pair potential atom types from pcof file.\n");
 
         // Find the atom types to which these coefficients relate
-        auto at1 = dissolve.coreData().findAtomType(parser.argsv(0));
-        if (!at1)
+        auto at1It = std::ranges::find_if(atomTypes, [&](const auto *atomType)
+                                          { return DissolveSys::sameString(parser.argsv(0), atomType->name()); });
```
rprospero, line 149: This was never intentional, but code like this makes me glad we waited for C++20 before getting rid of CoreData

rprospero (approved):

src/gui/models/atomTypeModel.h
trisyoungs: Happy to be of service!

src/gui/models/ffSortFilterModel.cpp
trisyoungs: Quite a heavy one most likely.

PR #2410, opened 2026-04-16
refactor: Use DoubleKeyedMap in PotentialSet

### LINKED ISSUE #2407: PotentialSet needs to use DoubleKeyedMap
`PotentialSet` stores `AtomType` pointers and needs to serialise them, so must be converted to use a `DoubleKeyedMap` of resolvables. This is necessary in order to allow the removal of `CoreData`.

trisyoungs:
This PR removes the dependency of raw `AtomType` pointers from `PotentialSet` and replaces them with a `DoubleKeyedMap` in line with other classes.  The changes break the `EPSRManagerModule` but that particular refactor depends on our chosen route with `EPSRModule`.

Closes #2407.


rprospero (approved):
This does feel much cleaner

PR #2414, opened 2026-04-21
feat: EPSR Node

RobBuchananCompPhys:


src/classes/configuration_contents.cpp
```diff
@@ -90,6 +90,16 @@ std::vector<const AtomType *> Configuration::atomTypeVector() const
     return result;
 }
 
+// Search for AtomType by name
+std::shared_ptr<const AtomType> Configuration::findAtomType(std::string_view name) const
+{
+    auto it = std::find_if(atomTypeVector().begin(), atomTypeVector().end(),
```
trisyoungs: This won't work as expected because a new vector is returned each time.  You'll need to store the vector in a local var and then do the `find_if` on it.

src/nodes/neutronSQ/neutronSQ.h
```diff
@@ -74,6 +76,8 @@ class NeutronSQNode : public Node
      * Functions
      */
     public:
+    // Return xRay weights
```
trisyoungs:
```suggestion
    // Return neutron weights
```

trisyoungs (commented):
Have had a cursory look through this. Some of the necessary changes feel a bit clunky - absolutely not your fault, just what you've had to engineer to get it working in its new context.  Hopefully we can do some refactoring once the basic integration is successful.

Spotted a probable bug (noted in a comment).

src/templates/keyedVector.h
```diff
@@ -30,6 +30,15 @@ template <class KeyClass, class ValueClass> class KeyedVector
...
     public:
+    // Convert to a map
+    static inline std::map<KeyClass, ValueClass> toMap(const KeyedVector<KeyClass, ValueClass> &instance)
+    {
+        std::map<KeyClass, ValueClass> map;
+        for (const auto &[key, value] : instance)
+            map.insert({key, value});
+
+        return map;
+    }
```
trisyoungs, line 41: Curios why this is a `static` function taking an instance to a `KeyedVector` when it could just be a `const` member function?

PR #2415, opened 2026-04-21
feat: Export trajectory node

rprospero:
This adds a node to export trajectories, mirroring the import trajectory node.

There is an included unit test to confirm that the exported trajectories match the imported trajectories.  It is not an exact match, but will be correct for the first set of coordinates.

As part of this, I've also added a framework for temp files that delete themselves when we're done with them.  I've added a compile time flag to enable keeping the files, but I could change this to checking an environment variable.  That would enable skipping the recompilation, but I also know that running a command with custom environment variables isn't always trivial on Windows.


src/nodes/exportTrajectory.h
```diff
@@ -0,0 +1,45 @@
...
+     */
+    private:
+    // File path
+    std::string filePath_;
+    // File format
+    TrajectoryExportFileFormat::TrajectoryExportFormat format_{TrajectoryExportFileFormat::TrajectoryExportFormat::XYZ};
+    // Last read file position (as int)
+    std::streampos filePosition_;
+    // Target configuration
+    Configuration *configuration_;
```
trisyoungs: Just for cleanliness.
```suggestion
    Configuration *configuration_{nullptr};
```

tests/tempFile.h
```diff
@@ -0,0 +1,47 @@
...
+    constexpr static bool delete_files = true;
+
+    // Generate random names for files
+    static std::string randomName()
+    {
+        std::string result(10, ' ');
+        std::ranges::generate(
+            result, [] { return std::string_view("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz").at(rand() % 52); });
+        return result;
+    }
```
trisyoungs, line 62: Seems like this function is starting to warrant being centralised somewhere. `DissolveSys` would be the obvious place in amongst the string-related functions., although that object is probably also worth a refactor for Dissolve2 - perhaps moving it all into a proper `Dissolve` namespace with a bunch of other stuff? Obviously not for this PR!  But worth a discussion.

tests/ff/water-1000.cpp
```diff
@@ -12,7 +13,8 @@ namespace UnitTest
 TEST(Water1000EnergyTest, Full)
 {
     auto water = createWater();
-    std::ofstream file("water.toml");
+    TempFile water_file("water.toml");
+    std::ofstream file(water_file);
```
trisyoungs: Could `TempFile` inherit `std::ofstream` publicly to make this sort of usage a one-liner?

trisyoungs (commented):
Looks good,

src/nodes/exportTrajectory.h
rprospero: Good call

tests/ff/water-1000.cpp
rprospero: I decided against directly inheriting, but a bit of composition did get rid of this boilerplate.

src/nodes/exportTrajectory.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    // Serialisable data
+    // addSerialisable("filePosition", filePosition_);
+}
+
+std::string_view ExportTrajectoryNode::type() const { return "ExportTrajectory"; }
+
+std::string_view ExportTrajectoryNode::summary() const
+{
+    return "Export configuration coordinates from sequential frames of a trajectory.";
```
trisyoungs:
```suggestion
    return "Export configuration coordinates as sequential frames of a trajectory.";
```

src/nodes/exportTrajectory.h
```diff
@@ -0,0 +1,44 @@
...
+
+    /*
+     * Data
+     */
+    private:
+    // File path
+    std::string filePath_;
+    // File format
+    bool extended_;
+    // Last read file position (as int)
```
trisyoungs:
```suggestion
    // Last read file position
```

tests/ff/water-1000.cpp
trisyoungs, line 1: Was this just a test of `TestFile` and should be deleted, or does it offer something else to this specific unit test?

tests/tempFile.h
```diff
@@ -0,0 +1,61 @@
...
+    operator std::string() const { return path; }
+    operator std::filesystem::path() const { return path; }
+
+    private:
+    // The actual path of the temp file
+    std::filesystem::path path;
+    // A compile time switch to control deleting the temp files.  Set
+    // this to false while debugging to read keep the file so you can
+    // analyse the output.
+    constexpr static bool delete_files = true;
```
trisyoungs: Could / should this just be `static` member and not `constexpr`? Then, for example, it could be set locally within individual test functions without requiring a more expansive recompile, or it could be set in the `testGraph` constructor (as we do for `Node::echo`) to give some broad applicability across tests if required

trisyoungs: Also:

```suggestion
    // A compile time switch to control deleting the temp files.  Set
    // this to false while debugging to read keep the file so you can
    // analyse the output.
    constexpr static bool deleteFiles = true;
```

trisyoungs (commented):
Looking good. Couple of small suggestions.  There's a merge conflict which I suspect is because of my recent merge in of `Atom` -> `ConfigurationAtom`, so apologies for that!

tests/ff/water-1000.cpp
rprospero: The answer is rather yes to both.

The unit test writes out a file which always shows up in my git log as a new file I haven't committed.  Making it a TempFile clears up the lingering file.  On the other hand, I could have just added it to the .gitignore, but it also made for a nice test of TempFile.

trisyoungs: Hmmm.  So did I put that file output in at some point I wonder?  I may have committed it by accident!  Happy for it to go if you deem it so.

tests/tempFile.h
rprospero: @trisyoungs What I've ultimately decided to do is make `deleteFiles_` local to the `TempFile` and set it in the constructor.  When an explicit `false` argument is added (e.g. changing `TempFile out("data.txt")` to `TempFile out("data.txt", false)`), the file will persist.  I thought that this might be a bit more convenient than trying to change a static variable.

PR #2416, opened 2026-04-22
refactor: Rename Atom to ConfigurationAtom

trisyoungs:
For clarity, this PR renames the `Atom` class to `ConfigurationAtom` (in line with `SpeciesAtom` being owned by a `Species`). It touches a lot of places, so this is done in isolation ahead of further work on a common atom class.

The changes are literally a rename and a format, nothing more.


RobBuchananCompPhys (approved):
👍

PR #2417, opened 2026-04-24
feat: Export coordinates node

rprospero:
This adds an ExportCoordinatesNode to go along side the ExportTrajectoryNode.


src/nodes/exportCoordinates.h
```diff
@@ -0,0 +1,60 @@
...
+     */
+    private:
+    // File path
+    std::string filePath_;
+    // File format
+    CoordinateExportFormat format_;
+    // Whether to tag (suffix) the filename with the current iteration index
+    bool tagWithIteration_{false};
+    // Last read file position (as int)
+    std::streampos filePosition_;
```
trisyoungs:
```suggestion
```

src/nodes/registry.cpp
```diff
@@ -81,6 +82,7 @@ void NodeRegistry::instantiateNodeProducers()
                   {"Derivative", makeDerivedNode<DerivativeNode>()},
                   {"DotProduct", makeDerivedNode<DotProductNode>()},
                   {"Energy", makeDerivedNode<EnergyNode>()},
+                  {"ExportCoorinates", makeDerivedNode<ExportCoordinatesNode>()},
```
trisyoungs:
```suggestion
                  {"ExportCoordinates", makeDerivedNode<ExportCoordinatesNode>()},
```

src/nodes/exportXYZConfiguration.h
```diff
@@ -0,0 +1,60 @@
...
+    /*
+     * Data
+     */
+    private:
+    // File path
+    std::string filePath_;
+    // File format
+    CoordinateExportFormat format_;
+    // Whether to tag (suffix) the filename with the current iteration index
+    bool tagWithIteration_{false};
```
trisyoungs, line 33: Now this is an interesting point - we don't really have the notion of a global iteration index any more, so this should probably be a local counter for the number of times the node has been run (which will also need serialising as part of the node's "restart" data.

trisyoungs (changes requested):
Well, `std::format_to` is a nice thing! This all looks fine, but I think we should go a step further and have two nodes, each focussing on a specific format (I also think we should refer to "Configurations" rather than "Coordinates" as we have historically done) - `ExportXYZConfiguration` and `ExportDLPolyConfiguration`.  This means we separate out options for specific file formats (not that we have any here, but the future is looming....) and remove the `CoordinateExportFormat` completely.

src/nodes/exportXYZConfiguration.h
rprospero: That forms a very interesting conundrum, because we would like to serialise this as part of the restart data, but presumably *not* as part of the regular saving of the file.

trisyoungs: Yes, it can be declared with `addSerialisable()`.

rprospero: With `addSerialisable`, the value will be serialised in both the input file and in the restart file.  I'm realising that this distinction is important, since, for instance, we are currently using addSerialisable for the md node, which is serialising all the velocities.  Is this our desired behaviour, or should that only be in the restart file?

trisyoungs: I'm not sure that's true - the `addSerialisable()` just tags a bit of data and adds it to `Node::serialisables_`. The intention for that data is that it will eventually be written out to the project directory, but at present the responsible function (`Node::deserialiseData()`) is not called anywhere.

src/nodes/exportDLPolyConfiguration.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+ExportDLPolyConfiguration::ExportDLPolyConfiguration(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput<Configuration *>("Configuration", "Configuration from which Coordinates will be exported", configuration_);
+
+    // Options
+    addOption<std::string>("FilePath", "File path", filePath_);
+    addOption<bool>("TagWithIteration", "Whether to tag (suffix) the filename with the current iteration index",
+                    tagWithIteration_);
```
trisyoungs, line 18: I wonder if we shouldn't mention, or use, "iteration" at all in future? This is not something for consideration in this PR, but this option could morph into something like an "Overwrite" option, which if true just keeps writing to the same (non-suffixed) file, but if false seeks out the next suffix number to use based on the pre-existing files.

src/nodes/exportDLPolyConfiguration.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+std::string_view ExportDLPolyConfiguration::type() const { return "ExportCoordinates"; }
```
trisyoungs:
```suggestion
std::string_view ExportDLPolyConfiguration::type() const { return "ExportDLPolyConfiguration"; }
```

src/nodes/exportDLPolyConfiguration.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+std::string_view ExportDLPolyConfiguration::summary() const
+{
+    return "Export configuration coordinates from sequential frames of a coordinates in DLPoly format.";
```
trisyoungs:
```suggestion
    return "Export a configuration in DL_POLY format";
```

src/nodes/exportDLPolyConfiguration.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+NodeConstants::ProcessResult ExportDLPolyConfiguration::process()
+{
+
```
trisyoungs:
```suggestion
```

src/nodes/exportXYZConfiguration.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "nodes/exportXYZConfiguration.h"
+#include "main/dissolve.h"
+#include "nodes/constants.h"
+#include <fstream>
+
+ExportXYZConfiguration::ExportXYZConfiguration(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput<Configuration *>("Configuration", "Configuration from which Coordinates will be exported", configuration_);
```
trisyoungs:
```suggestion
    addInput<Configuration *>("Configuration", "Configuration to be exported", configuration_);
```

src/nodes/exportDLPolyConfiguration.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "nodes/exportDLPolyConfiguration.h"
+#include "data/atomicMasses.h"
+#include "main/dissolve.h"
+#include "nodes/constants.h"
+#include <fstream>
+
+ExportDLPolyConfiguration::ExportDLPolyConfiguration(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput<Configuration *>("Configuration", "Configuration from which Coordinates will be exported", configuration_);
```
trisyoungs:
```suggestion
    addInput<Configuration *>("Configuration", "Configuration to be exported", configuration_);
```

src/nodes/exportXYZConfiguration.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+    // Options
+    addOption<std::string>("FilePath", "File path", filePath_);
+    addOption<bool>("TagWithIteration", "Whether to tag (suffix) the filename with the current iteration index",
+                    tagWithIteration_);
+    addSerialisable("Iteration", iteration_);
+}
+
+std::string_view ExportXYZConfiguration::type() const { return "ExportCoordinates"; }
```
trisyoungs:
```suggestion
std::string_view ExportXYZConfiguration::type() const { return "ExportXYZConfiguration"; }
```

src/nodes/exportXYZConfiguration.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+std::string_view ExportXYZConfiguration::summary() const
+{
+    return "Export configuration coordinates from sequential frames of a coordinates.";
+}
+
+NodeConstants::ProcessResult ExportXYZConfiguration::process()
+{
+
```
trisyoungs:
```suggestion
```

src/nodes/exportXYZConfiguration.h
```diff
@@ -0,0 +1,51 @@
...
+{
+    public:
+    ExportXYZConfiguration(Graph *parentGraph);
+    ~ExportXYZConfiguration() override = default;
+
+    enum class CoordinateExportFormat
+    {
+        XYZ,
+        DLPOLY
+    };
```
trisyoungs:
```suggestion
```

src/nodes/exportDLPolyConfiguration.h
```diff
@@ -0,0 +1,45 @@
...
+     */
+    private:
+    // File path
+    std::string filePath_;
+    // Whether to tag (suffix) the filename with the current iteration index
+    bool tagWithIteration_{false};
+    // Target configuration
+    Configuration *configuration_{nullptr};
+    // Iteration count
+    int iteration_ = 1;
```
trisyoungs:
```suggestion
    int iteration_{1};
```

src/nodes/exportXYZConfiguration.h
```diff
@@ -0,0 +1,51 @@
...
+     */
+    private:
+    // File path
+    std::string filePath_;
+    // Whether to tag (suffix) the filename with the current iteration index
+    bool tagWithIteration_{false};
+    // Target configuration
+    Configuration *configuration_{nullptr};
+    // Iteration count
+    int iteration_ = 1;
```
trisyoungs:
```suggestion
    int iteration_{1};
```

src/nodes/registry.cpp
```diff
@@ -81,6 +83,8 @@ void NodeRegistry::instantiateNodeProducers()
                   {"Derivative", makeDerivedNode<DerivativeNode>()},
                   {"DotProduct", makeDerivedNode<DotProductNode>()},
                   {"Energy", makeDerivedNode<EnergyNode>()},
+                  {"ExportDLPolyConfiguration", makeDerivedNode<ExportDLPolyConfiguration>()},
+                  {"ExportXYZConfiguration", makeDerivedNode<ExportXYZConfiguration>()},
                   {"ExportTrajectory", makeDerivedNode<ExportTrajectoryNode>()},
                   {"Forcefield", makeDerivedNode<ForcefieldNode>()},
```
trisyoungs, line 89:
```suggestion
                  {"ExportDLPolyConfiguration", makeDerivedNode<ExportDLPolyConfiguration>()},
                  {"ExportTrajectory", makeDerivedNode<ExportTrajectoryNode>()},
                  {"ExportXYZConfiguration", makeDerivedNode<ExportXYZConfiguration>()},
                  {"Forcefield", makeDerivedNode<ForcefieldNode>()},
```

trisyoungs (approved):
Bunch of  nitpicky suggestions, but all in good spirit! 😃

trisyoungs:
@rprospero Just noticed that the source files also need renaming ("Coordinates" -> "Configuration") too.

Issue #2418, opened 2026-04-24
Remove ConfigurationAtom::globalIndex()


trisyoungs:
We should do this, because its slow and involves a lookup / pointer arithmetic in the parent `Molecule`, and use the  `Atom::index_` member instead. Need to update the indices for the configuration at a suitable point, alongside the atom type indices probably?

PR #2419, opened 2026-04-24
refactor: Common Atom Class

trisyoungs:
This PR reintroduces `Atom` as a base class for `ConfigurationAtom` and `SpeciesAtom`.  There are likely to be comments on function / member variable namings I'm sure!  Every simulator refers to position as `r` and charge as `q` so I think we should retain these, but they don't translate to beautiful setter names (e.g. `setR()`). In the name of consistency, though...

One thing not tackled in this PR for reasons of size was getting rid of the horrible `Configuration::globalIndex()` function - this is a new issue #2418.

### DIFF
--- a/src/classes/atom.h
+++ b/src/classes/atom.h
@@ -9,6 +9,10 @@
 // Basic Atom
 class Atom
 {
+    public:
+    Atom() = default;
+    virtual ~Atom() = default;
+
     /*
      * Properties
      */
--- a/src/classes/atom.h
+++ b/src/classes/atom.h
@@ -31,19 +31,24 @@ class Atom
     public:
     // Set basic properties
     virtual void set(Elements::Element Z, const Vector3 &r, double q = 0.0);
+    // Set coordinates
+    void setR(const Vector3 &r);
+    // Adjust coordinates
+    void adjustR(const Vector3 &delta);
     // Return coordinates
-    Vector3 &r();
     const Vector3 &r() const;
+    // Set atomic element
+    void setZ(Elements::Element z);
     // Return atomic element
-    Elements::Element &Z();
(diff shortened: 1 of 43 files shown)


src/classes/atom.h
```diff
@@ -0,0 +1,56 @@
...
+    // Charge
+    double q_{0.0};
+    // Index in parent
+    int index_{-1};
+    // Atom type index in parent object
+    int atomTypeIndex_{AtomType::Ignore};
+
+    public:
+    // Set basic properties
+    virtual void set(Elements::Element Z, const Vector3 &r, double q = 0.0);
```
rprospero, line 33: It's probably a sign that I wrote too much Haskell in my life, but I genuinely love the fact that all of our main variables for this type are single letters.

src/classes/atom.h
```diff
@@ -0,0 +1,56 @@
...
+    // Set coordinates
+    void setR(const Vector3 &r);
+    // Adjust coordinates
+    void adjustR(const Vector3 &delta);
```
rprospero:
```suggestion
    // Adjust coordinates
    void operator+=(const Vector3 &delta);
```

Just a thought that a little operator overloading here might be clearer than the adjustR name.  Totally optional.

rprospero (approved):
I like this setup.  I've made a suggestion about a possible replacement for `adjustR` but everything else looks nice.

src/classes/atom.h
trisyoungs: Yes, let's do this.  I really don't like the `adjustR()` function name!

trisyoungs: @rprospero Ended up making a `-=` operator too.

PR #2420, opened 2026-04-28
refactor: Add general Structure class

trisyoungs:
With plans in mind for introducing the capability for a `Species` to be created within a graph from source geometry and forcefield terms, this PR introduces a fairly lightweight `Structure` class. It is envisaged that this will be used to store, e.g. data read from XYZ files, data read from CIF files, etc. Since its focus is on utility rather than strictly protected data relationships (as is the case with `Species`) both atoms and bonds are represented as vectors of unique ptrs, rather than trying to emulate the reference-based system in `Species` which requires a bunch of copy / move constructor magic to maintain it.  Thoughts on this welcome!

A little pre-emptive tidying of `Species` is done along the way to remove some unused stuff.


trisyoungs:
I have just realised that a number of functions were declared but not implemented.  I'll be moving these over in a follow-up PR.

src/classes/atom.cpp
```diff
@@ -23,6 +27,16 @@ void Atom::setZ(Elements::Element z) { Z_ = z; }
 // Return atomic element
 Elements::Element Atom::Z() const { return Z_; }
 
+// Return presence of atom
+Atom::Presence Atom::presence() const
+{
+    return Z_ == Elements::Phantom ? Presence::Phantom : Presence::Physical;
+    ;
```
RobBuchananCompPhys:
```suggestion
    
```

src/classes/structure.cpp
```diff
@@ -0,0 +1,231 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#include "classes/structure.h"
+#include "classes/species.h"
+
+Structure::Structure() { box_ = std::make_unique<SingleImageBox>(); }
```
RobBuchananCompPhys: This could all go into the initialiser list, in `structure.h`
```suggestion
Structure::Structure() : box_(std::make_unique<SingleImageBox>()) {}
```

RobBuchananCompPhys (changes requested):

PR #2424, opened 2026-04-30
refactor: Test species as TOML

trisyoungs:
## Follows #2422 

This PR removes the code-based generation of test species in `speciesData.h` and replaces it with loading of TOML species now that `AtomType` and `CommonTerm` data is local to the `Species` class.


tests/nodes/angle.cpp
```diff
@@ -16,7 +16,8 @@ TEST(AngleNodeTest, Water)
 {
     // Set up the test graph
     TestGraph testGraph;
-    testGraph.createConfiguration("Box", {{createWater, 267}}, 0.1);
+    testGraph.createConfiguration("Box", {{[]() { return TestGraph::loadTOMLSpecies("species/water-dlpoly.toml"); }, 267}},
```
RobBuchananCompPhys: I think this goes back to a question I asked in a PR some time ago - do we need to revisit the use of a lambda here (which in this case takes up code space with curly braces and empty lambda/arg capture)? If we still need the option to operate further on the created species, then this might be useful, otherwise, perhaps we could look at setting this argument to a simple `std::pair<std::unique_ptr<SpeciesNode>, int>`, or even just have an extra optional argument `const std::optional<std::function<SpeciesNode *>()>> &speciesInitialiser = {}` which is there as a back-up in case we want to call it further down the line like
```
if (speciesInitialiser)
    *speciesInitialiser(sp.get());
```

src/classes/species_intra.cpp
```diff
@@ -437,19 +433,24 @@ OptionalReferenceWrapper<const SpeciesImproper> Species::getImproper(int i, int
...
-}
 
-// Return whether the attached atoms lists have been created
-bool Species::attachedAtomListsGenerated() const { return attachedAtomListsGenerated_; }
+    // If this is a periodic species, we're done
+    if (box_->type() != Box::BoxType::NonPeriodic)
+    {
+        attachedAtomListsGenerated_ = false;
+        return;
+    }
```
RobBuchananCompPhys, line 451: Could this be a more simple bouncer pattern?
```suggestion
    attachedAtomListsGenerated_ = box_->type() != Box::BoxType::NonPeriodic;
    
    if (attachedAtomListsGenerated_)
    {
       ... // everything from line 455 to 561
    }
```

RobBuchananCompPhys (approved):

RobBuchananCompPhys (approved):

PR #2425, opened 2026-04-30
refactor: Dissolve2 CIF 

RobBuchananCompPhys:
This PR lays the ground work for refactoring the CIF node ecosystem into a more condensed form, where a (most likely) single node - `ImportCIFStructureNode` - returns a `Structure *` created from the species atoms and any connectivity parsed from the import `.cif` file. Virtually all of the Dissolve1 functionality for managing `.cif` import and parsing has been migrated into the `node/cif/io` folder. This includes the cif-relevant ANTLR4 classes.


trisyoungs (commented):
Two significant comments to make:
1) We could remove all of the old "secondary" CIF nodes which just adjust things in the main context since we need to reimplement them in node form anyway. That would break unit tests for the time being, but its probably easier to do that than try to maintain both while we work it up.
2) My thought was that we could completely remove use of `Configuration` and `Species` and generate a `Structure` directly from the contents of the CIF.

src/nodes/cif/importCIFStructure.cpp
```diff
@@ -0,0 +1,548 @@
...
+#include <vector>
+
+ImportCIFStructureNode::ImportCIFStructureNode(Graph *parentGraph) : Node(parentGraph)
+{
+    // Outputs
+    addPointerOutput<Structure>("Structure", "Structure containing atoms and connectivity", structure_);
+
+    // Option
+    addOption<std::string>("FilePath", "File path", filePath_);
+    addOption<bool>("PreventMetallicBonds", "Whether to prevent metallic bonding", preventMetallicBonds_);
```
trisyoungs: Another option that can be gotten rid of.

src/nodes/cif/importCIFStructure.h
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#pragma once
+
+#include "classes/structure.h"
+#include "data/spaceGroups.h"
+#include "math/matrix4.h"
+#include "neta/neta.h"
```
trisyoungs, line 9: Presumably not needed any more...
```suggestion
```

src/nodes/cif/importCIFStructure.cpp
```diff
@@ -0,0 +1,548 @@
...
+    return error("Failed to read contents of CIF file");
+}
+
+/*
+ * CIF I/O
+ */
+
+/*
+ * Raw Data
+ */
```
trisyoungs, line 61:
```suggestion
/*
 * Basic CIF Data
 */
```

src/nodes/cif/importCIFStructure.h
```diff
@@ -0,0 +1,136 @@
...
+    // Whether to prevent metallic bonding
+    bool preventMetallicBonds_{true};
+
+    /*
+     * CIF I/O
+     */
+
+    /*
+     * Raw Data
+     */
```
trisyoungs, line 43:
```suggestion
    /*
     * Basic CIF Data
     */
```

src/nodes/cif/importCIFStructure.h
```diff
@@ -0,0 +1,136 @@
...
+    private:
+    // CIF strucutre
+    Structure structure_;
+    // Space group ID
+    SpaceGroups::SpaceGroupId spaceGroup_{SpaceGroups::SpaceGroupId::NoSpaceGroup};
+    // CIF filepath
+    std::string filePath_;
+    // Whether to prevent metallic bonding
+    bool preventMetallicBonds_{true};
```
trisyoungs:
```suggestion
```

src/nodes/cif/importCIFStructure.h
```diff
@@ -0,0 +1,136 @@
...
+    bool hasBondDistances() const;
+    // Return whether a bond distance is defined for the specified label pair
+    std::optional<double> bondDistance(std::string_view labelI, std::string_view labelJ) const;
+
+    /*
+     * Creation
+     */
+    private:
+    // Temporary atom types used for unique atom labels
+    std::vector<std::shared_ptr<AtomType>> atomLabelTypes_;
```
trisyoungs, line 105: I wrote this - I don't like it any more! It's a strange use of `AtomType`.  I'll flag this on the Dissolve2ToDo as something to refactor.

src/nodes/cif/importCIFStructure.h
```diff
@@ -0,0 +1,136 @@
...
+    bool createStructure(SpaceGroups::SpaceGroupId sgid, double overlapTolerance);
+
+    public:
+    // Set overlap tolerance
+    void setOverlapTolerance(double tol);
+
+    /*
+     * End CIF I/O
+     */
+
```
trisyoungs:
```suggestion
```

src/gui/createGrapheneSpeciesDialog.cpp
trisyoungs, line 1: How come the graphene species dialog got removed here?  I don't recall there being any dependence on CIF in there!  This is definitely destined for a node, too...

RobBuchananCompPhys: Looks like I got a bit trigger-happy with anything that had an apparent CIFContext include... I've looked at the code in more depth, and the dialog doesn't appear to actually make use of the class (now removed), so we could in all likelihood reinstate it (without including `nodes/cif/io/cifContext.h`)!

src/nodes/cif/importCIFStructure.h
RobBuchananCompPhys: We should be able to use the new capability for string name on `StructureAtom` that you added in #2427 to link the newly added structural atom to the relevant atom type bonding pairs information. That way, we shouldn't need the `atomLabelTypes_` vector.

src/nodes/cif/importCIFStructure.cpp
```diff
@@ -0,0 +1,542 @@
...
+        auto [indexI, indexJ] = pair;
+        if (indexI == indexJ)
+            continue;
+
+        auto i = structure_.atomAt(indexI);
+        auto j = structure_.atomAt(indexJ);
+
+        // Prevent metallic bonding?
+        if (Elements::isMetallic(i->Z()) && Elements::isMetallic(j->Z()))
+            continue;
```
trisyoungs: This is an option for a separate node, should we want to reimplement it.
```suggestion
```

trisyoungs (approved):
Looks excellent.  Just one small deletion of a thing which is no longer relevant.

PR #2427, opened 2026-05-05
refactor: Import structure nodes

trisyoungs:
This PR effectively moves all code from `CoordinateImportFileFormat` into individual nodes which output a `Structure` object. All of the new unit tests are updated accordingly.

The old `CoordinateImportFileFormat` is _not_ removed as it is still depended on by `TrajectoryImportFileFormat` (to be converted in a follow-up PR).

PR #2431, opened 2026-05-13
refactor: Import trajectory nodes

trisyoungs:
Following #2427 this PR does the same thing for trajectory import. There's also a bunch of tedious renaming and a final removal of the original import modules (stuff in `io/` to follow later).


src/nodes/node.h
```diff
@@ -50,9 +50,9 @@ class Node : public Serialisable<>
     virtual std::string_view summary() const = 0;
 
     /*
-     * Node message
+     * Node Messagine
```
RobBuchananCompPhys:
```suggestion
     * Node Messaging
```
?

src/nodes/importDLPOLYTrajectory.cpp
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#include "nodes/importDLPOLYTrajectory.h"
+#include "nodes/importDLPOLYStructure.h"
+
+ImportDLPOLYTrajectoryNode::ImportDLPOLYTrajectoryNode(Graph *parentGraph) : Node(parentGraph)
+{
+    volatile_ = true;
```
RobBuchananCompPhys:
```suggestion
ImportDLPOLYTrajectoryNode::ImportDLPOLYTrajectoryNode(Graph *parentGraph) : Node(parentGraph), volatile_(true)
{
```

src/nodes/graph.cpp
```diff
@@ -32,6 +32,11 @@ std::string_view Graph::summary() const { return "A node which contains its own
 // Perform processing
 NodeConstants::ProcessResult Graph::process()
 {
+    // Outdate any volatile nodes
+    for (auto &node : std::views::values(nodes_))
+        if (node->isVolatile())
+            node->setUpdateRequired();
```
RobBuchananCompPhys, line 38: Did this actually solve the issue with multiple frame reading by RDF nodes per iteration? Just looking at this I'd imagine you get one volatile update per graph process which should be OK?

RobBuchananCompPhys (commented):
A couple of suggestions, also would be good to clarify whether we no longer encounter the problem of multiple frame reads by RDF per loop?

src/nodes/graph.cpp
trisyoungs: Ah, yes - I forgot to mention this in the PR overview!  This solved things very nicely, and very generally.

rprospero (approved):

PR #2433, opened 2026-05-14
chore: Remove small molecule test species

trisyoungs:
Removes yet more code-constructed test species and converts them to TOML.  Also adds `Species::load()` to load a species directly, useful in unit tests.

This work removes most of the usages of modifying functions in the `Species` class, making refactor it a lot easier!


src/classes/species.cpp
```diff
@@ -331,3 +331,19 @@ void Species::deserialise(const SerialisedValue &node)
...
+void Species::load(std::string_view tomlFile)
+{
+    clear();
+
+    SerialisedValue contents = toml::parse(std::string(tomlFile));
+    if (contents.contains("species"))
+    {
+        deserialise(contents["species"]);
+        auto name = contents["species"]["name"].as_string();
+        setName(name.str);
```
RobBuchananCompPhys, line 345:
```suggestion
if (contents.contains("species"))
    deserialise(contents["species"])
```
I feel that my preference would be to have the species name set inside `deserialise` - this would contain the responsibility to fully initialise the species from TOML within a single method.

Then, this would become ^

Is there any reason we don't do this already?

RobBuchananCompPhys (changes requested):

src/classes/species.cpp
trisyoungs: To be honest I've been ignoring this a bit, but ultimately the `Species` class doesn't need a `name_` member as this should be the `SpeciesNode` name.  However, the relationship between those two things is not quite clear to me at present. I agree that the name could / should be set in the `deserialise()`, so I'll make that change.

PR #2434, opened 2026-05-15
fix: Clear bond *if* `hasBond(i, j)` is true

RobBuchananCompPhys:


src/classes/structure.cpp
```suggestion
    for (auto &bond : bonds_)
    {
        bond->i()->removeBond(bond.get());
        bond->j()->removeBond(bond.get());
    }
    bonds_.clear();
}
```

trisyoungs (approved):

PR #2436, opened 2026-05-15
refactor: Immutable Species

trisyoungs:
This PR removes all structure-related editability from the `Species` class as we move towards a regime where `Species` are either loaded or created from a `Structure`. The main thing to note is the `template`-ing of the `Atom` and `Bond` classes in order to allow each to reference the relevant classes for where they are used.  Note also that a follow-up PR (#2449) adds in a concrete `BaseAtom` class to improve functionality,


src/classes/fragment.h
```diff
@@ -0,0 +1,41 @@
...
+            // Is this either of the excluded bonds?
+            if (exclude == bond || excludeToo == bond)
+                continue;
+
+            // Get the partner atom in the bond and select it (if it is not selected already)
+            auto *j = bond->partner(&i);
+            if (std::find(indices.begin(), indices.end(), j->index()) == indices.end())
+                getIndicesRecursive(atoms, indices, j->index(), exclude, excludeToo);
+        }
+    }
```
RobBuchananCompPhys, line 30:
```suggestion
    static void getIndicesRecursive(const std::vector<AtomClass> &atoms, std::vector<int> &indices, int index,
                                    const std::pair<BondClass *, BondClass *> excluded)
    {
        // Loop over bonds on indexed atom
        indices.emplace_back(index);
        const auto &i = atoms.at(index);
        auto &[excludedBondA, excludedBondB] = excluded;
        for (const auto *bond : i.bonds())
        {
            // Is this either of the excluded bonds?
            if (excludedBondA == bond || excludedBondB == bond)
                continue;

            // Get the partner atom in the bond and select it (if it is not selected already)
            auto *j = bond->partner(&i);
            if (std::find(indices.begin(), indices.end(), j->index()) == indices.end())
                getIndicesRecursive(atoms, indices, j->index(), excludedBondA, excludedBondB);
        }
    }
```
Persionally I find the separate bond arguments a little obscure - passing them in as a pair might be slightly more instructive as to these bonds being treated equally.

src/classes/fragment.h
```diff
@@ -0,0 +1,41 @@
...
+
+    public:
+    // Return the fragment (vector of indices) containing the specified atom
+    static std::vector<int> get(const std::vector<AtomClass> &atoms, int startIndex, const BondClass *exclude = nullptr,
+                                const BondClass *excludeToo = nullptr)
```
RobBuchananCompPhys: As above

RobBuchananCompPhys (approved):
This looks good, just one suggested change to think about

src/classes/fragment.h
trisyoungs: I see your point.  Passing as a `std::pair` would force the occasional `{bond, nullptr}` in the code, which is a bit ugly, but we could take a `std::vector<BondClass *>` instead and generalise it completely.

PR #2443, opened 2026-05-20
chore: Convert species structure unit test into Structure-based one

trisyoungs:
This reworks the existing `Species`-based unit test into a `Structure`-related one, given that we are working towards species immutability. It also exposes the useful part of `CalcualteBondingNode` as a static function for convenience.

Ground work for #2436


RobBuchananCompPhys (approved):
Looks good 👍

trisyoungs:
Mistakenly committed as part of #2446

PR #2446, opened 2026-05-20
chore: Species clean up

trisyoungs:
This PR removes a _lot_ of old functions from the `Species` class, almost all of which are related to mutating the class in some way but which is now either unnecessary or covered by `Structure` and specific new nodes.

Ground work for #2436


src/nodes/calculateBonding.h
```diff
@@ -38,4 +38,7 @@ class CalculateBondingNode : public Node
     private:
     // Run main processing
     NodeConstants::ProcessResult process() override;
+
+    public:
+    static void calculate(Structure &structure, double tolerance = 1.1, bool clearBefore = true, bool preventMetallic = false);
```
RobBuchananCompPhys, line 43:
```suggestion
    private:
    static void calculate(Structure &structure, double tolerance = 1.1, bool clearBefore = true, bool preventMetallic = false);
```
Just want to double check - do we actually want this to public? Is it intended for use by external objects?

trisyoungs: Yes. Otherwise we would have to re-implement the bonding algorithm in at least one other place. Just re-using code!

RobBuchananCompPhys (approved):

PR #2449, opened 2026-05-21
refactor: Atom base class

trisyoungs:
This PR adds a `BaseAtom` class from which the templated `Atom` class inherits in order to provide generic (non-templated) access to atom connectivity.


src/classes/atom.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Set coordinates
+void BaseAtom::setR(const Vector3 &r) { r_ = r; }
+
+// Return coordinates
+const Vector3 &BaseAtom::r() const { return r_; }
+
+// Set atomic element
+void BaseAtom::setZ(Elements::Element z) { Z_ = z; }
```
RobBuchananCompPhys:
```suggestion
void BaseAtom::setZ(Elements::Element Z) { Z_ = Z; }
```
Any chance we could refactor this to be a big `Z` when dealing with elements? Seems like a nice way to distinguish from the Cartesian coordinate `z` (of course, unlikely to be mistaken in practice but it could help when glancing over the code IMO)

src/classes/atom.h
```diff
@@ -3,21 +3,18 @@
...
 #include "classes/atomConstants.h"
-#include "classes/atomType.h"
+#include "classes/bond.h"
+#include "data/elements.h"
 #include "math/vector3.h"
 
-// Basic Atom
-template <typename BondClass> class Atom : public Serialisable<>
+// Base Atom
+class BaseAtom
```
RobBuchananCompPhys:
```suggestion
class AtomBase
```
I think this class should be `AtomBase`, rather than `BaseAtom` in line with existing Dissolve base classes (and most standard C++ syntax - I think ?) - what do you reckon?

RobBuchananCompPhys (changes requested):
Looks good 👍 Aside from the fact this code largely (and nicely) carries over from existing `SpeciesAtom` and should be fine throughout, I do have one or two suggested variable and type name changes you might want to consider.

src/classes/atom.cpp
trisyoungs: Absolutely - mistake on my part.

src/classes/atom.h
trisyoungs: Do you know, I hadn't even though of this at the time, but you're absolutely right - will rename for consistency.

RobBuchananCompPhys (approved):

PR #2450, opened 2026-05-21
refactor: NETA for all

trisyoungs:
This PR completes the long journey of immutable species, structure, and generic NETA usage. The dependence on `SpeciesAtom` is removed from the NETA classes and replaced by `BaseAtom`.

### DIFF
--- a/src/classes/speciesRing.cpp
+++ b/src/classes/speciesRing.cpp
@@ -6,7 +6,7 @@
 #include "classes/speciesBond.h"
 #include "data/elements.h"
 
-SpeciesRing::SpeciesRing(const std::vector<const BaseAtom *> &atoms) : atoms_(atoms) {};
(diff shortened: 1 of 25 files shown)


src/classes/structure.h
RobBuchananCompPhys, line 33: Interestingly, it appears to me that we have boiled `StructureAtom` down to the point that it's not clear that it's necessary for it to be a type of its own! I did wonder if there is a case to be made for simply replacing instances of `StructureAtom` with instances of `Atom<Bond<Atom>>` but then the inner `Atom` will also need a template! That leads me to the conclusion that this non-templated type is in fact very necessary for solving that recursion, but it does seem like there should be a slightly more "efficient" way of doing things here.

RobBuchananCompPhys (approved):

PR #2454, opened 2026-05-22
build: Specifically request windows-2022

trisyoungs:
Until June when `windows-latest` will use VS2026, this PR keeps us on the older version until we fix the nasty issues it brings with it.

https://github.com/actions/runner-images/issues/14017


RobBuchananCompPhys (approved):
Nice!

PR #2455, opened 2026-05-27
refactor: Forcefield tidy up

trisyoungs:
This PR wages code war on the `Forcefield` class with the intent to streamline it ahead of updating the existing `ForcefieldNode` to reproduce the behaviour of the `AddForcefieldDialog` / wizard, a UI element which is not being directly in the Dissolve2 GUI. To that end, this PR removes it.


rprospero (approved):
This is also a nice step toward eventually getting rid of OptionalReferenceWrapper

PR #2457, opened 2026-05-27
fix: Directly return atomType index from atomTypeIndex

rprospero:
This PR ensures that the `atomTypeIndex` method on `SpeciesAtom` returns the actual index of the atom type.  This will eventually be removed, but is needed at the moment to pass the assertions in Array2D

### DIFF
--- a/src/classes/species.cpp
+++ b/src/classes/species.cpp
@@ -246,6 +246,7 @@ void Species::deserialise(const SerialisedValue &node)
                             at->deserialise(data);
                             at->setIndex(atomTypes_.size() - 1);
                         });
+
     Serialisable::toVector(node, "atoms", [this](const SerialisedValue &atom) { atoms_.emplace_back(this).deserialise(atom); });
 
     Serialisable::toMap(node, "commonBonds", [this](const std::string &name, const SerialisedValue &bond)
--- a/src/classes/species.cpp
+++ b/src/classes/species.cpp
@@ -240,13 +240,13 @@ void Species::deserialise(const SerialisedValue &node)
 {
     setName(toml::find<std::string>(node, "name"));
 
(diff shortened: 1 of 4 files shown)


RobBuchananCompPhys (approved):
Looks good 👍 (Since this sees `SpeciesAtom` returning the base `atomTypeIndex_`, this shouldn't conflict with the design decisions mentioned in standup this morning @trisyoungs ?)

src/classes/species.cpp
```diff
@@ -243,6 +243,10 @@ void Species::deserialise(const SerialisedValue &node)
     Serialisable::toMap(node, "atomTypes", [this](const std::string &name, const auto &data)
                         { atomTypes_.emplace_back(std::make_shared<AtomType>(name))->deserialise(data); });
 
+    int idx = 0;
+    for (auto &a : atomTypes_)
+        a->setIndex(idx++);
+
```
trisyoungs:
```suggestion
Serialisable::toMap(node, "atomTypes",
                    [this](const std::string &name, const auto &data)
                    {
                        auto &at = atomTypes_.emplace_back(std::make_shared<AtomType>(name));
                        at->deserialise(data);
                        at->setIndex(atomTypes_.size() - 1);
                    });
```

trisyoungs: I think I would rather write this as suggested. Also, this in itself is enough to fix the failing assertions, so I would recommend removing the other changes since they will be reverted just as soon as `AtomType::index_` is removed and we switch to using `AtomBase::atomTypeIndex_` properly for `Species` energy/force calculations.

trisyoungs (changes requested):

trisyoungs (approved):

PR #2458, opened 2026-05-27
refactor: Remove dissolve object from graphs

rprospero:
This should wait until after #2457

This breaks the connection between the `DisolveGraph` node and the old `Dissolve` object.  `DissolveGraph` contains no `Dissolve` or `CoreData` objects.

## Direct access

There were two places where the old `Dissolve` object was directly being accessed.  The first was in getting the iteration number and the second was accessing the pair potentials.  There was already a `pairPotentials_` member, so I used that for the second.  As for the first, I've currently added a global iteration count to the `DissolveGraph`, but we had also discussed possibly giving each node its own iteration count.

## Indirect Access

I spent far more time than I should have tracking down the fact that the constructor on `CoreData` globally sets the pair potential charge source.  Thus, removing the `CoreData` from the `DissolveGraph` would cause certain tests to fail, despite not accessing `CoreData`.  I have currently added that command into the constructor on `TestGraph`, but it might be more appropriate for `DissolveGraph`.

### DIFF
--- a/src/main/dissolve.cpp
+++ b/src/main/dissolve.cpp
@@ -10,7 +10,7 @@ Dissolve::Dissolve(CoreData &coreData) : coreData_(coreData)
     // Set core simulation variables
     restartFileFrequency_ = 10;
 
-    graphNode_ = std::make_unique<DissolveGraph>(*this);
+    graphNode_ = std::make_unique<DissolveGraph>();
 
     // Clear everything
     clear();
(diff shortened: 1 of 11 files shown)


trisyoungs (approved):
I dislike the need for the `pairPotential()` retrieval function, but this is not your fault!  As we redesign `EPSRNode` moving forwards, perhaps this can be eliminated, but we shall see.

trisyoungs:
Otherwise, seeing `Dissolve` and `CoreData` start disappearing is very pleasing!

PR #2460, opened 2026-05-28
feat: Optional data parameters

trisyoungs:
The best way to describe this PR is that it adjusts the `Parameter` class to make `std::optional` "first-class datatypes". Such inputs are required for the upcoming changes to the `SpeciesNode` class.  I also extend the unit test for parameters and add in some useful error reporting in `Edge`.


RobBuchananCompPhys (approved):
Looks good 👍

PR #2462, opened 2026-05-28
chore: Reinstate forcefield assignment tests

### LINKED ISSUE #2393: Test species as TOML files
Dissolve2 unit tests have largely been written around test species data constructed via code in `tests/speciesData.h`, however this quickly becomes a real pain for even small molecules.

When possible we should generate a set of species-containing TOML input files that can be loaded in and used when necessary.

This enables the Moscito unit tests to be re-enabled (see #2387)

### LINKED ISSUE #2435: Refactor forcefield assignment tests to use Species-from-Structure creation

trisyoungs:
This PR makes use of the functionality introduced in #2461 to reinstate the forcefield assignment tests that were disabled due to us not being able to read in XYZ files and apply a forcefield because...... I forget exactly the reasons why.

One test is now disabled - assignment with UFF4MOF - as it requires us to set the cell on the `Structure` which we currently can't do (#2469 will enable this).

Closes #2435
Closes #2393 (the Moscito tests part)


tests/ff/cos-n.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+    // Get the configuration and create an energy kernel
+    auto cfg = testGraph.fetchHead()->getOutputValue<Configuration *>("Configuration");
+    auto kernel = testGraph.createEnergyKernel(cfg);
+
+    // Check consistency between production and test energies
+    auto productionEnergy = checkEnergyConsistency(kernel);
+
+    // Intramolecular energy: 183.4801   # (2.866876 per molecule) * 64 molecules
+    EXPECT_NEAR(183.4801, productionEnergy.geometry.torsionEnergy, 1.0e-2);
```
RobBuchananCompPhys, line 35: Just checking - do we want public getters for `.geometry` and `.torsionEnergy` as this looks like we are accessing the raw members (which we may want to do).

trisyoungs: In this instance I elected to keep these "energy return" objects just simple `struct`s with public members for ease of use, otherwise the energy calculation routines would need lots of `addTorsionEnergy(double x)`-type functions which just makes things ugly.

RobBuchananCompPhys (approved):

PR #2464, opened 2026-05-28
feat:Now commandline for Dissolve version 2

rprospero:
This PR converts the `dissolve` command line command to the new node graph architecture.  The primary interface requires giving both the file name for the input file and also the name for the node that will be iterated.  We'll likely want to rethink this going forward.

Much of the old restart file code has been commented out until such time as we have a proper restart file setup in place for Dissolve 2.

Also, by popular demand, it is now possible to show the graph in mermaid mode directly on the command line

Screenshot from 2026-05-28 17-20-26


src/base/serialiser.h
```diff
@@ -11,6 +11,8 @@
 #include <map>
 #include <vector>
 
+using ParserErrorMessage = std::string;
```
trisyoungs: As far as I can see this typedef is only used once - do we need it? Or, conversely, what am I missing?

trisyoungs (approved):
Looks OK to me, although quite what "iterating" means in Dissolve2 is still to be decided and the loop implemented here may soon change!

src/base/serialiser.h
rprospero: Truth be told, we may not need it.  The truth is that it will only every be used in two places - loading a file in the CLI and loading a file in the GUI.  My concern was that the GUI might want a more complex error type than just a string.  The idea was to use a typedef for now, but possibly create a real ParserErrorMessage class in the future.

This is a bunch of speculative abstraction, so I'm more than happy to just switch back to strings for now and make the complicated type when we actually need it.

PR #2466, opened 2026-05-29
chore: Remove lazy species intra accessors

trisyoungs:
A little (!) tidy-up PR to remove functions like `indexI()` and `indexJ()` from `SpeciesIntra`-derived classes - those functions were just "beautification wrappers" hiding calls to (e.g.) `i_->index()` - since they are used in hot loops I reasoned that we were probably better off being honest!


rprospero (approved):
As much as I enjoyed the `angle.index(column)` pattern, the code is so much cleaner without it.

PR #2467, opened 2026-05-29
chore: Remove more modules

trisyoungs:
With #2464 disabling the increasingly hollowed-out example unit tests, there is an opportunity to remove a few more modules (one of which - `EPSRModule` blocks the removal of `AtomType::index_`...)


src/base/serialiser.h
RobBuchananCompPhys: This looks potentially like overkill?

src/main/cli.cpp
RobBuchananCompPhys, line 22:
```suggestion
    auto entryNodeOption = app.add_option("entryNode", entryNode_, "Run the graph from the specified node.");
```
I would suggest a slightly higher level of descriptiveness here, given that we want to be very clear with the user about what can be accomplished, and what this particular node object corresponds to (i.e. the graph "entrypoint")

src/main.cpp
RobBuchananCompPhys, line 132: This is just a hunch (naturally, since we already spent a fair bit of time on iteration _within_ graphs), but I wonder if there could be a neat way to incorporate `IterableGraph` here as the main Dissolve loop driver. It may not be necessary at this point, but if the iteration requirements begin to converge with what we have already done we may want to look into basing `DissolveGraph` off `IterableGraph` and use either "private" option values or public setters to pass in the information about entry node, restart file, number of iterations etc to the underlying iterator graph and run it there.

RobBuchananCompPhys (approved):

PR #2469, opened 2026-05-31
feat: Variant parameters

trisyoungs:
This PR attempts to allow `std::variant` as a workable `Parameter` data type, driven from the need to use `SetCellNode` to work on both `Configuration`s as well as `Structure`s. To neaten things up at the end I introduced `VariantParameterData` (see `nodes/parameter.h`) to tidy things up and force the use of `std::monostate` as the first alternative in the variant.

### TODO
- [x] Assigning a `std::variant`-based `Parameter` currently fails - needs a refactor to be able to switch to using a setter on the variant (source) `Parameter` rather than a getter from the destination `Parameter`.
- [x] Empty type (because our `std::variant` can be empty?
- [x] Unit tests

### DIFF
--- a/src/nodes/parameter.h
+++ b/src/nodes/parameter.h
@@ -23,7 +23,7 @@ template <typename T> class Parameter;
 template <typename T> class SerialisableParameter;
 
 // Return whether the type_index provided matches one of the supplied variant's allowed types
-template <class... Ts> bool is_one_of(std::type_index id, const std::variant<Ts...> &x)
+template <class... Ts> bool is_variant_alternative(std::type_index id, const std::variant<Ts...> &x)
 {
     return ((id == typeid(Ts)) || ...);
 };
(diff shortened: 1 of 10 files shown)


src/nodes/parameter.h
```diff
@@ -208,6 +216,16 @@ std::shared_ptr<ParameterBase> createSerialisable(Node *parent, std::string_view
 }
 }; // namespace ParameterFactory
 
+// Put the data of the supplied parameter into the supplied variant
+template <class... Ts> bool put_into(ParameterBase *other, std::variant<Ts...> &x)
```
RobBuchananCompPhys:
```suggestion
template <class... Ts> bool emplace(ParameterBase *other, std::variant<Ts...> &x)
```

I think this might be nicer!

src/nodes/parameter.h
```diff
@@ -112,15 +109,20 @@ class ParameterBase : public Serialisable<>
     void clearDataInParent() const;
     // Mark edges for re-pull in parent node
     void markIncomingEdgesForPull() const;
+    // Perform any updates after a successful setData()
+    void updateAfterSet() const;
```
rprospero, line 113: This was a great addition

src/nodes/parameter.h
```diff
@@ -208,49 +193,75 @@ std::shared_ptr<ParameterBase> createSerialisable(Node *parent, std::string_view
...
+        return std::visit(
+            [](auto &&arg) -> bool
+            {
+                using T = std::decay_t<decltype(arg)>;
+                if constexpr (std::is_pointer_v<T>)
+                {
+                    return arg == nullptr;
+                }
+                return false;
+            },
```
rprospero, line 229: I briefly considered suggesting using dispatch to put each type into its own, correct lambda, but I honestly think that it's cleaner the way that you wrote it.

rprospero (approved):
Getting rid of context for right now is the right call.  We'll eventually need some framework like that for some of the GUI work, but we don't want it slowing us down right now.

PR #2472, opened 2026-06-05
refactor: AxisAngle node

trisyoungs:
Converts the AxisAngleModule into a Node.


src/nodes/axisAngle.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "math/mathFunc.h"
+
+AxisAngleNode::AxisAngleNode(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput("Configuration", "Target configuration for the calculation", configuration_)->setFlags({ParameterBase::Required});
+
+    // Outputs
+    addOutput<Configuration *>("Configuration", "Output configuration", configuration_);
+
```
rprospero: Odd question: does the configuration need to be an output? We never modify the configuration, so it's just a pass-through. Basically, instead of 

```mermaid
graph LR;
Configuration --> AxisAngle --> OtherAnalysis;
``` 

we could do 


```mermaid
graph LR;
Configuration --> AxisAngle;
Configuration --> OtherAnalysis;
```

I would then imagine that the outputs would be `rABNormalised`, `aABNormalised`, and`dAxisAngleNormalised`, which could then either be passed to an export or sent for further analysis

rprospero (approved):
One concern, but otherwise looks good

src/nodes/axisAngle.cpp
trisyoungs: Absolutely right to question this - the `Configuration` doesn't need to be an output as we can "parallelise" as you show in the bottom graph.

PR #2473, opened 2026-06-05
refactor: HistogramCN node

trisyoungs:
Converts the HistogramCN analysis module to a node.

PR #2474, opened 2026-06-06
refactor: ModifierOSites node

trisyoungs:
## Needs code in #2473 - do not merge until #2473 is complete!

Another module converted to a node, this time ModifierOSites.


src/nodes/modifierOSites.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    // Options
+    addOption("Modifier", "Set the modifier site(s) for which the distribution of oxygens should be calculated",
+              modifierSpeciesSites_);
+    addOption("BondingOxygen", "Set the site(s) 'BO' which are to represent the bonding oxygen", bridgingOxygenSpeciesSites_);
+    addOption("NetworkFormer", "Set the site(s) 'NF' for which the distribution around the modifier sites should be calculated",
+              networkFormerSpeciesSites_);
+    addOption("DistanceRange", "Distance range (min, max) over which to calculate from the central site", distanceRange_);
+    addOption("ModifierDistanceRange", "Distance range (min, max) over which to calculate from the central site",
+              modifierDistanceRange_);
```
rprospero, line 23: Now I feel like I have the opposite issue: I think that we need more outputs on this node.  Specifically, I think that we should have output parameters for all of the Data1D nodes (i.e. totalOxygens_, distanceMFO_, distanceMNBO_, distanceMBO_, and distanceMOtherO_).  That would allow us to export these to other nodes and allow them to be graphed or export to a text file.

rprospero (approved):
Again looks good, with one (opposite!) comment about output

PR #2475, opened 2026-06-06
refactor: MoleculeTorsion node

trisyoungs:
Moves the MoleculeTorsion module to a node.


src/nodes/moleculeTorsion.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+    // Inputs
+    addInput("Configuration", "Set target configuration for the module", configuration_)->setFlags({ParameterBase::Required});
+
+    // Options
+    addOption("Species", "Target species containing the torsion", species_);
+    addOption("I", "Index of atom i in the torsion i-j-k-l", i_);
+    addOption("J", "Index of atom j in the torsion i-j-k-l", j_);
+    addOption("K", "Index of atom k in the torsion i-j-k-l", k_);
+    addOption("L", "Index of atom l in the torsion i-j-k-l", l_);
```
rprospero, line 18: Again, I think that we need an output for `frequency_`

rprospero (approved):
Looks good, besides that comment I've had on everything else.

PR #2328, opened 2026-03-02
feat: AvgMol Node

rprospero:
This PR ports the AvgMol module into a new node.  I've done my best to make the transition as direct as possible, while trying to be idiomatic to the new node setup.

Issue #2488, opened 2026-06-09
Setting certain parameters must call clearData()


trisyoungs:
Setting some parameters must necessarily clear any already-calculated data from the `Node` (e.g. the `SpeciesSite` in `AverageMoleculeNode`).

We might also consider introducing the concept of "locking" such parameters to prevent users from accidentally modifying a number which will can all of their current calculate data in the node.

PR #2490, opened 2026-06-09
refactor: AverageMolecule node

trisyoungs:
Heavy refactor of `AvgMolModule` into `AverageMoleculeNode`.

Obsoletes #2328.


src/nodes/averageMolecule.cpp
```diff
@@ -0,0 +1,110 @@
...
+    {
+        message("Initialising arrays for average molecule: size = {}\n", requiredSize);
+
+        sampledX_.emplace().initialise(requiredSize);
+        sampledY_.emplace().initialise(requiredSize);
+        sampledZ_.emplace().initialise(requiredSize);
+
+        // Copy current geometry of species to set up atom / bond relationships in the Structure
+        structure_ = targetSpecies->asStructure();
+    }
```
rprospero, line 67: If I change the target species in the GUI, does sampledX_ get resized to the new required size?  The current logic seems to only check if the vector has every been initialised at all.

I'm wondering if we might be better off using the raw vector instead of an optional one.  Our test would then be whether the vector has the correct length, instead of checking whether the optional has been filled.

rprospero (approved):
I have a question/suggestion about how we're handling the sampled coordinates, but everything else looks good

src/nodes/averageMolecule.cpp
trisyoungs: So this is effectively captured by issue #2488 which I put in yesterday - changing the input `Site` (== `Species`) should absolutely force the data to be cleared and reset.  We should handle this in a general way at the `Node`/`Parameter` level, and we already have `virtual Node::clearData()`. What we don't currently have is the flag on `Parameter` to call that function if the parameter data changes, hence #2488.

PR #2491, opened 2026-06-10
chore: Remove Averaging-related restart files.

trisyoungs:
Removes two restart files that were added to the repo at some point.


rprospero (approved):
Looks like adding those was my fault.

PR #2492, opened 2026-06-10
chore: Remove some version tests

trisyoungs:
This PR remove one version test that didn't really do anything, and one which checked the semantic version of an old input file read in and converted to TOML. The latter is now irrelevant, and I'm not sure if it needs to be replaced by anything for Dissolve2?


rprospero (approved):
We'll need a proper version stability test before we release Dissolve 2, but this is not that test.  That actually brings up an interesting question that ties into some of my current work - should a user expect restart files to still work after upgrading Dissolve?  Does this answer change for patch vs minor vs major releases?

trisyoungs:
> We'll need a proper version stability test before we release Dissolve 2, but this is not that test. That actually brings up an interesting question that ties into some of my current work - should a user expect restart files to still work after upgrading Dissolve? Does this answer change for patch vs minor vs major releases?

Excellent question. I think this warrants a slightly longer discussion, but my instinct is that we should be as permissive as possible when it comes to restart data, or at least make it a better experience for the user when bits fail to load in.  With Dissolve2 separating out restart data per-`Node`, that obviously becomes a lot easier!  I suppose we could tag every bit of restart data with the code version and give ourselves the option to handle differences at the point of deserialisation...

trisyoungs:
At any rate, I don't think the data written out will change too much for a given existing `Node` even between major versions.

PR #2498, opened 2026-06-11
chore: Remove remaining modules

trisyoungs:
## Follows #2495 - do not merge until #2495 is complete

This PR removes the remaining modules and tidies up the remaining old input files. We begin!

Also note number of removed lines - over 100k!  Is that a record? :)


src/modules/registry.cpp
```diff
@@ -2,44 +2,8 @@
...
-    registerProducer<NeutronSQModule>(ModuleTypes::NeutronSQ, "Calculate neutron-weighted S(Q)", "Correlation Functions");
-    registerProducer<SQModule>(ModuleTypes::SQ, "Transform g(r) into unweighted S(Q)", "Correlation Functions");
-    registerProducer<TemperatureScheduleModule>(ModuleTypes::TemperatureSchedule,
-                                                "Adjust the temperature of a configuration during a simulation", "Evolution");
-    registerProducer<TestModule>(ModuleTypes::Test, "Development Module");
-    registerProducer<TRModule>(ModuleTypes::TR, "Calculate the partial and total t(r)", "Correlation Functions");
-    registerProducer<XRaySQModule>(ModuleTypes::XRaySQ, "Calculate x-ray-weighted S(Q)", "Correlation Functions");
-}
+
+ModuleRegistry::ModuleRegistry() {}
```
RobBuchananCompPhys, line 6:
```suggestion
ModuleRegistry() = default;
```

Any chance we can make this constructor default in a header file (and remove the empty one here)?

trisyoungs: We could, but it all gets deleted in #2500!

rprospero (approved):
This is certainly our largest net loss of lines.  I think we did once have a PR that changed more lines, but it was +300k/-300k.  I'm really curious how much this will improve compile times.

PR #2499, opened 2026-06-11
feat: `Structure` incorporates `LocalMolecule` manipulations

RobBuchananCompPhys:


src/classes/structure.cpp
```diff
@@ -26,7 +27,8 @@ Structure &Structure::operator=(const Structure &source)
         addBond(bond->i()->index(), bond->j()->index());
 
     // Copy source box
-    createBox(source.box_.axisLengths(), source.box_.axisAngles(), source.box_.type() == Box::BoxType::None);
+    if (source.box_)
+        createBox(source.box_->axisLengths(), source.box_->axisAngles(), source.box_->type() == Box::BoxType::NonPeriodic);
```
trisyoungs: This will need a slight refactoring following #2507

src/classes/structure.cpp
trisyoungs, line 125:
```suggestion
// Return positional instances of the root structure
```

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
     box_ = Box::generate(lengths, angles);
 }
 
+/*
+ * Manipulations
+ */
+
+// Recursive function for general manipulation
+void Structure::recurseLocal(std::vector<bool> &flags, const Box *box, int indexI, ManipulationFunction action)
```
trisyoungs: Don't need the `Box` argument here since we should use the local one. This is true of all the functions moved in from `Molecule`.

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
...
+            continue;
+
+        action(j.get(), box->minimumImage(j->r(), rI));
+
+        // Recurse into bound neighbours
+        recurseLocal(flags, box, indexJ, action);
+    }
+}
+
+// General manipulation function working on reassembled molecule
```
```suggestion
// General manipulation function working on reassembled fragments
```

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
...
+void Structure::traverseLocal(const Box *box, ManipulationFunction action)
+{
+    std::vector<bool> flags(atoms_.size(), false);
```
trisyoungs: Since we will be potentially calling this function many times - e.g. when unfolding multiple molecules in a box - perhaps this could be a `std::set<int>` (or even `std::set<StructureAtom*>`) instead?

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
...
+    recurseLocal(flags, box, 0, action);
+}
+void Structure::traverseLocal(const Box *box, ConstManipulationFunction action) const
+{
+    std::vector<bool> flags(atoms_.size(), false);
+    action(atoms_[0].get(), atoms_[0]->r());
+    recurseLocal(flags, box, 0, action);
+}
+
+// Un-fold molecule so it is not cut by box boundaries
```
trisyoungs, line 326: Here we need to think bigger in terms of what the `Structure` might contain - a `Molecule` for sure only had one bound fragment in it, but a `Structure` can have many. So, `unFold` will need to unfold all fragments it can find. Similar to the main loop in "DetectMolecule", we'll have to keep track of which atoms have been selected and transformed as part of fragments (perhaps `traverseLocal` could return the `std::set` it creates?) and operate until there are no un-folded fragments remaining.

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
...
+    Vector3 newR;
+    const auto cog = centreOfGeometry();
+
+    // Apply transform
+    for (auto n = 0; n < nAtoms(); ++n)
+    {
+        newR = box_->minimumVector(atom(n)->r(), cog) + newCentre;
+        atom(n)->setR(newR);
+    }
+}
```
trisyoungs: Makes sense for a single molecule, but doesn't for a periodic box, so I'd probably remove this function.

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
...
+
+// Calculate and return centre of geometry over supplied atom indices
+Vector3 Structure::centreOfGeometry(const std::vector<int> &indices) const
+{
+    const auto ref = atoms_[indices.front()]->r();
+    return std::accumulate(std::next(indices.begin()), indices.end(), ref,
+                           [&](const auto &acc, const auto idx) { return acc + box_->minimumImage(atoms_[idx]->r(), ref); }) /
+           indices.size();
+}
+
```
trisyoungs: Similarly, these aren't meaningful across all possible `Structure` contents so should be removed.

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
...
+    const auto ref = atoms_[indices.front()]->r();
+    auto sums = std::accumulate(std::next(indices.begin()), indices.end(), std::pair<Vector3, double>(ref * mass, mass),
+                                [&](const auto &acc, const auto idx)
+                                {
+                                    auto mass = AtomicMass::mass(atoms_[idx]->Z());
+                                    return std::pair<Vector3, double>(
+                                        acc.first + box_->minimumImage(atoms_[idx]->r(), ref) * mass, acc.second + mass);
+                                });
+    return sums.first / sums.second;
+}
```
trisyoungs: And here!

src/classes/structure.cpp
```diff
@@ -255,6 +255,188 @@ void Structure::createBox(const Matrix3 &axes)
...
+    for (auto n = 0; n < nAtoms(); ++n)
+        *atom(n) += delta;
+}
+
+// Translate specified atoms by the delta specified
+void Structure::translate(const Vector3 &delta, const std::vector<int> &targetAtoms)
+{
+    for (const auto i : targetAtoms)
+        *atom(i) += delta;
+}
```
trisyoungs: Could remove these for now too - not sure how useful they are or will be.

trisyoungs (commented):
Few things to sort out here. Also it is quite confusing for this to be based off the detectmolecules branch as there are changes from all sorts of places coming in. I've focussed only on the `Structure` class here.

PR #2500, opened 2026-06-11
chore: Remove module class

trisyoungs:
This PR removes the `Module` class itself, plus the `modules/` and `module/` directories, and associated code, models, and keywords.


RobBuchananCompPhys (approved):

src/gui/models/dissolveModelImageProvider.cpp
```diff
@@ -3,7 +3,6 @@
 
 #include "gui/models/dissolveModelImageProvider.h"
 #include "gui/models/dissolveModel.h"
-#include "gui/models/moduleLayerModel.h"
 
 DissolveModelImageProvider::DissolveModelImageProvider(DissolveModel *dissolveModel)
```
rprospero, line 7: Honestly, I think that this entire class can go.  It's only ever used once in the OverviewTab and I don't think that we will need it in the new QML GUI.

rprospero (approved):
Looks good.  One small, optional suggestionl

src/gui/models/dissolveModelImageProvider.cpp
trisyoungs: @rprospero There are no doubt a bunch of other models that we will no longer need or are no longer relevant, so I'll leave this in place for now and address it with the others.

PR #2505, opened 2026-06-12
build: OSX tests

trisyoungs:
This PR enables the test suite on OSX builds and fixes a couple of failing tests that cropped up as a result.


RobBuchananCompPhys:
Although this PR in particular is oriented towards OSX builds, it looks increasingly like we might be able to remove the conan `msvc_dev` option altogether. I'm just running some checks to see whether in light of this PR, we can do without it when building the Dissolve Visual Studio project. I could then branch off and bring in any required fixes.

rprospero (approved):

PR #2507, opened 2026-06-12
refactor:Concrete box class

rprospero:
The Box class is now a single type that handles everything with a switch statement, instead of using subclasses.  This makes it possible for a Box to trivially switch types.

This PR also eliminated the NonPeriodic box type and uses SingleImage where that had been used.

There was some concern that requiring the box to check its own type in a hot loop would be a speed issue, but the benchmark below shows the opposite.
fig

To explain, the top left plot is the benchmark timings (in ns) is the speed of the new code, while the top right is the speed of the old code.  The bottom left is the percentage speed up that the new code gives and the bottom right is the absolute speedup in ns.  The results show a signficant speed up in our fastest tests and no test increasing its speed by more than a nanosecond, so I feel that we're not slowing out hot loop.


src/classes/box.h
```diff
@@ -22,18 +22,18 @@ class Box : public Serialisable<>
...
     {
-        NonPeriodic,     /* Non-periodic system - cubic box, but no minimum image calculation */
         Cubic,           /* Cubic box with A == B == C, alphe == beta == gamma == 90 */
         Orthorhombic,    /* Orthorhombic box with A != B != C, alphe == beta == gamma = 90 */
         MonoclinicAlpha, /* Monoclinic box with A != B != C, alpha != 90, and beta == gamma == 90 */
         MonoclinicBeta,  /* Monoclinic box with A != B != C, beta != 90, and alpha == gamma == 90 */
         MonoclinicGamma, /* Monoclinic box with A != B != C, gamma != 90, and alpha == beta == 90 */
-        Triclinic        /* Triclinic box with A != B != C, alpha != beta != gamma != 90 */
+        Triclinic,       /* Triclinic box with A != B != C, alpha != beta != gamma != 90 */
+        SingleImage      /* Box with only one item */
```
trisyoungs: So in reality the `SingleImage` options corresponds to the simplest case of no periodicity in any direction, and for clarity I think we should call it as such. I think rename `SingleImage` to `None` and make it the first in the list to make things clear?

tests/classes/structure.cpp
```diff
@@ -53,6 +53,7 @@ TEST(StructureTest, Molecule2)
     s.addAtom(Elements::C, {1.241877e+00, 5.832093e-01, -3.044739e-02});
     CalculateBondingNode::calculate(s);
 
+    std::cout << "Type: " << (int)s.box()->type() << std::endl;
```
trisyoungs: Bit of debugging output?

trisyoungs (changes requested):
Excellent. I'm pleasantly surprised by the benchmark results! Overall this tidies things up no end. I think we should rename `SingleImage` to `None` as I mentioned in a comment, and the source files related to the specific types of `Box` (e.g. `boxCubic.cpp`) can now all be removed.  Finally, the only reason `Box` ever existed as a `std::unique_ptr` in `Configuration` and `Species` was because of it's inheritance.  Now that this is all removed, it can go back to being a simple member variable.  I appreciate that might be a bit of a chore to follow through the code, but I think it's worth doing!

tests/classes/structure.cpp
rprospero: Oops

src/classes/box.h
rprospero: Agreed

rprospero:
@trisyoungs I've moved the Box into just a regular object instead of a unique_ptr and switched most of the calls to using references.  This has *vastly* expanded the scope of the PR, but most of these changes are just using references instead of pointers.

Fun fact: it would have been possible for the original C language to unify member access for both structures and pointers, but it would have broken compatibility with some BCPL features that they had considered adding in the future (but were never actually added).

trisyoungs:
@rprospero - indeed it has, but the result is excellent and I'm very happy to see the back of `Box*`!

trisyoungs (approved):

PR #2509, opened 2026-06-12
refactor: Remove Data1DImportFileFormat

trisyoungs:
This PR removes `Data1DImportFileFormat` and passes responsibility for loading external XY or histogram data on to `ImportData1DNode` (renamed from `Data1DImportNode`).


src/nodes/data1DImport.cpp
RobBuchananCompPhys, line 1: That didn't last long! 😄

RobBuchananCompPhys (approved):

src/nodes/data1DImport.cpp
trisyoungs: Well, we didn't have a naming convention back then! :smile:

PR #2512, opened 2026-06-13
refactor: Benchmarks with Graphs

trisyoungs:
This PR overhauls our benchmarks to make them `Node`/`Graph`-centric rather than `Module`-centric, and removes it's dependency on `Dissolve` and `CoreData`.


benchmark/common/problems.h
```diff
@@ -8,68 +8,95 @@
...
+                species_->createAtomic(Elements::element(DissolveSys::beforeChar(speciesString, '|')),
+                                       {ShortRangeFunctions::Form::LennardJones, DissolveSys::afterChar(speciesString, '|')});
+        }
 
-        // Set the population of the test species in a predefined Parameters node in the generator
-        auto populationParameter = cfg->generator().rootSequence().parameterExists("Population");
-        populationParameter->setValue(populationSizes[population]);
+        // Add an Insert node for the Species
+        auto *insertNode = dynamic_cast<InsertNode *>(createNode("Insert"));
+        insertNode->setInput<Number>("Population", speciePopulation);
```
RobBuchananCompPhys:
```suggestion
        insertNode->setInput<Number>("Population", speciesPopulation);
```

Just checking - is the above correct?

benchmark/common/problems.h
```diff
@@ -8,68 +8,95 @@
...
+            {SystemType::Water1000, {"tests/data/species/water.toml", 1000, "tests/data/dlpoly/water1000/full.REVCON"}},
+            {SystemType::Hexane200, {"tests/data/species/hexane.toml", 200, "tests/data/dlpoly/hexane200/full.REVCON"}}};
 
-        // Population Sizes
-        static std::map<SpeciesPopulation, int> populationSizes = {{SpeciesPopulation::Single, 1},
-                                                                   {SpeciesPopulation::Small, 1000},
-                                                                   {SpeciesPopulation::Medium, 5000},
-                                                                   {SpeciesPopulation::Large, 10000}};
-        Messenger::setQuiet(true);
+        auto &[speciesString, speciePopulation, referenceCoordinates] = systemTypes[systemType];
```
RobBuchananCompPhys:
```suggestion
        auto &[speciesString, speciesPopulation, referenceCoordinates] = systemTypes[systemType];
```
See above

benchmark/ff/forces.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+// Single atom energy in different box shapes
+template <SystemType systemType> static void BM_CalculateForces_AtomicWorldForces(benchmark::State &state)
+{
+    GraphProblem<systemType> problemDef;
+    auto forceKernel = problemDef.createForceKernel(problemDef.configuration());
+    auto &i = problemDef.configuration()->atom(0);
+    std::vector<Vector3> ppForces, geometryForces;
+    for (auto _ : state)
+        forceKernel->totalForces(ppForces, geometryForces);
```
RobBuchananCompPhys, line 18: Iterating but not using the output? I imagine something different is going on under hood with each iteration, otherwise I don't see the use of this.

trisyoungs: Not sure what you mean here?  We're only calling the function to assess the speed, not to do anything useful with the results.  Because we're not returning and storing a local variable I don't think the compiler can optimise this loop out so it should be valid.

RobBuchananCompPhys (approved):

PR #2518, opened 2026-06-15
build: Refactor VS dev setup

RobBuchananCompPhys:


develop.ps1
```diff
@@ -32,12 +30,13 @@
         Generator to use (options are "Visual Studio 17 2022", "Ninja").
     .PARAMETER release
         Flag - install packages for release, otherwise debug.
+    .PARAMETER benchmarks
+        Flag - build benchmarks (only applicable in debug).
```
trisyoungs: Is this true? Benchmarks need to be run in Release mode really...

develop.ps1
```diff
@@ -58,6 +57,7 @@ $binSuffix = "d"
 
 if ($release) {
     $build = "Release"
+    $benchmarks = $false
```
trisyoungs: Again, setting this to false for a Release build feels like the opposite to what we should do!  More so, it should really be a user-chosen option, rather than a build-type related option, if you see what I mean.

trisyoungs (approved):

PR #2525, opened 2026-06-22
chore: Node tidy

trisyoungs:
It's a contender for most boring, unreviewable PR of 2026, but here is a `Node`-wide tidy up PR.  The goal was to make all the node definitions consistent in terms of formatting, layout, commenting etc.  Bit OCD, but it's done now!  I have also removed commented code and a few other small bits of unused / irrelevant stuff here and there, and done a little bit of renaming, but nothing of note.

All the `Node`s now look like this:
```
class MyNode : public Node
{
    public:
    MyNode(Graph *parentGraph);
    ~MyNode() override = default;

    /*
     * Definition
     */
    public:
    // Return type of the node

    /*
     * Data
     */
    private:
    /* Specific inputs, options, outputs, and data generated by the node

    public:
    // Clear any local data
    void clearData() override;      /* If required, typically analysis nodes, plus GR, SQ etc. */

    /*
     * Processing
     */
    private:
    /* Local functions implemented by the node */

    protected:
    // Perform processing
};
```


RobBuchananCompPhys (approved):
Very cool, glad to see those nodes broken out of sub-dirs too!

PR #2528, opened 2026-06-22
refactor: Detect molecules

RobBuchananCompPhys:


src/classes/fragment.h
trisyoungs, line 1: Does this still need a rebase? This was in #2530 ...

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,232 @@
...
+#include "classes/species.h"
+#include <algorithm>
+#include <format>
+
+DetectMoleculesNode::DetectMoleculesNode(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput<Structure>("Structure", "Input structure", inputStructure_);
+}
+
```
trisyoungs, line 17:
```suggestion

/*
 * Definition
 */
 
```

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,232 @@
...
+std::string_view DetectMoleculesNode::type() const { return "DetectMolecules"; }
+
+std::string_view DetectMoleculesNode::summary() const { return "Detect molecular species within a structure"; }
+
```
trisyoungs, line 21:
```suggestion

/*
 * Processing
 */
 
```

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,232 @@
...
+std::vector<std::vector<int>> DetectMoleculesNode::findMolecularFragments(const Structure &structure) const
+{
+    std::vector<std::vector<int>> fragments;
+
+    auto fragment = [structure](int i) { return Fragment<StructureAtom, Bond<StructureAtom>>::get(structure.atoms(), i); };
+
+    for (int i = 0; i < structure.nAtoms(); i++)
+        fragments.emplace_back(fragment(i));
```
trisyoungs, line 29: Shouldn't this function start from the next un-selected atom? Need something like a `std::set` to track which atoms have been detected within instances (a bug that doesn't show up for the atomic NaCl or MgO cases!)

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,232 @@
...
+    return fragments;
+}
+
+// Run main processing
+NodeConstants::ProcessResult DetectMoleculesNode::process()
+{
+    detectedStructures_.clear();
+
+    // Return all discovered molecular fragment index vectors
+    auto allFragmentIndices = findMolecularFragments(inputStructure_);
```
trisyoungs, line 40: If you're pre-detecting all bound fragments here (which is a good idea)....

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    // Try selecting within the species from the first atom - if this captures all atoms we have a bound framework...
+    if (allFragmentIndices[0].size() == inputStructure_.nAtoms())
+        return error(
+            "Can't create molecular definitions since this unit cell appears to be a continuous framework/network. Consider "
+            "adjusting the bonding options in order to generate molecular fragments.\n");
+
+    std::set<const StructureAtom *> atomMask;
+
+    for (int i = 0; i < inputStructure_.nAtoms(); i++)
```
trisyoungs, line 50: ...why is this a loop over structure atoms and not fragments?

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,232 @@
...
+                    // Replace the current match if there are more bonds on the current atom.
+                    if (fragmentAtom->nBonds() > rootAtoms.front()->nBonds())
+                        better = true;
+                }
+
+                if (better)
+                {
+                    bestNETA = neta;
+                    rootAtoms = currentRootAtoms;
+                }
```
trisyoungs, line 138: This whole NETA creation part needs to run on each atom within a fragment, reason being to find the best, most concrete way of describing uniquely the bonding / atoms within it.  Then it can be applied to each fragment (of the same size) in turn to see if the NETA matches.

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,232 @@
...
+            if (!instances.empty())
+                detectedStructures_
+                    .emplace_back(copyStructureAtomsAndBonds(inputStructure_, detectedMolStructure, fragmentIndices))
+                    .instances() = instances;
+        }
+    }
+
+    // Unfold all detected structures
+    for (auto &structure : detectedStructures_)
+        structure.unFold();
```
trisyoungs, line 182: This step needs to happen before you store the coordinates of the instances at the end of the loop above. Could happen immediately after the fragment detection at the top of the routine.

trisyoungs (commented):
This is excellent and a whole lot cleaner than the old one, but I think there are some looping / workflow errors to address!

trisyoungs:
Superseded by #2556

PR #2530, opened 2026-06-23
refactor: `Structure` manipulation and `SupercellConfigurationNode`

RobBuchananCompPhys:


src/classes/structure.cpp
```diff
@@ -246,6 +246,85 @@ void Structure::createBox(const Matrix3 &axes)
...
+}
+std::set<StructureAtom *> Structure::traverseLocal(ConstManipulationFunction action) const
+{
+    std::set<StructureAtom *> flags;
+    action(atoms_[0].get(), atoms_[0]->r());
+    recurseLocal(flags, 0, action);
+    return flags;
+}
+
+// Un-fold molecule so it is not cut by box boundaries
```
```suggestion
// Un-fold bound fragments in the structure
```

src/classes/structure.h
```diff
@@ -116,6 +117,24 @@ class Structure : public Serialisable<>
...
+    using ConstManipulationFunction = std::function<void(const StructureAtom *j, Vector3 rJ)>;
+    // Recursive function for general manipulation
+    void recurseLocal(std::set<StructureAtom *> &flags, int indexI, ManipulationFunction action);
+    void recurseLocal(std::set<StructureAtom *> &flags, int indexI, ConstManipulationFunction action) const;
+    // General manipulation function working on reassembled molecule
+    std::set<StructureAtom *> traverseLocal(ManipulationFunction action);
+    std::set<StructureAtom *> traverseLocal(ConstManipulationFunction action) const;
+
+    public:
+    // Un-fold molecule so it is not cut by box boundaries, returning the centre of geometry
```
trisyoungs:
```suggestion
    // Un-fold bound fragments in the structure
```

src/classes/structure.cpp
```diff
@@ -246,6 +246,85 @@ void Structure::createBox(const Matrix3 &axes)
...
+void Structure::unFold()
+{
+    std::set<StructureAtom *> flagged;
+
+    while (flagged.size() <= nAtoms())
+        flagged.merge(traverseLocal([](StructureAtom *j, Vector3 rJ) { j->setR(rJ); }));
```
trisyoungs: This won't work as expected since the first unflagged atom will need to be passed to `traverseLocal()` - it currently assumes the zeroth atom as the starting point.

src/classes/structure.cpp
```diff
@@ -246,6 +246,85 @@ void Structure::createBox(const Matrix3 &axes)
...
+
+        action(j.get(), box_.minimumImage(j->r(), strAtI->r()));
+
+        // Recurse into bound neighbours
+        recurseLocal(flags, indexJ, action);
+    }
+}
+
+// General manipulation function working on reassembled fragments
+std::set<StructureAtom *> Structure::traverseLocal(ManipulationFunction action)
```
trisyoungs: Tempted by a rename here - `traverseLocal()` doesn't really convey what's happening at all!
```suggestion
// Return atoms in the same fragment as the specified atom, unfolding the fragment at the same time
std::set<StructureAtom *> Structure::getUnfoldedFragment(StructureAtom *containing, ManipulationFunction action)
```

src/classes/structure.cpp
```diff
@@ -246,6 +246,85 @@ void Structure::createBox(const Matrix3 &axes)
...
+{
+    std::set<StructureAtom *> flags;
```
trisyoungs: Another rename suggestion...
```suggestion
    std::set<StructureAtom *> fragmentAtoms;
```

src/classes/structure.cpp
```diff
@@ -246,6 +246,85 @@ void Structure::createBox(const Matrix3 &axes)
     box_ = Box::generate(lengths, angles);
 }
 
+/*
+ * Manipulations
+ */
+
+// Recursive function for general manipulation
+void Structure::recurseLocal(std::set<StructureAtom *> &flags, int indexI, ManipulationFunction action)
```
trisyoungs: I also realise now that the reason for passing `indexI` rather than just a `StructureAtom *` was because previously the flags object was an array sized to the number of atoms. Now we're using a `std::set` we can forget about indices and just pass the pointer.

src/nodes/supercellConfiguration.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    // Inputs
+    addInput<Configuration *>("Configuration", "Target configuration", targetConfiguration_);
+
+    // Outputs
+    addPointerOutput<Configuration>("SupercellConfiguration", "Supercell configuration", supercellConfiguration_);
+
+    // Options
+    addOption<Vector3i>("SupercellRepeat", "Integer coefficients by which unit cell will be repeated along its dimensions",
+                        supercellRepeat_);
```
trisyoungs: Can remove the explicit template args to clean these lines up.

src/nodes/supercellConfiguration.cpp
```diff
@@ -0,0 +1,70 @@
...
+std::string_view SupercellConfigurationNode::summary() const
+{
+    return "Create a repeated instance (supercell) of a configuration";
+}
+
+/*
+ * Processing
+ */
+
+// Run main processing
```
trisyoungs: Following #2525...
```suggestion
// Perform processing
```

src/nodes/supercellConfiguration.cpp
```diff
@@ -0,0 +1,70 @@
...
+SupercellConfigurationNode::SupercellConfigurationNode(Graph *parentGraph) : Node(parentGraph)
+{
+    // Inputs
+    addInput<Configuration *>("Configuration", "Target configuration", targetConfiguration_);
+
+    // Outputs
+    addPointerOutput<Configuration>("SupercellConfiguration", "Supercell configuration", supercellConfiguration_);
+
+    // Options
+    addOption<Vector3i>("SupercellRepeat", "Integer coefficients by which unit cell will be repeated along its dimensions",
```
trisyoungs:
```suggestion
    addOption<Vector3i>("SupercellRepeat", "Integer coefficients by which unit cell will be repeated along its axes",
```

src/nodes/supercellConfiguration.cpp
```diff
@@ -0,0 +1,70 @@
...
+        const auto *sp = mol->species();
+
+        // Loop over cell images
+        for (auto ix = 0; ix < supercellRepeat_.x; ++ix)
+        {
+            for (auto iy = 0; iy < supercellRepeat_.y; ++iy)
+            {
+                // Create and translate molecule
+                for (auto iz = 0; iz < supercellRepeat_.z; ++iz)
+                    supercellConfiguration_.addMolecule(sp)->translate(Vector3(ix, iy, iz));
```
trisyoungs: The translation vector needs to be multiplied by the original box axes matrix.

src/nodes/supercellConfiguration.h
```diff
@@ -0,0 +1,41 @@
...
+class SupercellConfigurationNode : public Node
+{
+    public:
+    SupercellConfigurationNode(Graph *parentGraph);
+    ~SupercellConfigurationNode() override = default;
+
+    /*
+     * Definition
+     */
+
```
trisyoungs:
```suggestion
```

src/nodes/supercellConfiguration.h
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#pragma once
+
+#include "classes/configuration.h"
+#include "nodes/node.h"
+
+// SupercellConfiguration Node
```
trisyoungs:
```suggestion
```

src/nodes/supercellConfiguration.h
```diff
@@ -0,0 +1,41 @@
...
+    public:
+    std::string_view type() const override;
+    std::string_view summary() const override;
+
+    /*
+     * Data
+     */
+    private:
+    // Input configuration
+    Configuration *targetConfiguration_;
```
trisyoungs:
```suggestion
    Configuration *targetConfiguration_{nullptr};
```

src/classes/structure.cpp
```diff
@@ -246,6 +246,86 @@ void Structure::createBox(const Matrix3 &axes)
...
+ * Manipulations
+ */
+
+// Recursive function for general manipulation
+void Structure::recurseLocal(std::set<StructureAtom *> &fragmentAtoms, StructureAtom *i, ManipulationFunction action)
+{
+    if (fragmentAtoms.contains(i))
+        return;
+
+    // Set the flag for indexI and get some necessary values
```
```suggestion
```

src/classes/structure.cpp
```diff
@@ -246,6 +246,86 @@ void Structure::createBox(const Matrix3 &axes)
...
+        // Recurse into bound neighbours
+        recurseLocal(fragmentAtoms, j, action);
+    }
+}
+void Structure::recurseLocal(std::set<StructureAtom *> &fragmentAtoms, StructureAtom *i, ConstManipulationFunction action) const
+{
+    if (fragmentAtoms.contains(i))
+        return;
+
+    // Set the flag for indexI and get some necessary values
```
```suggestion
```

src/classes/structure.cpp
```diff
@@ -246,6 +246,86 @@ void Structure::createBox(const Matrix3 &axes)
...
+    while (fragmentAtoms.size() < nAtoms())
+    {
+        auto atomIt = std::find_if(atoms_.begin(), atoms_.end(),
+                                   [&fragmentAtoms](const auto &atom) { return !fragmentAtoms.contains(atom.get()); });
+        if (atomIt == atoms_.end())
+            break;
+
+        fragmentAtoms.merge(getUnfoldedFragment(atomIt->get(), [](StructureAtom *j, Vector3 rJ) { j->setR(rJ); }));
+    }
+}
```
trisyoungs: To avoid the full search over the `atoms_` array each time you could move the loop over atoms to the outside (haven't tested this, so hopefully my syntax is correct):
```suggestion
    for (auto &atom : atoms_)
    {
        if (fragmentAtoms.contains(atom.get())
            continue;

        fragmentAtoms.merge(getUnfoldedFragment(atom.get(), [](StructureAtom *j, Vector3 rJ) { j->setR(rJ); }));
    }
}
```

src/nodes/supercellConfiguration.cpp
```diff
@@ -0,0 +1,69 @@
...
+/*
+ * Processing
+ */
+
+// Perform processing
+NodeConstants::ProcessResult SupercellConfigurationNode::process()
+{
+    supercellConfiguration_.empty();
+
+    const auto box = targetConfiguration_->box();
```
trisyoungs:
```suggestion
    const auto &box = targetConfiguration_->box();
```

tests/classes/structure.cpp
```diff
@@ -68,4 +70,24 @@ TEST(StructureTest, Molecule2)
...
+
+    // After unfolding, the distances between C and H should all be unity
+    for (const auto &bond : structure.bonds())
+    {
+        auto iR = bond->i()->r();
+        auto jR = bond->j()->r();
+        auto distance = abs((iR - jR).magnitude());
+        ASSERT_NEAR(distance, 1, 10e-9);
+    }
+}
```
trisyoungs, line 104: From what is written here, I think this test returns a false positive. Reason is that there are no bonds in the structure because there is no `CalculateBonding` node applied to the structure!

trisyoungs (approved):
Looking really good - one significant comment made on the unit test though!

PR #2535, opened 2026-06-24
refactor: De-template Serialisable

trisyoungs:
This PR takes the opportunity to remove the now-unnecessary context on `Serialisable`, allowing it to be simplified down to a normal `class` with templated functions. I am hopeful that this, along with a liberal sprinkling of missing `override` specifiers, will make the output on OSX builds terse enough to actually interpret.

### DIFF
--- a/src/base/serialiser.h
+++ b/src/base/serialiser.h
@@ -3,11 +3,11 @@
 
 #pragma once
 
-#include <toml11/toml.hpp>
 #include "templates/keyedVector.h"
 #include "templates/orderedMap.h"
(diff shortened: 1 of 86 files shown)


RobBuchananCompPhys (approved):
🙌

rprospero (approved):
It's outside the scope of this PR, but, since Serialisable is no longer a template class, we can more most of the method implementations into a corresponding `cpp` file and cut down on our compile time.

There's also come interesting things we could do, now that we're on C++20, where we make serialisable a concept instead of a class, but that is something to look at *after* Dissolve 2.0 is out.

PR #2537, opened 2026-06-25
feat: Data class serialisers

trisyoungs:
This adds in TOML serialisation for a bunch of data-related classes that were completely missing it.

PR #2538, opened 2026-06-25
chore: Mark missing serialisable data

trisyoungs:
This PR modifies a fair number of nodes to include the necessary `addSerialisable()` calls for their "restartable" data (i.e. that which was tagged with `InRestartFile` in Dissolve1). It will be in draft until #2537 is complete as it depends on the serialisers implemented therein. There are several outstanding "exotic" serialisers that need to be written in order to make the coverage 100%, but these are captured in a separate issue (#2539) to be addressed in the very near future.


RobBuchananCompPhys (approved):
👍

PR #2542, opened 2026-06-26
chore: Clean up testing framework

trisyoungs:
This is a big clean-up PR of the test classes we had, hopefully making things a little clearer and setting up for follow-on work with `Node` test fixtures.


tests/testing.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+// Test simple double
+[[nodiscard]] bool testDouble(std::string_view quantity, double A, double B, double threshold)
+{
+    auto delta = fabs(A - B);
+    auto isOK = delta <= threshold;
+    Messenger::print("Reference {} delta with correct value is {:15.9e} and is {} (threshold is {:10.3e})\n", quantity, delta,
+                     isOK ? "OK" : "NOT OK", threshold);
+    return isOK;
+}
```
rprospero, line 26:
```suggestion
// Test simple double
[[nodiscard]] testing::AssertionResult testDouble(std::string_view quantity, double A, double B, double threshold)
{
    auto delta = fabs(A - B);
    if (delta <= threshold) {
        return testing::AssertionSuccess();
    } else {
        return testing::AssertionFailure() << std::format("{} values {} and {} have a difference of {}, which exceeds {}", quantity, A, B, delta, threshold);
    }
}
```

Using [predicate assertions](https://google.github.io/googletest/advanced.html#using-a-function-that-returns-an-assertionresult), we can attach extra data to the failed tests, but ensure that the test output still gives the correct line number for where the test fails.  We would want to do something similar for each of the test functions.

tests/testing.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    return !notOK;
+}
+// Test Vec3 data
+void testVec3(const Vector3 &A, const Vector3 &B, double tolerance)
+{
+    EXPECT_NEAR(A.x, B.x, tolerance);
+    EXPECT_NEAR(A.y, B.y, tolerance);
+    EXPECT_NEAR(A.z, B.z, tolerance);
+}
```
rprospero: One pain point that I inflicted on myself while doing the parsers work is putting EXPECT statements in a wrapper function.  The problem is that, when the test fails, it will just tell you that a line in the wrapper function failed without telling you which line in your actual test called that wrapper function.  Returning a predicate assertion, like I explained above, ensure that the test runner reports back the actual place that the line failed.

rprospero (commented):
I've left a couple of comments about [predicate assertsions](https://google.github.io/googletest/advanced.html#using-a-function-that-returns-an-assertionresult) that might give us cleaner results from our tests.

tests/testing.cpp
trisyoungs: It's funny, while working on the node text fixture stuff I was looking through the gtest docs and saw the predicate assertion stuff. I think my thought process was "that sounds useful, but not what I need right now, so I'll read about it later".  That "later" would be now!

tests/testing.cpp
```diff
@@ -0,0 +1,466 @@
...
+        auto arr2 = toml2.as_array();
+        ASSERT_EQ(arr.size(), arr2.size()) << location << std::endl << "Expected" << std::endl << toml;
+        for (int i = 0; i < arr.size(); ++i)
+            compareToml(std::format("{}[{}]", location, i), arr[i], arr2[i]);
+    }
+    else
+    {
+        EXPECT_EQ(toml, toml2) << location;
+    }
+}
```
rprospero, line 448: It's outside the scope of this PR, but this should definitely be reworked to return an assertion predicate.

rprospero (approved):

RobBuchananCompPhys (approved):

PR #2544, opened 2026-06-29
refactor: Start removing line parser

rprospero:
This PR removes the LineParser dependency in `ImportDLPOLYStructureNode`, `ImportDLPOLYTrajectoryNode`, `ImportDLPUtilsPDensNode`, `ImportDLPUtilsSurfaceNode`, and `ImportMoscitoStructureNode`. 

It also fixes up `MDNode` to use the cleaner =std::format_to= for file output.

The one bit that I'm not happy about is that I needed to adjust the tolerance on `CosNTorsionForcesTest::POE` in order to pass the tests.  I only had to change the tolerance from 0.006 to 0.0062 and the actual value is in the thousands, but I'm always concerned when a test like this needs changing.  I know that the change comes from the new Moscito parser and I suspect that it's a slight difference in how doubles are parsed, which can result in some differences in the lowest bits.

### DIFF
--- a/src/base/parserLibrary.cpp
+++ b/src/base/parserLibrary.cpp
@@ -54,6 +54,9 @@ Parser<double> real()
 // A parser that accepts a 3-vector of floating point numbers
 Parser<Vector3> vector3()
 {
+    // Annoyingly, the Moscito file format does not guarantee spaces
+    // between the numbers when the numbers are negative.  Luckily, we
+    // can still parse this scenario.
     return (real() & maybe(spaces()) >> real() & maybe(spaces()) >> real())
         .apply([](double x, double y, double z) { return Vector3(x, y, z); });
 }
--- a/src/base/parserLibrary.cpp
+++ b/src/base/parserLibrary.cpp
@@ -57,6 +57,12 @@ Parser<Vector3> vector3()
     return (real() & spaces() >> real() & spaces() >> real())
         .apply([](double x, double y, double z) { return Vector3(x, y, z); });
 }
+// A parser that accepts a 3-vector of integers
+Parser<Vector3i> vector3i()
+{
+    return (integer() & spaces() >> integer() & spaces() >> integer())
+        .apply([](auto x, auto y, auto z) { return Vector3i(x, y, z); });
+}
 
 // A parser that accepts a 3x3 matrix of floating point numbers
 Parser<Matrix3> matrix3()
--- a/src/base/parserLibrary.cpp
+++ b/src/base/parserLibrary.cpp
@@ -54,13 +54,13 @@ Parser<double> real()
 // A parser that accepts a 3-vector of floating point numbers
 Parser<Vector3> vector3()
 {
-    return (real() & spaces() >> real() & spaces() >> real())
+    return (real() & maybe(spaces()) >> real() & maybe(spaces()) >> real())
         .apply([](double x, double y, double z) { return Vector3(x, y, z); });
 }
 // A parser that accepts a 3-vector of integers
 Parser<Vector3i> vector3i()
 {
-    return (integer() & spaces() >> integer() & spaces() >> integer())
+    return (integer() & maybe(spaces()) >> integer() & maybe(spaces()) >> integer())
         .apply([](auto x, auto y, auto z) { return Vector3i(x, y, z); });
 }
 
--- a/src/base/parserLibrary.cpp
+++ b/src/base/parserLibrary.cpp
@@ -58,4 +58,12 @@ Parser<Vector3> vector3()
         .apply([](double x, double y, double z) { return Vector3(x, y, z); });
 }
 
+// A parser that accepts a 3x3 matrix of floating point numbers
+Parser<Matrix3> matrix3()
+{
+    return (maybe(inlineSpaces()) >> vector3() << spaces() & maybe(inlineSpaces()) >> vector3() << spaces() &
+            maybe(inlineSpaces()) >> vector3() << spaces())
+        .apply([](const auto m1, const auto m2, const auto m3) { return Matrix3(m1, m2, m3); });
+}
+
 } // namespace Parsers
--- a/src/nodes/importDLPOLYStructure.cpp
+++ b/src/nodes/importDLPOLYStructure.cpp
@@ -2,6 +2,9 @@
 // Copyright (c) 2026 Team Dissolve and contributors
 
 #include "nodes/importDLPOLYStructure.h"
+#include "base/applicative.h"
+#include "base/parserLibrary.h"
+#include "data/elements.h"
 
 ImportDLPOLYStructureNode::ImportDLPOLYStructureNode(Graph *parentGraph) : Node(parentGraph)
 {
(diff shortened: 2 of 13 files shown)


src/base/parserLibrary.cpp
```diff
@@ -54,8 +54,22 @@ Parser<double> real()
 // A parser that accepts a 3-vector of floating point numbers
 Parser<Vector3> vector3()
 {
-    return (real() & spaces() >> real() & spaces() >> real())
+    return (real() & maybe(spaces()) >> real() & maybe(spaces()) >> real())
```
trisyoungs, line 60: Confused as to how there can "possibly be spaces" between real numbers? Would the match fail if there were zero spaces?

src/nodes/importDLPOLYStructure.cpp
```diff
@@ -82,52 +77,63 @@ NodeConstants::ProcessResult ImportDLPOLYStructureNode::read(LineParser &parser,
...
+            structure.addAtom(Elements::Unknown, position);
+            forces.push_back(*force);
         }
-
-        if ((nAtoms > 0) && (structure.nAtoms() == nAtoms))
-            break;
     }
 
-    // Copy forces out?
+    // // Copy forces out?
```
trisyoungs:
```suggestion
    // Copy forces out?
```

src/nodes/importDLPOLYTrajectory.cpp
```diff
@@ -38,35 +40,41 @@ std::string_view ImportDLPOLYTrajectoryNode::summary() const
...
     // Seek to the next file position
-    parser.seekg(filePosition_);
+    infile.seekg(filePosition_);
 
     // Read first line:  'timestep    <stepNo>  <nAtoms>  <keytrj>  <imcon>
-    if (parser.getArgsDelim(LineParser::Defaults) != LineParser::Success)
+    auto head = header().parse(infile);
+    if (!head)
         return NodeConstants::ProcessResult::Failed;
+    auto &[stepno, natoms, keytrj, imcon, _] = std::get<0>(*head);
```
trisyoungs, line 60: I mean, I think this is really nice!

trisyoungs: Just a thought for discussion (or shooting down entirely!) but it would be nice if the parse result was somehow just contained in the parser object so we could do something like this:

```suggestion
    if (!header().parse(infile))
        return NodeConstants::ProcessResult::Failed;
    auto &[stepno, natoms, keytrj, imcon, _] = std::get<0>(*header());
```
... if you see what I mean. Loses a line and a temporary. But I've already talked myself out if this silly idea, thinking of the construction of the `matrix()` parser which quite rightly re-uses the `vector3()` parser....

src/nodes/importDLPUtilsSurface.cpp
```diff
@@ -42,44 +43,34 @@ NodeConstants::ProcessResult ImportDLPUtilsSurfaceNode::process()
 // Read data specified
 bool ImportDLPUtilsSurfaceNode::read(Data2D &data, std::string filePath)
 {
+    using namespace Parsers;
```
trisyoungs, line 46: You know, I've never seen a `using namespace` anywhere other than at the top of a source file. The locality makes sense though!

src/nodes/importMoscitoStructure.cpp
```diff
@@ -2,6 +2,8 @@
 // Copyright (c) 2026 Team Dissolve and contributors
 
 #include "nodes/importMoscitoStructure.h"
+#include "base/parserLibrary.h"
+#include <fstream>
```
trisyoungs: We didn't need the `<fstream>` include in the other import nodes?

src/nodes/importMoscitoStructure.h
```diff
@@ -20,6 +22,15 @@ class ImportMoscitoStructureNode : public Node
     std::string_view type() const override;
     // Return short summary of the node's purpose
     std::string_view summary() const override;
+    // parse the header of a moscito file
+    static Parsers::Parser<std::tuple<Vector3, int>> header();
+    // parse an atom from a moscito file
+    static Parsers::Parser<std::tuple<std::string, int, Vector3, Vector3, Vector3>> atom();
+    // parse an molecule from a moscito file
```
trisyoungs:
```suggestion
    // Parse the header of a moscito file
    static Parsers::Parser<std::tuple<Vector3, int>> header();
    // Parse an atom from a moscito file
    static Parsers::Parser<std::tuple<std::string, int, Vector3, Vector3, Vector3>> atom();
    // Parse an molecule from a moscito file
```

trisyoungs: Organisationally, these `static` parser definitions belong in the `private` "Processing" domain below.

tests/ff/cos-n.cpp
```diff
@@ -61,7 +59,7 @@ TEST(CosNTorsionForcesTest, POE)
 
     // Check agreement with external reference forces
     checkReferenceForceConsistency(pairPotentialForces, geometryForces,
-                                   importNode->getOutputValue<std::vector<Vector3>>("Forces"), 6.0e-2);
+                                   importNode->getOutputValue<std::vector<Vector3>>("Forces"), 6.2e-2);
```
trisyoungs, line 62: I can live with this.  As you suggested, floating-point numbers are fickle beasts when converted from text.

tests/io/applicative.cpp
```diff
@@ -159,3 +162,75 @@ TEST(ApplicativeTest, Helium)
 }
 
 } // namespace UnitTest
+
+TEST(ApplicativeTest, DLPOLYStructure)
+{
+    using namespace Parsers;
```
trisyoungs: Here, wouldn't this be better at the top of the file rather than repeated in every `TEST`?

trisyoungs (approved):
On the one hand, this is an easy win because `LineParser` was truly awful, but I think this has turned out excellent and makes our input parsing feel very proper indeed!

Bunch of small comments, mostly formatting etc.

src/base/parserLibrary.cpp
rprospero: I've added a comment that explains the issues with Moscito files that I mentioned during the standup.

src/nodes/importDLPOLYTrajectory.cpp
rprospero: When I return next week, I'm planning on doing a refactoring on the parsers that gives an error message on a bad parse, instead of merely failing.  During this, I might look at seeing if there's a more ergonomic way of handling this.  I'll admit that, since seeing your post, I've already tried three different ways of cleaning up this code, but each of them still wound up at the same number of lines.  Still, it does feel like there's a simpler system trying to escape here.

If C++ had the `use` keyword from [Gleam](https://gleam.run/), there would be a wonderfully concise way of handling things.

```c++
   use auto &[stepno, natoms, keytrj, imcon, _] = header().parse(infile).and_then;
```

Maybe we'll get that around c++50

src/nodes/importMoscitoStructure.h
rprospero: I've moved the code into the processing domain, but I've kept the functions public so that the test functions can access the parsers.

PR #2545, opened 2026-06-30
feat: Node test fixtures

trisyoungs:
This implements a sort of fixture-based approach to more full-range testing of Graphs, particularly those involving analysis nodes calculating a specific result. The idea is to define the setup, run, and test functions individually, and then let the fixture setup, run, test. This automation then allows serialisation round-trip testing for free, as well as restart-data reading / testing when we are ready.

### DIFF
--- a/tests/testGraphFixture.h
+++ b/tests/testGraphFixture.h
@@ -48,6 +48,8 @@ class TestGraphFixture : public testing::Test
     virtual void prepareTestData() = 0;
     // Perform graph construction
     virtual void constructGraph() = 0;
+    // Run graph
+    virtual void runGraph() = 0;
     // Perform tests on generated data
     virtual void performTests() = 0;
     // Find specified node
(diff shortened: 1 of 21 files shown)


tests/testGraphFixture.h
```diff
@@ -0,0 +1,93 @@
...
+    // Serialised graph TOML
+    SerialisedValue graphTOML_;
+
+    protected:
+    // Test graph
+    TestGraph testGraph_;
+    // Deserialised graph
+    DissolveGraph deserialisedGraph_;
+    // Current graph target
+    OptionalReferenceWrapper<DissolveGraph> currentGraph_;
```
rprospero, line 26: Perhaps go ahead and make this DissolveGraph* ?  I know that eventually removing OptionalReferenceWrapper is on the RoadMap.

rprospero (approved):

PR #2553, opened 2026-07-08
refactor: End of LineParser

rprospero:
The PR removes the last calls to LineParser.  Most of the files are just removing a vestigal header, but I did replace a few parsers and serialisers.


trisyoungs (approved):
Just looking at the EPSR Ato code gives me a flash of anxiety!  Good work.

PR #2554, opened 2026-07-09
feat: More Monte Carlo

trisyoungs:
This PR brings in `MCNode` (== MolShake) and `IntraMCNode` (== IntraShake).

There is an opportunity to improve their functionality by implementing per-Species step sizes (a feature long missing from Dissolve) but unfortunately this requires a bit of thought on serialisation of the required `ResolvableKeyedVector` as a piece of restartable data.  So I leave that for a follow-up.


src/nodes/mcCommon.h
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#pragma once
+
+namespace MCCommon
```
RobBuchananCompPhys: Oh, a non-class in with the nodes! Should we look at grouping all these mc-related node/namespace files into a single subdir (as we do with `cif` stuff)?

RobBuchananCompPhys (approved):

src/nodes/mcCommon.h
trisyoungs: I think I'd be inclined not to and stick to the single h/cpp file layout - each of the MC nodes is doing a specific job, after all.  This little namespace could be moved out to the `math/` dir as a very general-purpose thing I suppose?

RobBuchananCompPhys: Yes that sounds good.

PR #2555, opened 2026-07-09
fix: Some <cmath> includes

trisyoungs:
Add some `<cmath>` includes.


rprospero (approved):
It's always weird to me how system dependent these include issues can be.  The C++ modules system could have fixed all of this, if it hadn't been such a disaster in its own right.

PR #2556, opened 2026-07-09
refactor: Detect molecules (Version 2)

RobBuchananCompPhys:


src/classes/structure.cpp
```diff
@@ -119,6 +124,10 @@ const StructureAtom *Structure::atom(int i) const { return atoms_[i].get(); }
 const std::vector<std::unique_ptr<StructureAtom>> &Structure::atoms() const { return atoms_; }
 std::vector<std::unique_ptr<StructureAtom>> &Structure::atoms() { return atoms_; }
 
+// Return molecular species coordinates
```
```suggestion
// Return positional instances
```

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,297 @@
...
+    {
+        if (!instances.empty())
+        {
+            detectedStructures_.emplace_back(copyStructureAtomsAndBonds(inputStructure_, detectedMolecularStructure, fragment))
+                .instances() = instances;
+
+            return true;
+        }
+        return false;
+    };
```
trisyoungs: I'm not convinced this needs to be a separate lambda: the check on the instances being empty isn't really useful (I'm not sure when that would ever occur?) and the return value isn't ever checked.

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,297 @@
...
+            for (const auto &fragmentAtom : fragmentAtoms)
+            {
+                if (atomMask.contains(fragmentAtom))
+                    continue;
+
+                /*
+                 * Best NETA definition
+                 */
+
+                // Set up the return value and bind its contents
```
trisyoungs:
```suggestion
                // Find the best NETA definition for this fragment
```

src/nodes/parameter.h
```diff
@@ -52,6 +52,7 @@ class ParameterBase : public Serialisable
         ClearData, /* Indicates that any local data should be cleared if the parameter is changed */
         Input,     /* Indicates that the parameter is meant to be a sink for data and not a source */
         Output,    /* Indicates that the parameter is meant to be a source of data and not a sink */
+        Dynamic,   /* Indicates that the parameter is meant to be a dynamic source of data and not a sink */
```
trisyoungs: Not used any more?
```suggestion
```

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    return getFragmentAtoms(structure, indices);
+}
+
+// Find all molecular fragments
+std::map<int, DetectMoleculesNode::FragmentVector> DetectMoleculesNode::findMolecularFragments(const Structure &structure)
+{
+    std::map<int, FragmentVector> map;
+
+    auto fragment = [structure](int i) { return Fragment<StructureAtom, Bond<StructureAtom>>::get(structure.atoms(), i); };
```
trisyoungs: This lambda is only used once, so it may as well be inline.

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    for (int i = 0; i < structure.nAtoms(); i++)
+    {
+        auto element = fragment(i);
+        const int size = element.size();
+        if (!map.contains(size))
+            map.emplace(size, FragmentVector{});
+        auto &targetFragments = map[size];
+        targetFragments.push_back(element);
+    }
```
trisyoungs, line 81: I don't believe this loop has the desired effect for anything other than pure atomic systems. You'll need to maintain a map of which indices in the structure have already been detected as part of a fragment, and skip over those in the loop. Otherwise, for e.g. liquid water you're going to get three times the number of detections.

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,297 @@
...
+        {
+            detectedStructures_.emplace_back(copyStructureAtomsAndBonds(inputStructure_, detectedMolecularStructure, fragment))
+                .instances() = instances;
+
+            return true;
+        }
+        return false;
+    };
+
+    // Try selecting within the fragment from the first atom - if this captures all atoms we have a bound framework...
```
trisyoungs:
```suggestion
    // Check for a single, bound framework fragment
```

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,297 @@
...
+                break;
+            }
+
+            for (const auto &fragmentAtom : fragmentAtoms)
+            {
+                if (atomMask.contains(fragmentAtom))
+                    continue;
+
+                /*
+                 * Best NETA definition
```
trisyoungs: The whole loop over atoms to determine the best NETA definition needs to be separate from the application of that definition to the fragments.  As written, it finds a NETA definition and always applies it to the atoms in the fragment, regardless of whether it is the best or not (in fact `bestNETA` is set but never used).

src/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,297 @@
...
+                }
+
+                // Loop over matched unit cell atoms, retrieving instances
+                for (const auto &matchedUnitCellAtoms : matchedUnitCellAtomSets)
+                {
+                    if (matchedUnitCellAtoms.empty())
+                        continue;
+
+                    addInstance(instances.emplace_back(), matchedUnitCellAtoms);
+                }
```
trisyoungs: Wouldn't this be more clearly-written as a loop over individual groups (`std::set`s) of instance atoms, applying the NETA to see if it is a match, rather than grabbing a huge vector of atoms all together and having to do the manual tracking?  You could go one step further, and use the erase/remove idiom on the vector of instances, with the lambda performing the check responsible for matching and storing the instance if required.

src/nodes/detectMolecules.h
```diff
@@ -0,0 +1,65 @@
...
+                                                 const std::vector<int> fragmentIndices);
+    // Add fragment molecular instance
+    static void addInstance(std::vector<Vector3> &targetInstance, const AtomCollection &instanceFragmentAtoms);
+    // Get fragment atoms from either a single set of fragment indices, or in its overloaded form, a vector of fragments
+    static std::vector<const StructureAtom *> getFragmentAtoms(const Structure &structure,
+                                                               const std::vector<int> &fragmentIndices);
+    static std::vector<const StructureAtom *> getFragmentAtoms(const Structure &structure,
+                                                               const FragmentVector &fragmentIndices);
+    // Find all molecular fragments
+    static std::map<int, FragmentVector> findMolecularFragments(const Structure &structure);
```
trisyoungs: Since these are all `private`, making them `static` doesn't really make much sense I don't think. For example, `getFragmentAtoms` only ever takes the `inputStructure_` as the structure argument, so why not just make it `const` and operate implicitly on `inputStructure_`?

src/nodes/insert.cpp
```diff
@@ -8,20 +8,25 @@
...
 InsertNode::InsertNode(Graph *parentGraph) : Node(parentGraph)
 {
     // Inputs
     addInput("Configuration", "Target configuration to insert into", configuration_);
     addOutput("Configuration", "Modified configuration", configuration_);
-    addInput("Species", "Species to add - all resulting molecules will have identical geometry", species_);
-    addInput("MoleculeSet", "MoleculeSet to use as the source", moleculeSet_);
     addInput("Population", "Population of the target to add", population_);
     addInput("Density", "Density at which to add the target", density_);
+    addInput("Species", "Source species or molecule set to add - all resulting molecules will have identical geometry",
```
trisyoungs:
```suggestion
    addInput("Species", "Source species or molecule set to add",
```

src/nodes/insert.cpp
```diff
@@ -183,6 +235,33 @@ NodeConstants::ProcessResult InsertNode::process()
...
+            for (int i = 0; i < mol->nAtoms(); i++)
+                mol->atom(i)->setR(atomicCoords[i]);
+
+            if (insertionComplete)
+                continue;
+        }
+
         // Randomise position of Molecule over the whole box
         auto newCentre = box.getReal({DissolveMath::random(), DissolveMath::random(), DissolveMath::random()});
         mol->setCentreOfGeometry(box, newCentre);
```
trisyoungs, line 247: I think this would be a lot clearer and easier to follow if it could be rewritten as a `switch` or similar.

src/nodes/insert.cpp
```diff
@@ -175,6 +198,35 @@ NodeConstants::ProcessResult InsertNode::process()
...
+            if (alreadySampled.contains(sampledIndex))
+                continue;
+
+            index += sampledIndex;
+            alreadySampled.insert(index);
+            break;
+        }
+
+        return index;
+    };
```
trisyoungs: If we're sampling from a set of instances there's no need to ensure that each instance is  used before bailing out. In fact, this goes fully against the random nature of the process, plus will force the routine to bail out if more molecules are requested than exist in the instances available.

src/nodes/insert.cpp
```diff
@@ -183,6 +235,33 @@ NodeConstants::ProcessResult InsertNode::process()
...
+            }
+            else
+            {
+                auto instancesIndex = sampleInstances();
+                if (instancesIndex < 0)
+                    break;
+            }
+
+            // Update molecular atomic coordinates
+            for (int i = 0; i < mol->nAtoms(); i++)
```
trisyoungs:
```suggestion
            for (auto i = 0; i < mol->nAtoms(); ++i)
```

tests/nodes/parameters.cpp
```diff
@@ -359,4 +359,70 @@ TEST_F(ParametersTest, OptionalPointerToVariant)
     EXPECT_EQ(std::get<Configuration *>(b_->variant().data), &a_->optionalConfiguration().value());
 }
 
+TEST_F(ParametersTest, DynamicOutput)
+{
+    ASSERT_TRUE(testGraph_.addNode(std::make_unique<TestNode>(&testGraph_), "Sender"));
+    ASSERT_TRUE(testGraph_.addNode(std::make_unique<TestNode>(&testGraph_), "RecieverA"));
```
trisyoungs: `Receiver` - here and elsewhere.

trisyoungs (changes requested):

src/nodes/detectMolecules.cpp
RobBuchananCompPhys: Yep, quite right. I'll put in a check on a `std::set<int>` to skip those fragments

trisyoungs (approved):

PR #2559, opened 2026-07-22
refactor: Graph & IterableGraph TOML roundtrip

RobBuchananCompPhys:
There's a couple of things going on in this PR, all with the purpose of fixing deserialisation of `IterableGraph` from TOML:

- The problem was in the first instance that deserialisation builds out the graph in the 'opposite' direction to how we typically construct them manually*, such that when it comes to parsing a 'sub'-graph (such as any graph containing an instance of `IterableGraph`), the edges that depend on external nodes connected to the sub-`InputsNode` basically fail to materialise, since they have no awareness of any such data source. Upon closer analysis it was found that this was a problem occuring for deserialisation of _any_ `Graph` node, iterable or otherwise. This issue has been fixed in this PR, so a TOML roundtrip is possible for the `SubGraph` test. This exact problem would have been the root cause of the `IterableGraph` loop edges not being parsed, however...
- ...it turns out that no specialised de/serialisation code had ever been written for `IterableGraph` and `LoopEdge` classes, so the `loopEdges_` were never parsed anyway. This has also been fixed, and there are new tests for this.

*it should be noted that after implementing these changes, it should in theory be possible for a user to manually construct a nested graph from the inside-out without having yet emplaced the relevant external nodes! It may however require some further GUI work where, for instance, hovering an edge over a graph's `InputsNode` triggers the edge creation, whether or not any (proxy) inputs exist or not.

### DIFF
--- a/src/nodes/edge.cpp
+++ b/src/nodes/edge.cpp
@@ -318,6 +318,7 @@ void Edge::deserialise(const SerialisedValue &node)
 void LoopEdge::serialise(std::string tag, SerialisedValue &target) const
 {
     definition().serialise(tag, target);
+    target[tag]["targetNode"] = "LoopBacks";
     target[tag]["analogue"] = analogue_;
 }
 
--- a/src/nodes/edge.cpp
+++ b/src/nodes/edge.cpp
@@ -312,3 +312,13 @@ void Edge::deserialise(const SerialisedValue &node)
     throw std::runtime_error("Cannot directly deserialise edges.  Please contact the Dissolve development team if you are "
                              "seeing this error - this is a bug and NOT your fault.\n");
 }
+
+// Express as a serialisable value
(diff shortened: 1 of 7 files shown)


rprospero (approved):
This is going to conflict with my #2560, but I can do a rebase.

src/nodes/edge.cpp
```diff
@@ -31,18 +32,37 @@ class EdgeConstructor : public Edge
 // Create an edge from the supplied definition
 std::unique_ptr<Edge> Edge::create(Graph *parent, const EdgeDefinition &definition)
 {
+    // Get target node
+    auto targetNode = parent->findNode(definition.targetNode);
```
trisyoungs, line 36: If the check for the `targetNode` needs to move here then the check for finding a valid `Node` should move here along with it.

trisyoungs (approved):
One heavily suggested change, otherwise looks OK.

PR #2560, opened 2026-07-24
refactor:Serialisible concept

rprospero:
This PR replaces the old Serialisable class with class with a pair of concepts.  This provides the following advantages

1) The concepts can be in different namespaces.  Instead of the old, confusing `fromVector` and `toVector` methods, we now have `Serialisable::vector` and `Deserialisable::vector` methods
2) The handling of classes and raw types are now unified.  Previously, we had many helper functions calling `foo.serialise`, which would fail if `foo` was a string or a double.  The new `serialiseOnto` method works equally well with classes and raw types
3) There's no need to inherit from a Serialisable class to produce toml.  This also creates smaller classes, as none of the virtual member function pointers are needed
4) The serialisation helper functions are now in a separate header (serialiserLibrary.h), which is often only needed in the .cpp file and not the .h file.

The fundamental theory of this library is that the overloaded `serialiseOnto` function serialises things and `deserialiseOnto` deserialises things.  Making a type serialisable merely involves adding another overload to these two functions.  For simplicity, I've added a templated overload that works on any type that had the old `serialise` and `deserialise` methods that we used before, so most of the code is largely unchanged.

I have the following changes I would like to perform before I pull this PR out of draft

- [x] Remove all direct references to toml outside of the serialiser files.  This will make it easier to replace the TOML11 library in the future
- [x] Remove import of serialiserLibrary.h from an header files.  This should prevent a larger library from many downstream files
    - [X] src/classes/atom.h
    - [x] src/classes/speciesIntra.h

As a guide for preliminary reviews, the best files to look at will be:
- `src/base/serialiser.h`  is the new minimal header and establishes the new namespaces and the `serialiseOnto` and `deserialiseOnto` method signatures
- `src/base/serialiserLibrary.h` contains both the actual serialisation concepts and the old helper functions
- `src/classes/speciesSite.cpp` gives a good example on how the new library works (mostly the same with cleaner names).


src/classes/speciesSites.cpp
```diff
@@ -67,27 +67,29 @@ std::vector<const SpeciesSite *> SpeciesSites::getSpeciesSites() const
 // Express as a serialisable value
 void SpeciesSites::serialise(std::string tag, SerialisedValue &target) const
 {
+    using namespace Serialisable;
```
trisyoungs: I'm with Google on forbidding the use of `using namespace` to make members visible. Especially so when the functions we're exposing have very "standard" names like `map`, `vector` etc., IMO.

src/classes/speciesSites.cpp
```diff
@@ -67,27 +67,29 @@ std::vector<const SpeciesSite *> SpeciesSites::getSpeciesSites() const
...
-              toMap(sites, [&](const std::string &siteName, const SerialisedValue &population)
-                    { set[siteName] = population.as_floating(); });
-          });
+    map(node, "set",
+        [&](const std::string &speciesName, const SerialisedValue &sites)
+        {
+            auto &set = sites_[speciesName];
+            map(sites, [&](const std::string &siteName, const SerialisedValue &population)
+                { set[siteName] = population.as_floating(); });
+        });
```
trisyoungs: I still have a problem with clarity here. I think this refactoring illustrates it pretty well - I'm supposedly serialising a `vector`, but deserialising a `map`. Admittedly this might be caused from the underlying type of `sites_` being `ResolvableKeyedVector`, but you see my point!

trisyoungs (commented):
Direction is good.  Made a couple of comments.

src/classes/speciesSites.cpp
rprospero: I've switched all the call sites to using explicitly namespaced calls.

src/base/serialiserLibrary.h
```diff
@@ -0,0 +1,275 @@
...
+#include <iterator>
+#include <map>
+
+namespace Serialisable
+{
+
+// We need a way at compile time to detect all the types of smart
+// pointers for things that can be serialised
+template <typename T>
+concept SerialisablePointer = requires(T a, std::string tag, SerialisedValue target) { a->serialise(tag, target); };
```
trisyoungs, line 18: I actually wonder if this is even required now?  In Dissolve1 a module had to write out, e.g., a `Species*` target which it did by just serialising the name.  In Dissolve2, I don't think there are any remaining examples of us needing to serialise a pointer value - all of our data is either a concrete member or passed via an `Edge` (bypassing the need for direct serialisation in the Dissolve1 style).  The only possible example might be in the `ForcefieldNode`, but this could be factored out (if I haven't done so already).

src/base/serialiserLibrary.h
```diff
@@ -0,0 +1,275 @@
...
+
+template <typename T>
+concept SerialisableClass = requires(T a, std::string tag, SerialisedValue &target) { a.serialise(tag, target); };
+
+// template <typename T>
+// concept SerialisableFromInto = requires(T a,
```
trisyoungs: Old WIP code?

src/base/serialiserLibrary.h
```diff
@@ -0,0 +1,275 @@
...
+    return temp["inner"];
+}
+// A helper function to add the elements of a map to a node under a name
+template <typename K, typename V> void map(const std::map<K, V> &map, std::string name, SerialisedValue &node)
+{
+    SerialisedValue result;
+    for (auto &[key, value] : map)
+        serialiseOnto(value, std::format("{}", key), result);
+    if (!map.empty())
+        node[name] = result;
```
trisyoungs: For consistency with other functions (plus I think it makes more sense this way):

```suggestion
    if (map.empty())
        return;

    SerialisedValue result;
    for (auto &[key, value] : map)
        serialiseOnto(value, std::format("{}", key), result);
    node[name] = result;
```

src/base/serialiserLibrary.h
```diff
@@ -0,0 +1,275 @@
...
+
+} // namespace Serialisable
+
+namespace Deserialisable
+{
+template <typename T>
+concept DeserialisibleClass = requires(T a, SerialisedValue &node) { a.deserialise(node); };
+
+// template <typename T>
+// concept DeserialisableFromInto = requires(T a,
```
trisyoungs: Another one. I realise these might be "helper comments" to illustrate the function signature, but if that's the case I'd prefer to have it written as that more explicitly (so I understand it in the future!).

src/classes/atom.cpp
```diff
@@ -135,3 +136,17 @@ AtomBase::AtomGeometry AtomBase::geometry() const
...
+{
+    target[tag] = {{"index", index_}, {"z", Serialisable::ser(Z_)}, {"r", Serialisable::ser(r_)}, {"q", q_}};
+}
+// Read values from a serialisable value
+void AtomBase::deserialise(const SerialisedValue &node)
+{
+    index_ = Deserialisable::de<int>(node.at("index"));
+
+    set(Deserialisable::de<Elements::Element>(node.at("z")), Deserialisable::de<Vector3>(node.at("r")),
+        Deserialisable::de_or<double>(node, "q", 0));
```
trisyoungs: OK, so dumb question - I don't see `de` or `de_or` obviously declared anywhere? Presumably the `de` is for `deserialise`, but I'm not sure I like the brevity in this case as it harms readability.

trisyoungs (approved):
Looking really good I think.  Couple of moderate comments to consider.

src/base/serialiserLibrary.h
rprospero: It is still required, but usually indirectly, as in that we're usually serialising concrete objects, but some of those objects then contain pointers. As a quick example, the `atomTypes_` vector in `Species` is a vector of `shared_ptr`.

src/base/serialiserLibrary.h
rprospero: Yep.  That was part of an attempt to apply the serialisable concept to vectors of serialisable, but you can't recurse concepts like that.

src/base/serialiserLibrary.h
rprospero: Same failed experiment mentioned above.

src/base/serialiserLibrary.h
rprospero: Agreed.

src/base/serialiserLibrary.h
trisyoungs: Yep, fair comment.  Didn't consider the atom type pointers.

src/classes/atom.cpp
rprospero: `de` is defined on line 190 of `serialiserLibrary.h`.  I'd originally went with the shorter name hoping that the brevity might *improve* legibility, becoming essentially invisible against the structure of the data.  However, that only worked while we were importing the namespaces.  Without that, the benefits of brevity are lost.  I'm going to move to `deser` and `deser_or`, as a parallel to `ser`.

PR #2561, opened 2026-07-27
refactor: DetectMolecules TY

trisyoungs:


src/nodes/detectMolecules.cpp
```diff
@@ -276,3 +148,114 @@ NETADefinition DetectMoleculesNode::bestNETADefintion(const std::vector<int> &fr
...
+
+            detectedStructures_.emplace_back(structure);
+
+            continue;
+        }
+
+        // Loop over fragments of this size
+        while (!fragments.empty())
+        {
+            printf("FRAGs now %li\n", fragments.size());
```
RobBuchananCompPhys: Using
```
const auto nFragments = fragments.size()
```
after line 174, we could make this abit nicer?
```suggestion
            printf("Found %li fragments\n", fragments.size());
```

RobBuchananCompPhys: Of course, `nFragments - fragments.size()` is the running number of fragments found within _this_ current size class.

RobBuchananCompPhys: So, perhaps a better log would be
```
message("Found {} fragments of size {} \n", nFragments - fragments.size(), _);
```

src/nodes/detectMolecules.cpp
```diff
@@ -28,199 +28,71 @@ std::string_view DetectMoleculesNode::summary() const { return "Detect molecular
...
-
-    return NodeConstants::ProcessResult::Success;
-}
-
-/*
- * Helpers
- */
-
-// Copy atom and bond information from one structure to another
-Structure &DetectMoleculesNode::copyStructureAtomsAndBonds(Structure &target, const std::vector<int> fragmentAtomIndices) const
```
RobBuchananCompPhys, line 136: This is a small thing, but did `copy...(` not approach what is going on here as well as `duplicate...(` ?
Personally I prefer "copy", just feels more idiomatic.

src/nodes/detectMolecules.cpp
trisyoungs: Ah, it's just a debug line I left in - I'll remove it now.

src/nodes/detectMolecules.cpp
trisyoungs: I think I called it `duplicate...` while I was reworking stuff to keep it separate, but either way happy to revert to `copy`.

PR #2566, opened 2026-07-30
refactor: QML gui

RobBuchananCompPhys:
**Please read!**

This has been an exercise in pushing the Dissolve2 QML GUI towards something we can use in a basic way, up to and including running graphs. Along the way I've also addressed some "beautification" issues, and just adding some essential utilities here and there. Build it and try it out!

#--PART-1----------------------------------------------------------------------#
- [x] Clean up and refactor GUI directory structure

(For this bit I basically threw away as much Dissolve1 `gui` code as possible, mainly model srcs/headers and other Qt6-related files. I wanted to get things down to a minimally "compilable" environment for our new QML gui. This inevitably led to some hanger-on models which I couldn't compile without, and couldn't be 100% sure we'd want to remove, but probably will at some point. They currently reside in a folder marked `models/legacy`. I know this name sounds a bit silly, but its just to keep them clearly grouped together so we know where they are. Also, you might notice the renaming of the `gui` dir to `gui2` and `models2` etc... This was just to aid in transferring things between folders and keeping CMake targets separate, and can be changed back to `gui` and `models` if we decide to stick with this structure rather than resurrect the old one! Also, I've moved to designated QML modules, using CMake, for modularising the QML code. This is to give a suggestion as to how a module system could look, and probably doesn't strictly reflect the optimal way of dividing up the code.)

#--PART-2----------------------------------------------------------------------#
- [x] Fixup nodeGraph UI components

(I then delved into the QML code for the node-graph and started to try and fit things together in such a way that they can be interacted with in a clean and clear way. After a quick clean up/slight enhancement of the main menu and tabs, I introduced a node search dialog ("Find Node"), accessible from the "Graph" menu. This uses our C++ node-name-by-wildcard to allow you to start typing and select a relevant node from the list, which goes directly in the graph. Also introduced a global `NodeRegistryModel` to serve up info about all available nodes. I then refactored the QML for `NodeBox` and `GraphDelegate` to space out components, and make them draggable etc.)

#--PART-3----------------------------------------------------------------------#
- [x] Fix UI edges

(The code for controlling the rendering of edges was frankly quite confusing, and didn't seem to connect them to the exact locations of their endpoints. I basically started from scratch here, with a simple linear edge `ShapePath`, and created a new C++ model (`ParameterEndPointsModel`) which belongs to every node in the UI, and records all the available inputs and ouputs and their corresponding QML items - the place where you actually drag/drop the edge. This model then stores direct references to pairs of QML components corresponding to edge-linked parameters between nodes, and updates the linear path whenever the position of those areas changes, and if either source/target node is deleted, they are automatically destroyed.)

#--PART-4----------------------------------------------------------------------#
- [x] Descriptive nodes, message logging

(Nodes have been made richer/ more "descriptive":
- a naming convention (i.e. "<NodeType>_0", "<NodeType>_1", ... ) for newly created nodes. This allows the user to see immediately where the newest node yet to be named resides in the graph, as well as its type. A dynamic tally in the `NodeRegistryModel` of how many of each type of node is already in existence
- Nodes can display logs of the following levels: `Info`, `Warning`, and `Error`, accessed from a popup menu inside the node. The presence of errors are indicated in the corner of the node. Graph process results are displayed in the message log of the node which was run.
- Standard tooltips associated with inputs, outputs, and options were obscuring the node itself. There is now a banner at the bottom of the node that shows the description when hovered.

#--CHEAT-SHEET----#
To get things working
1. Go to "Graph" menu -> "Find Node", and add the new node
2. Drag it around by pressing and holding the space around the node's header (this is actually a bit tricky as the "grab" area is not very extensive at all, so something to fix in due course)
3. Connect edges by dragging/dropping from parameters
4. Run the graph either from the "Graph" menu, and typing the name of the node to run from in the popup, or simply right click a node and click "Run".

#---TODO----#

- [x] I'd like to see node categories managed by Dissolve itself, for instance an extra field in the `NodeRegistry`, rather than something hardcoded into the UI, which is the current state.
- [x] Spline edges. I reverted to lines to try and get something going quickly, but I don't see how the new state couldn't be updated to support curving edges.
- [ ] **Important caveat**: Subgraph/`IterableGraph` _proxy_ I/O doesn't show up yet. Ideally, you should be able to make arbitrary connections to the proxy inputs, for instance. This will probably be a blocker on immediately using this code to run graphs will looping behaviour, unfortunately!


src/gui2/models/legacy/types.cpp
```diff
@@ -0,0 +1,30 @@
...
+    //qmlRegisterType<SpeciesModel>(PROJECT, 1, 0, "SpeciesModel");
+    //qmlRegisterType<ConfigurationModel>(PROJECT, 1, 0, "ConfigurationModel");
+    //qmlRegisterType<AtomTypeModel>(PROJECT, 1, 0, "AtomTypeModel");
+    //qmlRegisterType<CommonAngleModel>(PROJECT, 1, 0, "CommonAngleModel");
+    //qmlRegisterType<CommonBondModel>(PROJECT, 1, 0, "CommonBondModel");
+    //qmlRegisterType<CommonImproperModel>(PROJECT, 1, 0, "CommonImproperModel");
+    //qmlRegisterType<CommonTorsionModel>(PROJECT, 1, 0, "CommonTorsionModel");
+    //qmlRegisterType<ModifyChargesModel>(PROJECT, 1, 0, "ModifyChargesModel");
+    //qmlRegisterType<SortFilterProxy>(PROJECT, 1, 0, "SortFilterProxy");
+    qmlRegisterType<GraphModel>(PROJECT, 1, 0, "GraphModel");
```
rprospero: If we can truly keep things down to only registering a single type, I will be obscenely happy.

src/gui2/models/nodeRegistrySortFilterModel.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+QString &NodeRegistrySortFilterModel::searchText() { return searchText_; }
+
+void NodeRegistrySortFilterModel::setSearchText(QString &text)
+{
+    // Expect 62 rows and 1 column in source model
+    searchText_ = text;
+    invalidateFilter();
+    Q_EMIT searchTextChanged();
+}
```
rprospero: With the newer Qt releases, there's now a proper QML [SortFilterProxyModel](https://doc-snapshots.qt.io/qt6-6.10/qml-qtqml-models-sortfilterproxymodel.html) which may remove the need for this class completely.

PR #2571, opened 2026-08-04
fix: Scattering matrix data was not correctly updated

trisyoungs:
This PR fixes a _horrible_ bug introduced by me at the end of last year whereby the data in the scattering matrix was never updated correctly. It was set correctly on the first run of the EPSR module, but after that everything but the first dataset remained static.  Unfortunate bug, but good refinements still have been achievable by several people depending on their individual circumstances.


rprospero (approved):

RobBuchananCompPhys (approved):
Good fix 👍

PR #2578, opened 2026-08-18
fix: Return cubic shape to edges

rprospero:
This should make the edge connections cubic curves again.  Unfortunately, since I currently don't have a good way of adding nodes and connections, I haven't been able to test this code, so it might be worth trying out before merging.


RobBuchananCompPhys (approved):
Excited to see this!

PR #2581, opened 2026-08-20
refactor: Use comparison operator in comparison overloads

rprospero:
C++20 added the comparison operator (sometimes referred to as the spaceship operator), from which all of the other comparison operators can be derived.  I've used this to cut down on a bunch of the boiler plate in our iterator overloads and for our `Number` type.

There's one change in the handling of NaN for the Number class. For any floating point value *x*, the following statements are *always* true under the new setup, but were *always* false under the old setup:

```
a < NaN
Nan < a
```

This was done to allow strong ordering of Number, despite floating point values only having weak ordering.  This convention was chosen to preserve the the following three relations, which I considered more fundamental:

```
a != NaN
NaN != a
NaN != Nan
```

### DIFF
--- a/src/classes/fullPairIterator.h
+++ b/src/classes/fullPairIterator.h
@@ -49,10 +49,11 @@ class FullPairIterator
     FullPairIterator operator-(difference_type backward) const;
 
     // Operators : comparison
-    bool operator==(const FullPairIterator &other) const;
-    bool operator!=(const FullPairIterator &other) const;
-    bool operator<(const FullPairIterator &other) const;
-    bool operator>(const FullPairIterator &other) const;
-    bool operator<=(const FullPairIterator &other) const;
-    bool operator>=(const FullPairIterator &other) const;
+    bool operator==(const FullPairIterator &other) const = default;
+    bool operator!=(const FullPairIterator &other) const = default;
+    bool operator<(const FullPairIterator &other) const = default;
+    bool operator>(const FullPairIterator &other) const = default;
+    bool operator<=(const FullPairIterator &other) const = default;
+    bool operator>=(const FullPairIterator &other) const = default;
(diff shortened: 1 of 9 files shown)


RobBuchananCompPhys (approved):
This looks interesting, I hadn't come across this yet so it's nice to learn about a new feature.

src/classes/fullPairIterator.h
```diff
@@ -49,10 +49,11 @@ class FullPairIterator
...
-    bool operator<(const FullPairIterator &other) const;
-    bool operator>(const FullPairIterator &other) const;
-    bool operator<=(const FullPairIterator &other) const;
-    bool operator>=(const FullPairIterator &other) const;
+    bool operator==(const FullPairIterator &other) const = default;
+    bool operator!=(const FullPairIterator &other) const = default;
+    bool operator<(const FullPairIterator &other) const = default;
+    bool operator>(const FullPairIterator &other) const = default;
+    bool operator<=(const FullPairIterator &other) const = default;
+    bool operator>=(const FullPairIterator &other) const = default;
```
trisyoungs: Doesn't the use of `<=>` mean that these `default` declarations are now unnecessary (also in `Array3DIterator`, `PairIterator` and `Number`) or have I misunderstood?

trisyoungs (approved):

src/classes/fullPairIterator.h
rprospero: They *should* be unnecessary.  However, GCC was complaining that == was not supported, so I manually asked for the default values for each operator.

After you comments, I dug into it a bit deeper.  If I default the == and < operators, then GCC does still correctly generate the rest.  Thus, I've removed the other specific overloads, ultimately removing more lines.

PR #2582, opened 2026-08-20
refactor: Categorical `NodeRegistry` and updated QML Graph menu

RobBuchananCompPhys:
This PR puts the responsibility on the C++ `NodeRegistry` to store Node categories (which had previously been hardcoded into a QML menu) by an enum type. This `categoricalProducers_` map is then flattened for more general use throughout the application, for retrieving any node. The categories are used to generate the context menu that appears when right-clicking in the Graph, and nodes can be created in that way.

### DIFF
--- a/src/gui2/models/enumOptionsModel.cpp
+++ b/src/gui2/models/enumOptionsModel.cpp
@@ -16,14 +16,6 @@ void EnumOptionsModel::setData(std::shared_ptr<const EnumOptionsBase> options)
 /*
  * QAbstractItemModel overrides
  */
-
-QHash<int, QByteArray> EnumOptionsModel::roleNames() const
-{
-    QHash<int, QByteArray> roles;
-    roles[Qt::DisplayRole] = "option";
-    return roles;
-}
-
 int EnumOptionsModel::rowCount(const QModelIndex &parent) const
 {
     Q_UNUSED(parent);
--- a/src/gui2/models/enumOptionsModel.cpp
+++ b/src/gui2/models/enumOptionsModel.cpp
@@ -17,6 +17,13 @@ void EnumOptionsModel::setData(std::shared_ptr<const EnumOptionsBase> options)
  * QAbstractItemModel overrides
  */
 
+QHash<int, QByteArray> EnumOptionsModel::roleNames() const
+{
+    QHash<int, QByteArray> roles;
+    roles[Qt::DisplayRole] = "option";
+    return roles;
+}
+
 int EnumOptionsModel::rowCount(const QModelIndex &parent) const
 {
     Q_UNUSED(parent);
--- a/src/gui2/qml/nodeGraph/GraphView.qml
+++ b/src/gui2/qml/nodeGraph/GraphView.qml
@@ -28,41 +28,42 @@ Pane {
         Menu {
             id: contextMenu
 
-            Instantiator {
+            Repeater {
                 model: nodeRegistry.categories()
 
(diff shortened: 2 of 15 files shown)


src/gui2/models/enumOptionsModel.cpp
```diff
@@ -17,6 +17,13 @@ void EnumOptionsModel::setData(std::shared_ptr<const EnumOptionsBase> options)
  * QAbstractItemModel overrides
  */
 
+QHash<int, QByteArray> EnumOptionsModel::roleNames() const
+{
+    QHash<int, QByteArray> roles;
+    roles[Qt::DisplayRole] = "option";
+    return roles;
+}
+
```
rprospero: Is there a reason to rename the display role as "option" instead of just using "display" in the QML?

src/gui2/qml/nodeGraph/GraphView.qml
```diff
@@ -24,133 +23,51 @@ Pane {
...
-
-                    onClicked: forcefieldDialog.open()
-
-                    ForcefieldDialog {
-                        id: forcefieldDialog
-
-                        graphModel: graphRoot.rootGraphModel
-                        posx: Math.round(ctxMenuCatcher.mouseX)
-                        posy: Math.round(ctxMenuCatcher.mouseY)
+            Instantiator {
```
rprospero: I'm a bit more inclined to use a `Repeater` here.  The Instantiator is useful when you're adding and removing nodes at runtime.  Since the number of nodes and categories are fixed, we can just use a repeater.  As an added bonus, we can drop the `onObjectAdded` and `onObjectRemoved`

src/nodes/registry.cpp
```diff
@@ -67,84 +67,119 @@
...
-std::map<std::string_view, NodeProducer> NodeRegistry::producers_;
-
 // Makes unique pointer to derived node instance
 template <typename T> NodeProducer makeDerivedNode()
 {
     return [=](Graph *parent) -> std::unique_ptr<Node> { return std::make_unique<T>(parent); };
 }
 
+// Static Singletons
+const std::map<NodeRegistry::Category, ProducerMap> NodeRegistry::categoricalProducers_{
```
rprospero, line 77: I'll admit that I had considered a slightly different architecture.  Specifically, I had though that we would add a virtual `category` method to the `Node` class which returns a `Node::Category` (since this would require that the `Category` enum be moved out of `NodeRegistry` into `Node`).  This map would then just be a vector and the NodeRegistryModel would filter the list by the category of each node type to ensure that everything was listed in the proper categories.

Honestly, though, I'm not sure that my architecture is any better, but rather just different.  The main advantage *would be* that it would be impossible to forget to assign a category to a new Node class, since the unimplemented virtual method would be a compiler error.  However, it would still be possible to forget to add the node to the registry vector, which gets us the very same problem.  The only real advantage is that Nodes would directly know their own category, but I'm not sure that will every be useful.

rprospero (approved):

src/gui2/models/enumOptionsModel.cpp
RobBuchananCompPhys: I didn't know you could do that, but it sounds sensible and would make this cleaner (no need for a custom `roleNames` override)

src/gui2/qml/nodeGraph/GraphView.qml
RobBuchananCompPhys: Good point👍

PR #2596, opened 2026-08-24
feat: Cbor restart file

rprospero:
This PR puts in the framework for saving and loading restart files/directories.  Nodes can overload virtual functions to save information to a restart file (which is named after the Node).

Currently, only the state of the configuration is saved, but other nodes will likely need restart information as well.  Also, there's a larger question about *when* to save and/or load restart files.  I suspect that it will need to go into IteratorGraph, but there are still some questions about how loading will work.


tests/nodes/restart.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+    auto configNode = dynamic_cast<ConfigurationNode *>(testGraph.findNode("Box"));
+    ASSERT_TRUE(configNode);
+    std::vector<ConfigurationAtom> before = configNode->configuration().atoms();
+
+    testGraph.loadRestart(path);
+    std::vector<ConfigurationAtom> after = configNode->configuration().atoms();
+
+    auto getR = [](const auto x) { return x.r(); };
+    EXPECT_TRUE(std::ranges::equal(before | std::views::transform(getR), after | std::views::transform(getR)));
```
trisyoungs, line 37: This is a lovely little construction!

trisyoungs (commented):
The intent here is good, I think, but it brings up a discussion regarding the data sources we need to serialise into a node restart file (incidentally, I think we should move away from calling this data "restart" data, and think about it just in terms of "data the node needs to save to ensure its state").

What you have here focuses quite rightly on the core member variable data of  `Configuration` - although the finished product needs `Molecule`-centred information rather than `Atom`ic data which brings its own complications. This data isn't calculated per se - it's inherent to the object. Other nodes specifically calculate data - this is already captured to a large degree with the `addSerialisable()`, `serialiseData()` and `deserialiseData()` functions, so to generate a full restart state packet for a node in general both the inner and "calculated" data must be written.

PR #2598, opened 2026-08-28
refactor: Instances in species

trisyoungs:
This PR adds instance storage to the `Species` class, introduces the `InstantiateNode` for adding detected molecule replicas into a configuration, and performs some refactoring on the CIF unit tests.


src/nodes/detectMolecules.cpp
```diff
@@ -214,46 +221,50 @@ NodeConstants::ProcessResult DetectMoleculesNode::process()
...
-            continue;
-
-        addOutput(paramName, "Detected molecular structure", detectedStructures_[i]);
+        // Update output data if it already exists, otherwise create new
+        auto outputIt = outputs_.find(name);
+        if (outputIt == outputs_.end())
+            addOutput(name, "Detected molecular structure", structure);
+        else
+        {
+            // TODO Need to update the underlying ParameterBase here, but how?
```
RobBuchananCompPhys, line 266: Is this outlying task going to be a blocker on the proper functioning of the node? Why/in what way exactly do we need to update the ParameterBase?

src/classes/configuration.cpp
```diff
@@ -66,4 +67,5 @@ void Configuration::deserialise(const SerialisedValue &node)
     setTemperature(Deserialisable::deser_or<double>(node, "temperature", defaultTemperature_));
     requestedSizeFactor_ = Deserialisable::deser_or<double>(node, "sizeFactor", defaultSizeFactor_);
     requestedCellDivisionLength_ = Deserialisable::deser_or<double>(node, "cellDivisionLength", defaultCellDivisionLength_);
+    setBox(Box(Deserialisable::deser<Vector3>(node.at("lengths")), Deserialisable::deser<Vector3>(node.at("angles"))));
```
RobBuchananCompPhys, line 70: Hmm, there's something up with the fact we use exactly the same code inside the actual Box `deserialise` method, but then duplicate it here to construct a new `Box` in place. I guess the de/serialisation code needs an existing object to operate, so there simply may be no clean way to reduce the duplication but...
I think there could be a static factory method to perform `Box` construction, inline like this

I'd maybe go with
```
static Box Box::read(const SerialisedValue &node)
{
      return  Box(Deserialisable::deser<Vector3>(node.at("lengths")), Deserialisable::deser<Vector3>(node.at("angles"));
}
```

and then in 
```
void Box::deserialise(const SerialisedValue &node)
{
    (*this) = read(node);
}
// using a copy assignment, which could call initialise(otherBox.lengths, otherBox.angles) 
```
Just a thought!

src/nodes/replicatedConfiguration.cpp
```diff
@@ -0,0 +1,64 @@
...
+                auto delta = box.axes() * Vector3(ix, iy, iz);
+
+                // Copy all molecules in the original configuration
+                for (auto &mol : configuration_->molecules())
+                    replicated_.copyMolecule(*mol)->translate(delta);
+            }
+
+    replicated_.updateObjectRelationships();
+
+    message("Created {}x{}x{} replica - {} atoms total.\n", repeat_.x, repeat_.y, repeat_.z, replicated_.nAtoms());
```
RobBuchananCompPhys:
```suggestion
    message("Configuration replicated: New dimensions are {}x{}x{}, {} atoms total.\n", repeat_.x, repeat_.y, repeat_.z, replicated_.nAtoms());
```
Really pedantic suggestion here (!). Although replication sounds fine as a description of the process taking place, the idea of a "replica" doesn't quite sit right (sounds a little too much like "made an exact copy"!). I suggest we frame this message slightly differently.

src/classes/species.cpp
```diff
@@ -191,6 +192,10 @@ void Species::serialise(std::string tag, SerialisedValue &target) const
     Serialisable::vector<>(impropers_, "impropers", result);
     Serialisable::fromVectorToTable<>(isotopologues_, "isotopologues", result);
     Serialisable::fromVectorToTable<>(sites_, "sites", result);
+    if (box_.type() != Box::BoxType::None)
+        result["box"] = Serialisable::ser(box_);
+    Serialisable::vector<>(instances_, "instances", result, [](const auto &instance)
+                           { return Serialisable::vector(instance, [](const auto &v) { return Serialisable::ser(v); }); });
```
RobBuchananCompPhys, line 198: Line feels a bit busy and less readable due to the repeated namespace calls. Could we put a `using namespace Serialisable` just within this scope?

Additionally, I wonder if the wrapping `return Serialisable::vector(instance, [](const auto &v) { return Serialisable::ser(v);` is strictly necessary.
```suggestion
using namespace Serialisable;
// ...
    vector<>(instances_, "instances", result, [](const auto &instance)
                           { return { instance, [](const auto &v) { return ser(v); } }; });
```

RobBuchananCompPhys (changes requested):
Looks very good 👍 , particularly like the clean logic flow in `ReplicatedConfigurationNode`. Would be good to get your thoughts on a couple of suggestions!

src/nodes/detectMolecules.cpp
trisyoungs: So this was a realisation that I had - the `ParameterBase` stores a reference to the `Structure`, but if the `DetectMolecules` node runs again the `Structure`s are all recreated so the reference won't be valid any longer.  We can't just replace the reference with another one, so we would have to recreate the dynamic output which means losing any connectivity we had to it before.  Maybe this isn't a big issue in practice because - we need the UI to find out I guess!

src/classes/configuration.cpp
trisyoungs: I think I would prefer to leave as-is - `Box` construction like this is fairly commonplace in the code, and it's three extra lines to save one duplicate line.

src/nodes/replicatedConfiguration.cpp
trisyoungs: OK, what about "Configuration contents replicated ({}x{}x{}) - {} atoms in new configuration."?

src/classes/species.cpp
trisyoungs: This suggestion would be against the Google style guide which we more-or-less adhere to. Personally I still prefer the explicit namespace usage on functions as it leaves no doubt as to their origin!

RobBuchananCompPhys (approved):

PR #2599, opened 2026-09-03
fix: Properly deserialise loopback edges

rprospero:
This fixes the issue with deserialising the loopback edges.  I've also added a small test to confirm that the number of loopback edges are correct after the roundtrip.

The extra test is because I briefly thought that I had solved the problem, but found that I had merely prevented loopbacks from exporting, so both the original and the roundtrip showed no loopbacks and the test passed, despite the roundtrip not actually reconstructing the same graph.


RobBuchananCompPhys (approved):
I haven't tested this out locally, so I assume the tests added here are passing, but this looks spot on! 👍

trisyoungs (approved):

PR #2606, opened 2026-09-15
refactor: QML GUI stack [Part 2] -> UI components

RobBuchananCompPhys:


src/gui2/models/graphModel.cpp
```diff
@@ -0,0 +1,346 @@
...
+        dX = dist(rnG_);
+        dY = dist(rnG_);
+
+        // If we are outside the graph view's border area, continue
+        if ((x + dX < border || x + dX > (maxX - border)) || (y + dY < border || y + dY > (maxY - border)))
+            continue;
+
+        if (!occupied.contains({x + dX, y + dY}))
+            isOccupied = false;
+    }
```
trisyoungs, line 261: This feels a little risky! Is there a guarantee that a position will always be found?

src/gui2/models/graphModel.cpp
```diff
@@ -0,0 +1,346 @@
...
+    {
+        auto &[sourceDropArea, targetDropArea] = endPoints_[i];
+        auto sourceParentNode = sourceDropArea->property("parentNodeBox").value<QObject *>();
+        auto sourceParentNodeName = sourceParentNode->property("nodeName").toString().toStdString();
+        auto targetParentNode = targetDropArea->property("parentNodeBox").value<QObject *>();
+        auto targetParentNodeName = targetParentNode->property("nodeName").toString().toStdString();
+        return sourceParentNodeName == node->name() || targetParentNodeName == node->name();
+    };
+
+    for (int row = endPoints_.size() - 1; row >= 0; --row)
```
trisyoungs, line 286:
```suggestion
    for (auto row = endPoints_.size() - 1; row >= 0; --row)
```

src/gui2/models/graphModel.cpp
```diff
@@ -0,0 +1,346 @@
...
+        if (!occupied.contains({x + dX, y + dY}))
+            isOccupied = false;
+    }
+}
+
+void GraphModel::handleReset() { Q_EMIT(graphChanged()); }
+
+void ParameterEndPointsModel::add(QQuickItem *sourceDropArea, QQuickItem *targetDropArea)
+{
+    int row = endPoints_.size();
```
trisyoungs, line 268:
```suggestion
    auto row = endPoints_.size();
```

src/gui2/models/graphModel.h
```diff
@@ -0,0 +1,177 @@
...
+#pragma once
+
+#include "graphEdgeModel.h"
+#include "gui2/models/graphNodeModel.h"
+#include "gui2/models/nodeWrapper.h"
+#include "nodes/edge.h"
+#include "nodes/graph.h"
+#include <QAbstractListModel>
+#include <QQuickItem>
+#include <qtmetamacros.h>
```
trisyoungs, line 13: This isn't typically how Qt headers get included - is there an UpperCamelCase equivalent?

src/gui2/models/nodeMessages.cpp
```diff
@@ -0,0 +1,177 @@
...
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#include "nodeMessages.h"
+
+// Message store
+void NodeMessages::setMessageStore()
+{
+    auto sourceNode = graphModel_->graph()->findNode(nodeName_.toStdString());
+    messageStore_ = sourceNode->messages();
+}
```
trisyoungs, line 11: Would this be better done in `setNodeName()`? Also, there is no check for `sourceNode` being a valid node pointer.

src/gui2/models/nodeRegistryModel.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+// Instantiate node from registry
+void NodeRegistryModel::instantiateNode(int x, int y, QVariant type)
+{
+    increment(type.toString());
+    const auto count = tally(type.toString());
+    std::string prefix = type.toString().toStdString() + "_";
+    auto name = prefix + std::format("{}", count);
+    graphModel_->emplace_back(x, y, type, name, true);
+}
```
trisyoungs, line 53: I guess all of this is to generate unique names for newly-added nodes, but this way of doing it seems like quite a bit of effort. Also, it always adds a number to the node name, whereas I think I would prefer it if the first node of a given type was just called by its name without a number suffix.  Perhaps the unique name generation is best moved out of the NodeRegistryModel and ensured at the point of use - or, it could be done internally by the `Graph` itself when it gets added.

src/gui2/models/nodeRegistryModel.h
```diff
@@ -16,30 +17,44 @@ class NodeRegistryModel : public QAbstractListModel
...
-    using NodeRegistryDisplayElement = std::tuple<std::string_view, std::string_view>;
+    using NodeRegistryDisplayElement = std::tuple<QString, QString, int>;
 
     Q_OBJECT
 
     public:
     explicit NodeRegistryModel(QObject *parent = nullptr);
     ~NodeRegistryModel() override = default;
 
+    enum NodeDisplayRoles
```
trisyoungs, line 35: Can this be an `enum class` so that it is better highlighted in the code? Seeing `roles[Name]` in the code lacks context IMHO.

trisyoungs: Can Q_ENUM even handle `enum class`?

src/gui2/models/nodeRegistrySortFilterModel.cpp
```diff
@@ -0,0 +1,32 @@
...
+    auto fuzzyTypes = NodeRegistry::getNodeTypesFuzzy(searchText_.toStdString());
+
+    return std::find(fuzzyTypes.begin(), fuzzyTypes.end(), rowName) != fuzzyTypes.end();
+}
+
+QString &NodeRegistrySortFilterModel::searchText() { return searchText_; }
+
+void NodeRegistrySortFilterModel::setSearchText(QString &text)
+{
+    // Expect 62 rows and 1 column in source model
```
trisyoungs, line 28:
```suggestion
```

src/nodes/clearBonding.cpp
```diff
@@ -6,7 +6,7 @@
 ClearBondingNode::ClearBondingNode(Graph *parentGraph) : Node(parentGraph)
 {
     // Inputs
-    addOutput("Structure", "Input structure", inputStructure_);
+    addInput("Structure", "Input structure", inputStructure_);
```
trisyoungs, line 9: Hah - I've definitely made this change somewhere before too!  I wonder in which PR that got lost in...

trisyoungs (changes requested):

src/gui2/models/nodeMessages.cpp
RobBuchananCompPhys: Yes, that could work, renaming the node would force the re-binding to the `nodeName` from QML and it would pull its messages *then*, rather than where it's currently called (less immediately).

src/gui2/models/nodeRegistryModel.h
RobBuchananCompPhys: I don't think it needs to be its own `enum`/`enum class` as it's not used outside the `NodeRegistryModel` class. I'd just use the typical `Roles` pattern as elsewhere.

src/gui2/models/nodeRegistryModel.cpp
RobBuchananCompPhys: Let's keep this code inside the model for now, as moving it to the internal `Graph` would be quite complex (the record of how many types are instantiated is kept inside the model's list of tuples`, but yes we can definitely do a pure-type-as-name for the first node of a kind added.

trisyoungs: Also, note that we already have `Graph::uniqueNodeName()` that can be used instead.

src/gui2/models/graphEdgeModel.h
```diff
@@ -25,16 +25,20 @@ class GraphEdgeModel : public QAbstractListModel
...
-    bool dropEdge(Edge &edge);
+    bool deleteEdge(Edge &edge);
 
-    void deleteNode(std::string index)
+    void deleteByNode(std::string index)
     {
+        // TODO: Something weird happening here - the internal edges() appears to update *later* than the actual source
+        // graph_::edges even though they reference the same thing. This leads to the below iterator returning potentially empty
+        // edges, as the deletion has already happened, but apparently this model doesn't know about it. So, we check the edge
+        // isn't empty.
```
rprospero, line 38: We should create an issue to track that this is a problem, as it's likely a sign of a deeper issue.

src/gui2/models/graphNodeModel.cpp
```diff
@@ -83,6 +86,8 @@ QVariant GraphNodeModel::data(const QModelIndex &index, int role) const
             return QVariant::fromValue(item.options.get());
         case INNER_GRAPH:
             return item.hasInner();
+        case IS_ROOT_NODE:
+            return dynamic_cast<DissolveGraph *>(item.rawValue().parentGraph()) != nullptr;
```
rprospero, line 90: Small thing - the role is `IS_ROOT_NODE`, but it looks like it returns `false` for the root node and `true for other nodes.  Shouldn't this be an == instead of a !=?

src/gui2/models/graphModel.cpp
```diff
@@ -0,0 +1,312 @@
...
+{
+    // FIXME
+    return false;
+}
+// Public wrapper of disconnect_
+bool GraphModel::disconnect(std::string source, int sourceIndex, std::string destination, int destinationIndex)
+{
+    // FIXME
+    return false;
+}
```
rprospero, line 183: Do we still need these methods?  It feels like `addEdge` does the same job and these functions are currently stubs.

src/gui2/models/graphModel.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+// Return bool - true if node exists in graph
+bool GraphModel::isValidNode(QVariant nodeName) const { return graph_->findNode(nodeName.toString().toStdString()); }
+
+// Return bool - true if graph run was successful/unchanged, false if failed
```
rprospero, line 188: The comment explains that the function returns `true` or `false`, but the type is void

src/gui2/models/graphModel.cpp
```diff
@@ -0,0 +1,312 @@
...
+    auto name = nodeName.toString().toStdString();
+    auto node = graph_->findNode(name);
+    graphRunComplete(node->run(), name);
+}
+
+int GraphModel::indexByName(std::string_view name)
+{
+    // FIXME
+    return 0;
+}
```
rprospero, line 200: Again, could we possibly delete this function?

src/gui2/models/nodeMessages.h
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// Copyright (c) 2026 Team Dissolve and contributors
+
+#include "gui2/models/graphModel.h"
+#include "nodes/node.h"
+#include <QObject>
+#include <memory>
+
+class NodeMessageModel : public QAbstractListModel
```
rprospero, line 9: Might be worth a few comments on what this class is for.  I assume that it's collecting the log messages from all the nodes.  Or possibly a single node?

src/gui2/models/nodeMessages.h
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public:
+    int rowCount(const QModelIndex &parent = QModelIndex()) const override;
+    QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const override;
+    Qt::ItemFlags flags(const QModelIndex &index) const override;
+    QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const override;
+    QHash<int, QByteArray> roleNames() const override;
+};
+
+class NodeMessages : public QObject
```
rprospero, line 46: Again, could use a quick explanatory comment

src/gui2/models/nodeMessages.cpp
```diff
@@ -0,0 +1,175 @@
...
+QString NodeMessages::nodeName() { return nodeName_; }
+
+// Return the parent node
+void NodeMessages::setParent(QObject *parent) { parent_ = parent; }
+
+// Set the parent node
+QObject *NodeMessages::parent() { return parent_; }
+
+// Update all
+void NodeMessages::updateMessages()
```
rprospero, line 52: This is more of something that will need to be added as an issue, since it would be far outside the scope of this PR, but I think that having separate models for the different logging levels of messages is the wrong abstraction.

Specifically, this would lead us toward an architecture where we would likely display the info in one list, the warnings in a second, and the errors in a third.  This will make difficult, as it will be tricky to figure out the order in which the messages occur.  I believe that we are better off having a single list with each message tagged it its level, but with the messages otherwise freely mixed.  This also means that we don't need to handle the individual cases of there being no messages at each level.

src/gui2/models/nodeRegistryModel.cpp
rprospero: We don't have to move this into graph to make life simpler.  Instead of needing the `increment` and `decrement` members, we could have the code look like

```cpp
int count = 0
auto name = type.toString().toStdString();
while(graphModel_->containsNode(name))
  name = std::format({}+{}, type.toString().toStdString(), ++count);
```

Obviously, this would require creating the `containsNode` method, which feels like it will come in handy again, and we might not want to do the prefix increment trickery, but it does keep the logic here while simplifying the other code.

I'm also concerned about `increment` and `decrement` because I think that `decrement` won't work.  Imaine that I add seven number nodes `Number01` through `Number07`.  I then delete `Number03` and decrement gets called.  The next Number node I add would then be `Number07` (since the count was decremented), but that node already exists.  With the loop code above, the new node would be created as `Number03`, as that's the name the first name that hadn't already been taken.  It's also nice because, if I create seven nodes and then give them reasonable names, the next node that I create would be `Number01`, instead of `Number08` under the increment system.

src/gui2/models/nodeRegistryModel.h
```diff
@@ -16,30 +17,43 @@ class NodeRegistryModel : public QAbstractListModel
...
-     * - Description/Node::summary (string)
+     * - Name/Node::type (QString)
+     * - Description/Node::summary (QString)
      * - Category (enum) - currently unavailable
+     * - Icon path (QUrl) - currently unavailable
+     * - tally (int)
      *
      */
-    using NodeRegistryDisplayElement = std::tuple<std::string_view, std::string_view>;
+    using NodeRegistryDisplayElement = std::tuple<QString, QString, int>;
```
rprospero: Instead of a three element tuple, should we go ahead and convert this into a proper `struct`?  It'll be easier to access.  Generally, pairs and tuples are nice for anonymous, one-off data types, but, since we've already given this a name, we might as well make it a real type.

src/gui2/models/nodeRegistryModel.h
```diff
@@ -16,30 +17,43 @@ class NodeRegistryModel : public QAbstractListModel
...
     private:
+    // Increment node type
+    void increment(QString nodeType);
+    // Decrement node type
+    void decrement(QString nodeType);
+    // Return tally for node type
+    int &tally(QString nodeType);
     // Source node registry data
-    static std::vector<const NodeRegistryDisplayElement> elements_;
+    static std::vector<NodeRegistryDisplayElement> entries_;
```
rprospero, line 54: I'm a little wary about static members like this.  Is there a reason why two separate node registries should see the same entries?

src/gui2/qml/controls/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+qt_add_library(controls_module STATIC)
+
+qt6_add_qml_module(
```
rprospero, line 3: I'm very happy that we're dividing things up into proper QML modules.  Thank you for getting this work going!

src/gui2/qml/DissolveMain.qml
```diff
@@ -14,13 +21,53 @@ ApplicationWindow {
...
+
+            onAccepted: {
+                if (quickRunDialog.graphModel.isValidNode(startNode))
+                    graphModel.run(startNode)
+
+                quickRunDialog.close()
+            }
+            onRejected: quickRunDialog.close()
+        }
+    }
```
rprospero, line 35: We might want to split this dialog off into its own file.

src/gui2/qml/nodeGraph/NodeBox.qml
```diff
@@ -6,45 +6,187 @@ import Qt.labs.qmlmodels
...
+        }
+        NumberAnimation {
+            target: creationPulsedBorder
+            property: "border.width"
+            from: 5
+            to: 1
+            duration: 500
+            easing.type: Easing.OutQuad   
+        }
+    }
```
rprospero, line 67: I know that we have some neurodiverse users who would prefer less animations in their applications, so we will eventually need to add a setting to allow users to disable this animation.

That said, I am immensely pleased that you are taking advantage of these animation options.  I've been playing with them a bit on some of my own private projects and they really showcase the power of QML and what is possible with this sort of declarative setup.

src/gui2/qml/nodeGraph/NodeBox.qml
```diff
@@ -6,45 +6,187 @@ import Qt.labs.qmlmodels
...
-        Button {
-            contentItem: Image {
-                fillMode: Image.PreserveAspectFit
-                source: "qrc:/IconsModule/cross.svg"
-                sourceSize.height: titleLabel.height
-                sourceSize.width: titleLabel.height
+        implicitHeight: root.headerHeight
+        width: root.width
+
+        color: '#a9f0f4ff'
```
rprospero, line 75: We will need a color scheme singleton somewhere in our QML to keep track of these colours.  Beyond providing support for dark modes or visually impaired users needing high contrast displays, it will also save us a good deal of trouble on keeping these colours consistent across the application.

src/gui2/qml/nodeGraph/GraphDelegate.qml
```diff
@@ -3,32 +3,319 @@ import QtQuick.Controls
...
+                    }
+                    background: Rectangle {
+                        color: messageDelegate.hovered ? "#444444" : "transparent"
+                    }
+                }
+            }
+        }
+    }
+    Popup {
+        id: warnings
```
rprospero, line 217: This ties back into my other point about having a single, combined location for all errors levels.

rprospero (approved):
I've included a fairly lengthy list of issues, but they are all exist as future work (that you may have even covered higher in the stack) and no necessarily as blockers on this specific PR.

src/gui2/models/graphModel.cpp
RobBuchananCompPhys: This is handled in a later PR

src/gui2/models/graphModel.cpp
RobBuchananCompPhys: See part 4

src/gui2/models/nodeMessages.cpp
RobBuchananCompPhys: Very good point!

src/gui2/models/nodeRegistryModel.h
RobBuchananCompPhys: I don't think there's any reason there should be more than one instantiated node registries. Perhaps it should be registered and exposed as a `Q_SINGLETON_TYPE`? This might clarify the inclusion of `static` members here, but if we enforce only one object at runtime, we could get away with non-`static`.

src/gui2/qml/nodeGraph/NodeBox.qml
RobBuchananCompPhys: Yes, I'm sure this can be controlled by a simple `enabled` bool property for these kinds of animations. I do like them!

src/gui2/models/nodeRegistryModel.h
RobBuchananCompPhys: That could work, and I'm not entirely sure why I didn't for a `struct` in the first place. I like the idea, and we should probably go ahead with this.

src/gui2/models/graphModel.cpp
RobBuchananCompPhys: I'll tackle this in part 4, as this code is expanded later

PR #2607, opened 2026-09-15
refactor: QML GUI stack [Part 3] -> Categorical `NodeRegistry`

RobBuchananCompPhys:


src/nodes/registry.cpp
```diff
@@ -68,85 +68,120 @@
...
-std::map<std::string_view, NodeProducer> NodeRegistry::producers_;
-
 // Makes unique pointer to derived node instance
 template <typename T> NodeProducer makeDerivedNode()
 {
     return [=](Graph *parent) -> std::unique_ptr<Node> { return std::make_unique<T>(parent); };
 }
 
+// Static Singletons
+const std::map<NodeRegistry::Category, ProducerMap> NodeRegistry::categoricalProducers_{
```
trisyoungs, line 78: Excellent. I was going to suggest we did something like this!

src/nodes/registry.cpp
```diff
@@ -68,85 +68,120 @@
...
+                                                               {Category::Export, "Export"},
+                                                               {Category::Graphs, "Graphs"},
+                                                               {Category::Import, "Import"},
+                                                               {Category::Math, "Math"},
+                                                               {Category::Other, "Other (TODO: finish categories)"},
+                                                           });
+}
+EnumOptions<NodeRegistry::Category> getEnumOptions(NodeRegistry::Category) { return NodeRegistry::category(); }
+
+// Remove categories from categorial node producer map, returning a 'flat' producer map
```
trisyoungs, line 167:
```suggestion
// Remove categories from categorical node producer map, returning a 'flat' producer map
```

src/nodes/registry.h
```diff
@@ -7,15 +7,35 @@
...
+    public:
+    static const std::map<Category, ProducerMap> categoricalProducers_;
+
     private:
     // Available Node producers
-    static std::map<std::string_view, NodeProducer> producers_;
+    static ProducerMap producers_;
 
     private:
+    // Remove categories from categorial node producer map, returning a 'flat' producer map
```
trisyoungs, line 37:
```suggestion
    // Remove categories from categorical node producer map, returning a 'flat' producer map
```

trisyoungs (approved):
This looks great to me. Nice catches on the various incorrect `Node::type` entries!

src/gui2/models/nodeRegistryModel.cpp
```diff
@@ -42,16 +43,43 @@ int &NodeRegistryModel::tally(QString nodeType)
...
+    for (const auto &[name, _] : nodes[categoryEnum])
+    {
+        auto nodeName = QString::fromStdString(std::string(name));
+        auto descriptionIt = std::find_if(entries_.begin(), entries_.end(),
+                                          [&nodeName](const auto &entry) { return std::get<0>(entry) == nodeName; });
+        auto descriptionIdx = std::distance(entries_.begin(), descriptionIt);
+        auto nodeDescription = std::get<1>(entries_[descriptionIdx]);
+        QVariantMap nodeInfo;
+        nodeInfo[QString::fromStdString("name")] = nodeName;
+        nodeInfo[QString::fromStdString("description")] = nodeDescription;
```
rprospero:
```suggestion
        auto entry = std::find_if(entries_.begin(), entries_.end(),
                                          [&nodeName](const auto &entry) { return std::get<0>(entry) == nodeName; });
        QVariantMap nodeInfo;
        nodeInfo[QString::fromStdString("name")] = std::get<0>(entry);
        nodeInfo[QString::fromStdString("description")] = std::get<1>(entry);
```

I haven't tested the above code, but it seems like we should be able to directly access the entry from the iterator, instead of needed to get its index.

This is also a good example of why we might want to change the entries to a simple named struct, because it would be even nicer to write

```cpp
        nodeInfo[QString::fromStdString("name")] = entry.name;
        nodeInfo[QString::fromStdString("description")] = entry.description;
```

src/gui2/qml/nodeGraph/GraphView.qml
```diff
@@ -24,133 +23,51 @@ Pane {
...
-                }
-                MenuItem {
-                    text: "Multiply"
-
-                    onClicked: graphRoot.rootGraphModel.emplace_back(Math.round(ctxMenuCatcher.mouseX), Math.round(ctxMenuCatcher.mouseY), "Multiply", "New Node")
-                }
-                MenuItem {
-                    text: "Number"
+            Repeater {
+                model: nodeRegistry.categories()
```
rprospero, line 32: I'm much happier with this being auto-populated now

rprospero (approved):
I like the new registry, though I've added a suggestion to simplify the code.

src/gui2/models/nodeRegistryModel.cpp
RobBuchananCompPhys: Agreed, with the new `struct` version we can now do this

PR #2608, opened 2026-09-15
refactor: QML GUI stack [Part 4] -> More UI components

RobBuchananCompPhys:


src/gui2/models/legacy/simpleForcefieldModel.cpp
trisyoungs: Just a general comment here. Another issue I have been thinking about is how we handle forcefields in the new UI. I am tempted at this point to have each forcefield represented by a `Node` of its own, so there is no need for a central list / store of the available `Forcefield` objects.  This means that there can be less strict adherence to using the `Forcefield` class to implement a forcefield, which sounds odd to say, but one special type of forcefield we definitely need is one which takes the current molecule geometry (bonds, angles at least) and applies some sensible force constants to it.  This doesn't fit well in within the `Forcefield` class, but it could be implemented however we like if we just have a `Node` for it.

src/gui2/models/graphEdgeModel.cpp
```diff
@@ -18,44 +19,110 @@ GraphEdgeModel::GraphEdgeModel(GraphModel *parent, Graph *&graph) : parent_(pare
...
+        auto targetNodeName = edgePtr->targetNode().name();
+        return (nodeName == sourceNodeName || nodeName == targetNodeName);
+    };
+
+    auto &allEdges = edges();
+    for (auto edgeIt = allEdges.begin(); edgeIt != allEdges.end();)
+    {
+        if (toRemove(*edgeIt))
+        {
+            const int row = std::distance(allEdges.begin(), edgeIt);
```
trisyoungs:
```suggestion
            const auto row = std::distance(allEdges.begin(), edgeIt);
```

src/gui2/models/graphModel.cpp
```diff
@@ -115,42 +160,49 @@ void GraphModel::setCanvasDimensions(const QSizeF &canvasDimensions)
...
+void GraphModel::emplace_back(int x, int y, QString type, QString name)
 {
     if (!graph_)
         Messenger::exception(
             "GraphModel has no graph.  This should have been impossible.  Please let the Dissolve developers know about this.");
     nodes_.beginInsertRows({}, graph_->nodes().size(), graph_->nodes().size() + 1);
-    auto nodeType = type.toString().toStdString();
-    auto node = graph_->createNode(nodeType, name);
+    auto nodeType = type.toStdString();
+    auto node = graph_->createNode(nodeType, name.toStdString());
```
trisyoungs:
```suggestion
    auto node = graph_->createNode(type.toStdString(), name.toStdString());
```

src/gui2/models/nodeMessages.cpp
```diff
@@ -135,15 +219,29 @@ void NodeMessageModel::setMessages(std::vector<QString> messages)
...
+    auto &[status, msg] = messages_[index.row()];
+    switch (role - Qt::UserRole)
+    {
+        case Roles::Message:
+            return QString::fromStdString(msg);
+        case Roles::StatusColor:
+        {
+            if (status == Node::MessageStatus::Info)
+                return QColor("white");
+            if (status == Node::MessageStatus::Info)
```
trisyoungs:
```suggestion
            if (status == Node::MessageStatus::Warn)
```

src/gui2/models/parameterEndPointsModel.h
```diff
@@ -0,0 +1,67 @@
...
+
+    private:
+    // Vector of parameter endpoint QQuickItem * pairs
+    ParameterEndPoints endPoints_;
+
+    public:
+    // Return vector of parameter endpoint QQuickItem * pairs
+    ParameterEndPoints &endPoints();
+    const ParameterEndPoints &endPoints() const;
+    //
```
trisyoungs: Need a comment here!

src/nodes/detectMolecules.cpp
```diff
@@ -224,14 +224,16 @@ NodeConstants::ProcessResult DetectMoleculesNode::process()
                                                if (fragmentMatch.matched().empty())
                                                    return false;
 
-                                               std::cout << "Fragment:\n";
-                                               // Store this match as an instance
+                                               // std::cout << "Fragment:\n";
```
trisyoungs:
```suggestion
```

src/nodes/detectMolecules.cpp
```diff
@@ -224,14 +224,16 @@ NodeConstants::ProcessResult DetectMoleculesNode::process()
...
                                                for (const auto fragmentAtom : fragmentMatch.matched())
                                                {
-                                                   std::cout << std::format(
-                                                       "{} @ {} {} {}\n", Elements::symbol(fragmentAtom->Z()),
-                                                       fragmentAtom->r().x, fragmentAtom->r().y, fragmentAtom->r().z);
+                                                   /*
+                                                      std::cout << std::format(
+                                                          "{} @ {} {} {}\n", Elements::symbol(fragmentAtom->Z()),
+                                                          fragmentAtom->r().x, fragmentAtom->r().y, fragmentAtom->r().z);
+                                                          */
```
trisyoungs:
```suggestion
```

src/nodes/cif/importCIFStructure.cpp
```diff
@@ -498,10 +498,12 @@ bool ImportCIFStructureNode::createStructure()
 // Perform processing
 NodeConstants::ProcessResult ImportCIFStructureNode::process()
 {
+    auto filePath = filePath_.string();
+
```
trisyoungs: This doesn't really need to be put into a temporary and could be passed directly to `read()`.

src/nodes/graph.cpp
```diff
@@ -226,6 +226,47 @@ Edge *Graph::findEdge(const EdgeDefinition &definition) const
...
+    while (it != edges_.end())
+    {
+        edges.push_back(edges_[std::distance(edges_.begin(), it)].get());
+        it = std::find_if(it + 1, edges_.end(),
+                          [&](const auto &edge)
+                          {
+                              auto otherDefinition = edge->definition();
+                              return otherDefinition.sourceNode == sourceNode && otherDefinition.sourceOutput == sourceOutput;
+                          });
+    }
```
trisyoungs: Wouldn't this be better as a simple `for` loop over `edges_`? We're just assembling a list of matching edges...
```
    for (const auto &edge : edges_)
    {
        auto definition = edge->definition();
        if (otherDefinition.sourceNode == sourceNode && otherDefinition.sourceOutput == sourceOutput)
            edges.push_back(edge.get());
    }
```

src/nodes/iterableGraph.cpp
```diff
@@ -35,6 +32,9 @@ LoopBacksNode *IterableGraph::loopBacks() const { return loopBacks_; }
 // Return loop edges
 const Graph::Edges &IterableGraph::loopEdges() const { return loopEdges_; }
 
+//
+void IterableGraph::updateInputs() { setLoopBacks(); }
```
trisyoungs: If `updateInputs()` only does `setLoopBacks()`, can we just call `setLoopBacks()` where required?

src/nodes/parameter.h
```diff
@@ -82,6 +82,8 @@ class ParameterBase
     // Set node parent
     void setParent(Node *parent);
     // Return the parameter name
+    bool setName(std::string name);
```
```suggestion
// Set the parameter name
    bool setName(std::string name);
```

trisyoungs (approved):
Some minor comments and suggestions.  I haven't performed a live test of the code at this point - that will be my job for next week!

PR #2610, opened 2026-09-15
chore: Export cif in test to a tempfile

rprospero:
Small thing, but this ensures that the CIF test exports to a temp file, instead of dumping the output in the test data folder.


tests/nodes/cif.cpp
```diff
@@ -106,7 +107,8 @@ class CIFNodeTest : public ::testing::Test
         auto *cfg = lastNode->getOutputValue<Configuration *>("Configuration");
         if (!cfg)
             return testing::AssertionFailure() << "Failed to retrieve reconstructed configuration.";
-        ExportXYZConfigurationNode::exportConfiguration(cfg, "THIS.xyz");
+        TempFile temp("THIS.xyz");
+        ExportXYZConfigurationNode::exportConfiguration(cfg, temp);
```
trisyoungs: To be honest, this was left over from when I was debugging the `DetectMoleculesNode`, so it can be fully removed.

PR #2611, opened 2026-09-15
feat: Unit test for molecule ordering in DetectMolecules

trisyoungs:
This PR adds an explicit unit test for `DetectMoleculesNode` working with non-contiguous atom ordering within molecules.


tests/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    for (const auto &structureAtom : structure.atoms())
+    {
+        auto r = fold ? configuration->box().fold(structureAtom->r()) : structureAtom->r();
+        if (std::ranges::find_if(configuration->atoms(),
+                                 [&structureAtom, r](const auto &cfgAtom)
+                                 {
+                                     return structureAtom->Z() == cfgAtom.Z() && fabs(r.x - cfgAtom.r().x) < 1.0e-6 &&
+                                            fabs(r.y - cfgAtom.r().y) < 1.0e-6 && fabs(r.z - cfgAtom.r().z) < 1.0e-6;
+                                 }) == configuration->atoms().end())
```
RobBuchananCompPhys: Quite a large inline AND like this one could benefit from being split up into discrete values. This could also be useful when printing/debugging.
```suggestion
auto isSameAtom = structureAtom->Z() == cfgAtom.Z();
auto isCloseX = fabs(r.x - cfgAtom.r().x) < 1.0e-6;
auto isCloseY = fabs(r.y - cfgAtom.r().y) < 1.0e-6;
auto isCloseZ = fabs(r.z - cfgAtom.r().z) < 1.0e-6;
                                     return isSameAtom && isCloseX && isCloseY && isCloseZ;
                                 }) == configuration->atoms().end())
```

RobBuchananCompPhys (changes requested):
Looks good, just a single suggested change to consider.

src/nodes/detectMolecules.h
```diff
@@ -18,6 +18,10 @@ class DetectMoleculesNode : public Node
     DetectMoleculesNode(Graph *parentGraph);
     ~DetectMoleculesNode() override = default;
 
+    /*
+     * Definition
+     */
+
```
rprospero: Do we need this comment twice? It looks like the same comment is right below.

tests/nodes/detectMolecules.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <string>
+
+namespace UnitTest
+{
+testing::AssertionResult compareContents(const Structure &structure, const Configuration *configuration, bool fold = true)
+{
+    for (const auto &structureAtom : structure.atoms())
+    {
+        auto r = fold ? configuration->box().fold(structureAtom->r()) : structureAtom->r();
+        if (std::ranges::find_if(configuration->atoms(),
```
rprospero: The call to `find_if` could be replaced with `none_of` and not need the comparison to against `end`

rprospero (approved):
A couple of minor stylistic comments, but otherwise looks good

src/nodes/detectMolecules.h
trisyoungs: We absolutely do not - duplicate removed!

tests/nodes/detectMolecules.cpp
trisyoungs: Well, for my sins I did not know `none_of` existed.

rprospero: I just learned about it last month.

